PR #826, opened 2014-06-08
adding orchestrator

ashley-taylor:
Hello

This is my approach to solving this use case
https://groups.google.com/forum/#!topic/rundeck-discuss/2Ck5dByfsGM

Will also supply a example orchestrator that just makes sure only a maximum percentage of the nodes get processed at once regardless of how many threads are configured.

Will supply this example once we agree on the plugin interfaces

### DIFF
--- a/core/src/main/java/com/dtolabs/client/services/RundeckAPICentralDispatcher.java
+++ b/core/src/main/java/com/dtolabs/client/services/RundeckAPICentralDispatcher.java
@@ -286,7 +286,7 @@ public class RundeckAPICentralDispatcher implements CentralDispatcher {
             data.put("scriptURL", scriptURL);
         }
         if (!isExec && args.size() > 0) {
-            params.put("argString", OptsUtil.join(args));
+            data.put("argString", OptsUtil.join(args));
         }
         addLoglevelParams(params, iDispatchedScript.getLoglevel());
         addAPINodeSetParams(params, iDispatchedScript.isKeepgoing(), iDispatchedScript.getNodeFilter(),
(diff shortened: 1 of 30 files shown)


gschueler:
@ashley-taylor optimistically putting this in 2.5 milestone :pray:

holycow001:
This would be useful for me. We currently use chef to spin up our multiple job servers using chef. Then we execute the job with filter like jobserver*. It would be nice to have a round-robin or load strategy for running the job once and not having the job run on all the servers.

gschueler:
@ashley-taylor finally had a chance to try this out.

I found an issue which is that with a threadcount of T,  if an orchestrator plugin dishes out fewer than T nodes (before returning null from getNode()), then any extra threads will stop any further processing, even if the orchestrator later has T or more nodes to be processed at once.  For your MaxPercentage example this is not a problem since the number of nodes available from the orchestrator at any time is monotonically decreasing.  I tested a plugin which tries to implement the "tier" behavior from #1160 and saw that if e.g. the first "tier" has only 1 node in it, then even if threadcount is 3, the nodes will only be processed one at a time in later tiers. 

I have a change which I believe solves this issue: The main thread get the nodes from the orchestrator and uses a BlockingQueue to send them to the threads for processing.  Here is a gist of the difference https://gist.github.com/gschueler/cce194596439b875650b/revisions

gschueler:
adding one more change, introducing an `isComplete()` method to the Orchestrator interface, to let `null` result from `getNode()` just mean no nodes are available, and `isComplete()==true` means the orchestrator has finished.

PR #972, opened 2014-10-16
Add MSSQL Support

jasonhensler:
I have been able to get mssql to work with rundeck after a few minor code changes. I've tested with the internal H2 DB, MSSQL2012, and Mysql 5.6.15 in a test instance using the war file and tomcat/jdk7. For MSSQL I used the microsoft JDBC driver and for Mysql I used the bundled driver. I couldn't test against oracle however; these changes use hibernate to choose the sql type rather than a hard coded one which should work properly with oracle.


chinmaica:
Hello,

Can anyone please give me the detail steps to connect rundeck with MS sql server db?

Thanks in advance

PR #1029, opened 2014-11-20
ssh-agent forwarding limited to job execution

mathpl:
This will start /usr/bin/ssh-agent on job execution (if enabled), inject the private key the job would normally use in the agent, make the agent available on the newly created jsch connection and enable agent forwarding on the connection. Once the job is done the ssh-agent is killed. Optionally ssh-agent can be started with "-t <ttl>" which will guarantee the private key will be flushed at that time if by some chance it's not killed.

Obviously with this enabled you cannot rely on Rundeck to limit what hosts are accessible to your users, proper key management is required on all hosts.

New variables are:
### framework.properties:

```
framework.local.ssh-agent=<true|false>
framework.local.ttl-ssh-agent=<time in sec>
```
### project.properties:

```
project.local.ssh-agent=<true|false>
project.local.ttl-ssh-agent=<time in sec>
```
### Node properties:

```
local-ssh-agent=<true|false>
local-ttl-ssh-agent=<time in sec>
```

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/common/NodeEntryImpl.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/common/NodeEntryImpl.java
@@ -48,7 +48,6 @@ public class NodeEntryImpl extends NodeBaseImpl implements INodeEntry, INodeDesc
     public static final String NAME = "nodename";
     public static final String TAGS = "tags";
 
-
     /**
      * All attribute names for predefined properties.
      */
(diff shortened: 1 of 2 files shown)


catalinafox:
hello @mathpl, I know this PR is pretty old, but I tried to enable this recently and it didn't work as expected: the ssh-agent was started, but the private key wasn't injected. Did you run any extra commands or add any other extra configuration to make it work as described? 
Thanks

PR #1037, opened 2014-11-28
Added include and exclude node filters into execution data provided to notification plugins

hbakkum:
Hi,

We'd like to be able to display any node filters (specifically, tags) used in a HipChat notification (via hipchat notification plugin) so I've added the include and exclude node filter maps to the execution data map that is passed to a notification plugin. I've also added the list of matched nodes to the execution data too.

Cheers,
Hayden


gschueler:

Thank you!

PR #1124, opened 2015-03-06
JettyPamLoginModule: supplementalRoles split regex requires whitespace

vinzent:
" +" matches "one or more whitespace".

Correctly it should be " *" which matches "zero or more whitespace"


gschueler:
:+1: thanks

PR #1127, opened 2015-03-09
Add default storage encryption plugin

gschueler:
Add a default storage encryption plugin

* [x] use jasypt
* [x] include bouncycastle provider
* [x] algorithm, provider, password can be configured directly or via Env vars or Java system props


gschueler:
added documentation

PR #1134, opened 2015-03-11
Refactor authorization into components, add preauth attribute role

gschueler:
Add ability to use pre-authenticated and authorized proxied requests, e.g. from Apache httpd.

* Requires disabling the container authentication mechanism (remove the `<auth-constraint>` from web.xml)
* Accepts the remote_user as username, and a list of roles defined in a request attribute, the name and delimiter can be configured.


gschueler:
- [ ] documentation

PR #1149, opened 2015-03-23
 Fix LDAP nested groups for Active Directory

### LINKED ISSUE #748: [2.0.4] LDAP authentication does not support nested groups
``` 
Version     : 2.0.4                             Vendor: (none)
Release     : 1.15.GA                       Build Date: Sat 19 Apr 2014 01:29:46 AM CEST
```
When defining the `<security-role>` to a group whose members are other groups, as in this example taken from a `freeIPA` deployment:
```
# ipa group-show rundeck_users
  Group name: rundeck_users
  Description: Rundeck users
  GID: 1168000072
  Member groups: developers, operators
  Indirect Member users: john, jane, bob, lynn, george, matt
  Indirect Member groups: operators_1, operators_2, operators_3
```
authentication does not work, triggering the `HTTP ERROR 403 Problem accessing /rundeck/. Reason: !role` error.

Defining direct members into the group does work as expected.

knifhen:
fixes #748 
caused by case-sensitivity bug in regexp pattern that failed for
ActiveDirectory but worked for openldap


mgarces:
great work! Thanks guys

PR #1244, opened 2015-05-18
Fix: Add a missing space in example

egeland:
The example as given would add a single group with name `"user,readonly"` rather than two groups: `"user"` and `"readonly"`. Adding the space after the comma fixes this problem.


gschueler:
thanks for the fix, what version of rundeck are you using? that issue was fixed in 2.2.0 i believe https://github.com/rundeck/rundeck/commit/f6f3509429d0d2621f01f8c8b321c07baef122d5#diff-734b79ecd8f7d358d5c7c0620ae1547a

egeland:
This happened on 2.4.2-1. So either a regression or a fix-that-wasn't-a-fix..?
I'll check on 2.5.1 later..

gschueler:
no longer necessary

Issue #114, opened 2012-06-23
Add job option to have nodes default to un-checked


gschueler:
It would be useful to have a job option that causes the dispatch nodes to be unchecked by default, such that the job executor must enable the desired hosts at job runtime.  This would allow node selection at runtime, without the risk of the job executor accidentally executing on all nodes defined in the job.

[original LH ticket](http://rundeck.lighthouseapp.com/projects/59277/tickets/470)

Original Creator: Andrew Crawford

jasonhensler:
+1.

gschueler:
trello card: https://trello.com/c/01DpT2jR

richiereynolds:
I can't get that trello link to work I'm afraid. Any ETA on this?

gschueler:
does this work: https://trello.com/c/01DpT2jR

dave-knight:
+1

richiereynolds:
Yes, that link works, thanks.
Is the scope creep into all sort of clever ideas here stopping what would be a very useful, very simple, feature being implemented? Simply having no nodes checked by default instead of all nodes, sounds like a far safer and more sensible default to me for all jobs. If not the default then a simple option on job creation to say "no nodes selected" for this job would be fine too.

gschueler:
@richiereynolds no, this issue is just backlogged, sorry :(  Another related request is to disable the ability to change the target nodes at all, so in the end these two changes might be combined.  The other ideas in that trello card are not a high priority

richiereynolds:
Thanks!

smags:
+1

PR #1245, opened 2015-05-18
add a job option to disable the automatic selection of all nodes

francois-travais:
Nowadays when a user wants to run a predefined job all nodes are automatically selected. Thus the user only need to click on "Run Job Now" to run the job on all those nodes.
In our case we are running jobs on business critical machines and we want to force the user to explicitly select nodes before running the job.

This commit add an option in the job definition to disable this automatic selection of nodes:
scheduled_execution_form

This feature is also visible in the "Definition" tab:
run_job_definition

When this feature is enable for a job, the user has to select node explicitly:
run_job_no_node_selected


gschueler:
thanks for the pull request for this feature, it has been requested often!

related issue  #114

ismailfaruqi:
+1 for this feature :smile_cat:

rangsan:
+1 , can't wait for this feature.

PR #1280, opened 2015-06-09
Fix next execution time on non-scheduled cluster member

noahhl:
In a clustered rundeck installation, the cluster member that isn't the assigned node for a job will display "never" for the next scheduled execution time.

This turns out to be because `trigger.getNextFireTime()` returns `null` if the trigger hasn't been added to a schedule (or more precisely, if the start time hasn't been set yet). Using `trigger.getFireTimeAfter(new Date())` will properly calculate the next execution.


gschueler:
thanks!

PR #1309, opened 2015-06-30
Document filtering on custom attribute in the API

andrewpwade:
I could not find this documented anywhere. Thanks to @gschueler for pointing this feature out on the mailing list.


gschueler:
thanks! sorry, I didn't realize that was not already present in the docs!

Issue #1342, opened 2015-07-16
Add DD/MM/YYY and 24h format for dates and hour


danifr:
![screenshot from 2015-07-16 10-49-20](https://cloud.githubusercontent.com/assets/7135482/8718980/a4c520ba-2ba8-11e5-9b49-47e03b63e81a.png)

danifr:
So, the change is very simple. 
It's just a matter of editing `server/exp/webapp/WEB-INF/grails-app/i18n/messages.properties` [more info](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/rundeck-discuss/wwtxPUf4gBY/5Nh1WPDl28AJ)

PR #1343, opened 2015-07-16
DD/MM/YYYY and 24h format for dates and time

danifr:
https://github.com/rundeck/rundeck/issues/1342

I added this new messages file. Since I only modified date and time format, I named it messages_**en-gb**.properties (english + _european_ date format + 24h time format)

If there is a better way to do it please let me know and I'll be happy to change it.

### DIFF
--- a/rundeckapp/grails-app/i18n/messages_en-gb.properties
+++ b/rundeckapp/grails-app/i18n/messages_en-gb.properties
@@ -1,0 +1,909 @@
+default.doesnt.match.message=Property [{0}] of class [{1}] with value [{2}] does not match the required pattern [{3}]
+default.invalid.url.message=Property [{0}] of class [{1}] with value [{2}] is not a valid URL
+default.invalid.creditCard.message=Property [{0}] of class [{1}] with value [{2}] is not a valid credit card number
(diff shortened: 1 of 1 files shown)


gschueler:
Does it work to include only the modified messages, i.e. just the date format entries, instead of duplicating all of the other messages as well?

danifr:
Hi Greg,

I'm not sure about it. Unfortunately, all my Rundeck servers are using the en_US locale, so I'm modifying directly `/var/lib/rundeck/exp/webapp/WEB-INF/grails-app/i18n/messages.properties`

For this test, I changed the machine's locale to en-GB but nothing changed after restating Rundeck, so I guess there is something wrong with my config.

gschueler:
@danifr I've noticed that grails will actually render pages using the locale requested by the web client browser, so if your browser says your locale is US, that's what response it will use.  You can try adding `?lang=en-GB` to the URL in your browser

ssbarnea:
Have a look at https://github.com/rundeck/rundeck/issues/1682#issue-133928883 -- which must be fixed ASAP. Unless RunDeck aims on being a US-only product.

gschueler:

One change we added was support for selecting the user language in the GUI (under the User > Profile) link.

Thank you!

Issue #1367, opened 2015-07-31
OpenJDK 64-Bit Server VM warning: ignoring option MaxPermSize=256m; support was removed in 8.0


mathieuchateau:
Hello,

I have this warning when I start any cli command:

OpenJDK 64-Bit Server VM warning: ignoring option MaxPermSize=256m; support was removed in 8.0

Can I safely remove it from profile file in rundeck?

ahonor:
Yes, you can. We will need to remove it from the template, too. 


     On Friday, July 31, 2015 6:47 AM, Mathieu Chateau <notifications@github.com> wrote:
   

 Hello,I have this warning when I start any cli command:OpenJDK 64-Bit Server VM warning: ignoring option MaxPermSize=256m; support was removed in 8.0Can I safely remove it from profile file in rundeck?—

gschueler:
@mathieuchateau  we are requiring java 8 in 2.8 release, so I will reopen this to track adding this change to the profile for that release

senselesss64:
yes

PR #1371, opened 2015-08-01
Issue 1367

mathieuchateau:
https://github.com/rundeck/rundeck/issues/1367

OpenJDK 64-Bit Server VM warning: ignoring option MaxPermSize=256m;
support was removed in 8.0


packaging/root/etc/rundeck/profile
```diff
@@ -36,13 +36,15 @@ export RDECK_JVM="-Djava.security.auth.login.config=/etc/rundeck/jaas-loginmodul
 #
 # Set min/max heap size
 #
-RDECK_JVM="$RDECK_JVM -Xmx1024m -Xms256m -XX:MaxPermSize=256m -server"
+RDECK_JVM="$RDECK_JVM -Xmx1024m -Xms256m -server"
```
danifr: But Java 7 is still supported, right? I don't think removing the option is the best solution.

Btw, Java 8 brings a new flag:
>MaxMetaspaceSize: [...] allowing you to limit the amount of native memory used for class metadata. If you don’t specify this flag, the Metaspace will dynamically re-size depending of the application demand at runtime.

[http://javaeesupportpatterns.blogspot.ch/2013/02/java-8-from-permgen-to-metaspace.html](http://javaeesupportpatterns.blogspot.ch/2013/02/java-8-from-permgen-to-metaspace.html)

Cheers,
Daniel.

gschueler:
ideally, it would check for java 7, and only include the `-XX:MaxPermSize` in that case, since we still support it

mathieuchateau:
I will update to push good settings depending on java version

Envoyé de mon iPad

Le 27 août 2015 à 18:33, Greg Schueler <notifications@github.com> a écrit :

ideally, it would check for java 7, and only include the -XX:MaxPermSize in
that case, since we still support it

—
Reply to this email directly or view it on GitHub
<https://github.com/rundeck/rundeck/pull/1371#issuecomment-135486364>.

mathieuchateau:
too old

PR #1403, opened 2015-08-17
Update rundeck.spec

robertdebock:
Both the rundeck and the rundeck-config RPM try to add a user and group. Rundeck requires rundeck-config to be installed, so removed the user/group add from the rundeck RPM.


gschueler:
was it causing an error?  shouldn't `getent group rundeck >/dev/null || groupadd rundeck` be idempotent if the group already exists?

robertdebock:
It won't cause an error, but to add a group twice, increases complexity unnecessarily.

gschueler:

Thank you!

PR #1412, opened 2015-08-19
UI Improvments: Nodes page

gschueler:
## Nodes Page

- [x] global tags list needs to be correct
-  default view in Nodes page
	- [x] select a filter as default
	- [x] select all nodes as default
	- [x] show tag/filter list for browsing by default
	- [x] Tab view:
		* use two tabs, a Summary and a Results tab
		* summary is default unless a filter/tag is selected
	* [x] browser forward/back works
	* [x] filter bar rearranged


gschueler:
builds on #1398

Issue #1391, opened 2015-08-11
Remote URL Option Provider is getting triggered too frequently


rophy:
RunDeck 2.6.0-SNAPSHOT

I have a remote URL option provider looks like this:

http://lsc626:8888/?arg0=${option.VMServer.value}&arg1=${option.VMUsername.value}&arg2=${option.VMPassword.value}

Issue: Remote URL is triggered as soon as I keyed in something in the first option, before all 3 options have a value.  As I continue to type more chars, it seems the remote URL is continuously triggered, throwing multiple errors due to incomplete values.

gschueler:
are the other options marked as required?  in that case, maybe the option URL should not be triggered unless all required options are available.  I think it currently triggers the option URL using blank if an option dependency is not set

rophy:
All 3 options are marked as required. See the screenshot below.
VMServer, VMUsername and VMPassword are required,
ResourcePool, Datastore and Location all depends on VMServer, VMUsername and VMPassword.
Ss soon as I type 'd' in VMServer, the 3 remote options are triggered.

![image](https://cloud.githubusercontent.com/assets/592046/9265416/7adffdb4-4265-11e5-9667-6a290ee12c2c.png)


```
- id: 8aa4e99e-097f-417d-ba52-bbe0edb5bec3
  project: GSSC
  loglevel: INFO
  nodesSelectedByDefault: false
  multipleExecutions: true
  sequence:
    keepgoing: false
    strategy: node-first
    commands:
    - type: gssc.rundeck.STAFCommandNodeStep
      nodeStep: true
      configuration:
        command: powershell -File Z:\vmware\Create_VM.ps1 "${option.VMServer}" "${option.VMUsername}" "${option.VMPassword}" "${node.name}" "${option.ResourcePool}" "${option.Datastore}" "${option.Location}" "${option.NumCpu}" "${option.MemoryMB}" "${option.DiskGB}" "${option.GuestId}" "${option.DiskStorageFormat}"
        executionNodeName: lsc626
  description: ''
  name: Create VM
  uuid: 8aa4e99e-097f-417d-ba52-bbe0edb5bec3
  nodefilters:
    dispatch:
      threadcount: 5
      keepgoing: true
      excludePrecedence: true
      rankOrder: ascending
    filter: .*
  group: common/vmware
  options:
    VMServer:
      sortIndex: 1
      required: true
    VMUsername:
      sortIndex: 2
      required: true
    VMPassword:
      sortIndex: 3
[...]

PR #1415, opened 2015-08-24
Prevents remote option spam in issue 1391 by using onChange event

rophy:
Issue #1391 describes 2 problems about remote URL option:

1. Remote option triggered before all required deps are available
2. Remote option (seems to be) repeatly triggered on every character typed, causing spam to the remote server. Ideally remote option should only be triggered after I switch to another field.

I tried to fix Problem 2 by using Event.observe(id, 'change', handler) instead of Form.Element.Observer(), as Form.Element.Observer() seems to get triggered on every character typed.

I'd like to fix Problem 1 as well, but looking through the js file, I couldn't find a simple way to determine if an option is "required" or not.


rophy:
Implemented Problem 2 as well (by checking visibility of the reqwarning icon)

PR #1458, opened 2015-09-30
Update Active Directory example configuration.

paul-krohn:
Cribbed from http://meinit.nl/connect-rundeck-active-directory


gschueler:
thank you.

Issue #1485, opened 2015-10-16
Scheduler status resets during recycle


ahonor:
We recently discovered an issue when restarting Tomcat which involves the new Enable/Disable Schedule feature in 2.6.0. When restarting Tomcat, any unscheduled jobs are rescheduled. The job editor as well as the job xml state that the scheduler is disabled, but in reality it is enabled and will execute.

gschueler:
closing as dupe of #1475

Issue #1475, opened 2015-10-14
When disabling a crontab style schedule and restarting the rundeck service, the schedule is enabled again after service restart.


oovoo:
Steps to reproduce:
1. create crontab style schedule
2. disable the schedule
3. restart the service
4. schedule is enabled again -> error

snebel29:
Would be useful to know the version you're using

oovoo:
Sorry,
2.6.0-1 with MySQL DB.

PR #1501, opened 2015-10-21
changes implementation of flipScheduleEnabled and flipExecutionEnable…

miguelantonio:
solving issue 1468

### DIFF
--- a/test/api/test-job-flip-scheduleEnabled-scheduler-bug.sh
+++ b/test/api/test-job-flip-scheduleEnabled-scheduler-bug.sh
@@ -1,0 +1,167 @@
+#!/bin/bash
+
+#test /api/jobs/import
+
+DIR=$(cd `dirname $0` && pwd)
+source $DIR/include.sh
+args="echo hello there"
+
+projectName=""
+jobName=""
+jobId=""
+
+generate_projectName_and_jobName(){
+    projectName="project-$(date "+%s")"
+    jobName="job-$(date "+%s")"
+}
+
+create_proj_and_job(){
+    projname=$1
+    jobname=$2
+
+    xmlproj=$($XMLSTARLET esc "$projname")
+    xmljob=$($XMLSTARLET esc "$jobname")
+
+    cat > $DIR/proj_create.post <<END
+<project>
+    <name>$xmlproj</name>
+    <description>description for $xmlproj</description>
+    <config>
+        <property key="test.property" value="test value"/>
+    </config>
+</project>
+END
+
+    runurl="${APIURL}/projects"
+
+    # post (thus creating project)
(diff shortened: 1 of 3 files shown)


gschueler:
@miguelantonio can you add a unit test to verify the fix?

miguelantonio:
@gschueler test added!

miguelantonio:
@gschueler 1ee4bce added fix to issues #1485 and #1475 (restart enabled schedule toggle)

We test this first using the development branch and it reproduced, it was because at startup ir was using the incorrect funciton, now it uses the correct one, we tested with the fix and it doesn't schedule disabled scheduled jobs

test/api/test-job-flip-scheduleEnabled-scheduler-bug.sh
```diff
@@ -0,0 +1,167 @@
...
+    runurl="${APIURL}/jobs/export?project=${projname}&format=xml"
+    params=""
+
+    # get listing
+    docurl ${runurl}?${params} > $DIR/curl.out || fail "failed request: ${runurl}"
+
+    assert 1 $(xmlsel "count(/joblist/job/schedule/month)" $DIR/curl.out) "can't find tag job/schedule/month"
+    assert 1 $(xmlsel "count(/joblist/job/schedule/time)" $DIR/curl.out) "can't find tag job/schedule/time"
+    assert 1 $(xmlsel "count(/joblist/job/schedule/year)" $DIR/curl.out) "can't find tag job/schedule/year"
+    assert 1 $(xmlsel "count(/joblist/job/schedule/weekday)" $DIR/curl.out) "can't find tag job/schedule/weekday"
```
gschueler: need to verify the actual value of `weekday`

gschueler:
if i run the `test/api/test-job-flip-scheduleEnabled-scheduler-bug.sh` test script without applying any of these changes, it still succeeds, so it doesn't seem to verify the fix

miguelantonio:
Greg, we were testing using the API but the bug doesn't happen in that channel, it only does by using the Web, we are right now building the right test and checking days

basically we are now trying to execute a GET simulating the use of the web http://localhost:4440/project/<project>/job/flipScheduleEnabled/<job>?scheduleEnabled=true/false

will commit as soon as it reproduce with the test

Also, do you have any hint on logon as user with curl? (headers?)

gschueler:
@miguelantonio I see. In that case, don't worry about the api-level test. If you want I can just add a unit test to verify the correct service layer method is called for that action

miguelantonio:
oh, thank you, that's great

gschueler:
ok, here's a trial: I committed a broken test.  merging this PR should fix it :)

miguelantonio:
alright :)

PR #1620, opened 2016-01-11
Added CODE renderingType to allow plugins to render textareas with th…

bryonwilliams:
…e code-style editor


ahonor:
Thanks! This is very useful.

PR #1635, opened 2016-01-17
Passive mode - more explicit for already running job

mathieuchateau:
Explicit that already running job are not affected


gschueler:
thanks!

Issue #1597, opened 2015-12-15
Querying 5k nodes leads to very long queries on the webui


kamaradclimber:
When searching for nodes in the web UI, rundeck currently returns data as html.

The data volume is really high, for instance on our 5573 nodes, it represent around 302MB whereas the initial data is only 0.5MB.

It has many downsides, one of them being that rundeck dedicate one core to build this html leading to 100% cpu on that core (effectively limiting the data transfer to 10Mb/s).

Overall, the interface feels very slow when searching nodes.

What could be done to mitigate that issue?

kamaradclimber:
More figures depending on the various projects we have:

Node in the project | Time to fetch data
-------------------------|------------------------
382                      | 20 secs
1668                     | 76 secs
2773                     | 150 secs
5573                     | 4 minutes (timeout)

Those project node sources are completely independant and have no filter applied.

ahonor:
@kamaradclimber what kind of model source is this?

kamaradclimber:
@ahonor URL Source, with cache result set and timeout to 120.

It seems that the fetch of the source done very quickly:
- a tcpdump show the whole source is sent to rundeck server in a few seconds
- the beginning of the data (time to first byte is 1 second) sent to my browser contains the number of nodes which can only be known if the whole source has been parsed by the rundeck server.

ahonor:
@kamaradclimber for comparison sake, can you save the result of a URL query containing the 5k nodes to a file and configure a File Model Source? I am curious to see how the UI responds.

kamaradclimber:
Result is the same: timeout after 4 minutes (exactly) and 299 MB of transferred data.

ahonor:
@kamaradclimber where does the timeout occur? For example, if you go to the Nodes page and list all, it will say load next 20 or list remaining. Does it timeout when you say list remaining?

kamaradclimber:
It appears on the Commands page when setting the `.*` filter.
![2015-12-16-164713_926x155_scrot](https://cloud.githubusercontent.com/assets/503537/11845532/afacf64e-a414-11e5-840f-0023dd1263af.png)

kamaradclimber:
Yes on the node page, it does that "Next 20… Load 5553 remaining…"

ahonor:
@kamaradclimber what is the maxheap size setting (i.e., -Xmx)?

kamaradclimber:
XmX is 16384m

kamaradclimber:
[...]

PR #1651, opened 2016-01-29
Node page UI updated for large node sets

gschueler:
Update Nodes page to better handle large node sets (15K+).  for #1597 

- [x] replace remote html ajax with json data and use knockout for UI
- [x] truncate node list data for too-large sets in embedded nodes view (e.g. commands tab, job editor, job preview)
- [x] add paging controls to node list

todos:
- [x] back/forward when paging through nodes page results doesn't work
- [ ] job run page, "change target nodes" is slow with many nodes


gschueler:
moved final todo to separate issue https://github.com/rundeck/rundeck/issues/1653

PR #1677, opened 2016-02-12
Documentation Proposed Change log4j

ltamaster:
In this case (https://simplifyops.desk.com/agent/case/105) the client ask to specify the way to setting the log4j properties related with the web API.

We propose to change the documentation to specify where that file is located, because the current one doesn’t explain that the file it is only to configure the logging of the CLI tools.


ahonor:
thanks!

Issue #1368, opened 2015-07-31
Log Output gone mad in 2.5.2


richiereynolds:
Hi, Just upgraded to 2.5.2 from 2.4.2 and some of the Log Output now have newlines after virtually every character.
E.g. where I used to have:
			
20:19:29			LOG_TRACE : "Tue Jul 21 20:19:29.521 2015" [App:App_Initialise]
20:19:29			Process id: 39913
etc.

I now have:

15:56:10	LO.LO1.tpdev-perf1l.tpsecm	 1. Starts the server if currently CLOSED.	Starting server
15:56:10			
15:56:10			L
15:56:10			O
15:56:10			G
15:56:10			_
15:56:10			T
15:56:10			RA
etc.

I think it's not the display as the new version displays the old version's log files correctly so it must be how the server is storing the logs. Logs form the rundeck server itself are ok too so I suspect it's something to do with capturing remote output.

richiereynolds:
Yes, it's the log files that now have all the extra lines:
^text/x-rundeck-log-v2.0^
^2015-07-31T14:56:08Z|stepbegin||{node=localhost|step=1|stepctx=1|user=Richard.Reynolds}|^
^2015-07-31T14:56:08Z|nodebegin||{node=LO.LO1.tpdev-perf1l.tpsecm|step=1|stepctx=1|user=tpsecm}|^
^2015-07-31T14:56:10Z|||{node=LO.LO1.tpdev-perf1l.tpsecm|step=1|stepctx=1|user=tpsecm}|Starting server^
^2015-07-31T14:56:10Z|||{node=LO.LO1.tpdev-perf1l.tpsecm|step=1|stepctx=1|user=tpsecm}|^
^2015-07-31T14:56:10Z|||{node=LO.LO1.tpdev-perf1l.tpsecm|step=1|stepctx=1|user=tpsecm}|L^
^2015-07-31T14:56:10Z|||{node=LO.LO1.tpdev-perf1l.tpsecm|step=1|stepctx=1|user=tpsecm}|O^
^2015-07-31T14:56:10Z|||{node=LO.LO1.tpdev-perf1l.tpsecm|step=1|stepctx=1|user=tpsecm}|G^
^2015-07-31T14:56:10Z|||{node=LO.LO1.tpdev-perf1l.tpsecm|step=1|stepctx=1|user=tpsecm}|_^
[...]

PR #1700, opened 2016-02-22
fix issue 1368: log output newlines

gschueler:
* does not flush all output when a flush() is sent
* flushes remaining output when child threads complete


gschueler:
for #1368

PR #1751, opened 2016-03-16
Change init script to use start-stop-daemon and provide proper PID

penekk:
Original init script puts pid of su process rather than java's, which is what you'd rather like to find in a pidfile, especially if its being used elsewhere (like pulling process info for monitoring).

Feel free to correct me since these are my first steps in contributing! ;)


gschueler:

Thank you!

PR #1806, opened 2016-04-14
API Documentation

ltamaster:
It was reported in desk.com (https://simplifyops.desk.com/agent/case/207)
that when it is used some REST client (different than curl) it is necessary to add the following headers, in order to get  a JSON response of the system/info API call:

Content-Type: `application/json`
User-Agent: `curl/7.19.7`

If this headers are not set, the API response will be in a XML format.
So, we propose to add this to the API documentation.


gschueler:
cannot reproduce the error

PR #1819, opened 2016-04-20
Place Jetty work directory outside of the installation tree

sebastianopilla:
Added a new system property "rundeck.server.workDir" to specify a temporary directory for the Jetty web app context which can be configured to be outside of the installation tree.

When unset, a new directory "work" is created under the "rundeck.server.serverDir" directory.
When set, the embedded Jetty uses or creates the directory identified by the "rundeck.server.workDir" directory.


gschueler:
thanks!

isuftin:
Awesome! Thank you.

Issue #1811, opened 2016-04-18
Documentation: add job.executionType to context variables documentation


ctrlrsf:
Would be great if we had a job context variable we could check to see if the job was scheduled or not.

For example, `${job.wasScheduled}` which we could check from a plugin via environment variable `$RD_JOB_WASSCHEDULED` (don't care what variable name actually is)

This would help differentiate between jobs that ran on their set schedule, vs jobs that were ran ad-hoc by a Rundeck user.

If we can determine this some other way, would love to hear how.

If there's no opposition to this new context variable and you have some pointers, I can start working on a pull request.

ahonor:
+1

plambert:
+1

ctrlrsf:
Just created a PR to add a context variable called job.executionType. Depending on how the job was ran, we can expect this to be either "adhoc", "user", or "scheduled".

fuinyroot:
+1

gschueler:
you can use `${job.executionType}`

fuinyroot:
Based on http://rundeck.org/docs/manual/jobs.html#context-variables , there is no context variable named **job.executionType** 

Please re-open this ticket

gschueler:
@fuinyroot ok, it should be documented, sorry.  will retitle the issue

PR #1835, opened 2016-04-29
Add job.executionType context variable

ctrlrsf:
Currently there's no way to differentiate whether a job was ran by a user
and a job ran by the scheduler by looking at the job context variables.

With this change, we'll have a job.executionType context variable that will
indicate what invoked the job.

Possible values of job.executionType:
  adhoc = adhoc command
  user = job initiated by a user
  scheduled = job triggered by the scheduler

This is a first stab at this problem to get the ball rolling.

Please review for correctness and let me know if anything
else is needed.

Addresses issue https://github.com/rundeck/rundeck/issues/1811

Signed-off-by: Rene Fragoso ctrlrsf@gmail.com

### DIFF
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/ScheduledExecutionController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/ScheduledExecutionController.groovy
@@ -2669,6 +2669,7 @@ class ScheduledExecutionController  extends ControllerBase{
             inputOpts.putAll(params.extra.subMap(['nodeIncludeName', 'loglevel',/*'argString',*/ 'optparams', 'option', '_replaceNodeFilters', 'filter']).findAll { it.value })
             inputOpts.putAll(params.extra.findAll{it.key.startsWith('option.')||it.key.startsWith('nodeInclude')|| it.key.startsWith('nodeExclude')}.findAll { it.value })
         }
+        inputOpts['executionType'] = "user"
         def result = executionService.executeJob(scheduledExecution, authContext,session.user, inputOpts)
 
         if (result.error){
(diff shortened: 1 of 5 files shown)


gschueler:
thanks, looks like a good start!  setting the `execution.executionType` in ExecutionJob.groovy may need to be changed, ideally it should be set when the Execution is created, just so there is no need to do another .save() to the db. I think it will be easy, the executions are created by ExecutionJob for scheduled jobs only, for all other cases, the executions are created ahead of time, and so it should be easy to tell what value to use when.

ctrlrsf:
Hi @gschueler. I've updated my PR to set the executionType when Execution is created. Can you confirm it's on the right track? Thanks!

gschueler:
@ctrlrsf looks good so far.  the other place that `createExecution` is called is when an execution is retried here: https://github.com/rundeck/rundeck/blob/development/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy#L1595

however, retried executions are already marked as retried from previous executions, so not sure if executionType is necessary to set

ctrlrsf:
@gschueler thanks for the additional comments. I updated PR to pass through the executionType when a job is retried.

gschueler:
@ctrlrsf great, the only other thing I would request is if you could add some unit tests, but otherwise i think it looks good

ctrlrsf:
@gschueler I've added two tests to confirm executionType is being set properly. 

I also modified the code to be more straight forward. Now executionType is set to "scheduled" by default when an Execution object is created.

When a user runs a job from the web using Run Now button, ScheduledExecutionController.runJob() function will set an executionType input option to "user", which will eventually get passed as the executionType property when Execution object is later created.

Let me know if you have any further comments.

Issue #1781, opened 2016-04-05
Cron format for job and the reference is dead link.


shigemk2:
I'm looking for cron format for rundeck and I found this link: http://rundeck.org/docs/manual/jobs.html#scheduled-jobs

This link says, "The schedule can be defined in a simple graphical chooser or Unix crontab format."

But the reference about cron format is "Quartz Scheduler crontrigger" and this reference is dead link: 
http://www.quartz-scheduler.org/documentation/quartz-1.x/tutorials/crontrigger

Is the rundeck cron format not Unix but Quartz?

ahonor:
They appear to be reorganizing the site. Here is a java doc page I found http://www.quartz-scheduler.org/api/2.2.1/org/quartz/CronTrigger.html

shigemk2:
Thank you for your replying. 
So, the dead link will be http://www.quartz-scheduler.org/api/2.2.1/org/quartz/CronTrigger.html and crontab format is Quartz.
Can I send a pull-request for this? Or should I wait for their reorganizing the site?

kapil514:
It's again dead, please use this tool for more info: http://www.cronmaker.com/
Reference of the tool is from this comment: 
https://github.com/rundeck/rundeck/issues/2211#issuecomment-264697970

PR #1836, opened 2016-05-01
Fix deadlink quartz

shigemk2:
Refer #1781

There is a dead link on http://rundeck.org/docs/manual/jobs.html#scheduled-jobs and I fixed the dead links(both en document and jp document).


shigemk2:
Thank you!

PR #1866, opened 2016-05-26
Make moving/renaming of files more robust

### LINKED ISSUE #1702: Log storage fails to retrieve remote file if unable to rename temp file
ref https://github.com/rundeck-plugins/rundeck-s3-log-plugin/issues/6

should not use `File.renameTo`

jamieps:
- Fix #1702 by replacing `File.renameTo` with `Files.move`, which handles moving between different filesystems; use the JDK 7 `Files.copy` for copy operations.
- When creating temporary files, check the parent directory structure exists and create it if needed.
- Tidy up temporary files on exit to reduce clutter in `/tmp` after running tests.


gschueler:
awesome, thanks!

PR #1884, opened 2016-06-08
Adds global parameters

miguelantonio:
Adds global parameters using framework.properties defined parameters, in that file as framework.globals.VAR, in rundeck as ${globals.VAR}

Issue #1629, opened 2016-01-14
Environment variables feature request


Mapel88:
Hello dear devs!
I hope I'm posting this in the right place, please let me know if not :)

Is it possible to add environment variables that can be accessed from all job steps and nodes?
For example, I want to use rundeck ip while running scripts on remote nodes, but don't want to add another tag for it in resource.xml per node..

I thought of two options:
1. Add more variables like RD_JOB_SERVERURL (or ${job.serverUrl}) that will contain data of the rundeck server (OS, version, ip, hostname etc)
2. Load the data from framwork.properties (or any other configuration file) so it can be used as variables in scripts (this way users can add whatever environment variables they wish to use in the future)

Thanks!

cyrilthank:
Add an environment variable which can be used as placeholder to use/pass common values to all jobs

ahonor:
@Mapel88 Use a global configuration setting. http://rundeck.org/docs/administration/configuration-file-reference.html#global-execution-variables

Issue #1270, opened 2015-05-29
Make Project specific context variables like project.dir available to executions


andham:
I think it would be great if custom project wide context variables could be configured. It could for example be a common file path that used in several jobs.

Also, I think that it would be good if the project-specific stuff like project.dir also were available as context variables.

gschueler:
see #806 

also, people have requested job-specific variables that can be set

mathieuchateau:
+1

andham:
@gschueler Thanks for the pointer! So this ticket would then be more about making project-specific things like project.dir available as context variables.

ntikare:
+1

gschueler:
Project-global context variables were added in 2.6.8.

I will rename this issue to reflect desire for `project.dir`

PR #1890, opened 2016-06-10
Add framework and project global variables

gschueler:
some cleanup and additional tests from https://github.com/rundeck/rundeck/pull/1884

ref #1629 #1270


richiereynolds:
This is very useful.
Is there any way to use context variables for script paths? They don't seem to work there e.g.
${globals.my_workflow_scripts_dir}/someScript.sh
${globals.my_node_scripts_dir}/someScript.sh
etc.

puremourning:
There's a hack. You have to use a URL like file://. I'll IM you :)

PR #1900, opened 2016-06-16
Fix cascading options problems

### LINKED ISSUE #1895: GUI option dynamic refresh broken in 2.6.6 and 2.6.8
We have just upgraded from 2.6.4 and found that dynamic population of options has stopped working properly.

Our URL is for populating a "Nodes" multi-valued option is:
http://localhost:1080/tprundeck/getResourceNames.jsp?project=${job.project}&type=node&regionname=${option.Region.value}&tpname=${option.TP.value}&componentname=${option.Component.value}

We find that the Nodes option updates correctly when we select a "Component" but is not updated when we select a "TP" or a "Region".
This makes a previously working GUI unusable.

Note that TP and Region are optional values, Component is required and has "enforce from list" set.
We checked and found that even having "required" was not enough to trigger the cascade, it needs both required and enforce from list.

#1832 appears to suggest this might be fixed in 2.6.8 but we've just tried that version and it's still not working.

gschueler:
replace the mishmash of javascript and dhtml/ajax with knockout

fixes #1895


rundeckapp/grails-app/assets/javascripts/menu/job-remote-options.js
```diff
@@ -0,0 +1,271 @@
...
+            method: 'GET',
+            type: 'json',
+            url: _genUrl(self.url, params),
+            success: function (data, status, jqxhr) {
+                // self.addReloadRemoteOptionValues(opt,data);
+                opt.loading(false);
+            },
+            failure: function (data, status, jqxhr) {
+                opt.loading(false);
+                alert("error: " + status);
```
gschueler: oops

Issue #1360, opened 2015-07-24
Improve workflow step up/down graphics for usability


kgeis:
When I click on the up/down arrows to rearrange steps in a workflow, nothing happens.

gschueler:
those are not buttons, but a drag handle. you can drag the steps

kgeis:
I see the hover text now, but some other affordance would improve usability. Maybe an icon that looks like a single button, or on clicking, surrounding the row with a dotted blue border.

mathieuchateau:
Hello,

I have this issue when connecting from iPad, there some things that doesn't
work (at least I can't get them to), this one included.


Cordialement,
Mathieu CHATEAU
http://www.lotp.fr

2015-07-24 23:42 GMT+02:00 Ken Geis <notifications@github.com>:

gschueler:
:+1: on using glyphicons, the workflow editor GUI code is fairly old and could do with a few touch ups.

And for mobile UI buttons would probably be a better way to do it

gschueler:
fix in #1907

Issue #363, opened 2013-04-22
NTH: improve job option ordering to allow GUI drag and drop


gschueler:
supplement to issue #361

danifr:
+1

fpires82:
+1 to this request!

hmalinov:
+1

dfsis:
+1

gschueler:
fix in #1907

PR #1907, opened 2016-06-23
GUI support for reordering options

### LINKED ISSUE #1002: As a rundeck user, i'd like to have the possibility to sort rundeck job executions option on user interface.
In our rundeck jobs, we have some options that depend on over options. (options  which are generated with an option provider which take as url parameter a previous option)
It would be very useful to sort our job execution options on rundeck GUI, in order to choose option in an intuitive order for the job's user.   
job_option_order

As it's an enhancement, I've also post a request : https://groups.google.com/forum/#!topic/rundeck-discuss/G1_Fr_UKBKU

Thx

gschueler:
fix #1002 #363


gschueler:
also fix #1360

Issue #1699, opened 2016-02-22
Rundeck i18n - ability to select language in GUI


miguelantonio:
The spanish version merged works as tested with http://localhost:4440/menu/home?lang=es_419, but is there any way to select if for users? whats the correct way to do it?

if there is none, do you mind if we just pull request a simple button to toggle between languages?

gschueler:
what i understand is that Grails *should* set the locale based on the client (i.e. web browser). in the `Accept-Language` header in the web request (see [grails docs](https://grails.github.io/grails-doc/2.4.4/guide/i18n.html)).

It could be useful to have a way to switch to another language, tho.  Do you mean a button that just addes the `lang=` parameter? I think it would make sense to have a way to set the preferred language for a user that is persisted perhaps (either in the http session, or as a user preference stored in the DB).

gschueler:
oh according to grails docs, adding `lang=` will store it in a cookie, so maybe that's the easiest way

miguelantonio:
oh! awesome, yeah we'll do it that way thanks!

variacode95:
Hi @gschueler so returning to this matter as we are interested in getting rundeck in spanish we'll just create a combobox on /user/profile that has two languages (for now) English and Español, do you agree or do you think it's better in some other way?

Miguel

gschueler:
@miguelantonio i think that would be fine

PR #1919, opened 2016-06-28
Added drop down to switch language for application

variacode95:
1699: Added drop down that enables users to switch rundeck language from their profile page.


variacode95:
@gschueler @miguelantonio Guys, I messed up with the previous pull request for merge commits, so replaced the origin with upstream master and closed the old pull request. 

Created this new pull request, that only contains a single commit for all the changes for https://github.com/rundeck/rundeck/issues/1699.

gschueler:
👍

PR #1634, opened 2016-01-15
This adds dattepicker functionality.

bajacondor:
*A new job option type that is a date and uses the datepicker to select it.
*Datepicker widgets to filter the activity reports

PR #1987, opened 2016-08-05
Support for ad hoc scheduling of jobs

### LINKED ISSUE #150: Add AT time scheduling of jobs.
* Issue Type: Request/Enhancement
* Rundeck version: 1.4.3

---

Issue:

Would like the ability to schedule a job to run once at a specific point in time and then never again.
Thanks!

---

[original LH ticket](http://rundeck.lighthouseapp.com/projects/59277/tickets/520)

Original Creator: Michael Gibson

jamieps:
Fixes #150. 

I've used the bootstrap-datetimepicker for this, as it blended in well as Rundeck already uses Bootstrap. Since then I've noticed #1634 which uses the jQuery date picker. Do you have a preference on which one is used?

Some screenshots:

run_job_later_btn
schedule_popover
schedule_date
schedule_time

Scheduled jobs in the activity view:

activity_view

If the job supports multiple executions, then it's possible to schedule the same job multiple times. Otherwise only one execution (deferred or run immediately) is allowed.

 - Update Moment.js to 2.14.1, the later version is required for bootstrap-datetimepicker.
 - Added support for specifying the time to execute to the `/api/18/job/[ID]/run` endpoint. This must be a valid ISO 8601 date and time.

Some limitations...

 - Permissions are only checked when the job is scheduled, not when it is executed.
 - If Rundeck is restarted, all executions which were scheduled ad hoc will be killed and must be re-scheduled manually. This is the same behaviour as for executions which are run immediately.

We don't have access to a working cluster environment right now, though I am working towards getting one setup. Do you think takeover should affect jobs which are scheduled ad hoc?


gschueler:
This looks great, thank you!

> If Rundeck is restarted, all executions which were scheduled ad hoc will be killed and must be re-scheduled manually. This is the same behaviour as for executions which are run immediately.

I suppose if the scheduled start time is still in the future, it could be re-scheduled? The big issue with that is secure option input values, which are only passed into the scheduler context and are not stored.  Perhaps if the job has no secure input options it could be rescheduled, otherwise killed.  For now killing is fine.

> We don't have access to a working cluster environment right now, though I am working towards getting one setup. Do you think takeover should affect jobs which are scheduled ad hoc?

Good question. Right now if you takeover a job's scheduling on another cluster node, the original node will still execute the schedule, but as soon as it is triggered it will see schedule owner has changed and unschedule it locally and then not execute.  Ideally "runAt/ad-hoc" job executions would behave the same way.  To implement that the "takeover schedule" action will also have to look for the one-off scheduled executions to add to quartz. Alternatively, they could be ignored and be left to the original schedule owner. Thoughts @ahonor ?

ahonor:
Killing is fine but we should log which jobs it occurs for. It's confusing when a user expects a job to run and then gets no log of it when it doesn't execute.

As for takeover, I prefer it to act like other scheduled jobs for now.

jamieps:
@gschueler @ahonor 

> Killing is fine but we should log which jobs it occurs for. It's confusing when a user expects a job to run and then gets no log of it when it doesn't execute.

I've implemented rescheduling of ad hoc scheduled executions unless the job has secure input options; if a job does has secure input options, or it's not possible to reschedule the execution, an error is logged for each ad hoc scheduled execution and the execution is terminated.

> As for takeover, I prefer it to act like other scheduled jobs for now.

Unscheduling and takeover now also affects ad hoc scheduled executions.

I've had to migrate some unit tests to integration tests since the claim executions functions now use [criteria](http://docs.grails.org/latest/guide/single.html#criteria) and I'm not aware of a way of unit testing these.

I started with using GORM named queries but then I've since read that the Grails developers have 'deprecated' this in favour of where queries instead. Do you have any preference?

I'll merge in the changes from master and then update the PR shortly.

gschueler:
@jamieps awesome work, thanks!

jamieps:
@gschueler I've rebased against master and updated the PR

willemdh:
Is it possible to schedule a job like this with the api?

jamieps:
I noticed an issue with killing off executions which can't be rescheduled. 

I was relying on `cleanupRunningJobs()` being run at start up, however I've noticed that sometimes it completes before anything has been (re)scheduled, and in this case the executions are not killed properly. 

Now `cleanupRunningJobs()` gets called explicitly to make sure this always happens.

PR #2014, opened 2016-08-17
Schedule owner plus heartbeat

jtobard:
Code for cluster mode, enable a heartbeat every 5 minutes in a QuartzJob.
This heartbeat first insert or update a message with his `serverNodeUUID` and the datetime of the pulse, this way when the admin create or update a job can see the list of active nodes to select one and assign it to the job, instead of the actual behavior (the actual node is assigned by default).
By default the actual node came pre selected, emulating the actual behavior.
Also if the job is edited by API call, and the `serverNodeUUID` is omitted, the actual node is selected by default.

When the job is assigned to another node, is marked with `!scheduleOwnerClaimed` ans a message is sent to the message table, this way in the next heartbeat, every node check if are any node assigned to him and wit `!scheduleOwnerClaimed`, it changes the value in the job and schedule it as usual.

Also added  `nodeSummary.reload();` in **rundeckapp/grails-app/views/scheduledExecution/_edit.gsp**. This fix a problem in the edit Job view, no saved node filters are displayed.


gschueler:
question: once a job is switched to a new owner, should the Messaging object be deleted?

jtobard:
I don't deleted thinking in some kind of log history, but maybe its better to get the table clean

gschueler:
well, one problem will be the same Messaging result will be returned the next time it is queried, right?

jtobard:
No, because the job now got scheduleOwnerClaimed changed to true, so no action is taken

gschueler:
but if it gets transferred again to another node, now 2 nodes will see the messages and try to take over.

gschueler:
a unit test for that would be: transfer to node A, verify it was taken over, transfer to node B, verify node A did not take over again

jtobard:
You are right, I will add that test

PR #2025, opened 2016-08-24
Feature/api job info

gschueler:
Add a new endpoint `/api/18/job/[ID]/info`, returns metadata about the job, including `project` and other info already returned in job list and scheduler list info.

Fix issue: no way to know a Job's project based on ID


gschueler:
- [ ] add job average duration, description

PR #2056, opened 2016-09-07
Feature/node filter enhancements

jtobard:
Added ability to search Activity by nodes on which the task were executed or by the applied node filter.
newfield
Example of searching by nodename:
searchbynode

Example of searching by activities executed against a filter:
searchbyfilter

Also added the "editable filter" property on jobs, if set to true, the user can modify the filter on the execution page, applied to Run job Now and Run Job Later buttons.
On the create job screen:
createjob

Usage in execution:
execoverridenodefilter


gschueler:
rebased and merged

PR #2068, opened 2016-09-12
Escape HTML characters in ExecutionController.renderOutput

jrunningen:
renderOutput was not sanitizing task logs correctly for an HTML view. If
an attacker controlled a node, they could send Javascript code in the
response, which would be executed in the user's browser upon viewing the
page.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,27 @@
+# Bug report
+
+(Note: Please fill in the blank `___` about your rundeck installation below. You can remove this line.)
+
+**My Rundeck detail**
+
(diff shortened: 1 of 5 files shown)


gschueler:
Thank you: If the log message has no ansi escapes it is not being encoded properly.

If there are Ansi escapes, the encoder does escape chars `<` and `>` from the output, using the HTMLElement codec, but that should be replaced with `encodeAsHTML`. 

note the "renderOutput" action is used when the "html" or "text" view link is followed from the execution page.

gschueler:
correction: just the "html" view link.  the "text" link outputs `text/plain` content

gschueler:
👍

PR #2073, opened 2016-09-14
Issue with disable schedule job on cluster

jtobard:
Fix a problem with cluster mode, when you disable schedule a job on a different cluster node, the job still runs as scheduled on the owner node. Now when the job its executed, we detect if was disabled and prevent the execution and unschedule it.


rundeckapp/grails-app/jobs/rundeck/quartzjobs/ExecutionJob.groovy
```diff
@@ -281,6 +281,13 @@ class ExecutionJob implements InterruptableJob {
                     }
                     context.getScheduler().deleteJob(context.jobDetail.key)
                     return initMap
+                }else{
+                    //verify run on this node but scheduled disabled
+                    if(!initMap.scheduledExecution.scheduleEnabled){
```
gschueler: perhaps it should test `scheduledExecution.shouldScheduleExecution()` this will also fix it if the schedule is removed or the execution is disabled

gschueler:
fyi; the travis build failure is something i'm trying to fix, i will restart it it should work.  or you can try rebasing on the master branch, i tried to add a fix to the docker tests for that travis issue

rundeckapp/test/unit/rundeck/quartzjobs/ExecutionJobSpec.groovy
```diff
@@ -169,4 +169,62 @@ class ExecutionJobSpec extends Specification {
...
+                description: 'a job',
+                argString: '-a b -c d',
+                serverNodeUUID: serverUUID,
+                workflow: new Workflow(
+                        keepgoing: true,
+                        commands: [new CommandExec(
+                                [adhocRemoteString: 'test buddy', argString: '-delay 12 -monkey cheese -particle']
+                        )]
+                ),
+                scheduled: false,
```
gschueler: You can parameterize the `scheduled`, `executionEnabled` and `scheduleEnabled` values to test all of them being set to false

rundeckapp/test/unit/rundeck/quartzjobs/ExecutionJobSpec.groovy
```diff
@@ -169,4 +169,62 @@ class ExecutionJobSpec extends Specification {
...
+
+            getScheduler() >> quartzScheduler
+        }
+
+        when:
+        job.execute(context)
+
+        then:
+        1 * quartzScheduler.deleteJob(ajobKey)
+
```
gschueler: Here you could use:

~~~ {.groovy}
where:
isscheduled | isexecenabled | isscheduleenabled
false | true | true
true | false | true
true | true | false
~~~

rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
```diff
@@ -1518,6 +1518,7 @@ class ScheduledExecutionService implements ApplicationContextAware, Initializing
                 return [success: false, scheduledExecution: scheduledExecution,
                         errorCode: 'api.error.job.toggleSchedule.notScheduled' ]
             }
+            scheduledExecution.serverNodeUUID = frameworkService.isClusterModeEnabled()?frameworkService.serverUUID:null
```
gschueler: this change should also be tested if we expect the "update execution flags" to change the schedule ownership.  

e.g. If it was owned by another node, enabling the schedule would now reschedule the job on the current node, correct? I think that is the correct behavior, but it should be tested.

philippevidal80:
Have found the same behaviour on our 2.6.9-1 Rundeck Cluster with 2 nodes through the pop-up shortcut edition of a job.

The box "disable execution" and "disable schedule" are checked both on the 2 nodes but job continue to be executed at shceduled time on owned node.

Moreover, the indication of the schedule and node in charge of the job stay on the jobs list at the place of "Never" on the not owned node.

PR #2104, opened 2016-10-06
Add support for job reference node intersect

### LINKED ISSUE #2074: Run on failed nodes doesn't work on more complex workflow
# Bug report

**My Rundeck detail**
- Rundeck version: 2.6.9-1
- install type: war
- OS Name/version: CentOs 7.2

**Expected Behavior**

Let's have a job called: master-job with these specifics:
- node filter that matches 100 nodes. 
- and jobref steps (setup as workflow) that contains node filters: job_a (60 nodes); job_b (20 nodes); job_c (20 nodes)

When I run master-job, I expect that job_a will be executed on nodes that are intersection of these two sets (master-job nodes and job_a nodes). 

This is a big pain point when you have more complex workflow then a few simple commands, but composed from jobrefs, which works with actually large set of nodes but in separated way (like it's described above). Once you run such workflow against ~500 nodes, and 20 of them fails (10 in one job, 5 in another, ... ).. then feature Run failed node could save a lot of time (instead of investigating what jobs that are referenced in jobref to restart with what nodes).

**Actual Behavior**

No matter on how many jobs the master-job failed, the Run on failed nodes will run whole procedure.

**How to reproduce Behavior**
Described above in expected behaviour.

pgressa:
This resolves #2074 

Allows to trigger referenced job by nodes that
are intersection of upstream node filter and
referenced job defined or overridden node
filter.

Node intersection can be applied on node
filter defined in referenced job or when
referenced job node filter is overriden in
upstream job.

screen shot 2016-10-07 at 11 19 17

Motivation for this feature is ability of creation
of complex workflows composed from jobs that
contains many job-refs with differend node filter.
In such cases (hundreds of nodes and 10 job refs)
when something goes wrong on few nodes you want to
restart whole procedure by selecting 10 failed nodes
what is without this feature not feasible.


pgressa:
Hi @gschueler, are you gonna have a time to look at this please?

gschueler:
@pgressa i will review soon hopefully

pgressa:
@gschueler you think this will be part of 2.7.0 as it's still not reviewed yet.

PR #2180, opened 2016-11-16
Feature/jetty upgrade

gschueler:


gschueler:
Notes about upgrading:

* The Default JAAS login module class name changed to `org.eclipse.jetty.jaas.spi.PropertyFileLoginModule`

PR #2196, opened 2016-11-21
fix #2185 use double-@ to escape @ sign

### LINKED ISSUE #2185: 2.6.10: Inline script token expansion breaking perl
# Bug report

via https://github.com/rundeck/rundeck/issues/2021#issuecomment-261463049

~~~{.perl}
sub perlFunction {
    ...
    return \@files;
}
~~~

the script would now be generated without the backslash:

~~~{.perl}
sub perlFunction {
    ...
    return @files;
}
~~~

gschueler:


gschueler:
this should be changed; don't use a separate escape character: require `@@` to produce a `@` when not used for token expansion.

PR #2209, opened 2016-11-29
allow `?opt.name=value&opt.name2=value2` url params to set option values

gschueler:


jimr6007:
Does this still work in version 2.10?

derekslenk:
> 
> 
> Does this still work in version 2.10?

We are using it with 2.10.7-1 and it works great

PR #2214, opened 2016-12-01
Cleanup/remove node filter code

gschueler:
remove old unused node filter domain fields and code


gschueler:
will need to revist

PR #2236, opened 2016-12-12
Prevent csrf attacks

jtobard:
By default, preserve the actual state without CSRF filter, to activate the administrator need to add a property on _rundeck-config.properties_:
"rundeck.security.csrf"

It accepts the values: 
**_POST_**: recommended, only filter the POST methods, verifies Refer and url.
**_*_**: applies the filter to any type of request, including POST and GET, except the login screen wich need to be accesible without refer or with a redirect from another url (case of localhost resolved to hostname)
_**NONE**_: don't applies any filter.
Any other string or the absence of the property, resolves as NONE.

i.e.:
rundeck.security.csrf=POST

### DIFF
--- a/rundeckapp/grails-app/conf/RefererFilters.groovy
+++ b/rundeckapp/grails-app/conf/RefererFilters.groovy
@@ -2,7 +2,7 @@ class RefererFilters {
     def grailsApplication
 
     def filters = {
-        checkReferer(controller: '*', action: '*') {
+        checkReferer(controller: 'user', action: 'login',invert: true) {
             before = {
                 def csrf = grailsApplication.config.rundeck.security.csrf
                 if(csrf && csrf != 'NONE'){
@@ -16,12 +16,8 @@ class RefererFilters {
                     }else if(csrf == '*'){
                         def validRefererPrefix = "^${grailsApplication.config.grails.serverURL}".replace("http", "https?")
                         def referer = request.getHeader('Referer')
-                        if(request.method.toUpperCase() == 'GET' && (referer == null || referer.endsWith("user/login"))) {
-                            //excluding the user/login screen
-                            return true
-                        }else{
-                            return referer && referer =~ validRefererPrefix
-                        }
+                        return referer && referer =~ validRefererPrefix
+
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


rundeckapp/grails-app/conf/RefererFilters.groovy
```diff
@@ -0,0 +1,31 @@
...
+                        if (request.method.toUpperCase() == "POST") {
+                            // referer must match serverURL, optionally https
+                            def validRefererPrefix = "^${grailsApplication.config.grails.serverURL}".replace("http", "https?")
+                            def referer = request.getHeader('Referer')
+                            return referer && referer =~ validRefererPrefix
+                        }
+                    }else if(csrf == '*'){
+                        def validRefererPrefix = "^${grailsApplication.config.grails.serverURL}".replace("http", "https?")
+                        def referer = request.getHeader('Referer')
+                        if(request.method.toUpperCase() == 'GET' && (referer == null || referer.endsWith("user/login"))) {
```
gschueler: instead of checking referer.endsWith, should check the `controllerName` and `actionName` variables.  Or, even better, change line 5 to exclude that pattern

gschueler:
@jtobard if the csrf is set to `POST` then all API post requests are blocked.  if it is set to `*` then all API requests are blocked, plus you can't paste a rundeck URL and visit a certain page

gschueler:
note: POST requests from GUI forms require a synchronizer token (nonce).  Can check referrer header for these requests

gschueler:
behavior should be:
* [ ] require correct `Referer` header for POST requests for web app, excluding API requests.

gschueler:
rundeck-config.properties settings:

~~~{.properties}
#Set HTTP Method to filter based on Referer header.  Can be POST, or "*" for all methods. Default:
# NONE (disabled)
rundeck.security.csrf.referer.filterMethod=NONE|POST|*

# Allow /api/* requests without requireing matching Referer header. Default: true.
rundeck.security.csrf.referer.allowApi=true|false

# If server URL is HTTPS, Require referer header to be from HTTPS version of server URL, if false allow HTTP as
# well. Default: true.
rundeck.security.csrf.referer.requireHttps=true|false

~~~

PR #2244, opened 2016-12-20
$rundeckd is never set in init file

PascalBourdier:
use $prog and not $rundeckd var


burdandrei:
:+1:

PascalBourdier:
a line missing in profile file, no error in init script !

Issue #2256, opened 2017-01-02
'profile' missing RUNDECK_WITH_SSL env var to un-comment


bhamail:
# Bug report

**My Rundeck detail**

* Rundeck version: 2.7.1-1
* install type: deb (rpm,deb,war,launcher)
* OS Name/version: raspian (debian jessie)

**Expected Behavior**

'profile' mentions:

    # SSL Configuration - Uncomment the following to enable.  Check SSL.properties for details.

but no setting or example `RUNDECK_WITH_SSL` environment variable exists.

**Actual Behavior**

I added the line below to trigger SSL operation:

    RUNDECK_WITH_SSL=true

This worked, so I created the following patch that adds this line in a 'commented-out' form:

https://github.com/bhamail/rundeck/pull/1

or

https://github.com/rundeck/rundeck/pull/2257

**How to reproduce Behavior**

Without RUNDECK_WITH_SSL defined, rundeck only works on non-secured http.

# Enhancement request

(*For enhancements*: Please search the existing Issues and look at the [Trello board](https://trello.com/b/sn3g9nOr/rundeck-development) for your idea before posting.)

...

oriolsantos:
ok, but when you use these option (RUNDECK_WITH_SSL=true) the value of RDECK_JVM is overwrite  
 for : RDECK_JVM="$RDECK_JVM -Drundeck.ssl.config=$RDECK_SERVER_CONFIG/ssl/ssl.properties -Dserver.https.port=${RDECK_HTTPS_PORT} ${RDECK_SSL_OPTS}" ... and than don't use the variable
RDECK_JVM_SETTINGS

sjrd218:
Documentation for this setting is here: https://docs.rundeck.com/docs/administration/security/configuring-ssl.html

PR #2257, opened 2017-01-02
add RUNDECK_WITH_SSL env var to uncomment

bhamail:
Maybe I missed where RUNDECK_WITH_SSL is set elsewhere, but it appears the profile is missing an example line to set the RUNDECK_WITH_SSL environment variable in order to trigger SSL flags.

related issue report: #2256


gschueler:
The `rundeckd` init script now sources `/etc/sysconfig/rundeckd` (redhat) or `/etc/default/rundeckd` (ubuntu) before sourcing `/etc/rundeck/profile`.  the `/etc/{sysconfig,default}/rundeckd` is where customization of init properties should be defined, such as `RUNDECK_WITH_SSL`

dmitry-tiger:
But why there is no information about it in documentation  or stub file /etc/sysconfig/rundeckd in rpm with commented options?
http://rundeck.org/docs/administration/configuring-ssl.html seems outdated

Issue #2228, opened 2016-12-06
Value too long for column "SUCCEEDED_NODE_LIST VARCHAR(255)


lancepowell:
# Bug report

Upgraded to 2.7.1 from 2.6.11, jobs not completing

**My Rundeck detail**

* Rundeck version: 2.7.1
* install type: rpm (rpm,deb,war,launcher)
* OS Name/version: CentOS 7

**Expected Behavior**

Jobs finishing successfully. They do complete all the steps but rundeck stays in a running job state

**Actual Behavior**

Job stays running and the log shows:

Value too long for column "SUCCEEDED_NODE_LIST VARCHAR(255) ... All node names ... (14155)

rundeck-config.properties:

dataSource.url = jdbc:h2:file:/var/lib/rundeck/data/rundeckdb;MVCC=true
quartz.props.threadPool.threadCount = 30

lancepowell:
I also see it do this for FILTER_APPLIED VARCHAR(255) too.

jjethwa:
I'm also having the same issue.  However, my succeeded_node_list column is set to longtext.  I'm using mysql  Ver 14.14 Distrib 5.5.53, for debian-linux-gnu (x86_64) using readline 6.3

```
org.springframework.dao.DataIntegrityViolationException: Hibernate operation: could not execute statement; SQL [n/a]; Data truncation: Data too long for column 'succeeded_node_list' at row 1; nested exception is com.mysql.jdbc.MysqlDataTruncation: Data truncation: Data too long for column 'succeeded_node_list' at row 1
        at rundeck.services.ReportService.reportExecutionResult(ReportService.groovy:65)
        at rundeck.services.ExecutionService.logExecution(ExecutionService.groovy:843)
        at rundeck.services.ExecutionService$_saveExecutionState_closure61.doCall(ExecutionService.groovy:2227)
        at rundeck.services.ExecutionService.saveExecutionState(ExecutionService.groovy:2148)
        at rundeck.quartzjobs.ExecutionJob$_saveState_closure4.doCall(ExecutionJob.groovy:509)
        at rundeck.quartzjobs.ExecutionJob.withRetry(ExecutionJob.groovy:447)
        at rundeck.quartzjobs.ExecutionJob.saveState(ExecutionJob.groovy:519)
        at rundeck.quartzjobs.ExecutionJob.execute_internal(ExecutionJob.groovy:162)
[...]

PR #2260, opened 2017-01-04
Automatic database migration

jamieps:
This PR uses the grails database migration plugin (https://grails.org/plugin/database-migration) to provide automatic database migrations, which are not handled by Hibernate's automatic update mechanism (for example see #2228).

Using the standard Hibernate migration (`dbCreate = update`) will not switch types (e.g. from `varchar` to `text`) or change the size of an existing field (e.g. `varchar(100)` to `varchar(255)`).

Unfortunately, liquibase generates migrations which are specific to the database they are run against, so if run against H2, it will generate text fields as `varchar(MAX_INT)`. To work around this, I've borrowed from http://roosbertl.blogspot.co.uk/2013/01/grails-database-migration-plugin-db.html to modify the groovy on the fly, depending on which database is in use. I've gone through all of the type differences between MySQL, postgres and H2 and tried to use the best option for each (the MySQL/postgres types are based on https://dev.mysql.com/doc/workbench/en/wb-migration-database-postgresql-typemapping.html).

This has been tested on H2, with limited testing on postgres (I tested the migration of `varchar` to `text`). I've also tested the migrations in our MySQL cluster environment and they were successful.

When comparing our database schema 'vs' the groovy DSL I found quite a few changes were missing, including some indexes which hadn't been added, this database was created initially on 2.6.x.

Creating a new migration _should_ be as simple as:
```
grails dbm-update
grails dbm-gorm-diff add_something.groovy --add
```

Currently new migrations assume that they haven't been run and promptly fail when run on an existing database, so I have added in prerequisites manually for all of them so that if they have already been run they continue without failing.

I understand this is quite a big procedural change, so I understand if you don't want to merge it yet, however I thought I would propose it as a starting point.

### DIFF
--- a/packaging/debroot/etc/rundeck/rundeck-config.properties
+++ b/packaging/debroot/etc/rundeck/rundeck-config.properties
@@ -6,5 +6,5 @@ rdeck.base=/var/lib/rundeck
 rss.enabled=false
 # change hostname here
 grails.serverURL=http://localhost:4440
-dataSource.dbCreate = update
+dataSource.dbCreate = none
 dataSource.url = jdbc:h2:file:/var/lib/rundeck/data/rundeckdb;MVCC=true
--- a/packaging/root/etc/rundeck/rundeck-config.properties
+++ b/packaging/root/etc/rundeck/rundeck-config.properties
@@ -6,5 +6,5 @@ rdeck.base=/var/lib/rundeck
 rss.enabled=false
 # change hostname here
 grails.serverURL=http://localhost:4440
-dataSource.dbCreate = update
+dataSource.dbCreate = none
 dataSource.url = jdbc:h2:file:/var/lib/rundeck/data/rundeckdb;MVCC=true
--- a/rundeck-launcher/launcher/src/main/resources/templates/config/rundeck-config.properties.template
+++ b/rundeck-launcher/launcher/src/main/resources/templates/config/rundeck-config.properties.template
@@ -5,7 +5,7 @@ rdeck.base=${rdeck.base}
 #rss.enabled if set to true enables RSS feeds that are public (non-authenticated)
 rss.enabled=false
 grails.serverURL=http://${server.hostname}:${server.http.port}${server.web.context}
-dataSource.dbCreate = update
+dataSource.dbCreate = none
 dataSource.url = jdbc:h2:file:${server.datastore.path};MVCC=true
 
 # Pre Auth mode settings
--- a/rundeckapp/authfilter/grails-app/conf/DataSource.groovy
+++ b/rundeckapp/authfilter/grails-app/conf/DataSource.groovy
@@ -29,19 +29,19 @@ hibernate {
 environments {
     development {
         dataSource {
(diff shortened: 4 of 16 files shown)


gschueler:
thanks for doing this work, it will be very useful! I've wanted to figure out how to use the db-migration plugin for some time, the stumbling block had been exactly how to handle multiple db types. I dislike that we have to do that, but the templating mechanism you put in place seems like a workable solution.

To incorporate this into our release process, it sounds like the `grails dbm-update ; grails dbm-gorm-diff version-x.y.z.groovy --add` will have to be executed prior to releasing a new version.

I would like to add support of MSSQL too, perhaps we can add that in a second pass.

👍 awesome

jamieps:
I added the equivalent types for SQL Server but I don't have an environment to test it with, so I can't say whether it will work or not.

gschueler:
very odd ci failure

jamieps:
I realised that the Liquibase dependency is only required on Grails >= 3 (as Spring Boot depends on a different version), so I've removed it, which solves the problem. :smiley:

gschueler:
@jamieps ok, great

gschueler:
Finally trying this out, I've been able to verify most of it works, e.g. installing on top of 2.6.9 and migrating some data types in Mysql, with just the changelog you provided.  @jtobard created a number of docker-based "upgrade" tests which I will expand on to test the app upgrade process for a number of different database types.

Right now I'm trying to work through the procedure we will need to make changes to the app for new releases. Basically: adding or removing domain classes or fields, changing field types, renaming, etc. and then generating and incorporating the dbm changelogs for those changes.

I did these steps:

1. Add a new domain class field (type text)
2. revert to the baseline code, remove the h2 dev db
3. using dev db configuration, run `dbm-update` (and `dbm-db-doc` to look at the output)
4. checkout the modified code
5. run `dbm-gorm-diff` 

A few things odd about the diff, maybe this is specific to h2:

1. It seems like datatype used for h2 "text" fields is "clob" in the changelog.groovy/baseline.groovy, but the gorm diff wants to use "longvarchar", so it ends up generating a lot of `modifyDataType` entries to convert back to "longvarchar" 
        
        modifyDataType(columnName: "AUTH_ROLES", newDataType: "longvarchar", tableName: "AUTH_TOKEN")

    It is easy to remove those, but I wonder if we can avoid that?
2. In the diff the `addColumn` for the column I added, the type is `longvarchar` as well.

I then included the new log in `changelog.groovy`, and ran `dbm-update` again.

It added the `/* AUTO_REWORKED_MIGRATION_KEY */` text to the diff file, but I noticed it did not replace `longvarchar` with `${text.type}`.  

Looking at DbmCallbacks:

* there are regexes which convert data types to `${string.type}`, but `string.type` is not defined in changelog.groovy
* I notice the regexes will only update log entries for addColumn `type: ...` statements, not modifyDataType `newDataType: ..`
* regexes for `longvarchar` are not present

gschueler:

We plan to revisit using dbmigration soon.  We have just gone through the process of upgrading to grails 3, so we might need to start this again from scratch, however.

Thanks again.

Issue #1784, opened 2016-04-06
Different authorization levels for different "tokens" 2.6.1


snebel29:
Hi,
I'm trying to implement different authorization levels for different tokens "Each of them associated to different rundeck local users", however all my tests seems to point to the fact that this is not possible yet or this is not working, just api_token_group aclpolicy seems applied to all of them.

As I see this from a functional point of view, the token should in-heritate both authorization rules, the api_token_group aclpolicy plus the associated user aclpolicy stuff, at this point I'm not sure If I'm doing something wrong in my configuration or this is just not possible at this moment, I've tried to figure it out on my own by looking to

https://github.com/rundeck/rundeck/blob/v2.6.1/rundeckapp/grails-app/conf/rundeck/filters/AuthorizationFilters.groovy

but I'm lacking some Java/groovy knowledge, could some one confirm me if this is possible? If yes could you provide me some configuration example? If this is not possible, don't you think this would be a useful feature?

Regards

gschueler:
all token access uses the pseudo group `api_token_group`, and the username for the token's user.  So you can apply overall access for that group, and specific access for the username.

snebel29:
Hi @gschueler,
Thanks for you quick answer, this ishow I though this should work, however I'm unable to make it work, this is my use case:

I have a local "admin" user that belongs to "admin" group, I have aclpolicy files for for api_token_group and admin group.

When I test "admin" user using the GUI this works as expected I can do what I'm supposed to being able to do and I'm unable to do what I'm not supposed to do.

My api_token_group policy just allow "read" actions, I have a token associated to admin user through the tokens.properties file and admin user is created using real.properties file.

However when using the associated admin token I'm unable to get the admin authorization rules, looking to rundeck.audit.log I see the actions are denied and just api_token_group is mention on.

```
2016-04-04 04:23:43,238 - Evaluating Decision for: res<name:TEST, type:project> subject<Username:admin Group:api_token_group> action<admin> env<http://dtolabs.com/rundeck/env/application:rundeck>: authorized: false: REJECTED, reason: REJECTED, evaluations:        ACLRule</etc/rundeck/api_token_group.aclpolicy[2][type:resource][rule: 4]>{'Admin Application level access control, applies to creating/deleting projects, admin of user profiles, viewing projects and reading system information.' context={application='rundeck'} type='resource' equals , resource={kind=user} for: { group='api_token_group'} allow=[read]} REJECTED for action admin => REJECTED (0ms)
```
[...]

Issue #1550, opened 2015-11-15
Feature Request: web api should have same ACL as the api token user


rophy:
Rundeck 2.6.1.0-SNAPSHOT

I'm trying to create a hubot-rundeck, and found that when interacting with rundeck through web api, only one group is allowed: 'api_token_group'.

Basically this means I cannot implement a hubot interface with web api, since ACL is limited to a single group.

Is there any reason why web api cannot simply share the same ACL as the token owner?

davewongillies:
I'd love to know about this too :+1:

gschueler:
it would be possible to add this kind of behavior for tokens.  the reasons it currently does not work this way:

* user authorization roles are determined by the authentication method (e.g. LDAP) each time a user logs in, and is not accessible to Rundeck outside of a user login session, so Rundeck can't dynamically "look up" what roles a user/token should be given
* we could store the user role list when generating the token, but it's brittle/potentially a security issue:
    1. roles may change, i.e. a user may have some auth role added/removed, which won't be known to Rundeck
    2. a user may be removed from LDAP, e.g. no longer a valid user or no longer allowed to access rundeck, but since the API token could still be used with the original user roles, the token is still valid while the user can no longer log in.

It could still be useful, but we should make the risks explicit and make it an optional feature.

Some ideas for mitigating these problems;

* add expiration to tokens
* allow admin to determine a super set (whitelist) of allowed token roles.  generating a token would only allow the intersection of that whitelist and the user's roles

In the long term, if Rundeck had its own authentication system (v.s relying on servlet container/jaas as we do now), then we could fix these problems.

acasademont:
This would be really nice. Right now I can only have one set of permissions for API tokens, I would love to give limited API tokens to each project and also have un admin token.

hekaldama:
+1

Feel free to make some notes on this to start us in the right direction for a PR.

mbanton:
+1

mnhan3:
+1

lusis:
At this point, I can see a few simple things that would be a happy medium:

- Support the concept of user vs admin token groups
- Allow non-user associated tokens in some fashion

It might be "nice" if the policy could follow the token in a sense as well if that makes sense.

willemdh:
The suggestion of @lusis  would be a minimum requirement to get some security with authentication tokens. 

+1

BobVanB:
+1

doug2525:
+1

datinho:
[...]

PR #2358, opened 2017-03-08
Api token enhancement

jtobard:
# Api Token enhancement
Implemented the capability allow users to generate his own token with the configuration on aclpolicy:

```
context:
  application: 'rundeck'
for:
  resource:
    - equals:
        kind: user
      allow: [generate_self_token] 
```

This users can create tokens associated to his own user and can select a subset or all their roles to assign to the token.
And can be created with a custom duration in minutes, hours or days.
The max time allowed for this tokes is defined via property on the file `framework.properties`:
```
rundeck.user.tokens.duration=1h
```
Using the format: (integer number)(m/h/d)
For minutes, hours or days.

Also, users with admin privileges or with the configuration:

```
  application: 'rundeck'
for:
  resource:
    - equals:
        kind: user
      allow: [generate_service_token] 
```

Can create tokens for other others or create new users associated to the token with arbitrary groups .
The tokens can be created with a custom duration in minutes, hours or days, and the admin can create tokens without expiration time.

Its also possible create tokens from a custom file using `rundeck.tokens.file` on the same ways works now, but you can add more configuration to this new tokens:
```
username: token_string, grouplist
```
as example:
```
user: HW170LMS8w4rFwCj6qmYrUDrRYmm7H0d,dev,ops,qa
```
If you don't add groups, the default `api_token_group`  value is used as group.
This tokens are created without expiration as long as they are on the configuration file they can be used.


gschueler:
for issues #1550 #1784

PR #2380, opened 2017-03-21
Expose refresh node cache to plugin

jtobard:
Added capability to refresh the node cache to plugins, and added a plugin using it.


jimr6007:
how does this work for a plugin like Ansible, similar to AWS/EC2 Autoscaling, if my Ansinle inventory is a dynamic source, a call to the FreeIPA or Foreman API for example and my Ansible inventory is updated, how do I make Rundeck, the Rundeck node cache, be aware of that ASAP?

PR #2392, opened 2017-03-29
fix #921 add project.jobs.gui.groupExpandLevel configuration

### LINKED ISSUE #921: Collapse jobs by default
Usability suggestion:
Add an option [per project] to collapse all job groups by default. 
We have a use case in which our operations group have start stop jobs in one project by team, if it's collapsed navigating to the group they want is very simple.

gschueler:
for #921


gschueler:
project config property:

`project.jobs.gui.groupExpandLevel=1`

Issue #1814, opened 2016-04-19
Failed status when node filter does not match any node


ltamaster:
When scheduling a job with a node filter, if the node filter does not match any node during run, the job fails. The problem with this is the fact that, for this job, no node does not really means error - it is only an indication that the job has nothing to do.

One approach could be to add  a boolean flag on the job, that control the execution status for this case (not maches nodes), considering that in some cases the failure status is required because will indicate  a misconfiguration or a failure.

pgressa:
@gschueler I will fix this as this will be follow up when https://github.com/rundeck/rundeck/pull/2104 will be out - once there's no intersection then it will fail.

bbomgardner:
+1 makes sense as a filter intersection feature. Could definately use that flexibilty over no-match senarios.

PR #2456, opened 2017-04-21
Flag for success when node filter does not match any node

jtobard:
See #1814 
Add a flag to the job on create/edit for success and continue on a step without target.
This property is rendered only when `Dispatch to Nodes` is selected and is false by default.
Included in export/import job via xml/yaml using `successOnEmptyNodeFilter=true`.

Gui:
01_job_edit


jerome83136:
Hi,
I just installed the lastest version and I can't find "Success on a step without targed node ?" on the GUI

Has this feature been implemented ?

I hace checked "Dispatch to nodes"

Many thanks for your help
Best regards
Jérôme

jtobard:
@jerome83136 the message has changed, now it looks like this:
imagen

jerome83136:
Hi,
Thank you very much for this really quick reply.

Ok I see the message you mentioned, but my job still fails with a "node
filter empty" error
I have a dynamic node filter on my job witch is actually empty, but I guess
it should work thanks to this feature. Am I right ?
Regards
Jerome

Issue #2448, opened 2017-04-20
Simplify rdb queries for project home page statistics


gschueler:
Issue type: Enhancement Request

When there are many thousands of executions, the Project List home page does too many db queries to retrieve the statistics, and causes slow loading of the List page.

Issue #1547, opened 2015-11-12
Rundeck performance on 'Activity for this Job' pages


jds86930:
Our rundeck instance has a lot of jobs & a lot of executions ('execution' table is about 1 million rows large). We were noticing that performance got really slow on some pages, such as the 'Activity for this Job' pages. We were able to improve page load time quite a lot by adding the indexes listed below. These indexes were successfully applied on a MySQL 5.6.23 database. I wanted to post them here in case they can help anyone else out, or in case the developers want to integrate them into rundeck.

create index WORKFLOW_STEP__ID on workflow_step(id);
create index WORKFLOW_WORKFLOW_STEP__WORKFLOW_COMMANDS_ID on workflow_workflow_step(workflow_commands_id);
create index EXECUTION__PROJECT__DATE_COMPLETED on execution(project, date_completed);
create index EXECUTION__DATE_COMPLETED__SCHEDULED_EXECUTION_ID on execution(date_completed, scheduled_execution_id);
create index BASE_REPORT__CLASS__JC_JOB_ID__CTX_PROJECT on base_report(class, jc_job_id, ctx_project);

jimr6007:
I have been using http://rundeck.org/2.6.3/api/index.html#delete-an-execution
to cleanup my stuff but I also have 1 million plus rows in workflow and friends tables

I ended up here for the same reason, super slow "Activity for this Job" tabs on the UI

very annoying when you want to look at the last 10 executions for a job.

Any advice from the Rundeck people ?

Thanks !

jds86930:
Haven't seen any followup on it from the rundeck devs. We're still successfully using the indexes listed in the original post, which cut the time down by 10x for us. It'd be nice if they'd integrate that into rundeck (or otherwise optimize the queries and/or data model), but at least this isn't crippling us anymore.

truppert:
This is awesome!  I also saw a huge increase in speed by applying these.  The Rundeck folks should really integrate this.

nickrowlandboxuk:
We're running version 2.6.4-1 of Rundeck. Is this still an issue the current version?

Thanks for sharing this fix. We've seen a massive improvement in page load speed.

zsoca8711:
We have applied these indexes too on our Rundeck 2.7.3-1 and the improvement is massive.

PR #2457, opened 2017-04-24
Improve slow db queries related to activity page and home page

gschueler:
* Add some database index definitions
* simplify/improve some queries


gschueler:
ref #1547

gschueler:
fix #2448

PR #2480, opened 2017-05-03
Update postgresql connector to 42.0.0

nerzhul:
postgresql is very very very old (PostgreSQL 9.1). It's not compatible with postgresql 9.6, update to a postgresql 9.6 version with JDK 8 support.

See https://jdbc.postgresql.org/download.html

### DIFF
--- a/rundeckapp/grails-app/conf/BuildConfig.groovy
+++ b/rundeckapp/grails-app/conf/BuildConfig.groovy
@@ -123,7 +123,7 @@ grails.project.dependency.resolution = {
         compile("org.rundeck:rundeck-core:${rundeckVersion}")
         compile("org.rundeck:rundeck-storage-filesys:${rundeckVersion}")
 
-        runtime 'postgresql:postgresql:9.1-901.jdbc4'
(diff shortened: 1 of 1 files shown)


rundeckapp/grails-app/conf/BuildConfig.groovy
```diff
@@ -123,7 +123,7 @@ grails.project.dependency.resolution = {
         compile("org.rundeck:rundeck-core:${rundeckVersion}")
         compile("org.rundeck:rundeck-storage-filesys:${rundeckVersion}")
 
-        runtime 'postgresql:postgresql:9.1-901.jdbc4'
+        runtime 'postgresql:postgresql:42.0.0'
```
gschueler: note: i think the dependency group id changed to `org.postgresql` so it should be:

    runtime 'org.postgresql:postgresql:42.0.0'

nerzhul:
Fixed

gschueler:
thanks!

nerzhul:
no problem

PR #2482, opened 2017-05-03
Shared data context and data passing between steps

gschueler:
Work-in-progress. (target release: 2.9.0)

Goal: add ability for workflow steps to output data variables which can be used in later steps.

(This is a precursor to adding "data capture plugins" which can parse command output and extract useful data values for later steps.)

- [x] workflow engine refactoring to support a shared data context between steps
- [x] add a "scoped" shared data context for collating data in global, node step, and node contexts
- [x] step context should allow Output data to specify a scope, e.g. export to global scope instead of local scope.
- [x] add mechanism for expanding variable references like `${data.key@node}` and other scoped expansion (syntax work-in-progress)
- [x] variable expansion should resolve values by widening search scope if scope is not specified, otherwise do not widen scope if scope is explicit
- [ ] finalize variable syntax
- [x] use shared context variable expansion in place of previous mechanism in all step contexts


gschueler:
ref #116

gschueler:
Current syntax for arg strings and commands:

    ${step:group.key@node}

E.g.:
   
    ${data.mydata} - global value, or current node's value
    ${1:data.mydata} - value produced by step 1
    ${1:data.mydata@mynode1} - value produced by step 1, node "mynode1"
    ${data.mydata@mynode1} - value produced by node "mynode1"

For scripts, the "inline template" format using `@` chars as delimiters now looks like:

    @step:group.key/node@

jtobard:
Issue using a workflowstep plugin running a local execution or to a single node:
Having this definition:
screenshot at jun 15 13-21-04
Using this regex: `^.+?: (.+?)\s*=\s*(.+)$`

The output is captured as can be seen here:
screenshot at jun 15 13-21-50

But the data is not available to the next job.

Same behavior using a command job its ok:
screenshot at jun 15 13-22-18
screenshot at jun 15 13-22-34

gschueler:
premature merge to master

DarwinJS:
Is this work still active?

gschueler:
@DarwinJS this has been merged to master and will be 2.9.0 release

DarwinJS:
@gschueler - thanks - missed that.

Does this only apply to workflow steps or to all step types?  (The latter would bring it more on par with TeamCity)

gschueler:
@DarwinJS see https://github.com/rundeck/rundeck/issues/2577

gschueler:
@DarwinJS to answer your question, yes it applies to workflow or node steps.  The data captured in a Node context is associated with the node, and is available in later node steps on the same node, but can be referenced explicitly if necessary.  It is not supported for Job Reference steps

dhoffi:
is there a "how to use" example somewhere?

blajoie55:
@gschueler I'm not understanding something and hoping you can help

If passing data generated in a step (like an inline script step) to a job reference step isn't supported, then I don't understand why the issue here:
https://github.com/rundeck/rundeck/issues/116 
was closed. The description you wrote states such type of data passing as a requirement:

"Given the following job :
Job A :

step1 : script

step2 : job reference to Job B

It would be very useful to be able to pass a value generated in step1 as commandline argument value for Job B."

Issue #138, opened 2012-06-23
Scheduled job Time zone support


gschueler:
Is it possible to schedule jobs specifying the time zone? Our
environment consists in different servers supporting different time
zones, and we would like to have a single Rundeck to handle all the
scheduled tasks.

We could try to do the math and schedule all the jobs in local time,
but this would not support different daylight savings settings.

This would be a great feature for large organizations.

[original LH ticket](http://rundeck.lighthouseapp.com/projects/59277/tickets/257)

Original Creator: peppla

nilroy:
@gschueler Is this feature available yet? If not do you know when this could be?

gschueler:
@nilroy it is not available yet

added to our roadmap trello board, please vote for it there: https://trello.com/c/W4M03gnm

PR #2504, opened 2017-05-11
Time Zone support

### LINKED ISSUE #906: Support custom Timezone in all views
Hi,
I'm running Rundeck 2.2.1-1-GA on Ubuntu and I configured it with timezone EST5DST (https://groups.google.com/d/msg/rundeck-discuss/_NUBNfuS9tM/qUBT6G5HBG8J)

It's showing the right time for the activity at Job's level but at the project's activity page (URL/project/SOME_PROJECT/activity) its showing an hour less.
Could some please advice of any custom timezone settings that need to be done?

jtobard:
Fix #906 bug, now the execution time is displayed the same on Activity page and on the job detail.

And add the support for execute scheduled jobs on other time zone from #138

The Time zone now its part of the **Schedule to run repeatedly** section.
The schedule will be executed against this time zone, in case its empty or null, the actual implementation will be used, scheduling it against the jvm timezone.

A new input text with contextual help about how to complete it:
screenshot at may 05 16-15-14

With autocomplete of supported time zones:
pasted image at 2017_05_09 01_42 pm

The input isn't restricted to time zones to enter GMT+/- custom time zones.

In the case the entered time zone isn't valid a error is displayed:
image


rundeckapp/grails-app/views/execution/_wfstateSummaryLine.gsp
```diff
@@ -32,7 +32,7 @@
     </span>
     <span class="timerel">
         <g:message code="at" />
-        <span data-bind="text: formatTimeAtDate(endTime()), attr: {title: endTime() }">
+        <span data-bind="text: formatTimeAtDate('${execution.dateCompleted}'), attr: {title: endTime() }">
```
gschueler: not sure this is good, I think `Date.toString()` is dependent on locale, so the string might change.  I think it is better to use the javascript `endTime()`.  (also 2 other locations)

Fahri28:
Hi,
is this custom script already fixed?, and may i know what rundeck version you use, because i not found "ScheduledExecutionController.groovy" in /var/lib/rundeck

thank you

jtobard:
@Fahri28 This is not a custom script but this feature will be in the 2.9.0 release.

Fahri28:
@jtobard thank you for you replying, 
may i know when the new version 2.9 will be released as we check in rundeck.org currently only 2.8 version

thank you

jimr6007:
Can't believe I didn't notice this until now but it seems that it's defaulting to EST. The server is definitely at UTC and none of my jobs have the TZ specified on the job config page. Today I realized, what da..., my jobs are all running 4 hours earlier than I expected, or 4 hours later. Now I'm confused.

Issue #561, opened 2013-10-02
add a job reference from another project


schast:
Hi,

i would like to add job references to jobs in other projects.
then you could add global jobs in a global project and include them in other projects as a workflow or node step.

jcardali:
Bumping this because it would be a great feature, sadly doesn't seem like it will happen

hmalinov:
+1

kallqvist:
+1

sillkongen:
+1

michnmi:
+1

matejhasul:
+1

stuartTombola:
+1

PR #2519, opened 2017-05-18
Job Reference on another Project

jtobard:
Added capability to use Job Reference step referencing another project.
The current behavior is maintained leaving the field empty ensuring backward compatibility.

New create/edit GUI:
screenshot at may 18 13-51-57

Autocomplete of allowed projects:
screenshot at may 18 13-52-26
If the user has no permission to see another project will not be displayed on autocomplete.
Choose A Job... button will show the entered project jobs if the field is completed. On empty field, continue to show current project jobs.

If the project field is filled with an unauthorized project or a non existent one this error will be displayed:
screenshot at may 18 13-53-20

New text string representation on screen of Job reference from another project:
screenshot at may 18 13-54-09

See #561

### DIFF
--- a/rundeckapp/grails-app/assets/javascripts/jobedit.js
+++ b/rundeckapp/grails-app/assets/javascripts/jobedit.js
@@ -1070,7 +1070,7 @@ function jobChosen(name, group) {
     }
     hideJobChooser();
 }
-function loadJobChooser(elem, nameid, groupid) {
+function loadJobChooser(elem, nameid, groupid, projectid) {
     if (jQuery(elem).hasClass('active')) {
         hideJobChooser();
         return;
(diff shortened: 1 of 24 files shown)


gschueler:
the job reference execution works

there are a few things I noticed:

- in the workflow editor, when you type in a bad project name: the job picker when you click Select
- there could be a dropdown/select for the project name, instead of typing it in
- in the Job show/execution show page for a job with a job reference to another project, there are also some GUI issues:
    - job reference's workflow does not get used in the UI for steps.  This is likely a javascript issue

gschueler:
@jtobard can you look into those issues?

jtobard:
I'm on it

Issue #2498, opened 2017-05-09
Requesting to stop the service while jobs are running doesn't fully kill the service


dandunckelman:
Issue type: Bug report

**My Rundeck detail**
* Rundeck version: 2.8.2
* install type: deb
* OS Name/version: Ubuntu 16.04
* DB Type/version: h2

**Expected Behavior**
When I run `systemctl stop rundeckd`, I expect all Rundeck processes to stop running.

**Actual Behavior**
When I run `systemctl stop rundeckd`, the `bash` process that starts the java process is killed, but the java process continues to run.
- The `bash` process ID matches what is in `/var/run/rundeckd.pid`
- You'll also see that `systemctl status rundeckd` reports that it has stopped

**How to reproduce Behavior**
- Fresh install
- Start the service:  `systemctl start rundeckd`
- Wait until the service has started
- Confirm it's working via `curl -v localhost:4440`
- Stop the service:  `systemctl stop rundeckd`
- See that the bash process was killed, but the java process remains running
- Because of that, if you were to call `systemctl restart rundeckd`, a new `bash` process is started, but it then fails with this error from `/var/log/rundeck/service.log`:

        2017-05-09 17:52:59.725:WARN:oejuc.AbstractLifeCycle:main: FAILED ServerConnector@36e95b7b{HTTP/1.1}{0.0.0.0:4440}: java.net.BindException: Address already in use
- The only workaround I have is to kill all processes found from `ps aux | grep -i rundeckd` using `kill -9 PID`

dandunckelman:
To be extra sure of this, I explicitly ran all the commands for `stop` from `/etc/init.d/rundeckd` and got the same result.

dandunckelman:
This properly killed both the bash & java processes:  `pkill -P $(cat $PIDFILE)` but maybe there's a way to tweak `killproc` and or the value in `$rundeckd` to get it to work?

Dalma-TB:
same OS and rundeck version -> same is happening, although -9 is not necessary, simple kill is enough. So it is definitely reproducible.

dandunckelman:
To add more context to this:  it seems to happen when it tries to stop the service while one or more jobs are currently running.

This is what I see:

```Bash
# Stop service
systemctl stop rundeckd

# Check for running processes
root@rundeck-01:~# ps aux | grep -i rundeck
rundeck   3671  0.0  0.0  45248  3972 ?        Ss   17:13   0:00 /lib/systemd/systemd --user
[...]

PR #2532, opened 2017-05-29
Rundeck init scripts updates

hmalinov:


acasademont:
+1, the current scripts are pretty terrible, spawning 2 processes. Also, maybe a systemd unit file should be bundled instead of the old init script...

acasademont:
@gschueler this works like a charm btw, we've been using this init script for a while now

acasademont:
HI @gschueler any updates on this one? Been using it for over 2 months and works flawlessly (besides #2498 which doesn't have to do with this at all)

hmalinov:
Fixed the conflict.

jtobard:
I have tried and it works to me, I implement the changes to the last version on my master, run the tests, install it on a clean environment, it starts, restarts and stops as expected.

PR #2534, opened 2017-05-29
Passive mode for project

jtobard:
New project booleans that control:

- enable/disable schedule
- enable/disable executions

Similar to the global passive mode, can be edited on the project.properties:
```
project.disable.executions=true
project.disable.schedule=true
```
Or via project configuration page:
2_edit_configuration

If the flags change on the project config page, the jobs are scheduled or unscheduled automatically.


rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
```diff
@@ -3198,4 +3250,24 @@ class ScheduledExecutionService implements ApplicationContextAware, Initializing
...
+    def isProjectScheduledEnabled(String project){
+        def fwProject = frameworkService.getFrameworkProject(project)
+        def disableSe = fwProject.getProjectProperties().get("project.disable.schedule")
+        ((!disableSe)||disableSe.toLowerCase()!='true')
+    }
+
+    def shouldScheduleInThisProject(String project){
+        try {
+            return isProjectExecutionEnabled(project) && isProjectScheduledEnabled(project)
+        }catch (Exception e){
```
gschueler: why is this silently catching an exception?

rundeckapp/grails-app/views/framework/_editProjectForm.gsp
```diff
@@ -49,6 +49,30 @@
             <g:textField name="description" size="50"  value="${projectDescription}" class="form-control"/>
         </div>
     </div>
+    <div class="list-group-item">
```
gschueler: I think instead of hardcoding into the form, I will modify these config options to work via the `RundeckProjectConfigurable` interface.  (just making a note of this)

PR #2546, opened 2017-06-05
Copy job to another project

jtobard:
GUI enhancement to allow copy a job to another project, only if the user is authorized to create in the other project, uses the same copy method only overwriting the project for the selected one

**New item in dropdown menu "Duplicate this job to other Project":**
1_dropdown_menu

**Popup asking to choose the project in the autocomplete list:**
2_select_project

**If the user can't create in any other project, the user don't get the option on the menu:** 
3_dropdown_menu_wo_perm

**This is how it looks if the user can create in other project but only can read on the actual:**
4_dropdown_other_not_this

**If the user tries to duplicate to a project he can't see:**
5_error_cant_see_pr

**If the user tries to duplicate to a project he can see but cant create:**
6_error_noauth_create


gschueler:
issues:

1. does not work from the job show page.  results in 404 `/scheduledExecution/copyanother`
2. should use a Select not a text field for project select

gschueler:
also has the same problem #1 from the execution show page

jtobard:
problems fixed in last commit

Issue #2083, opened 2016-09-23
Delay between retries


daveres:
# Enhancement request

We want to implement retries in a failed job but in some circumstances, we prefer to wait a period between to retries because it's very likely that the job fails another time if we retry after a fail.

We could create a step to make this functionality but it's better to implement job definition.

The last request is to make the wait time exponential, for example, first time 5 minutes, second 10 minutes, third 15 minutes, etc. until job timeout.

truppert:
This would be awesome.  I also see times where an immediate retry fails, where a retry 10 minutes later would succeed because the thing being run needs a bit of time to reset.

apassy:
This would be great - The failure i get are OFTEN because the data on the systems I'm pulling from isn't ready - but I want to get it as soon as it is.  so an instant retry is useless.

ahonor:
+1

PR #2576, opened 2017-06-21
Delay between retries

jtobard:
**Time delay** 

Added in job edit/create page:
captura de pantalla 2017-06-21 a la s 10 40 20
Uses the same format for timeout option and accept option value as input too.

ref. issue: #2083


ahonor:
Update the job format reference docs (yaml/xml) and also update the manual page for Jobs.

jtobard:
Updated doc and new format for xml:
`<retry delay='1h1m1s'>2</retry>`

rundeckapp/grails-app/domain/rundeck/ScheduledExecution.groovy
```diff
@@ -328,8 +328,12 @@ class ScheduledExecution extends ExecutionContext {
             se.uuid = data.uuid
         }
         se.timeout = data.timeout?data.timeout.toString():null
-        se.retry = data.retry?data.retry.toString():null
-        se.retryDelay = data.retryDelay?data.retryDelay.toString():null
+        if(data.retry instanceof Map){
+            se.retry = data.retry.retry?data.retry.retry.toString():null
```
gschueler: just a style comment: but you can do this: 

~~~{.groovy}
se.retry = data.retry.retry?.toString() ?: null
~~~

gschueler: actually even this `data.retry.retry?.toString()` (same thing)

PR #2600, opened 2017-06-30
Move to Gradle 3.5

ReneKroon:
I am trying to find why our project pages are not loading that fast. Since i'm not too familiar with the source and Groovy, i decided to work from the outside in.

This pull-request updates gradle so that it becomes easier to start updating Grails as well. That in turn means that development will become easier since questions, issues and FAQ's will be up to date.

Note that this pull requests also fixes some unit tests which were disabled and gradle 3.5 did not allow it.

Please let me know if you would also appreciate a grails update.


gschueler:
@ReneKroon thank you! we do want to upgrade to grails 3, i have done some work on that front, however it is not complete.  There are a few major technical issues, I think.  If you want to give it a shot, take a look at this branch: https://github.com/gschueler/rundeck/tree/grails3  It is out of synch with master, and we have ongoing work in the prerelease-2.9.0 branch which will be merged to master on its release.  So I was postponing any further grails3 work until that is complete.

My process was sort of odd because i created a grails3 project from scratch and began piecemeal copying things over to the new project.  Maybe it would be better to take what I learned from that and apply it to the existing rundeckapp grails project.  If you have experience doing a grails3 upgrade I would appreciate the help!

ReneKroon:
Thanks for the update. Can this be merged already as separate PR? or do i need to target a different branch than master?

gschueler:
@ReneKroon I will merge it after we perform the next release

PR #2643, opened 2017-07-31
fix typo

chenl:
s/stype/step/


gschueler:
thanks for the typo fixes! for future reference: they could be combined into one pr, that would be fine as well.

chenl:
I was just reading the guide, and send fixes as I was reading, never
expecting the next typo...
Only at a fairly late stage I realize I should probably submit my PR at the
end of a chapter.

| Cheers,
| Chen.

gschueler:
no problem, thanks again!

PR #2648, opened 2017-07-31
typo: available

chenl:
s/availale/available/


chenl:
Found some more typos on the same line. Will submit a different PR to avoid conflicts.

Issue #1584, opened 2015-12-08
Add GUI for managing nodes.


MicahZoltu:
I am evaluating Rundeck and it appears that I can create projects via the GUI, I can create Jobs via the GUI, but in order to create nodes I have to manually edit a file on disk (on the Rundeck box) or manage YAML, JSON or XML files via some other means.

I would very much prefer a GUI that allows me to add/remove nodes.  Initially, perhaps only the file resource type supports GUI modifications but it would be nice if other mechanisms were supported as well (e.g., PUT/POST for URL backed, commit/push for Git backed, INSERT/DELETE for database backed, etc.).

Nhoya:
Push! And plase, a GUI module to setting up the mail server, Grails is a really, really bad thing

sebw:
I integrated Rundeck with iTop CMDB (http://www.combodo.com/itop). 

You add nodes in iTop and they appear in Rundeck immediately:
https://blog.wains.be/2014/2014-09-10-itop-rundeck-integration.md

Check that while waiting for the feature to appear (in a couple of years maybe :-) )

DerekTBrown:
This seems to be a well requested feature, starting a PR:

- [Request #1](https://groups.google.com/forum/#!searchin/rundeck-discuss/add$20nodes|sort:relevance/rundeck-discuss/M7eCei3i_1o/iw3xi_jUwzIJ)

- [Request #2](https://groups.google.com/forum/#!searchin/rundeck-discuss/add$20nodes|sort:relevance/rundeck-discuss/UTLBGyDigPw/_oXMHjh23sMJ)

- [Request #3](https://groups.google.com/forum/#!searchin/rundeck-discuss/add$20nodes%7Csort:relevance/rundeck-discuss/rRhUSa3_oyc/RLwM_h7Zlu4J)

DerekTBrown:
@gschueler this seems to be your domain.  I think the way to approach this is to write a ModelSourceFactory and ModelSource which has two options: a path and a tinymce window for editing the file in one of the accepted formats.  Thoughts?

ahonor:
@DerekTBrown it would be nice if the existing File resource model source plugin provided an inline editor that saved to the configured path (maybe using the ACE editor).

DerekTBrown:
@ahonor agreed.  Bigger question is where to store nodes when added by default

chenlingmin:
When do you support adding node from  GUI?

gschueler:
[...]

PR #2674, opened 2017-08-04
Added Inline Source

DerekTBrown:
Added an inline source option per:

https://github.com/rundeck/rundeck/issues/1584


gschueler:
Thanks for the PR, we have added a solution with https://github.com/rundeck/rundeck/pull/2753 I think.  The only problem with this PR is that the entire source data is then stored in the project configuration, due to the way the plugin configuration is stored.

gschueler:
I am closing this PR, if you want to create a separate project for the plugin, please do so, however the new "writeable" file based model sources already solve this issue for local files.  thanks.

PR #2725, opened 2017-08-30
Improve ISO 8601 date handling with timezone when scheduling ad-hoc job

krzwalko:
When user has a local timezone set to IST (+05:30) job is scheduled at wrong time. This is because ISO 8601 date format was not specified with long timezone format. Browsers (Chrome, Firefox and Safari) are sending runAtTime parameter with long TZ format:

```
------WebKitFormBoundaryxAWvOTSB4R0rNhMN
Content-Disposition: form-data; name="runAtTime"

2017-08-30T20:47:00+05:30

```

Fixed also test case 'should scheduleAdHocJob with alternative ISO 8601 date' where assertion was always true. 

Additional observation: when building and testing - local timezone should be set to GMT.

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
@@ -115,6 +115,20 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
     def fileUploadService
     def pluginService
 
+    static final ThreadLocal<DateFormat> ISO_8601_DATE_FORMAT_WITH_MS_XXX =
+        new ThreadLocal<DateFormat>() {
+            @Override
+            protected DateFormat initialValue() {
+				return new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSXXX")
+			}
+		}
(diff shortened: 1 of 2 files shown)


rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -1976,6 +1990,16 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
      */
     def Date parseRunAtTime(String runAtTime) {
         try {
+            return ISO_8601_DATE_FORMAT_XXX.get().parse(runAtTime)
+        } catch (ParseException e1) {
+
+        }
+        try {
+            return ISO_8601_DATE_FORMAT_WITH_MS_XXX.get().parse(runAtTime)
```
gschueler: Could be good to add a testcase for this format as well

krzwalko: I will add testcases to "should scheduleAdHocJob with alternative ISO 8601 date" and "should not scheduleAdHocJob with invalid ISO 8601 date". 

But I don't know when this ISO_8601_DATE_FORMAT_WITH_MS_XXX format is used. Browsers are sending in ISO_8601_DATE_FORMAT_XXX.

PR #2727, opened 2017-08-30
Fix nested Job Reference

### LINKED ISSUE #2718: Referenced jobs not found by job from another project
Issue type: Bug report/Enhancement Request

**My Rundeck detail**

* Rundeck version: 2.9.3
* install type: (rpm,deb,war,launcher?) RPM
* OS Name/version: Oracle Linux 7.2 (RHEL)
* DB Type/version: mysql? postgres? h2?..  H2

**Expected Behavior**
J1 job from Project P1 calls a SJ1 from another project (P2).  SJ1 calls SJ2 which is also in the P2 project.
Expect J1 to complete.

**Actual Behavior**
J1 fails unable to find job SJ2 in the P1 project.  

**How to reproduce Behavior**
Create Project P2
Create job SJ2 in P2
Create job SJ1 in P2 which calls SJ2

Create Project P1
Create job J1 to call SJ1 in P2

Looking at the XML for the jobs, the project tag is only added in when calling a job from a different project.  It seems like when J1 runs in P1 it can run SJ1 OK because the job step is recording the fact that SJ1 is in the P2 project.  But when it comes time to execute SJ2 that job can't be found in the P1 project so it fails.

If I export/import SJ2 into the P1 project then J1 will complete without error.

These jobs worked fine in 2.9.1

jtobard:
Fix #2718 
When the jobRef item has null project, the parent project is used instead.

Test case:
ProjectA - job1  -> *runs command*
ProjectA - job2 -> *ref job1*
ProjectB - job3 -> *ref ProjectA/job2*

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
@@ -3065,7 +3065,16 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
         def project = null
         if(jitem instanceof JobRefCommand){
             JobRefCommand jitemRef = (JobRefCommand) jitem
-            project = jitemRef.project
+            if(jitemRef.project){
+                project = jitemRef.project
+            }else{
+                //use parent job project before using context framework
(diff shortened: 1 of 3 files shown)


jtobard:
test case added

PR #2748, opened 2017-09-11
Fix Synch -> Sync typo

ashmckenzie:
Noticed a very minor typo for the word Sync :)


gschueler:
Thanks for the submission, however I'm not sure it's a typo! https://english.stackexchange.com/questions/681/synced-or-synched

Sounds like it could go either way, but we consistently use "synch" in user interface messages.  I'm inclined to just keep the current usage.

ashmckenzie:
Oh, ok :)

PR #2759, opened 2017-09-13
Add threadcount as a job context variable

tcsalameh:
This commit adds threadcount as a job context variable, so scripts and
commands called by rundeck can use the threadcount of a job.

For example, if you are running a job that involves taking some nodes in a
cluster offline, the threadcount vs. the total number of nodes in the
cluster could be used by called scripts to check that enough capacity
will be left.


rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -908,6 +908,7 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
         jobcontext.loglevel = textLogLevels[execution.loglevel] ?: execution.loglevel
         jobcontext.retryAttempt=Integer.toString(execution.retryAttempt?:0)
         jobcontext.wasRetry=Boolean.toString(execution.retryAttempt?true:false)
+        jobcontext.threadcount=Integer.toString(execution.nodeThreadcount)
```
gschueler: execution.nodeThreadcount *might* be null, so perhaps it should add `?:1` to guard against that case

tcsalameh: Thanks, just made the change so it will default to 1 in that case.

PR #2763, opened 2017-09-14
Allow building (without tests for now) in Windows

damageboy:
I'm sure I'm doing this horribly wrong in gradle as I'm a total n00b.

Right now I can't build rundeck in windows without something like this patch.

Generally speaking I can't run unit-tests at all (they fail miserably).

But this way, I can at least run:
```
gradlew.bat build -x test
```

And get something done...


gschueler:
Thanks

PR #2767, opened 2017-09-18
addSupplementalRoles if No user roles found

scollector65:
it should be addSupplementalRoles if No user roles found


scollector65:
My ldap hava no role, so I remove all role* attribute from JettyCombinedLdapLoginModule.
And I want give default role but supplementalRoles is not work.
I need supplementalRoles is working because of "ERROR 403 !role"
I expect that it will apply. Thank you!

scollector65:
[#2781](https://github.com/rundeck/rundeck/issues/2781)

scollector65:
I think solve #2781 bug just add 1 line.

PR #2772, opened 2017-09-21
ACL File editor

gschueler:
GUI support for ACL file editing

* separate project/system acl listings
* upload, edit, create, delete acl entries
* entries are validated before saving


jamesdelbarco:
are changes to ACLs in the GUI editor now also tracked by the SCM plugin? Would be great to track for auditing.

PR #2814, opened 2017-10-11
Fix link in manual/getting-help

shigemk2:
http to https


shigemk2:
duplicate #2813, so close.

Issue #1067, opened 2015-01-09
Feature Request: When using the retry option, allow option to suppress notifications until the last execution


smithtimamy:
I would like to submit a feature request. It would be nice to be able to allow a job that is using the "Retry" option to hold of on sending notifications until the last execution.

Current behavior for a job that is set to email on failure and set with a retry count of 2:
- Job starts
- Job fails
- Email sent
- Job retried
- Job fails
- Email sent
- Job retried
- Job fails
- Email sent

Requested behavior for a job that is set to email on failure with a retry count of 2 and option enabled to suppress notifications until the last execution.
- Job starts
- Job fails
- Job retried
- Job fails
- Job retried
- Job fails
- Email sent

NitroPye:
I'm running into this as well. Are you able to reproduce this on scheduled jobs?

WillLiu360:
I would also love this feature! Waking up to a failure email only to find it's okay on retry is not fun...

tumbl3w33d:
Is this planned to be implemented for a 2.7.x release yet or does it need a newer milestone assigned? I would also appreciate this feature. Was surprised it wasn't there yet and also that not more people asked for it here. I consider it essential, exactly for the reason mentioned above.

yazgoo:
+1, this feature is quite needed.

mlachal:
I'm joining in, it would make a great combination with the new retry delay feature.

brianjcarpenter:
+1 for this feature
+1 for retry delay

gschueler:
@brianjcarpenter retry delay was added in 2.9.0

sgrimm-sg:
As a workaround, you can add a second scheduled job with a "Job State Conditional" step and have it look at the status of the job you care about. There are two separate statuses you can test for, "Failed" and "Failed-with-retry." If the job still has retries remaining, the most recent status will be the second of those, so you can have the step check for plain "Failed" which will only match when the retries on the original job are exhausted. Then trigger your notification on the results of that second watchdog job rather than the original one.

This has a bunch of shortcomings (not least: once the original job goes to "Failed" state, the watchdog job will trigger a notification every time it fires) but it may be better than getting failure notifications for jobs that retry and succeed on their own.

[...]

PR #2864, opened 2017-10-19
Separate notifications for retryable failures

sgrimm-sg:
This addresses issue #1067 by adding a new notification trigger
alongside the existing succeeded, failed, started, and average
duration exceeded notifications.

The new notification is triggered when a job that's configured to
retry on failure fails with retries still remaining. Jobs without
retries enabled never get the new notification.

The "failed" notification is now only sent on the last attempt.

### DIFF
--- a/docs/en/developer/05-notification-plugins.md
+++ b/docs/en/developer/05-notification-plugins.md
@@ -12,6 +12,7 @@ Currently there are three conditions that can trigger notifications:
 * `onsuccess` - the Job completed without error
 * `onfailure` - the Job failed or was aborted
 * `onavgduration` - The Execution exceed the average duration of the Job
+* `onretryablefailure` - the Job failed but will be retried
 
 Rundeck has two built-in notification types that can be configured for Jobs:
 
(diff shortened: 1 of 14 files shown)


gschueler:
nb: travis build failure https://github.com/travis-ci/travis-ci/issues/8600

sgrimm-sg:
Should I update the PR to specify `openjdk8` instead of `oraclejdk8` in the Travis configuration? Not sure if the project has reasons to want to stick with the Oracle JDK.

gschueler:
@sgrimm-sg i'm hoping the error is resolved fairly soon.  no particular reason to use oracle vs openjdk. I will try a master pr using openjdk, if that works perhaps you can rebase

gschueler:
@sgrimm-sg well, no luck with that. the problem is in the docker containers we use inside the tests, not a problem with the travis environment, although it appears to have the same cause as the travis issue i linked to.

gschueler:
@sgrimm-sg can you look at this? https://github.com/rundeck/rundeck/issues/3086

PR #2940, opened 2017-11-15
Uplift variables to global

jtobard:
### New Workflow step that creates variables on global for the current execution. 
* Value: Value of the variable, can be a reference to another variable or a static value
* Group: Group on global of the new variable. The variables created on the group `export` are going to be exported to parent jobs.
* Name: Name of the new variable.
The new variable can be used in the way `${group.name}`

screenshot at nov 16 10-08-24

Variables created in global in a Job Reference step with the group `export`, are uplifted to the global on the parent execution context.

Variables created on global are added as state changes after every step for the workflow execution environment.


rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -3230,6 +3230,8 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
         }
         result.sourceResult = wresult
 
+        Map<String, Map<String, String>> data = ((WorkflowExecutionResult)wresult).getSharedContext().getData(ContextView.global())
+        executionContext.getOutputContext().addOutput(ContextView.global(),data)
```
gschueler: I think we should restrict this, to perhaps only `export.*`, so that the user can explicitly export variables (by renaming via the Uplift plugin)

PR #2952, opened 2017-11-20
Uplift variables to global (2)

jtobard:
### New Workflow step that creates variables on global for the current execution. 
* Value: Value of the variable, can be a reference to another variable or a static value
* Group: Group on global of the new variable. The variables created on the group `export` are going to be exported to parent jobs.
* Name: Name of the new variable.
The new variable can be used in the way `${group.name}`

screenshot at nov 16 10-08-24

Variables created in global in a Job Reference step with the group `export`, are uplifted to the global on the parent execution context.

Variables created on global are added as state changes after every step for the workflow execution environment.


gschueler:
@jtobard testbuild.groovy needs to be updated

jolbax:
It seams to be not possible to set the value without the "@nodename". Or I am missing something?
Is there any documentation about this new feature?

ev0ldave:
> It seams to be not possible to set the value without the "@nodename". Or I am missing something?
> Is there any documentation about this new feature?

The problem is the plugin is a workflow step, not a node step meaning it does not have any node context so we must explicitly define the node to read the data from.  In the dialogue it gives an example without an identifier so maybe I am missing something.

gschueler:
@ev0ldave use `${data.xyz*}` from within a workflow step, which collates all node values for `${data.xyz}` as a comma-separated string. You can also use `${data.xyz*-}` for example to use `-` as the delimiter

ev0ldave:
@gschueler Thanks, I thought I had tried that earlier but I must have had a typo.  It definitely works for me now :)

PR #2974, opened 2017-12-04
Add: Job filter as a context variable

adambrett:
We have a situation where we need to do the following:

1) have a common deployment job that takes a number of parameters and can deploy any application
2) have individual jobs that reference the common job with the parameters filled in for that specific application
3) have a series of steps performed on each server before the deployment
4) have the deployment happen sequentially
5) only have a single job reference in each of the "individual" jobs (for ease of maintenance and to ensure governance/compliance)
6) have node filters work normally

It looks something like this in reality:

```
Node 1
--- Run puppet
Node 2
--- Run puppet

Node 1
--- Drain load balancer
--- Deploy application
--- Undrain load balancer

Node 2
--- Drain load balancer
--- Deploy application
--- Undrain load balancer

Workflow step
--- Notify deployment metrics
--- Update changelog
```

```
Application Job (with node filter)
└── Deployment Job (option.NODE as node filter) [workflow step] step first ordering
          └── Puppet Job (option.NODE as node filter)
          └── Drain/Deploy/Undrain Job (option.NODE as node filter) node first ordering
                     └── Drain
                     └── Deploy
                     └── UnDrain
          └── Post Deployment Tasks
```

There is currently no way to achieve this in rundeck, the closest we have is to pass in the nodes as an option to the top level job, but then we lose the node filter functionality.

By adding the job.filter variable as a job context variable, and then passing this to the NODE option of each job it is possible. We need job.filter to be available to ensure that the normal filter functionality continues to behave as expected.

### DIFF
--- a/docs/en/manual/04-jobs.md
+++ b/docs/en/manual/04-jobs.md
@@ -1654,6 +1654,7 @@ Job context variables:
 * `job.retryAttempt`: A number indicating the attempt, if this execution is a [retry](#retry).
 * `job.wasRetry`: `true` if this execution is a retry, otherwise `false`. See: [retry](#retry).
 * `job.threadcount`: Threadcount (number of nodes run at once) of the Job
+* `job.filter`: The filter used to select the nodes for this job (if applicable)
 
 Node context variables:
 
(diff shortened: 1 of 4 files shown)


gschueler:
This seems like a good enhancement.  It would be useful to have a test case to verify the behavior

adambrett:
@gschueler I've added a simple unit test; there didn't seem to be anything already testing the context variable functionality that I could add it on to.

If you meant for me to add an integration or acceptance test creating all the jobs and linking them together...  I have no idea where to start with that.

I could easily provide some terraform manifests or job yaml files that are linked correctly but don't know how I'd hook that into the CI/build in order to test it.

PR #2984, opened 2017-12-08
Improvement to reduce MSSQL deadlocks

jtobard:
Added Database cache by default:
```
hibernate {
	cache.use_query_cache = true
	cache.queries = true
```
This can be removed and use directly from rundeck-config.properties using:
```
hibernate.cache.queries=true
hibernate.cache.use_query_cache=true
```

The count query of executions by job id is reported to create deadlocks too, so, according to Microsoft documentation this can be reduced using the lowest isolation level:
```
total = Execution.withTransaction([isolationLevel: TransactionDefinition.ISOLATION_READ_UNCOMMITTED]) {
                Execution.countByScheduledExecution(scheduledExecution)
            }
```

Also, on bulk delete removed the flush after each execution delete, that slightly reduces the deadlock meanwhile the bulk deleted is executed.

### DIFF
--- a/rundeckapp/grails-app/conf/DataSource.groovy
+++ b/rundeckapp/grails-app/conf/DataSource.groovy
@@ -7,7 +7,8 @@ dataSource {
 }
 hibernate {
 	cache.use_second_level_cache = true
-	cache.use_query_cache = false
+	cache.use_query_cache = true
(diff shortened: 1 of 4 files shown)


gschueler:
closing

PR #3047, opened 2018-01-08
Changes for supporting the spring security plugin on rundeck pro.

ahormazabal:


gschueler:
@ahormazabal it looks like several other changes not related to the config change were included

gschueler:
please rebase on master, i think the security checks are failing because it is not up to date

gschueler:
can you do a `git rebase master` instead of `git merge master`?

ahormazabal:
@gschueler just read your comments. Indeed there is one commit which doesn't belong here. For the sake of simplicity, i'll create a new PR based on the current master and just the needed changes for SSO to work. I'll send it to you ass soon as its ready.
You can close this PR if you want.

bluikko:
So if `web.xml` is now gone from Rundeck non-Pro, how do we use REMOTE_USER authentication from an SSL terminating reverse proxy?

Before 3.0.0 it was necessary to comment out the "auth-constraint" part from `web.xml`.
What is the equivalent on Rundeck 3.0.0?

PR #3122, opened 2018-02-07
Plugins control

ronave:
Implementation to check plugins disabled at project level.


ronave:
Plugin control uses a property under the project configuration that allows the user to disable job plugins.

EG: disabled.plugins=copyfile,localexec

gschueler:
note the format is `Service:provider,...` such as `WorkflowNodeStep:copyfile`

Issue #3139, opened 2018-02-15
SCM git-import plugin limits jobs per YAML file


GroovyCarrot:
Issue type: Bug report
* Rundeck version: 2.10.6-1
* install type: deb
* OS Name/version: Linux 4.4.0-112-generic
* DB Type/version: local + LDAP

The YAML job specification states you can have multiple jobs per file:
http://rundeck.org/docs/man5/job-yaml.html#structure

However due to these lines, it seems that you explicitly can't:
https://github.com/rundeck/rundeck/blob/3673ca9a6cd6e52a4129b460f89735b5fb00f07a/rundeckapp/grails-app/services/rundeck/services/scm/ScmJobImporter.groovy#L62-L65

gschueler:
Currently the SCM support in rundeck only supports a single job per file. Possibly the SCM Import feature could be modified to allow multiple jobs per file, however it would be more difficult to do that for the SCM Export feature because Jobs are mapped to a file path within the repo based on UUID or name.

PR #3140, opened 2018-02-15
Updated git-plugin to support multiple job definitions per YAML file.

GroovyCarrot:
Is a bit of a work around for #3139, I think you would have to do a bit of redesigning the job import plugin api to do it further up. You could potentially expose an `importMultipleFromStream` method, and that wouldn't be a breaking change? I quite liked that I could fix this without compiling the the entire application however. Feel free to reject this PR if it's hacky.


GroovyCarrot:
Ah just realised this needs more work for a job-diff to extract the right job from the yaml file, also I think parsing yaml at this stage breaks xml

Issue #2995, opened 2017-12-13
in the "Git Export: Commit Changes to Git" screen, I would like "Push Remotely" to be checked by default.


gjvc:
Issue type: Enhancement Request

**My Rundeck detail**
* Rundeck version:  Rundeck 2.9.1-1 caffe latte forestgreen sunglasses 2017-08-04
* install type: rundeck-2.9.1-1.35.GA.noarch
* OS Name/version: CentOS release 6.9 (Final)
* DB Type/version: h2

**Expected Behavior**
in the "Git Export: Commit Changes to Git" screen, I would like "Push Remotely" to be checked by default.

**How to Reproduce**
change a job definition
open the "Git Export: Commit Changes to Git" screen, "Push Remotely" is always unchecked and there is no option I can see in the configuration to change the default.

gjvc:
Thanks very much for including it in the latest release :-)

PR #3152, opened 2018-02-26
SCM usability enhancement

jtobard:
### Automatic synchronize on fetch on `git-export` plugin. 
Default `false`, useful in situations where the same repository and branch is used on other projects.
screenshot at feb 26 12-01-14

### Automatic pull on fetch for `git-import` plugin.
Default `true`, should be the default behavior, to be disabled for advanced git users.
screenshot at feb 26 12-02-20

### Clean button 
Only when the plugin is disabled, it performs a real cleanup of the plugin (in git plugin, deletes the repo folder to be created again when the plugin is enabled).

screenshot at feb 26 12-04-22
screenshot at feb 26 12-04-48

### Automatically "initialize-tracking" setup on file pattern

Git Import Setup on configuration
imagen

imagen

### New behavior on commit
The `Push Remotely?` input  is checked by default.
See #2995

imagen


gschueler:
@jtobard the `test-scm-git-plugin-inputs.sh` test failed

jtobard:
Tested upgrade from 2.10.6 to 2.10.7 with this changes without problem. Any previous problem should be solved using the new Clean button. All the configurations are successfully imported.

gschueler:
good work on docs 👍 @jtobard

jtobard:
Thanks! that was the last commit for this PR, now it's ready.

PR #3167, opened 2018-03-02
Accents into firstName and lastName

### LINKED ISSUE #1581: Profile firstName does not accepts accents
Not critical, just information.

First name in the profile section does not allow anything but: `[^[a-zA-Z0-9\s\.,\(\)-]+$]`

Error message:

`Property [firstName] of class [class rundeck.User] with value [Sebastián] does not match the required pattern [^[a-zA-Z0-9\s\.,\(\)-]+$]`

cheers!

jtobard:
Accented and foreign characters on `User` `firstName` and `lastName`.
Fix #1581


goostleek:
The issue still persists in `3.0.1-20180803`. I cannot enter my last name containing Polish diacritical characters.

PR #3183, opened 2018-03-10
Force to kill execution on database problem

### LINKED ISSUE #3155: Can't abort a job, message is "Unable to modify the execution"
Issue type: Bug report

The message appears if for some reason Rundeck has trouble performing a write to the DB on the specified Execution row, and causes the abort to fail before interrupting the process.

The interruption should happen regardless of the outcome of the DB write.

**My Rundeck detail**

* Rundeck version: 2.10.4

**Expected Behavior**

Aborting an execution should cause the execution to stop even if the DB write does not work.

**How to Reproduce**

The message appears if for some reason Rundeck has trouble performing a write to the DB on the specified Execution row.

jtobard:
Fix #3155 
Kill execution wrapped on try catch block, in case a database error is thrown, the job is forced to be killed, and later.
If was killed on the job execution page, ~~the next `ajaxExecState` will try to mark the execution as incomplete.~~ the kill button is replaced with a **Mark as Incomplete** button if you're still on the same execution page.

### DIFF
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/ExecutionController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/ExecutionController.groovy
@@ -1710,8 +1710,9 @@ class ExecutionController extends ControllerBase{
             return
         }
         ExecutionService.AbortResult abortresult
+        def Execution e
         try {
-            def Execution e = Execution.get(params.id)
+            e = Execution.get(params.id)
             if (!apiService.requireExists(response, e, ['Execution ID', params.id])) {
                 return
             }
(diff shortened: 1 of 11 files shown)


rundeckapp/grails-app/controllers/rundeck/controllers/ExecutionController.groovy
```diff
@@ -347,6 +349,13 @@ class ExecutionController extends ControllerBase{
             return render(contentType: 'application/json', text: [error: "Execution not found for id: " + params.id] as JSON)
         }
 
+        JobExecutionContext jexec = scheduledExecutionService.findExecutingQuartzJob(Long.valueOf(params.id))
+        if(!jexec && !e.dateCompleted){
+            log.debug("Mark killed execution ${params.id} as incomplete")
+            executionService.cleanupExecution(e,'incomplete')
```
gschueler: this should not happen in the ajaxExecState call

PR #3195, opened 2018-03-15
Changing RPM Spec file to add rundeck user and group within system UI…

aaronmaxlevy:
…D ranges instead of normal UID ranges.

Currently, it is adding a user within normal UID ranges. As a result, if you install Rundeck on a node with LDAP authentication, Rundeck will create a user for itself using the next available UID and GID, that can cause a UID/GID conflict the next time that an LDAP user is created / populated on the server via nslcd.


com-quoininc-gmanning:
there are also groupadd and useradd commands in the "%pre config" section further down

PR #3237, opened 2018-03-30
update guava from 15.0 to 24.1-jre

mickymiek:
this guava version was causing issues with grpc-protobuf
(even the oldest version of grpc-protobuf was expecting guava version to be at least 18.0)

### DIFF
--- a/rundeckapp/grails-app/assets/javascripts/executionControl.js
+++ b/rundeckapp/grails-app/assets/javascripts/executionControl.js
@@ -1407,9 +1407,21 @@ var FollowControl = Class.create({
                     $(icon).addClassName(status);
                 }
             }
-            if (this.updatepagetitle && !/^\[/.test(document.title)) {
-                document.title =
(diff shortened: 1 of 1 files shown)


gschueler:
what is the purpose of this change?

jbguerraz:
Hello @gschueler :)

Thank you for your feedback! 

Short answer: to get Rundeck GRPC friendly.

We're working on a Rundeck Step plugin that relies on GRPC.

In order to use the GRPC java implementation (https://github.com/grpc/grpc-java) with Rundeck a quite recent version of Guava is needed (18+ AFAIR), and, in addition to that, we probably have to update (and rename) that PR in order to also enforce the version 3.5.1 of java-protobuf (Rundeck uses version 2.5.0 for now) which is also required to get the GRPC java implementation working (CF https://github.com/rundeck/rundeck/compare/master...jbguerraz:guavaprotobuf)

mickymiek:
Closing in favor of #3251

PR #3251, opened 2018-04-06
GRPC ready

jbguerraz:
```
update deps versions to make Rundeck GRPC ready

- guava from 15 to 24.1-jre
- protobuf-java from 2.5.0 to 3.5.1
```
We're working on a Rundeck Step plugin that relies on GRPC.

In order to use the GRPC java implementation (https://github.com/grpc/grpc-java), Rundeck need a quite recent version of `guava` (18+ AFAIR), and  of `java-protobuf`

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -36,6 +36,13 @@ dependencies {
     compile group: 'com.google.protobuf', name: 'protobuf-java', version: '3.5.1'
     compile project(":rundeck-storage:rundeck-storage-api")
     compile project(":rundeck-storage:rundeck-storage-conf")
+    compile ('com.google.guava:guava:24.1-jre') {
+        exclude group:'org.codehaus.mojo', module: 'animal-sniffer-annotations'
+        exclude group:'org.checkerframework', module: 'checker-compat-qual'
+        exclude group:'com.google.errorprone', module: 'error_prone_annotations'
+        exclude group:'com.google.j2objc', module: 'j2objc-annotations'
+        exclude group:'com.google.code.findbugs', module: 'jsr305'
+    }
     compile 'org.apache.ant:ant:1.8.3',
         'log4j:log4j:1.2.17',
         'commons-codec:commons-codec:1.5',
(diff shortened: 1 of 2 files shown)


gschueler:
some more dependencies seem to be added due to the gradle change, the travis build fails because the rpm spec is not expecting those files:

>   /var/lib/rundeck/cli/animal-sniffer-annotations-1.14.jar
>   /var/lib/rundeck/cli/checker-compat-qual-2.0.0.jar
>   /var/lib/rundeck/cli/error_prone_annotations-2.1.3.jar
>   /var/lib/rundeck/cli/j2objc-annotations-1.1.jar
>   /var/lib/rundeck/cli/jsr305-1.3.9.jar
>   /var/lib/rundeck/cli/protobuf-java-3.5.1.jar

jbguerraz:
Thank you for the feedback @gschueler :)

We gonna have to add `protobuf-java` in a better way since it's not required by the CLI (other jars comes from guava, which is required by the CLI).

We need some extra time to dig into that gradle stuff and find out how to manage properly that protobuf-java update: we have to update that jar only `rundeck-launcher/launcher/build/launcher-contents/pkgs/webapp/WEB-INF/lib/protobuf-java-2.5.0.jar`. If you can provide any guidance, we'll be pleased!

gschueler:
reading here: <https://github.com/google/guava/wiki/UseGuavaInYourBuild#what-about-guavas-own-dependencies> it seems that all of guava's dependencies can be excluded.

So here is an example https://github.com/rundeck/rundeck/blob/master/core/build.gradle#L64 where the commons-httpclient dependency is added but excludes the 'junit' transitive dependency.  

So you can do the same thing to include guava but exclude the transitive dependencies that are not needed.

For the rundeck.spec, you should add the protobuf-java jar to the list of dependencies.

jbguerraz:
@mickymiek maybe you could tackle that one ? thx

mickymiek:
Done !

jbguerraz:
@gschueler @mickymiek fixed it (thx!), the checks have passed now :)

gschueler:
👍 Looks good to me!

jbguerraz:
@gschueler What's the next steps to get it merged ?

PR #3253, opened 2018-04-06
Project free form label

jtobard:
Free-form labels for projects.
New property `project.label` in the project configuration:
File configuration

Project configuration

This label is going to be rendered on the GUI instead of the project name in the project list page:
Old:
screenshot at apr 05 15-09-02

New:
screenshot at apr 05 15-09-17

In the project home page:
Project Home Page
In the project selector on the menu:
Project selector

On the API can be read using the `project.label` value:
`/api/21/project/<project>/config/project.label`


guice:
That could work. I don't suppose you can see about "name" being changed to "guid" or something that identifies it's an identifier?  It would be a BC change. It's workable as "name", at least until you get Project settings into a database table. ;)

Issue #1144, opened 2015-03-18
Feature Request: redirect to 'Log Output' tab after submitting a job instead of summary tab


TomGudman:
Hi, 

It would be very useful to have the possibility to immediately see the output of the job we have just triggered instead of displaying the default 'Summary' tab. 

The majority of our jobs are tasks we outsource to end users to be able to trigger operations themselves that returns useful information (e,g: credentials, list of connections, etc) they need straight away. Clicking on the 'Log output' tab is an extra step that could be removed to simplify the end user (our) workflow. 

This parameter could be global or per project or maybe even per job. I dunno from the code point of view how this can be managed though. Sorry I am not very fluent in Java

Version: 2.4.2
OS: Centos6

badarahmed:
+1

jedblack:
+1

gschueler:
a previous request to use `#output` URL fragment on the executio page to automatically select the output tab would I think be necessary, and then augment/change the behavior of the "follow output" checkbox when running a job

gschueler:
this previous request #1107

ahonor:
+1

TomGudman:
Is it possible to add that one in the 2.6.0 milestone? 
I think that could be handy for everybody especially when I see #1154

Thanks

dcorneli:
+1

TomGudman:
What's missing or hard to code/change to make this happen ? Maybe some extra details could help others to code it.

ahonor:
I think there are two ideas here:

1. job definition defines the default tab to follow
2. run page lets a job runner choose the tab

dcorneli:
+1

dirkniblick:
+1 - Adding this feature would be such a game changer.  It's hella annoying to try to flip backwards in the executions of a job trying to find the one that "failed" but didn't really *fail*.  I tried to find a way to get execution output from the rd-cli, but that doesn't seem to be an option.

0xAalaoui:
+1

gschueler:
@dirkniblick you can use `rd` to get the output, with `rd executions follow --restart -e $ID`

dirkniblick:
@gschueler, I may not have described the problem correctly.  I'm not looking to follow a particular job.

[...]

Issue #1107, opened 2015-02-13
Feature Request: job execution links with #output should take you to the actual tab


dbeckham:
When sharing a link to a Rundeck execution such as `https://rundeck.example.com/project/Production/execution/show/290439#output`, taken from the Log Output tab, the link should select the tab the link came from instead of defaulting to the summary.

gschueler:
good idea :)

PR #3304, opened 2018-04-25
Follow execution enhancements

jtobard:
Features #1144 and #1107 implemented.

For 1107:
If you enter the URL using the fragment `#output` you start on the `Log Output` tab.
output fragment

For 1144:
On the job edit page you get a new option to select the default tab:
Edit job

Also you can edit on the `Prepare and Run` page the default tab to be used for this execution.
prepare and run
prepare and run inline


Kaldor37:
That's a really nice update, thank you!
Could you tell me if there is some configuration key to set it by default for all jobs?
I have too many jobs to do it manually.

jtobard:
There's no default configuration for all jobs. 
I think as an option you can update all your jobs directly on the database, the field is `defaultTab` and you can set it to `output`

Kaldor37:
Ok I'll do it this way, thank you for the update :-)

Issue #3300, opened 2018-04-24
Enable Internationalization for Project config level settings


PeekLeon:
Hi,

I would like to internationalize the entire job creation form.

But I don't know how to do with .groovy files like shelduledExecutionService.groovy for example.

How could  integrated a "g.message" for replace the title text or description text.

PS : I not a developer java but i have some skills in development (PHP, javascript, C#)

Sorry for my english, i m french

Thank you

gschueler:
Hi, thanks for the help!

Do you mean you want to internationalize parts of the form that don't have it yet? If so you should look in the various files in `rundeckapp/grails-app/views/scheduledExecution/*.gsp`, they contain the template used to generate the form.  Most of the text is already i18n'ized using g.message, so the messages in messages.properties should apply to it.

PeekLeon:
Hi, thank you for your quick response

Sorry i made a mistake, I wanted to talk about the **Projects forms**  and not ~~Job forms~~ .

The job form are translated by views (*.gsp) there no problem but the project form contains hard text in the file ```rundeck\rundeckapp\grails-app\services\rundeck\services\ScheduledExecutionService.groovy```

Example :

![inter_project](https://user-images.githubusercontent.com/28193768/39241233-75bd77d8-4886-11e8-9c8e-3ffd4c864be3.PNG)

You can find the text here : 

```java
class ScheduledExecutionService implements ApplicationContextAware, InitializingBean, RundeckProjectConfigurable {
    public static final String CONF_GROUP_EXPAND_LEVEL = 'project.jobs.gui.groupExpandLevel'
    public static final String CONF_PROJECT_DISABLE_EXECUTION = 'project.disable.executions'
    public static final String CONF_PROJECT_DISABLE_SCHEDULE = 'project.disable.schedule'

    public static final List<Property> ProjectConfigProperties = [
            PropertyBuilder.builder().with {
                integer 'groupExpandLevel'
                title 'Job Group Expansion Level'
                description 'In the Jobs page, expand Job groups to this depth by default.\n\n' +
                                    '* `0`: collapse all Groups\n' +
                                    '* `-1`: expand all Groups.'
                required(false)
                defaultValue '1'
                renderingOption('projectConfigCategory', 'gui')
            }.build(),
            PropertyBuilder.builder().with {
                booleanType 'disableExecution'
                title 'Disable Execution'
                required(false)
                defaultValue null
                renderingOption('booleanTrueDisplayValueClass', 'text-warning')
[...]

PR #3316, opened 2018-04-27
Update of the French translation

PeekLeon:
Corrections of bug of display and add of translation for the forms of project.


PeekLeon:
Thanks for your problem solving #3300 Enable Internationalization for Project config level settings

PR #3338, opened 2018-05-02
Fix bulk delete button in french translation

PeekLeon:
Fix problem with bulk delete button in french translation.
This button doens't work because the variables default.paginate.rew/ff was false in the french translation


PeekLeon:
Is it worth reopening my pull request for the badge of environment or it is not interressant?

PeekLeon:
Yes indeed, I will test a UI Plug-in.
Thank you for your return.

gschueler:
@PeekLeon thanks for all of the French translation work!

PR #3341, opened 2018-05-03
API: Run execution with just failed nodes (retryExecId)

jtobard:
An endpoint for the API to retry a job's execution on failed nodes.

The new endpoint is: 
`POST /api/23/job/[ID]/retry/[EXECID]`

Optional parameters. 
All of this parameters are going to be populated with the execution values unless they are included in the call:

* `argString`: argument string to pass to the job, of the form: `-opt value -opt2 value ...`.
* `loglevel`: argument specifying the log level to use, one of: 'DEBUG','VERBOSE','INFO','WARN','ERROR'
* `asUser`: specifies a username identifying the user who ran the job. Requires `runAs` permission.
* `option.OPTNAME`: Option value for an option named `OPTNAME`. If any `option.OPTNAME` parameters are specified, the `argString` value is ignored.

If the request has `Content-Type: application/json`, then the parameters will be ignored,
and this format is expected in the content:

~~~~~ {.json}
{
    "argString":"...",
    "loglevel":"...",
    "asUser":"...",
    "options": {
        "myopt1":"value",
        ...
    }
}
~~~~~

The `options` entry can contain a map of option name -> value, in which case the `argString` is ignored.

### DIFF
--- a/docs/en/api/rundeck-api.md
+++ b/docs/en/api/rundeck-api.md
@@ -2302,6 +2302,49 @@ and this format is expected in the content:
 
 See [Listing Running Executions](#listing-running-executions).
 
+### Retry a Job on Failed Nodes
+
+Retry a failed execution only on failed nodes.
+This is the same functionality as the `Retry Failed Nodes ...` button on the execution page.
+
+**Request:**
+
+    POST /api/23/job/[ID]/retry/[EXECID]
+
+Optional parameters. 
+All of this parameters are going to be populated with the execution values unless they are included in the call:
+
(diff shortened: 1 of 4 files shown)


gschueler:
in the GUI you can do Retry Exec or Retry on Failed nodes...I think the API should allow both options as well.  Perhaps `?failedNodes=true/false` (default: `false`)

    POST /api/23/job/[ID]/retry/[EXECID]

    POST /api/23/job/[ID]/retry/[EXECID]?failedNodes=true

gschueler:
Also, you should bump the API version to `24` for this new feature

jtobard:
I ruin the branch by mistake, opening a new PR #3372

PR #3348, opened 2018-05-04
Fix issue 3309 : Missing i18n Texts

PeekLeon:


PeekLeon:
Hi,

The value "you" is not translatable :

image

I think this value is defined by the variable: **${execution.user}** in: ```rundeck/rundeckapp/grails-app/views/execution/_wfstateSummaryLine.gsp```

gschueler:
@PeekLeon i think that comes from here https://github.com/rundeck/rundeck/blob/master/rundeckapp/grails-app/taglib/rundeck/UtilityTagLib.groovy#L879

probably just need to change to `out << g.message(code:'you',default:'you')`

PR #3356, opened 2018-05-07
Fix #2976 

### LINKED ISSUE #2976: Project automatically added to jobs
Issue type: Bug report

**My Rundeck detail**

* Rundeck version: 2.10.0-1
* install type: rpm
* OS Name/version: Red Hat Enterprise Linux 6.8
* DB Type/version: mysql 14.14

**Behavior**
When we upgraded from Rundeck 2.9.2-1 to 2.10-1, several (but not all) of the jobs got the project name automatically added to the job steps.

I have managed to fix that, albeit temporarily, by extracting the job definition, manually removing the project name from job steps, then importing the updated job definition. However, the issue eventually reappeared on the same jobs.

It would not be an issue if we didn't use Rundeck API to export the job definitions jar file. We use this file to populate several differently-named projects on other Rundeck servers, and thus the import fails as they can't find the referenced project on those servers.

**How to Reproduce**
I have not yet found a way to reproduce this, as the issue only affects a subset of jobs, and after manually fixing them, the issue seems to reappear randomly.

jtobard:
Fix #2976 Project automatically added to job's definition after execution of nest referenced job.

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/ExecutionUtilService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ExecutionUtilService.groovy
@@ -241,9 +241,9 @@ class ExecutionUtilService {
             } else {
                 args = new String[0];
             }
-
+            def tmpProj = jobcmditem.jobProject
             if(!jobcmditem.jobProject && parentProject){
-                jobcmditem.jobProject = parentProject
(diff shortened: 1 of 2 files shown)


jtobard:
Missing unit test added

gschueler:
closing to try to restart travis

Issue #3366, opened 2018-05-09
Grails3/merge: Job xml/yaml downloads append full gsp to payload


sjrd218:
When the user downloads the job definition from the UI in either xml or yaml format, the gsp rendered html gets appended to the downloaded artifact.

PR #3367, opened 2018-05-09
Fix for #3366. 

sjrd218:
Job definition download should only contain desired format contents.


gschueler:
going to close and reopen

gschueler:
not sure why snyk is waiting for "rundeck-launcher" gradle files, since they no longer exist...

PR #3372, opened 2018-05-09
API: Run job based on execution (retryExecId)

jtobard:
An endpoint for the API to retry a job's execution on failed nodes.

The new endpoint is: 
`POST /api/23/job/[ID]/retry/[EXECID]`

Optional parameters. 
All of this parameters are going to be populated with the execution values unless they are included in the call:

* `argString`: argument string to pass to the job, of the form: `-opt value -opt2 value ...`.
* `loglevel`: argument specifying the log level to use, one of: 'DEBUG','VERBOSE','INFO','WARN','ERROR'
* `asUser`: specifies a username identifying the user who ran the job. Requires `runAs` permission.
* `option.OPTNAME`: Option value for an option named `OPTNAME`. If any `option.OPTNAME` parameters are specified, the `argString` value is ignored.
* `failedNodes` : `false` to run on the same nodes as the original execution, `true`or empty to run only on failed nodes.

If the request has `Content-Type: application/json`, then the parameters will be ignored,
and this format is expected in the content:

~~~~~ {.json}
{
    "argString":"...",
    "loglevel":"...",
    "asUser":"...",
    "options": {
        "myopt1":"value",
        ...
    }
}
~~~~~

The `options` entry can contain a map of option name -> value, in which case the `argString` is ignored.

### DIFF
--- a/docs/en/api/rundeck-api.md
+++ b/docs/en/api/rundeck-api.md
@@ -2302,6 +2302,49 @@ and this format is expected in the content:
 
 See [Listing Running Executions](#listing-running-executions).
 
+### Retry a Job on Failed Nodes
+
+Retry a failed execution only on failed nodes.
+This is the same functionality as the `Retry Failed Nodes ...` button on the execution page.
+
+**Request:**
+
+    POST /api/23/job/[ID]/retry/[EXECID]
+
+Optional parameters. 
+All of this parameters are going to be populated with the execution values unless they are included in the call:
+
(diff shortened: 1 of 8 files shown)


gschueler:
@jtobard the include.sh file in test/ dir needs to know the latest api version

gschueler:
these tests may be hardcoding/using the wrong version: `test-v23-project-sources-json.sh` and `test-v23-project-sources-xml.sh`

PR #3420, opened 2018-05-22
Enhancement: Using variable in Storage path job options

### LINKED ISSUE #2092: Using variable in Storage path job options
# Bug report

When specifying a variable instead of a hardcoded value in key storage path, option is not interpreted

**My Rundeck detail**
- Rundeck version: 2.6.9-1
- install type: deb
- OS Name/version: Ubuntu 14.04.5

**Expected Behavior**

What I want to do is using a variable when declaring a secure option in the key storage path like this : keys/${job.username}/mykey
The purpose is to use this option as an argument in a command line.

**Actual Behavior**

When executing the job, the path is just not interpreted because (I think) the variable doesn't exists in the process. This results by raising an error : "The ' characters around the executable and arguments are not part of the command".

**How to reproduce Behavior**

Complete workflow is : 
Assume that you already created a key called mykey which type is Password in your key storage. For example, Bob should have a key located in keys/Bob/mykey
- Create an option called userpass, select Secure as input type and then specify keys/${job.username}/mykey as Storage Path.
- Create a command line node step and write this : echo ${option.userpass}
- Run the job with Debug mode activated

...

jtobard:
Resolves #2092 

Add capability to use ${job.x} and ${node.x} on Storage Path options.

example

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/execution/ExecutionServiceImpl.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/execution/ExecutionServiceImpl.java
@@ -26,6 +26,8 @@ package com.dtolabs.rundeck.core.execution;
 import com.dtolabs.rundeck.core.CoreException;
 import com.dtolabs.rundeck.core.common.Framework;
 import com.dtolabs.rundeck.core.common.INodeEntry;
+import com.dtolabs.rundeck.core.data.SharedDataContextUtils;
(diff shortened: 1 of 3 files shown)


YYAK:
I am using Rundeck 3.0.9-20181127 and this feature still not available. I am very surprised that it is not implemented yet as it was core plan to propogate secrets to /dev/test/prod.

thomas-illiet:
Hello, Can you add support for global variables?

PR #3426, opened 2018-05-24
Fix #3423 by responding with error 

### LINKED ISSUE #3423: Create project via API with invalid project name does not return error
Issue type: Bug report

Using the `POST /api/11/projects` endpoint to create a project, with an invalid project name such as `my project` (space char), results in `201 Created` even though the project is not really created.

using `rundeck.projectsStorageType=db` results in silently failing to create the project.

using `rundeck.projectsStorageType=filesystem` results in creating a project directory on disk, but does not result in a functional project, it is not listed in the projects list.

**My Rundeck detail**

* Rundeck version: 2.11 and earlier
* install type: any

**Expected Behavior**

`HTTP 400` response

**How to Reproduce**

can reproduce with [`rd projects create -p 'my project'`](https://github.com/rundeck/rundeck-cli)

reference: https://github.com/rundeck/rundeck-cli/issues/180

sjrd218:
Respond with error when create project api is called with an invalid project name.


gschueler:
👍 Looks good to me!

PR #3440, opened 2018-05-28
Limit Multiple Executions

### LINKED ISSUE #1387: Feature request: Limit Multiple Executions
We have a fare amount of jobs that we like to have multiple executions on the same node but we need a cap due to resource constraints. 

Having a "Multiple Execution Limit" would allow us to cap a job to x consecutive executions. This will allow the job to grow up to x executions when enough work is required. 

Our somewhat workaround is using a external script to call into the API to see how many executions are currently running for the job. This is pretty ugly and something built in would be extremely appreciated. If there is currently a way to do this that we have overlooked please let me know.

Thanks!

jtobard:
Closes #1387

Adds the capability of establishing a maximum number of multiple executions.
Leave the value in blank or to zero means unlimited multiple executions. For current jobs with multiple executions, this will be the default behavior. 

Edit job:
edit

Job definition:
resume


rundeckapp/grails-app/i18n/messages_fr_FR.properties
```diff
@@ -488,13 +488,18 @@ scheduledExecution.property.nodeRankOrder.descending.label=D\u00e9croissant
 scheduledExecution.property.nodeIntersect.label=Intersection de n\u0153uds
 scheduledExecution.property.nodeIntersect.true=Oui, cr\u00e9ez une intersection de n\u0153uds avec les n\u0153uds de travail actuels
 scheduledExecution.property.nodeIntersect.false=Non, utilisez le filtre de n\u0153ud r\u00e9f\u00e9renc\u00e9 ou remplac\u00e9 par un travail r\u00e9f\u00e9renc\u00e9
+<<<<<<< HEAD
```
gschueler: merge remnant

PR #3442, opened 2018-05-28
ImportOptions to jobref in job definition

### LINKED ISSUE #3429: importOptions missplaced in yaml/xml export.
Hello,
I have a problem with the import options feature, when combined with the xml/yaml job export feature.

When I do a xml/yaml export of a job using the import option feature, the xml/yaml object for the import options is at the wrong place => the importOptions element is in the command element instead of the jobref element.

Issue type: Bug report

**My Rundeck detail**

* Rundeck version: 2.11.3-1
* install type: rpm
* OS Name/version:  centos 7
* DB Type/version: h2

**Expected Behavior**

A job using the "import options" feature should be exported/imported without the loss of this feature activation.

**How to Reproduce**

Create a Job with options.
Create a Job with previously created Job as a workflow step and activate  the "import options" feature.
Export this job as xml/yaml
Re-import this job.
=> The "import options" feature is not activated anymore.

jtobard:
Fix #3429

Job definition now it's exported correctly:

```
  sequence:
    commands:
    - jobref:
        group: ''
        importOptions: true
        name: test2
```

### DIFF
--- a/rundeckapp/src/test/groovy/rundeck/JobExecSpec.groovy
+++ b/rundeckapp/src/test/groovy/rundeck/JobExecSpec.groovy
@@ -219,7 +219,7 @@ class JobExecSpec extends Specification {
         null            | _
     }
 
-    def "from map with node intersect"() {
+    def "from map with importOptions"() {
         given:
         def map = [
                 jobref     : [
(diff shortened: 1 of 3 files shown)


rundeckapp/src/test/groovy/rundeck/JobExecSpec.groovy
```diff
@@ -167,4 +167,80 @@ class JobExecSpec extends Specification {
...
+        }
+
+        where:
+        importOption    | _
+        true            | _
+        false           | _
+        null            | _
+    }
+
+    def "from map with node intersect"() {
```
gschueler: test name seems wrong, should be `from map with importOption` ?

PR #3457, opened 2018-06-01
Implement Travis stages.

ProTip:
This PR introduces an initial Travis CI stages **"spike"**.  It introduces a few new concepts:

* Travis CI stages to bundle jobs that can be run in **parallel**
* Scripts re-bundled into a build stage, followed by a test stage with parallel jobs
* Artifact syncing to s3 for sharing outputs between stages

### Immediate benefits:
* Knocks a further **10+ minutes** off the existing build process
* Further increases build output readability by grouping outputs under jobs in the Travis UI
* Opens up s3 and other aws services to the build process

### Future enhancements:
* Building base docker containers and pushing them to dockerhub for reuse in downstream stages
* Speeding up the API test suite(possibly use a different runner from roundup that support parallel execution and a junit results formatter)
* Gradle caching(some info out there on how to do this with Travis)
* Exporting junit reports to a location for further visibility/analysis; could also pull all the junit files into a stage and spit some details out in travis)

### There are a few Travis CI oddities worth mentioning:
* Matrix expansion is not super friendly with stages(or at all TBH)
* The reported build time in notifications is based on the sum of all jobs, not the elapsed wall time of the entire build
* Travis CI does not support naming jobs; an environment variable is set to work around this and it is visible in the UI

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@
 **/*/build
 **/out
 **/*/out
+/artifacts
 /cobertura.ser
 /gradle-cache
 /rundeckapp/stacktrace.log
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,28 +6,85 @@ jdk:
 sudo: required
 services:
 - docker
+
+stages:
+- build
+- test
+
 before_install:
 - sudo apt-get update -qq
 - sudo apt-get install -y xmlstarlet jq
 - sudo chsh --shell $(which bash)
+- pip install awscli --upgrade --user
 install: true
 before_script:
 - export -f travis_nanoseconds
 - export -f travis_fold
 - export -f travis_time_start
 - export -f travis_time_finish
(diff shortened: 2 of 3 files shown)


gschueler:
looks great! One nitpick: do we need to sync all of the packaging/debdist/ contents? I think we only need the .deb: https://travis-ci.org/rundeck/rundeck/jobs/386484765#L2887

ProTip:
I don't think we do.. It just seemed like a simpler command(and s3 storage is only $2.30 for 100GB months).

I think I've come up with a different copy strategy that is all-around better based on your comment though :) It's baking now.

gschueler:
👌 perfect

gschueler:
[Improve travis build (stages)](https://trello.com/c/QdBEbcUS/13-improve-travis-build-stages)

PR #3463, opened 2018-06-04
Add stage for forks.

ProTip:
### What
This PR introduces a `fork` stage that consolidates all the tests into one stage/job. This is essentially the *slow* path.

### Why
Encrypted envars are not available to forks, so we are not able to pass artifacts between stages and jobs. In the future we can discuss running a subset of stages/tests on PRs coming through on forks. For instance, if we were using Sauce Labs we wouldn't want to expose credentials to forks and may skip the selenium tests or run them locally. As well, Travis is supposedly releasing artifact support this summer...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,11 +9,11 @@ services:
 
 stages:
 - name: build
-  if: (NOT fork)
+  if: fork = false
 - name: test
-  if: (NOT fork)
+  if: fork = false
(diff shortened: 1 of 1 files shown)


ProTip:
@gschueler This should be good to go..

gschueler:
👍 Looks good to me!

PR #3477, opened 2018-06-06
Set a max file size of 25mb configurable

### LINKED ISSUE #3476: Error 400 importing projects over 128kb
Error 400 importing projects over 128kb.

**My Rundeck detail**

* Rundeck version: 3.0
* install type: launcher
* OS Name/version: Mac OS
* DB Type/version: mysql

**To Reproduce**
1. Go to Project -> Import Archive
2. Try to import any file over 128kb
3. Error 400 on screen

**Expected behavior**
Correct import project.

**Screenshots**
imagen

**Additional context**
Grails 3 default is 128000 (~128KB).
https://github.com/grails/grails-core/blob/3.0.x/grails-plugin-controllers/src/main/groovy/org/grails/plugins/web/controllers/ControllersGrailsPlugin.groovy#L73

jtobard:
Fix #3476
Default max size for file upload. The grails 3 max by default is 128kb, but exported projects can be greater than that.
The new default max size is 25MB but can be overrided on the `rundeck-config.properties` file with the file size in bytes:
```
grails.controllers.upload.maxFileSize=26214400
grails.controllers.upload.maxRequestSize=26214400
```

### DIFF
--- a/rundeckapp/grails-app/conf/application.groovy
+++ b/rundeckapp/grails-app/conf/application.groovy
@@ -13,6 +13,9 @@ dataSource {
     password = ''
 }
 
+grails.controllers.upload.maxFileSize=26214400
+grails.controllers.upload.maxRequestSize=26214400
+
 environments {
     production {
 //        grails.serverURL = "http://www.changeme.com"


gjvc:
Boom.  That was it.

`-rw-rw-r--.  1 gjvc gjvc  30M Sep 12 12:05  MyProject-UAT-20180912-065815.rdproject.jar
`

Might I suggest bumping the default to a round number like 64MB or 128MB ? (This particular instance has been running for approximately two years and is not that busy.)   Bear in mind that as time passes, there will be a whole slew of people using the import feature to upgrade from rundeck 2 to 3, and this is likely to be a repeating problem support-wise.

PR #3498, opened 2018-06-08
Usage of option values for timeout

### LINKED ISSUE #3484: Job timeout with options is not working
**Describe the bug**

**My Rundeck detail**

* Rundeck version: 3.0
* install type: launcher
* OS Name/version: Mac OS
* DB Type/version: mysql

**To Reproduce**
1. Create/Edit job
2. Add an option, in this case, I'm going to call it `timeout`
3. Set job's timeout to `${option.timeout}`
4. Run the job (it can be a `sleep 100` command and the option timeout could be `2s`)
5. The timeout is ignored

**Expected behavior**
The job should translate the option and stop the job marking it as `Time Out`.

**Screenshots**
imagen

**Additional context**
On the execution definition, the timeout is translated correctly, but it doesn't work
imagen

jtobard:
Fix #3484
Get the options values from the executions if the timeout of the `scheduleExecution` starts with `${`.

### DIFF
--- a/rundeckapp/grails-app/jobs/rundeck/quartzjobs/ExecutionJob.groovy
+++ b/rundeckapp/grails-app/jobs/rundeck/quartzjobs/ExecutionJob.groovy
@@ -251,6 +251,12 @@ class ExecutionJob implements InterruptableJob {
             initMap.secureOpts=jobDataMap.get("secureOpts")
             initMap.secureOptsExposed=jobDataMap.get("secureOptsExposed")
             initMap.execution = Execution.get(initMap.executionId)
+            def jobArguments=initMap.frameworkService.parseOptsFromString(initMap.execution.argString)
+            if (initMap.scheduledExecution?.timeout && initMap.scheduledExecution?.timeout.contains('${')) {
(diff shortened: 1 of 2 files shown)


rundeckapp/grails-app/jobs/rundeck/quartzjobs/ExecutionJob.groovy
```diff
@@ -251,6 +251,12 @@ class ExecutionJob implements InterruptableJob {
             initMap.secureOpts=jobDataMap.get("secureOpts")
             initMap.secureOptsExposed=jobDataMap.get("secureOptsExposed")
             initMap.execution = Execution.get(initMap.executionId)
+            def jobArguments=initMap.frameworkService.parseOptsFromString(initMap.execution.argString)
```
gschueler: this needs to happen after the below "retry" section where it attempts to load the execution again, otherwise it may fail

gschueler: https://github.com/rundeck/rundeck/pull/3498/files#diff-7ec10a60f137d2c77c4da486dbc02452R281

PR #3593, opened 2018-06-26
Add ability to serve static user assets from a configurable location on the filesystem

sjrd218:
Allows custom logos to be served fixing #3589.


gschueler:
looks good, I suppose some admin docs will need updating?

PR #3621, opened 2018-06-28
Add date component to version.

ProTip:
Adds date in the format of `YYYYMMDD` to build versions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 SHELL=/bin/bash
 
+DATE=$(shell date +%Y%m%d)
 VNUMBER=$(shell grep version.number= $${PWD}/version.properties | cut -d= -f 2)
 ifndef TAG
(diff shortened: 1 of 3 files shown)


ProTip:
Only `release` builds produce versions with a date component now.

I also dove into a bit of reshaping/refactoring ahead of the beta release:
* Gradle is moved more towards being the version source of truth across different build processes/script; arguments are more just for testing overrides
* Makefiles get version info from Gradle and some cleanup/refactor
* The version in the `version.properties` is "roll-forward" now; no need to revert it back after release builds.
* deb/rpm versioning tweaked to use tilde consistently in the versioning

There is one simple toggle now to produce a snapshot vs release versioned build:
`./gradlew bootRepackage`
vs.
`./gradlew -Penvironment=release`

Still need to do some downstream cleanup and cross-project testing.

PR #3692, opened 2018-07-13
Add metrics to API, update to APIv25

gschueler:
paths: /api/25/metrics  - list of metric hrefs

/api/25/metrics/{metrics,ping,threads,healthcheck} - endpoints

configuration, the servlets must be enabled for each endpoint, plus
api must be enabled for each endpoint

rundeck.web.metrics.servlets.$endpoint.enabled=true
rundeck.api.metrics.$endpoint.enabled=true

TODO:

- [x] access control?
- [x] unit tests
- [x] api functional tests
- [x] documentation

### DIFF
--- a/rundeckapp/grails-app/conf/application.yml
+++ b/rundeckapp/grails-app/conf/application.yml
@@ -159,6 +159,16 @@ rundeck:
                 delimiter: ','
         useHMacRequestTokens: true
     storage:
+    execution:
+        finalize:
+            retryMax: 10
+            retryDelay: 5000
+        status:
+            retryMax: 5
+            retryDelay: 5000
+        logs:
+            fileStorage:
+                cancelOnStorageFailure: true
     mail:
         template:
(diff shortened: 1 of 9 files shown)


gschueler:
Previously we had configuration to enable each metrics *servlet*, like `rundeck.web.metrics.servlets.[name].enabled=true/false`, however the metrics servlets were not very useful by themselves without API access.  Now that I've enabled API access, should that config be replaced with an API specific one, and the previous use of servlets directly be removed?

gschueler:
Documentation pr: https://github.com/rundeck/docs/pull/33

lukasged:
by chance, is there a way to enable anonymous access to the `/api/25/metrics/healthcheck` endpoint?

monsdar:
Late to the party, but as I stranded here: Anonymous access to health-check is available at /health now. Docs: https://docs.rundeck.com/rundeck-cloud/administration/cluster/loadbalancer/health-check.html

PR #3697, opened 2018-07-16
Job not found message on definition tab

### LINKED ISSUE #3694: Wrong Job Reference link
In the description tab, when a job references another job that doesn't exist, either by a misspelled name or by a UUID that no longer exists, the link to the other job disappears, but it lacks a message to warn of this situation.

**My Rundeck detail**

* Rundeck version: 3.0.0
* install type: launcher
* OS Name/version: Mac OS
* DB Type/version: mysql

**To Reproduce**
1. Create a job.
2. Add a Job Reference step.
3. Select any job, change the UUID to something that doesn't exist, or remove the UUID and misspell the name.
4. Save the Job
5. Go to Definition tab.
6. The reference step is just black without any link. 

Another way to reproduce it is to delete the referenced job and see the definition of the parent job.

**Expected behavior**
A warning message about the situation, something like `(Missing Job)`.

**Screenshots**

Example of correct job reference:
jobrefok

Example of incorrect job reference:
jobrefnok

jtobard:
Fix #3694
Adds message `(Job not found)` on definition tab for wrong referenced jobs.

ss

### DIFF
--- a/rundeckapp/grails-app/i18n/messages.properties
+++ b/rundeckapp/grails-app/i18n/messages.properties
@@ -1655,3 +1655,4 @@ passwordUtility.output.label=Output
 passwordUtility.encoder.label=Encoder:
 
 execution.html.waiting=Waiting for Execution log...
+jobreference.wrong=(Job not found)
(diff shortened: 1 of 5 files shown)


gschueler:
maybe it should still include the job name/uuid in the display however

jtobard:
@gschueler the message is in bold, the name is still after the `)`

jtobard:
Actually I'm going to move the message to another line to make it clear

jtobard:
Red color to the message on the name:

name

And using just uuid:

uuid

PR #3698, opened 2018-07-16
fix scm plugin working checkboxes

### LINKED ISSUE #3696: SCM plugin checkbox not working
Using SCM plugin (import or export), on the `Commit Changes to Git` page the checkboxes doesn't work graphically. (Any browser)

**My Rundeck detail**

* Rundeck version: 3.0.0
* install type: launcher
* OS Name/version: Mac Os
* DB Type/version: mysql

**To Reproduce**
1. Setup git export plugin (or import)
2. Go to the project job list page
3. Job Actions ... commit changes to git
4. See the job select page with a list of jobs with checkboxes

**Expected behavior**
Checkboxes that work graphically.

**Screenshots**
commit

jtobard:
Working checkboxes on git plugins
git-export

### DIFF
--- a/rundeckapp/grails-app/views/scm/performAction.gsp
+++ b/rundeckapp/grails-app/views/scm/performAction.gsp
@@ -110,9 +110,10 @@
                                                 <g:if test="${jobstatus?.synchState?.toString() != 'CLEAN'}">
 
                                                     <g:checkBox name="id" value="${job.extid}"
+                                                                id="id${job.extid}"
                                                                 checked="${selected?.contains(job.extid)}"/>
                                                 </g:if>
-                                            <label>
+                                            <label for="id${job.extid}">
                                                 <g:render template="statusIcon"
                                                           model="[iscommit          : true,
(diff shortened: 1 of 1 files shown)


gschueler:
when the `<label>` doesn't surround the `<input>` it should have a `for=".."` attribute using the `id` of the `<input>`

rundeckapp/grails-app/views/scm/performAction.gsp
```diff
@@ -191,8 +192,9 @@
 
                                         <div class="checkbox col-sm-12">
                                                 <g:checkBox name="deletePaths" value="${path}"
+                                                            id="deletePaths${path}"
```
gschueler: the value of `path` is likely to be like `project/X/some-uuid.xml` which is not going to be a valid HTML ID attribute value.  In the `<g:each` you can add a `status="somevar"` which will create a variable `somevar` as the index of the item in the list... then you could use `id="deletedPath${somevar}"` for the id

PR #3714, opened 2018-07-20
Create new branch on git-export

jtobard:
Add new configuration options to create a new branch on the `git-export` SCM plugin.

config

If the branch doesn't exist on the remote repo, it will create a new one with the `branch` name. It will be based on the `base branch`.

If the `base branch` doesn't exist on remote, it will throw an error:

imagen

### DIFF
--- a/plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/GitExportPlugin.groovy
+++ b/plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/GitExportPlugin.groovy
@@ -139,44 +139,6 @@ class GitExportPlugin extends BaseGitPlugin implements ScmExportPlugin {
         inited = true
     }
 
-    private void createBranch(ScmOperationContext context, String newBranch, String baseBranch){
-        def createCommand = git.branchCreate()
-                .setName(newBranch)
-                .setStartPoint("${REMOTE_NAME}/${baseBranch}")
-                .setForce(true)
-
-        try {
-            createCommand.call()
-        } catch (Exception e) {
-            logger.debug("Failed creating branch ${newBranch}: ${e.message}", e)
-            throw new ScmPluginException("Failed creating branch ${newBranch}: ${e.message}", e)
-        }
-        def pushb = git.push()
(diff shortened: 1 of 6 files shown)


plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/GitExportPlugin.groovy
```diff
@@ -118,7 +120,21 @@ class GitExportPlugin extends BaseGitPlugin implements ScmExportPlugin {
...
+            if(config.shouldCreateBranch()){
+                if(config.baseBranch && existBranch("refs/remotes/${this.REMOTE_NAME}/${config.baseBranch}")){
+                    createBranch(context, config.branch, config.baseBranch)
+                    cloneOrCreate(context, base, config.url)
+                }else{
+                    logger.debug("Non existent remote branch: ${config.baseBranch}")
+                    throw new ScmPluginException("Non existent remote branch: ${config.baseBranch}")
+                }
+            }else{
+                throw new ScmPluginException("Non existent remote branch: ${this.branch}")
```
gschueler: Maybe the error message in this case should say something like `Could not clone the remote branch: ${this.branch}, because it does not exist. To create it, you need to set the Create Branch option to true.` So that the user knows how to fix the problem.

plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/config/Export.groovy
```diff
@@ -80,6 +80,35 @@ Template".
...
+            description = "If the entered branch doesnt exist on remote repo, create a new one. If false, it will fail if the branch doesn't exist",
+            defaultValue = 'false',
+            required = false
+    )
+    @SelectValues(values = ['true', 'false'])
+    @RenderingOption(
+            key = StringRenderingConstants.GROUP_NAME,
+            value = "Git Repository"
+    )
+    String createBranch
```
gschueler: Can this be a `boolean` ?

gschueler:
👍 Looks good to me!

PR #3743, opened 2018-07-27
Fix PAM login issues

sjrd218:
Update libpam4j dependency. Add a password callback to Jetty support to retrieve the password from the form and pass it to the PAM module. Add docker test for PAM auth.


gschueler:
oops, i see you added pam docker tests, but they were not included in travis config

sjrd218:
My bad, I can do that.

PR #3821, opened 2018-08-10
Disabled plugins on Job

### LINKED ISSUE #3820: Plugin control disabled plugins
After disabling a Plugin using the project properties, the Plugin still can be selected as a Node or Workflow step for a new job. After running the job an error message appears that the Plugin is disabled.

**My Rundeck detail**
* Rundeck version: 3.0.1
* install type: any
* OS Name/version: Mac OS
* DB Type/version: any

**To Reproduce**
1. Disable any plugin in the project.properties file, as example `disabled.plugins=WorkflowNodeStep\:stub-node-step`
2. Create a new job or edit one
3. Add step
4. The plugin is on the list

**Expected behavior**
Is expected to the disabled plugins disappear from the list of plugins on create/edit job to avoid confusion on the users.

jtobard:
Fix #3820 
Disabled plugins are no longer in the plugin list of the create/edit job page.


jtobard:
closing because ugly rebase

Issue #3855, opened 2018-08-20
Rundeck LDAP Group Synchronization


shehzadshaikh:
**Describe the bug**
We have deployed **Rundeck 3.0.1** and configured it to authenticate via LDAP (Active Directory). I am able to authenticate with individual users. Now i want my user to act as admin user for rundeck hence, i have added the group name which i belong to (for instance myOrgAdminGroup) in admin.aclpolicy file under 
**by:
  group: myOrgAdminGroup**
above change doesn't help and produces an error 
You have no authorized access to projects.
Contact your administrator. (User roles: ROLE_myOrgAdminGroup, ROLE_myUserName, and ROLE_all other group i belong to.
If i add another username explicitly in admin.aclpolicy file such as below
**by:
 group: myOrgAdminGroup
 username: myUserName**
then it works and i become an Admin user for rundeck.
 
Here is my **admin.aclpolicy** file
description: Admin, all access.
context:
  project: '.*' # all projects
for:
  resource:
    - allow: '*' # allow read/create all kinds
  adhoc:
    - allow: '*' # allow read/running/killing adhoc jobs
  job:
    - allow: '*' # allow read/write/delete/run/kill of all jobs
  node:
    - allow: '*' # allow read/run for all nodes
by:
  group: admin
  #username: myUserName

---

description: Admin, all access.
context:
  application: 'rundeck'
for:
  resource:
    - allow: '*' # allow create of projects
  project:
    - allow: '*' # allow view/admin of all projects
  project_acl:
    - allow: '*' # allow admin of all project-level ACL policies
  storage:
    - allow: '*' # allow read/create/update/delete for all /keys/* storage content
by:
  group: admin
  #username: myUserName

and here is my **jaas-activedirectory.conf**
activedirectory {
  com.dtolabs.rundeck.jetty.jaas.JettyCachingLdapLoginModule required
        debug="true"
        contextFactory="com.sun.jndi.ldap.LdapCtxFactory"
  providerUrl="ldaps://AD_HOSTNAME:3269"
  bindDn="AD_USER@AD_HOSTNAME"
  bindPassword="PASSWORD"
  authenticationMethod="simple"
  forceBindingLogin="true"
  userBaseDn="DC=example,DC=example,DC=com"
  userRdnAttribute="sAMAccountName"
  userIdAttribute="sAMAccountName"
  userPasswordAttribute="unicodePwd"
  userObjectClass="user"
  roleBaseDn="DC=example,DC=example,DC=com"
[...]

PR #3883, opened 2018-08-22
LDAP authentication fixing

ltamaster:
This is a fix for a bug getting the roles from LDAP.
1)  By default, rundeck3 use the role prefix `ROLE_` which can break the use of LDAP on 2.x
2) Also, rundeck3 add the username as a role.

1) Replacing the default role prefix with an empty value.
That value can be modified using the parameter `rundeck.security.jaasRolePrefix=` on `rundeck-config.properties`

I tried using  `rundeck.security.jaasRolePrefix=` on  `rundeck-config.properties` to pass an empty value, but that is not posible using  `rundeck-config.properties` (maybe using the groovy format)

**Additional context**
The use of `ROLE_` as prefix can produce this kind of errors

screenshot 2018-08-22 15 51 34

This can fix this issue: https://github.com/rundeck/rundeck/issues/3855

### DIFF
--- a/rundeckapp/grails-app/conf/spring/resources.groovy
+++ b/rundeckapp/grails-app/conf/spring/resources.groovy
@@ -397,7 +397,7 @@ beans={
 
     rundeckUserDetailsService(RundeckUserDetailsService)
     rundeckJaasAuthorityGranter(RundeckJaasAuthorityGranter){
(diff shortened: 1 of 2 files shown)


gschueler:
@ltamaster can you add docs to @rundeck/docs regarding the rundeck.security.jaasRolePrefix setting?

PR #3900, opened 2018-08-23
Object storage capabilities

sjrd218:
Add a tree implementation that stores data in an amazon s3 compatible object store.
Add a storage plugin that uses the object storage tree to store the data.


sjrd218:
Restructured object store as a plugin. No need for this.

PR #3979, opened 2018-09-05
Use grails asset sass plugin

ProTip:
**Note**
Even if this looks good I would suggest we hold off until the selenium image diff stuff up and running so we can detect visual regressions between this and master more easily before merging.

**Enhancement.**
Moves the `scss` compilation into the grails build pipeline and cleans up the `nodejs` plumbing from `/assets`

* Utilizes the Grails `sass-asset-pipeline` plugin
* Uses `app.css` as the reference as the plugin will not convert `app.scss.scss` to `app.scss.css`
* Renames `app.less` to `dead.less` to prevent conflict


gschueler:
can we do a css diff somehow on the generated css from before vs. after? perhaps that would be sufficient to determine if the result is equivalent

ProTip:
On of the best approaches I've seen for that is to compile the CSS into an AST and then do a comparison from there. It's quite involved and I don't believe I'd recommend the approach here necessarily, but there are some tools to help out with it these days.

https://codeascraft.com/2015/02/02/transitioning-to-scss-at-scale/

ProTip:
With the selenium image regression in I have some level of confidence that this isn't changing the looking of the app. Going to merge it in 😄

Issue #2877, opened 2017-10-23
Hibernate errors on running jobs in latest RunDeck


dsteffa:
Issue type: Bug report

**My Rundeck detail**

* Rundeck version: 2.9.4 / 2.10.0
* install type: rpm
* OS Name/version: centos-release-6-9.el6.12.3.x86_64
* DB Type/version: mysql

**Expected Behavior**

If I run a job against multiple target nodes it should dispatch to all nodes and run all the steps.

**Actual Behavior**

When I run the job who run other job(s), each of them can run further job(s), against multiple target nodes the job fails with error:

> Failed dispatching to node DVW-WEB-MULTI08: org.springframework.orm.hibernate4.HibernateOptimisticLockingFailureException: Object of class [rundeck.JobExec] with identifier [616]: optimistic locking failed; nested exception is org.hibernate.StaleObjectStateException: Row was updated or deleted by another transaction (or unsaved-value mapping was incorrect) : [rundeck.JobExec#616]

Those jobs were running fine for a long period of time, up to 2.9.4 version upgrade, and I haven't seen this type of error before. I've got the same errors running the same jobs using fresh RunDeck server.

**How to reproduce Behavior**

The jobs are failing from time to time, sometimes they just run fine. I didn't find a way to create a job that would always return with an error. 

If you need more information please let me know and I'll do my best to dig it up for you.

Thanks!

ajxb:
I've been able to reproduce this problem with the logs set to debug.  Unfortunately it's sporadic across our instances of RunDeck, so I can't provide exact steps to reproduce.  It seems to happen when the jobs in question haven't been used for a while.  In the cases I've seen of this issue, one node runs to completion, all of the others fail.  Output is as follows:

```
[...]

PR #3985, opened 2018-09-05
fix jobs failing over lock exception on mysql and mariadb

jtobard:
Backport for #3875 version 2.11.x


puremourning:
does this resolve https://github.com/rundeck/rundeck/issues/2877 ?

PR #4005, opened 2018-09-12
user filterPref to session

jtobard:
`user.filterPref` to `session.filterPref` on login and refresh on `UserController.addFilterPref` method.


rundeckapp/grails-app/controllers/rundeck/controllers/UserController.groovy
```diff
@@ -743,6 +743,7 @@ class UserController extends ControllerBase{
             if(result.error){
                 return renderErrorFragment(result.error)
             }
+            session.filterPref=result.user?.filterPref
```
gschueler: I'd like to have the session.filterPref be the Map data, so you should use `UserService.parseKeyValuePref`  that way the gsp can use things like `${session.filterPref.mykey}`

PR #4008, opened 2018-09-13
Add @DirtyCheck annotation to ScheduledExecution superclasses

sjrd218:
Recent versions of Gorm change the way dirty checks work on domain class superclasses. You either need to use setter methods or annotate the super class with @DirtyCheck.

https://github.com/grails/grails-data-mapping/issues/1097


gschueler:
good find

PR #4035, opened 2018-09-18
Fix: LDAP user credentials should not be exposed

gschueler:

Fix: user LDAP credentials potentially exposed in server log.

Don't log user credentials.

**Additional context**

User credentials may be exposed in the rundeck server log in these conditions:

* using JettyCachingLdapLoginModule or JettyCombinedLdapLoginModule
* using `forceBindingLogin="false"` in the jaas.conf

If user credentials in the LDAP server are stored unencrypted, then the server log may print them in plainttext.

Workarounds:

* use `forceBindingLogin="true"` in jaas conf for the JettyCachingLdapLoginModule


gschueler:
Another workaround when forceBindingLogin="false":

Use `-Dcom.dtolabs.rundeck.jetty.jaas.LEVEL=ERROR` on the JVM args

PR #4046, opened 2018-09-21
Activity tab default to 1d configurable

jtobard:
Enhancement:

In projects with too many executions, the current behavior of the activity page is suboptimal, the default to show executions of all times brings too many info and it could be unnecessary.

To help with this, now if you enter to the activity page using the menu link, it will load only 1 day of info by default. 
This can be configured by modifying project properties or framework properties.

You can configure a global on `framework.properties` and then override for only one project on that `project.properties`

framework.properties:
```
rundeck.activityDefault=all
``` 

project.properties:
```
project.activityDefault=2d
``` 
The value can be any number + the letter for hours `h`, days `d`, weeks `w`, months `m` or years `y`.

If you want to preserve executions for all time you can use `all`


rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
```diff
@@ -104,12 +105,25 @@ class ScheduledExecutionService implements ApplicationContextAware, Initializing
                 defaultValue null
                 renderingOption('booleanTrueDisplayValueClass', 'text-warning')
             }.build(),
+            PropertyBuilder.builder().with {
+                string 'activityDefault'
+                title 'Activity Default'
```
gschueler: I think the title should be more descriptive, like "Activity Page Default Time Period"

gschueler:
@jtobard is this still an issue in latest 3.1.0 snapshot? the activity page is updated and uses 1d filter by default.. it could be made configurable however

jtobard:
these changes are no longer working with the new GUI so I'm going to make a new PR from scratch to cover this particular enhancement

PR #4103, opened 2018-10-09
Referenced Job successOnEmptyNodeFilter

### LINKED ISSUE #4077: "continue on empty node set" doesn't work in a referenced job
**Describe the bug**
A referenced job ignores the "continue" setting for an empty nodeset

**My Rundeck detail**

* Rundeck version: 2.11.4
* install type: n/a
* OS Name/version: n/a
* DB Type/version: n/a

**To Reproduce**
1. Create job A
2. Configure job A to continue if the nodeset is empty, rather than fail
3. Create job B
4. Add a step to B which references A, and filters the passed nodeset further
5. Execute job B against a set of nodes which will all be filtered out in the reference to A, leaving an empty node set

**Expected behavior**
The reference step in job B to job A should succeed, and do nothing

**Screenshots**
none available

n/a

**Additional context**
n/a

jtobard:
Referenced Job respect `successOnEmptyNodeFilter` when a set of empty nodes is passed.

Fix #4077


jtobard:
If there's a new release it could be included.

Issue #3459, opened 2018-06-01
Slack Notification Plugin required properties not allowing project/framework defaults


cwaltherf:
**Describe the bug**
The plugin is not taking the global properties from framework.properties and project.properties

https://github.com/rundeck-plugins/slack-incoming-webhook-plugin

**My Rundeck detail**

* Rundeck version:2.11.3
* install type: launcher
* OS Name/version: ubuntu 16.04
* DB Type/version: H2

**To Reproduce**
1.Set Slack Webhook URL and Token in framework.properties and project.properties files as by plugin description

WebHook Base URL:
Slack Incoming WebHook Base URL
configure project: project.plugin.Notification.SlackNotification.webhook_base_url=https://hooks.slack.com/services
configure framework: framework.plugin.Notification.SlackNotification.webhook_base_url=https://hooks.slack.com/services
Default value: https://hooks.slack.com/services
WebHook Token:
WebHook Token, like T00000000/B00000000/XXXXXXXXXXXXXXXXXXXXXXXX
configure project: project.plugin.Notification.SlackNotification.webhook_token=••••••••
configure framework: framework.plugin.Notification.SlackNotification.webhook_token=••••••••

2.Start Rundeck
3.create a job and set Slack Webhook Notifications

**Expected behavior**
While creating a job after selecting Slack Webhook notification the Webhook URL and Token fields should be populated with the values set in framework.properties and or project.properties but they aren't. The user is obliged to set it again for each job.

**Desktop (please complete the following information):**
 - Ubuntu 16.04
 - Browser Firefox
 - Version 60.0.1 (64-bit)

PR #4117, opened 2018-10-12
Notification Plugin required properties not allowing project/framework defaults

ltamaster:
For notification plugins, fix the issue when the plugin properties are defined at project or framework level. 
For https://github.com/rundeck/rundeck/issues/3459


rundeckapp/grails-app/services/rundeck/services/NotificationService.groovy
```diff
@@ -392,10 +392,17 @@ public class NotificationService implements ApplicationContextAware{
                     execMap.job=jobMap
                     execMap.context=context
                     Map config= n.configuration
+                    println(config.toString())
```
gschueler: remove debugging output

ltamaster: done

PR #4121, opened 2018-10-12
fix: grails-spa build, don't include extraneous assets

gschueler:

the rundeckapp/grails-spa build was including extraneous files (jars) in the assets dir

Update node based gradle build for grails-spa to use gradle base plugin, define archives, and update the rundeck app grails build to only copy the correct assets.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -118,7 +118,6 @@ allprojects {
 subprojects {
     // set the eclipse project naming convention to rundeck:<path>:<projectName>
     // so it matches the logical hierarchy more closely
-    apply from: "${rootDir}/gradle/java.gradle"
     apply plugin: 'eclipse'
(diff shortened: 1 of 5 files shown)


ProTip:
@gschueler I just noticed these artifacts popping up when I re-based and tested a different branch. Do you know if these are getting built as part of a different, recent change to the build scripts?

gschueler:
@ProTip i'm not sure when it started.  I think it is happening because of the gradle/java.gradle file which is applied to all subprojects in the root build.gradle, causing extra jars to be created in rundeckapp/grails-spa build dir.

gschueler:
ok i think i fixed the gradle issues... now the deb install test is failing because it can't install oracle jdk 8 😢

ProTip:
Looks like possibly a transient error. Saw a similar error last time oracle/webupd8team updated the package..

PR #4140, opened 2018-10-19
Use sync tokens to allow API access from web pages

sjrd218:


rundeckapp/grails-app/controllers/rundeck/interceptors/ApiVersionInterceptor.groovy
```diff
@@ -60,6 +64,16 @@ class ApiVersionInterceptor {
     }
 
     boolean before() {
+        if(params[SynchronizerTokensHolder.TOKEN_KEY]) {
+            tokenVerifierController.withForm {
+                tokenVerifierController.refreshTokens()
+                session.api_access_allowed = true
```
gschueler: i think we should change this so that the api_access_allowed is set only for the request, not the session.  otherwise this will persist in the session, right?

PR #4143, opened 2018-10-20
Add env variables for preauthenticated mode

### LINKED ISSUE #4142: Can't use preauthenticated mode while running Docker images.
Since the entry point overrides the file `rundeck-config.properties` it's not useful to mount a custom file, and there're no mapped env variables for configuring this feature.

Also, the documentation is wrong, since version 3.0, editing the file `WEB-INF/web.xml` inside the `rundeck.war` is not needed, and the file does not exist anymore, as explained on the [changelog for version 3.0.0](https://github.com/rundeck/rundeck/blob/master/CHANGELOG.md#release-300)

I have created a PR on the documentation, to remove the part of "modifying the web.xml file" [here](https://github.com/rundeck/docs/pull/137)

gargrag:
In order to use the official container image with an oauth2 proxy
Or any other authentication layer, we need to add this environment
variables to the rundeck-config.properties template, since the
entrypoint of the image overwrites the rundeck-config.properties if
mounted. Also that's for adopting a better practice.

fixes  #4142

### DIFF
--- a/docker/official/README.md
+++ b/docker/official/README.md
@@ -118,3 +118,14 @@ Set to anything enables audit logging. This can be very verbose so use with caut
 
 Configuration options for key storage providers and converts. These map to the
 [Storage Facility Docs](http://rundeck.org/docs/administration/configuration/storage-facility.html).
+
+### `RUNDECK_PREAUTH_ENABLED=false`
+### `RUNDECK_PREAUTH_ATTRIBUTE_NAME=REMOTE_USER_GROUPS`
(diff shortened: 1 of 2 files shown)


ProTip:
Hi @gargrag , thanks for this PR!

This looks really good but I am concerned that the defaults are too implementation specific. What would you think about putting the whole section under a conditional that `rundeck/preauth/enabled == true` and then requiring the rest of the configuration keys be set?

ProTip:
Ah, I had not noticed those defaults. LGTM 👍

PR #4151, opened 2018-10-25
add env for LDAP auth in docker  and fix Misspelled words 

gudeyang1:
for non-docker configuration document https://rundeck.org/docs/administration/security/authenticating-users.html#login-module-configuration , there's`roleUsernameMemberAttribute="memberUid"` in `Login module configuration`, but not found in docker. it's needed to map LDAP group to rundeck roles.

fix the Misspelled words of `module`

### DIFF
--- a/docker/official/remco/templates/jaas-loginmodule.conf
+++ b/docker/official/remco/templates/jaas-loginmodule.conf
@@ -22,7 +22,9 @@
         rolePrefix="{{ getv("/rundeck/jaas/ldap/rolePrefix", "") }}"
         cacheDurationMillis="{{ getv("/rundeck/jaas/ldap/cachedurationmillis", "600000") }}"
         reportStatistics="{{ getv("/rundeck/jaas/ldap/reportstatistics", "true") }}"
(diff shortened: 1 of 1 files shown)


docker/official/remco/templates/jaas-loginmodule.conf
```diff
@@ -22,6 +22,7 @@
         rolePrefix="{{ getv("/rundeck/jaas/ldap/rolePrefix", "") }}"
         cacheDurationMillis="{{ getv("/rundeck/jaas/ldap/cachedurationmillis", "600000") }}"
         reportStatistics="{{ getv("/rundeck/jaas/ldap/reportstatistics", "true") }}"
+	roleUsernameMemberAttribute="{{ getv("/rundeck/jaas/ldap/rolessernamememberattribute", "memberUid") }}"
```
ProTip: We may need to wrap this in an `if exists` as the documentation states it overrides `roleMemberAttribute`
```suggestion
    {% if exists("/rundeck/jaas/ldap/roleusernamememberattribute") -%}
        roleUsernameMemberAttribute="{{ getv("/rundeck/jaas/ldap/roleusernamememberattribute") }}"
    {% endif %}
```

ProTip:
@gudeyang1 Great, thanks for the PR!

Issue #3961, opened 2018-08-30
NotificationPlugin not receiving correct state for Job-Reference steps on fail or success


elijah-ward:
**Describe the bug**
Seemingly related to Issue #3791. While using a Groovy Slack notification plugin: https://github.com/bdclark/rundeck-slack-notification, notifications for succeeded or failed job-reference step executions have a status of "Started" even though they have completed.

**My Rundeck detail**

* Rundeck version: Rundeck 3.0.5-20180828
* install type: deb
* OS Name/version: ubuntu 16.04

**To Reproduce**
1. Install the https://github.com/bdclark/rundeck-slack-notification plugin
2. Configure a webhook to receive slack messages from the plugin
3. Run a job which contains steps that are references to other jobs
4. See "Started" slack messages in the channel after the job reference steps have completed and message should read "Failed" or "Succeeded".

**Expected behavior**
After the change related to #3791 I would expect to see a success/failed status for each job reference step and not just once the parent job has completed.

**Screenshots**
I modified the plugin Groovy script to print the execution state that the plugin receives during an `onfailure` event to be the following:
```
  onfailure { Map executionData, Map configuration ->
    System.out.println("ON FAIL EXECUTION MAP: "+executionData)
      triggerMessage(executionData, configuration, 'danger')
      true
  }
```

See in this screenshot that the `onfailure` event is received but the status within the execution details passed into the plugin lists the state as `running`

![screen shot 2018-08-30 at 1 51 15 pm](https://user-images.githubusercontent.com/17422956/44872916-c4026b00-ac64-11e8-86b8-05c2af62755a.png)

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: Chrome

vighter:
Hello,
The same issue appears in:

- Rundeck version: Rundeck 3.0.6-20180917
- install type: rpm
- OS Name/version: CentOS 7.5.1804

unicolet:
I can confirm I am seeing the same behavior with https://github.com/higanworks/rundeck-slack-incoming-webhook-plugin

Besided the wrong status, it appears `dateEnded`, `failedNodeListString` or `succeededNodeListString` are missing as well. This used to work with RD 2.x. Thanks for looking into this.

p.s.: I've tried looking into the code but it's quite a lot to process :-)

driseley:
[...]

PR #4152, opened 2018-10-25
Remove refresh of Execution state from database on Notification rundeck/rundeck#3961

driseley:
Rundeck issue rundeck/rundeck#3961 shows an out of date execution status being sent to Notification plugins.  I believe that this is due to the call to `Execution.get(e.id)` which refreshes the execution state from the database, however I think the end state of the execution hasn't yet been committed to the database, so the previous running state is fetched.

This passes all tests , and in my local testing addresses the issue - although I can't be sure why this was originally added.


driseley:
@gschueler  - sorry no I haven't.  I did debug the execution state after the change and it all looks correct to me - but I'm not sure what to look for that could be used by the email templates that might break ( but the notifications are pretty broken anyway at the moment as it sends the incorrect status )

PR #4192, opened 2018-11-09
rd3: Add validation to check if a key exists on secure options

ltamaster:
When a key is called form a secure option, a "check exists" validation is called before the key is read.
The job will fail if the secure option is required. Otherwise, a warning message will be print on logs.


rundeckapp/src/main/groovy/com/dtolabs/rundeck/server/plugins/storage/KeyStorageTreeImpl.groovy
```diff
@@ -64,6 +64,22 @@ class KeyStorageTreeImpl extends TypedStorageTreeImpl implements KeyStorageTree
         readPrivateKey(PathUtil.asPath(path))
     }
 
+    @Override
+    boolean hasPassword(String path) {
+        println("checking password path ${path}")
```
gschueler:
```suggestion
```

PR #4193, opened 2018-11-09
Job state conditional plugin throws an exception on failure

### LINKED ISSUE #4178: job state conditional is not working since rundeck 2.9.x
**Describe the bug**
Job state conditional is not working as expected.
Last working version was rundeck-launcher-2.8.4.jar

**My Rundeck detail**

* Rundeck version: since rundeck-launcher-2.9.1.jar
* install type: launcher
* DB Type/version: h2

**To Reproduce**
See screenshots.

**Expected behavior**
Not keep running next steps after the job conditional fails to check condition.

**Screenshots**
JOB STEP CONDITION CONFIG:
captura de pantalla de 2018-11-05 13-07-07

ls script command should not be run:
captura de pantalla de 2018-11-05 13-01-30

jtobard:
Fix #4178 

if the `fail` condition is checked, and the status doesn't match. An error is thrown and the flow is interrupted.

`Halt`and `status` custom message properties not affected.

### DIFF
--- a/plugins/job-state-plugin/src/main/java/org/rundeck/plugin/jobstate/JobStateFailureReason.java
+++ b/plugins/job-state-plugin/src/main/java/org/rundeck/plugin/jobstate/JobStateFailureReason.java
@@ -1,0 +1,10 @@
+package org.rundeck.plugin.jobstate;
+
+import com.dtolabs.rundeck.core.execution.workflow.steps.FailureReason;
+
+public enum JobStateFailureReason implements FailureReason {
+    /**
+     * State not match and failed marked true
+     */
+    ExecutionStateNotMatch
+}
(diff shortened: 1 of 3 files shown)


gschueler:
I don't think this is the correct change.  The Job State Conditional has two flags: `halt` and `fail`.  If `halt` is `false`, it won't halt.  If `halt` is `true` it *will* halt the workflow (using flow control HALT message).  When it halts, `fail=true` means it fails, and `fail=false` means it succeeds. This has always been the behavior of this plugin.

From the #4178 it seems they are not using `halt`.

gschueler:
@jtobard I think the problem is happening in BaseWorkflowExecutor, and is more fundamental than I expected. I will send a PR

gschueler:
created https://github.com/rundeck/rundeck/pull/4235

PR #4204, opened 2018-11-13
Fix #3413. Bump jetty version to 9.4.13.v20181111 for java 9+ support.

### LINKED ISSUE #3413: JDK10 support
Enable Rundeck to work on JDK 10

- [ ] jaxb dependency needs to be included explicitly
- [ ] grails mail plugin issue: https://github.com/grails3-plugins/mail/issues/31 
- [ ] Do not pass null as the credential to UserInfo in the JettyCachingLdapLoginModule class after binding login. A null value causes a stacktrace in JDK 10.

SpencerMalone:
fixes #3413

@gschueler did great work for 2/3rds of that issue, but I recently wanted to try out RD with java 11, as i'm having some perf issues, and decided to deep dive on this. Turns out, Jetty realized this was a breaking change in https://github.com/eclipse/jetty.project/commit/83b2abebb7cdc4411ffd7a285ee968041f0e77d5, and un-broke the change. All I did was update, ran a build, watched the tests, scp'd to my stage instance of rundeck, and verified LDAP worked for me while on java 11

Nil. Could have figured out a workaround for this, poked around at that, it seemed more involved that this.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -25,6 +25,6 @@ gormVersion=6.1.10.RELEASE
 hibernateVersion=5.1.13.Final
 gradleWrapperVersion=3.5
 groovyVersion = 2.4.15
-jettyVersion=9.4.11.v20180605
+jettyVersion=9.4.13.v20181111
 mavenCentralUrl = http://repo1.maven.org/maven2/
 grailsCentralUrl = http://grails.org/plugins
--- a/rundeckapp/build.gradle
+++ b/rundeckapp/build.gradle
@@ -130,6 +130,9 @@ dependencies {
     compile "org.eclipse.jetty:jetty-jaas:${jettyVersion}"
     compile "org.eclipse.jetty:jetty-util:${jettyVersion}"
     compile "org.eclipse.jetty:jetty-security:${jettyVersion}"
+    compile "org.eclipse.jetty:jetty-server:${jettyVersion}"
+    compile "org.eclipse.jetty:jetty-http:${jettyVersion}"
+    compile "org.eclipse.jetty:jetty-io:${jettyVersion}"
     compile 'org.grails.plugins:spring-security-core:3.2.1'
     compile 'org.kohsuke:libpam4j:1.10'
     compile 'ca.juliusdavies:not-yet-commons-ssl:0.3.17'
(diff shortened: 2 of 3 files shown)


SpencerMalone:
Hrm, having trouble replicating what this failing test is looking for. Lemme dig into the build types?

SpencerMalone:
Ah, I had missed that it was peeking into the war itself, lemme poke around there.

SpencerMalone:
Looks like something is causing unmatched jetty package versions, I can dig into those more tomorrow, initial glance seems maybe it's caused by `sprint-boot-starter-jelly` have preset versions, and rundeck not specifying specific versions in the `rundeckapp/build.gradle`?

```
|    +--- org.springframework.boot:spring-boot-starter-jetty -> 1.5.15.RELEASE
|    |    +--- org.eclipse.jetty:jetty-servlets:9.4.11.v20180605
|    |    |    +--- org.eclipse.jetty:jetty-continuation:9.4.11.v20180605
|    |    |    +--- org.eclipse.jetty:jetty-http:9.4.11.v20180605 (*)
|    |    |    +--- org.eclipse.jetty:jetty-util:9.4.11.v20180605 -> 9.4.13.v20181111
|    |    |    \--- org.eclipse.jetty:jetty-io:9.4.11.v20180605 (*)
````

from `./gradlew rundeck:dependencies`, which is also what I see what I unpack the war.

Would you prefer I explicitly set some of the packages we are checking the versions for in `testbuild.groovy`, such as `WEB-INF/lib-provided/jetty-http-${versions.jetty}.jar`, or change the checks to support two sets of jetty versions, one from rundeck, one from `spring-boot-starter-jetty`?

SpencerMalone:
In leui of input, I explicitly pulled in the implicit jetty packages we were version checking for. Willing to change that, if we want?

sjrd218:
After looking into this issue further, it looks like we're going to need to wait on Grails to update their dependency before we can update the underlying Jetty dependency. If the Jetty dependency is updated as you've done it introduces competing versions of Jetty into the classloader which almost invariably causes problems. I also tried excluding the Jetty dependencies that are required by spring boot and adding them manually with the more recent Jetty versions, but that broke other things.

Really appreciate your work on this, but I think we'll have to hold off on it for now.

SpencerMalone:
Bummer, oh well! Thank you for digging in, excited for grails to update. My testing on Java 11 made it seem 💯!

PR #4211, opened 2018-11-15
Implementing job that allow to clean old executions

carlosrfranco:
This job will be executed according to what is configured in the project settings.


carlosrfranco:
1 - List of executions:
selecao_007

2 - Project settings to enable history cleaner:
selecao_003

3 - History cleaner settings:
 property 1 -> Maximum of days to keep the executions
 property 2 -> When the cleaner will be executed. Is needed to make sure it will not generate a performance issue
selecao_004

4 - Project config file:
selecao_010

5 - List of the executions cleaned:
selecao_011

carlosrfranco:
Log output when cleaner executions history is executed:

selecao_012

selecao_014

miguelantonio:
final review

miguelantonio:
Hi @gschueler just following up on this merge, is there anything you'll like to change before merging?

carlosrfranco:
Is needed to set a feature `rundeck.feature.cleanExecutionsHistoryJob.enabled = true`

rundeckapp/grails-app/conf/application.groovy
```diff
@@ -36,6 +36,7 @@ environments {
         application.refreshDelay=5000
         grails.profiler.disable=false
         feature.incubator.'*'=true
+        rundeck.feature.cleanExecutionsHistoryJob.enabled = true
```
jtobard: This feature should be disabled by default by now.

rundeckapp/grails-app/init/rundeckapp/BootStrap.groovy
```diff
@@ -324,6 +324,13 @@ class BootStrap {
          if(!maxLastLines || !(maxLastLines instanceof Integer) || maxLastLines < 1){
              grailsApplication.config.rundeck.gui.execution.tail.lines.max = 500
          }
+         if(grailsApplication.config.rundeck.feature.cleanExecutionsHistoryJob.enabled){
+             println("cleanExecutionsHistoryJob ${grailsApplication.config.rundeck.feature.cleanExecutionsHistoryJob.enabled}")
```
jtobard: this `println` should be a `log.debug` or `log.info`

ahonor:
* All fields should be required so no blank fields accepted.
* Default days to keep executions: 60
*  Minimum executions to keep: 50

image

lukasged:
thanks for this feature 👍 

I was testing it in Rundeck 3.1 and it works very fast: only takes some seconds to delete 1K jobs and without requiring too much CPU.

But when I check the folder with the logs, I see that the `*.rdlog` and `*.state.json` files are cleaned up correctly, but the `*.execution.xml` are all still there (I had around 3K jobs and after cleanup, there are 3K of these xml files). 

Is this the expected behaviour?

Issue #4058, opened 2018-09-25
Allow GIT import from SCM without "update" access in the ACL


mlamutt:
**Describe the bug**
First off, I am not sure if this is a bug, or if I am trying to do something that is not allowed.  We have a SCM process where we have a DEV rundeck server where a group is allowed to develop new rundeck jobs.  Once the job is built and tested, the jobs are commit to GIT. once commit to master, they are then Git Exported to a Production rundeck server to be imported.  We want jobs on this server to not be editable, but only allow for importing tested jobs through the SCM process.   I have not been able to find an ACL combination that allows me to "import" a job without also having the "update" access also allowed.  

**My Rundeck detail**

* Rundeck version: [ e.g. 2.11.3]
* install type: [launcher]
* OS Name/version: [Windows 2012 R2]
* DB Type/version: [choose: internal ]

**To Reproduce**

**Expected behavior**
Ability to import a job, when the "Import needed" flag is set without having "Update" access:
![image](https://user-images.githubusercontent.com/30131225/46031744-00d04f00-c0b7-11e8-95bf-23bbfb212254.png)
results in the errors:
Git Import failed
Failed importing:  jobpath.xml: Unauthorized: Update Job 135

When I add the updated access it works: 
![image](https://user-images.githubusercontent.com/30131225/46031678-ca92cf80-c0b6-11e8-8635-fe35d368ec78.png)

ACL:
![image](https://user-images.githubusercontent.com/30131225/46031973-a1bf0a00-c0b7-11e8-9c68-b973784f8f28.png)

Adding "update" to the job -allow works. 

Thanks

gschueler:
we should add a new ACL action like "scm_updated" perhaps, that allows update via SCM plugins, but not via normal edit procedure

PR #4220, opened 2018-11-16
scm_update acl special permission

jtobard:
Implements #4058 
A user with `scm_update` job permission can import SCM changes over a particular job.
A user with `update` permission still can edit or import scm changes normally.

acl example:
```
context:
  project: '.*'
for:
  job:
    -allow: [read,view,kill,scm_update]
```

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/cli/acl/AclTool.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/cli/acl/AclTool.java
@@ -974,7 +974,8 @@ public class AclTool extends BaseTool {
                     ACLConstants.ACTION_KILLAS,
                     ACLConstants.ACTION_CREATE,
                     ACLConstants.ACTION_TOGGLE_EXECUTION,
-                    ACLConstants.ACTION_TOGGLE_SCHEDULE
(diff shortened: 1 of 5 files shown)


gschueler:
1) In this code the check for `scm_update` applies any time a "loadJobs" is called.  This can happen via SCM import, file upload (GUI) or API import as well.  This should be changed so "loadJobs"  only allows `scm_update` action in the case of SCM import.  You can use `changeinfo['method']=='scm-import'` to check for this case.  See ScmJobImporter.groovy: this is what will call the loadJobs method, and passes the changeinfo with that value.

2) I think we need to also add `scm_create` and `scm_delete`, for jobs created or deleted via SCM import plugin. (This should also only be allowed  when the `changeinfo['method']=='scm-import'` is true.)

However for job `create` auth check, there are actually two checks.  One where resource kind=='job', and one for the specific job.  The auth check in `loadJobs` method checks only the first one.  Then in the `_dosave` method it checks the second one.  we should support `scm_create` in both cases.

gschueler:
changes i made: 

* check `scm_delete` / `scm_create` are allowed for resource kind `job` as alternative to `delete` and `create`

PR #4238, opened 2018-11-21
Adding templates for canonical roles for ACLs and users

ltamaster:
adding standard roles:
- job_runner
- job_writer
- job_reader
- job_viewer
- project_admin

and standard users:
- job-runner
- job-writer
- job-reader
- job-viewer
- project-admin


gschueler:
- [ ] instead of default policy files, they should become default policy templates, by adding `.template` to filename

ProTip:
@ltamaster @gschueler Is there a way to remove these if the user doesn't want them? For instance, in an auto-deployed scenario(Docker or otherwise) these would get added back to the file-system every time?

Packaging changes need to be made to: https://github.com/rundeck/packaging . This changes to this repo would not have carried over to enterprise in the past either.

gschueler:
we should also delete the "apitoken.aclpolicy" file

PR #4243, opened 2018-11-26
Disable toggle off of first run message

ltamaster:
Adding a config setting that prevents the first run message from being hidden so it remains visible.

If this setting is adding on rundeck-config, then the first run message is not hidden:

```
rundeck.startup.alwaysFirstRun=true
```


gschueler:
@ltamaster FYI i modified the config prop name. it is now `rundeck.startup.alwaysFirstRun=true`

ltamaster:
got it @gschueler

PR #4282, opened 2018-12-04
add messages_ja.properties

s-tokutake:
- add messages_ja.properties


sjrd218:
@s-tokutake Thanks for your contribution!
Have you run across other places in the app that need 
translation that are not internationalized? If so, please let us know
so we can update those places.

Is there anything we can do to make Rundeck better for you?

We would love to send you some gear as a thank you for your commit.
If you contact aziz@rundeck.com he will make sure some Rundeck gear heads your way.

PR #4295, opened 2018-12-09
Added retry logic around git lock exceptions in clusterFixJobs

### LINKED ISSUE #4281: Semi-regular JGitInternalException LockFailedException from /api/21/project/<project>/jobs endpoint
**Describe the bug**
We have a slack bot service that regularly populates it's model of what our rundeck jobs look like through the API. That service regularly (~20-50 times per day? It doesn't sync that often though, this code path is on a spaced out cron) gets thrown 500s from rundeck when getting a project overview through the API (`/api/21/project/<project-name>/jobs`). When digging through the logs, the error was correlated with the following: https://gist.github.com/SpencerMalone/b21f649c5275650535292d6844bb84df

Summarized, that is...
```
org.eclipse.jgit.api.errors.JGitInternalException: Cannot lock /var/rundeck/projects/<redacted>/scm/.git/index
	at org.eclipse.jgit.api.CheckoutCommand.call(CheckoutCommand.java:316)
	at org.eclipse.jgit.api.CheckoutCommand.call(CheckoutCommand.java:126)
	at org.rundeck.plugin.scm.git.GitExportPlugin$_clusterFixJobs_closure2.doCall(GitExportPlugin.groovy:529)
	at org.rundeck.plugin.scm.git.GitExportPlugin.clusterFixJobs(GitExportPlugin.groovy:519)
	at rundeck.services.ScmService.fixExportStatus(ScmService.groovy:1332)
	at rundeck.services.ScmService.exportStatusForJobs(ScmService.groovy:1069)
```

**My Rundeck detail**

* Rundeck version: Rundeck 3.0.7-20181008
* install type: rpm, clustered setup behind nginx LB
* OS Name/version: Centos 7.3
* DB Type/version: mysql

**To Reproduce**
Not sure yet? I assume when two users (or services) are interacting with the SCM at the same time?

**Expected behavior**
A 200 should return with project info. It seems like `exportStatusForJobs` is trying to "fix" the export status, but I don't entirely understand the purpose of that, or if we can try/catch that failure?

**Additional context**
I'll be poking at this further to try and figure out what's up

SpencerMalone:
This is a bugfix, which fixes #4281 

The original exception was...
```
org.eclipse.jgit.api.errors.JGitInternalException: Cannot lock /var/rundeck/projects/<redacted>/scm/.git/index
	at org.eclipse.jgit.api.CheckoutCommand.call(CheckoutCommand.java:316)
	at org.eclipse.jgit.api.CheckoutCommand.call(CheckoutCommand.java:126)
	at org.rundeck.plugin.scm.git.GitExportPlugin$_clusterFixJobs_closure2.doCall(GitExportPlugin.groovy:529)
	at org.rundeck.plugin.scm.git.GitExportPlugin.clusterFixJobs(GitExportPlugin.groovy:519)
	at rundeck.services.ScmService.fixExportStatus(ScmService.groovy:1332)
	at rundeck.services.ScmService.exportStatusForJobs(ScmService.groovy:1069)
```

Both the rm here, and the checkout can run into issues when multiple requests to the project are happening at once in a clustered environment, as they all end up calling `exportStatusForJobs`, which calls `fixExportStatus`, which calls this, and jgit will throw lock errors if you try to run stuff like checkout concurrently in a single repo on disk. In the end, I wrapped the jgit logic here in a while try/catch. Not the most elegant direction, but I'm not sure the best way to do this without rearchitecting the scm plugin to maintain backup repos on disk for this edge case, which sounds like a mess.

In the end I wrapped more logic than just the original `checkout().addPath().call` logic after further testing. I managed to poke my instance enough to get a project to throw the same error from the `git.rm()` code path, so I figured that wrapping this around both of those would be safe, especially with such a targeted `catch`

Note that the exception being caught is described as...

> During command execution a lot of exceptions may be thrown. Some of them represent error situations which can be handled specifically by the caller of the command. But a lot of exceptions are so low-level that is is unlikely that the caller of the command can handle them effectively. The huge number of these low-level exceptions which are thrown by the commands lead to a complicated and wide interface of the commands. Callers of the API have to deal with a lot of exceptions they don't understand.
> 
> To overcome this situation this class was introduced. Commands will wrap all exceptions they declare as low-level in their context into an instance of this class. Callers of the commands have to deal with one type of low-level exceptions.
- (source) http://download.eclipse.org/jgit/site/5.1.3.201810200350-r/apidocs/index.html 

I figure retrying other stuff thrown from jgit here would be safe, if they expect these to be so low level that we shouldn't be trying to deal with the individual exceptions? I can get the exception cause if we wanna limit it to a `LockFailedException`. Let me know if that's desired

The other alternative I thought of is either...

1. Letting it fail as is. This "works", but is frustrating for users who want to rely on this, since this was just cropping up on an GET for the project from the API, and there's no way to avoid this code branch in a clustered configuration.
2. Try/catching the jgit lock exception and moving on from `exportStatusForJobs`. I kind of like this approach, but it's a tradeoff of giving you potentially out of date data, but with a lower maximum request duration. I'm down to go down this path if desired, but figured that a fix which still ensures the current expectation is best (that SCM data on a clustered environment will always be up to date)

**Additional context**
I struggled with fitting unit tests in here, but actually found that it hasn't gone past the first loop in my testing. This will add a maximum increase of 2s to a request which hits this code path, which seems reasonable. I've never had it actually get up to that so far, it always gets the lock after one loop, but I imagine a bigger instance might run into this?

### DIFF
--- a/plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/GitExportPlugin.groovy
+++ b/plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/GitExportPlugin.groovy
@@ -520,15 +520,29 @@ class GitExportPlugin extends BaseGitPlugin implements ScmExportPlugin {
             def storedCommitId = ((JobScmReference)job).scmImportMetadata?.commitId
             def commitId = lastCommitForPath(getRelativePathForJob(job))
             def path = getRelativePathForJob(job)
-            if(storedCommitId != null && commitId == null){
-                //file to delete-pull
-                git.rm().addFilepattern(path).call()
-                toPull = true
-                retSt.deleted.add(path)
-            }else if(storedCommitId != null && commitId?.name != storedCommitId){
-                git.checkout().addPath(path).call()
-                toPull = true
-                retSt.restored.add(job)
+            def gitAttempts = 0
+            while(gitAttempts < 10){
+                gitAttempts++
+                try{
+                    if(storedCommitId != null && commitId == null){
+                        //file to delete-pull
+                        git.rm().addFilepattern(path).call()
+                        toPull = true
+                        retSt.deleted.add(path)
+                    }else if(storedCommitId != null && commitId?.name != storedCommitId){
+                        git.checkout().addPath(path).call()
+                        toPull = true
+                        retSt.restored.add(job)
+                    }
+                    break
+                }catch (org.eclipse.jgit.api.errors.JGitInternalException e){
+                    if(gitAttempts < 10){
+                        log.warn("Failed to checkout due to jgit error, on attempt ${gitAttempts}: ${e.message}")
+                        sleep(200)
+                    }else{
+                        throw e
+                    }
+                }
             }
         }
         Status status = git.status().call()


SpencerMalone:
Bleh, closing for now, after poking around, these lock errors are cropping up all over the place. Frustrating. May need a wrapper class to check for this problem.

SpencerMalone:
Just to document, here is the other issue I found, from the import git scm:
```
2018-12-10 15:44:58.278 ERROR --- [168947686-44870] o.g.web.errors.GrailsExceptionResolver   : LockFailedException occurred when processing request: [GET] /api/21/project/jenkins/jobs
Cannot lock /var/rundeck/projects/jenkins-import/scm/.git/index. Stacktrace follows:

org.eclipse.jgit.api.errors.JGitInternalException: Exception caught during execution of merge command. org.eclipse.jgit.errors.LockFailedException: Cannot lock /var/rundeck/projects/jenkins-import/scm/.git/index
	at org.eclipse.jgit.api.MergeCommand.call(MergeCommand.java:413)
	at org.eclipse.jgit.api.PullCommand.call(PullCommand.java:322)
	at org.eclipse.jgit.api.PullCommand.call(PullCommand.java:78)
	at org.rundeck.plugin.scm.git.BaseGitPlugin.gitPull(BaseGitPlugin.groovy:277)
	at org.rundeck.plugin.scm.git.GitImportPlugin.clusterFixJobs(GitImportPlugin.groovy:651)
	at rundeck.services.ScmService.fixImportStatus(ScmService.groovy:1341)
	at rundeck.services.ScmService.importStatusForJobs(ScmService.groovy:1110)
	at rundeck.controllers.MenuController.jobsFragment(MenuController.groovy:525)
	at rundeck.controllers.MenuController.apiJobsList(MenuController.groovy:2623)
	at rundeck.controllers.MenuController.apiJobsListv2(MenuController.groovy:2839)
```

PR #4307, opened 2018-12-13
Node Health Check

miguelantonio:
- Toggle on job definition to enable health check
- Node Status Service with a cache to get the status of the node
- If the toggle is enabled on the job, the job filter will be modified to run only on nodes with successful health check


miguelantonio:
- Job toogle 
screenshot 2018-12-17 12 07 34

- Node Status Cache delay 
screenshot 2018-12-17 12 08 57

- Node Page

screenshot 2018-12-17 13 41 05
screenshot 2018-12-17 13 41 22

rundeckapp/grails-app/services/rundeck/services/NodeStatusService.groovy
```diff
@@ -151,78 +207,98 @@ class NodeStatusService implements RundeckProjectConfigurable, InitializingBean
...
             }
 
-        }catch(Exception e){
-            executorReachable="fail"
-            executorTimeout="false"
+        } catch (Exception e) {
+            executorReachable = "fail"
+            executorTimeout = "false"
             statusDescription = e.message
         }
```
gschueler: since you are binding the thread outputstream earlier, i think you should add `finally{` section to reset the threadbound outputstream. (see how it is done here https://github.com/rundeck/rundeck/blob/master/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy#L1153-L1156 ) in that case it only happens on exception because the thread is handled later when it completes, but in your case i think you want to do that reset inside `finally`

rundeckapp/grails-app/services/rundeck/services/NodeStatusService.groovy
```diff
@@ -118,28 +147,55 @@ class NodeStatusService implements RundeckProjectConfigurable, InitializingBean
...
     }
 
-    CacheNodeStatus getNodeStatus(String project, INodeEntry node, String user, String userRolelist){
+    void mergeNodeAttributes(String project, INodeSet nodeSet){
+        //merged attributes with health check cache
+        Map<String, CacheNodeStatus> cacheStatus = getCurrentStatus(project)
+        for (final INodeEntry iNodeEntry : nodeSet.getNodes()) {
+            if(cacheStatus.get(iNodeEntry.nodename)) {
+                def status =  cacheStatus.get(iNodeEntry.nodename)
+                iNodeEntry.attributes?.put("checkExecutor", status.executorReachable)
```
gschueler: i wish `checkExecutor` was a configurable attribute, or maybe we can just agree on some other name, like "healthCheckStatus" @ahonor ?

miguelantonio: is this OK @gschueler ?

screenshot 2018-12-24 13 46 24

gschueler:
what is the difference between `executorTimeout` and `executorReachable` they seem to reflect the same value (whether the command was successful).  If there is no difference, i think you should remove one.

miguelantonio:
yes @gschueler, I will remote executorTimeout

rundeckapp/grails-app/views/scheduledExecution/_execOptionsForm.gsp
```diff
@@ -172,6 +172,10 @@
                                                    class=" ${localNodeName && localNodeName == node.nodename ? 'server' : ''} node_ident"
                                                    id="${enc(attr:nkey)}_key">
                                                    <g:enc>${node.nodename}</g:enc>
+
+                                                   <g:if test="${node.attributes["healthCheckStatus:icon"]}">
+                                                       <i class="glyphicon ${node.attributes["healthCheckStatus:icon"]}"></i>
```
gschueler: this needs proper encoding
```suggestion
                                                       <i class="glyphicon ${enc(attr:node.attributes["healthCheckStatus:icon"]})"></i>
```

PR #4338, opened 2018-12-21
User profile information can be sync'd from LDAP

sjrd218:


sjrd218:
Documentation is here: https://github.com/rundeck/docs/pull/189

PR #4360, opened 2019-01-04
Plugin to load a keystorage value by variable path

### LINKED ISSUE #4321: Load a keystorage value by passing the path into an option
Request to load a key storage value by passing the path into the option.
The secure options expect the input to be a password. the "default storage path" allows that password to be loaded from key storage, but it doesn't support directly loading a password by sending a key path into the option of a child job. The option value can't be used anymore in a Ruleset once passed.

Possibly a plugin could be written to take a key path as input (via job option), and load the key storage value and do something with it.

jtobard:
A utility plugin to load a key storage value by path and it creates a global variable to be used in the workflow. The path can contain variables like option values or recently exported data.

flow

fix #4321


plambert:
We probably shouldn’t merge this until we have decided how to distribute it yet. It needs some clear caveats regarding the impact it could have on the secrecy of the data used, to ensure there are no surprises. These are temporary concerns that will be addressed by a future improvement to the plugin API; until then we probably want to leave it out of Core by default?

jtobard:
@plambert agree. 
I'm going to close the PR since it needs to rebase now anyway and we are going to discuss with the team for a new location.

jtobard:
moving to https://github.com/rundeck-plugins/ ✈️

PR #4366, opened 2019-01-08
UUID validation on jobref

### LINKED ISSUE #4337: UUID validation and Autocomplete in Job Reference Workflow step
**Is your feature request related to a problem? 
When adding a Job Reference Workflow Step, selecting the JobName makes the UUID corresponding to the selected Job to be updated automatically.  However, if you initially add the UUID to select the Job Reference, the Job Name IS NOT updated accordingly. And more, in case you update the Job Reference by changing the UUID, the selected Job ( JobName) remains unchanged and the UUID is not validated. This may cause the Job Reference to point to an invalid UUID/job or (worst scenario) to an incorrect Job ( if UUID is valid ) that may cause undesired effects.

Currently, the "jobName" field auto-completes according to the Jobs available in the selected project and the UUID corresponding to the selected job fills up the "UUID" field. A solution may be to have autocomplete work the other way around, i.e.  when selecting a UUID, valid options may show up and Job Name completes accordingly. 

Also, a confirmation may be desired if the current selection is changed to a valid/invalid UUID.

jtobard:
fix #4337 
New error message when the job name and the UUID of the job don't match, preventing users to enter a wrong UUID and rely on the job name field.

As always users can delete the UUID and reference jobs only using the jobname or delete the jobname and use only the UUID. But if both fields are filled they must correspond to the same job.
screenshot


gschueler:
Perhaps the job ref input, or the error message, should explain it as you just did "You can enter a job name or UUID, but if both UUID and job name are specified they must correspond to the same job"

jtobard:
imagen

PR #4379, opened 2019-01-11
UI plugin install status fix

sjrd218:
Fix a bug when a UI plugin is installed it is not flagged as installed when listing repository plugins.


sjrd218:
If we add UI plugins to the output of listPluginDetailed() they would show up on the list plugins UI. That list did not previously include UI plugins. Since I didn't know the reasoning behind leaving them out, I opted for the more conservative approach of keeping it the same.

gschueler:
ok we can revisit it later

PR #4393, opened 2019-01-17
Adds Plugin provider metadata

gschueler:
**Is this a bugfix, or an enhancement? Please describe.**

Enables plugin providers to define a map of metadata key/values. 
This can be used to support simple plugin icons using glyphicons/font-awesome, as well as other use cases.

The metadata can be defined for java, script, or groovy plugins

**Describe the solution you've implemented**

For Java; add new annotation types
For script: support new `plugin-meta` yaml entry
For groovy; support `metadata` map property inside plugin DSL

For front-end: include provider metadata in provider description responses for plugin detail and provider descriptions list endpoints

## TBD

* [x] add docs
* [x] support Log Filter plugins GUI display with new icons
* [x]  fix issue with NodeStepPlugin vs. RemoteScriptNodeStepPlugin loading

PR #4413, opened 2019-01-23
Fix Plugin list api by referencing correct plugin list information service

sjrd218:
Looks like old code was merged into this controller. This PR repoints it back to the correct service.


gschueler:
doh, i will have to update this then https://github.com/rundeck/rundeck/pull/4393

gschueler:
@sjrd218 nm, looks like i'm unaffected

PR #4441, opened 2019-01-30
Add API endpoint that allows a user to list their roles

sjrd218:
API endpoint for a user to list their roles.


sjrd218:
Docs here: https://github.com/rundeck/docs/pull/209

PR #4472, opened 2019-02-08
Job options hidden1

carlosrfranco:
A new boolean flag should be added to the Option that defines if the option should be shown in the Job Run page.

job_options_hidden


carlosrfranco:
Created a validation for hidden options that must have a default value.

validate_hidden_options

Issue #4443, opened 2019-01-31
Notification email template with log output: use blank when not included


gschueler:
Enhancement recommendation:

`${logoutput.data}` should be set to blank/null or "see attachment" if it is not chosen in the rundeck UI. Because we use the same email templates regardless of how the logout is set and seeing "${logoutput.data}" in the email when we did not choose to attach it is confusing.

example 
![image](https://user-images.githubusercontent.com/42989099/52067477-435fd780-2540-11e9-9e0a-d4cd69929c3e.png)

_Originally posted by @gtsteffaniak in https://github.com/rundeck/rundeck/issues/4440#issuecomment-459402134_

PR #4474, opened 2019-02-08
Notification email template with log output: use blank when not included

ltamaster:
Enhancement: https://github.com/rundeck/rundeck/issues/4443

Adding an empty string to ${logoutput.data} context variable when that variable is set on the template, and the flag attachLogInline is false.
 
**Additional context**
For issue https://github.com/rundeck/rundeck/issues/4443

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/NotificationService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/NotificationService.groovy
@@ -324,15 +324,23 @@ public class NotificationService implements ApplicationContextAware{
                         def template = new File(templatePath)
                         if (template.isFile()) {
 
-                            if(attachlogbody){
-                                if(template.text.indexOf('${logoutput.data}')>0){
-                                    outputBuffer=copyExecOutputToStringBuffer(exec,isFormatted)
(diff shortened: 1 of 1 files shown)


rundeckapp/grails-app/services/rundeck/services/NotificationService.groovy
```diff
@@ -324,15 +324,23 @@ public class NotificationService implements ApplicationContextAware{
...
-                                    outputBuffer=copyExecOutputToStringBuffer(exec,isFormatted)
-
-                                    //add to context in order to display output using a custom template ( using ${logoutput.data} )
-                                    def contextOutput=[:]
-                                    contextOutput['logoutput'] = ["data":outputBuffer.toString().encodeAsSanitizedHTML()]
-                                    context = DataContextUtils.merge(context, contextOutput)
-                                }
+                            //check if the custom template is calling ${logoutput.data}
+                            def templateLogOutput=false
+                            if(template.text.indexOf('${logoutput.data}')>0){
```
gschueler:
```suggestion
                            if(template.text.indexOf('${logoutput.data}')>=0){
```
0 is success result as well

PR #4509, opened 2019-02-15
Add secondary node filter for jobs: Exclude filter

miguelantonio:
New exclude filter on jobs. It will exclude nodes added on the filter.
Also it adds a flag that will deselect the excluded nodes on Run Now page 

A new filter and flag on job form.
The nodes that match with this filter will be exclude form the main filter of the job.

**Additional context**

**Job Form**
screenshot 2019-02-15 23 31 27

**Node Detail**
screenshot 2019-02-15 23 31 42

**Run Now**
screen shot 2019-02-15 at 5 35 49 pm


rundeckapp/grails-app/domain/rundeck/ScheduledExecution.groovy
```diff
@@ -257,6 +257,7 @@ class ScheduledExecution extends ExecutionContext {
         map.scheduleEnabled = hasScheduleEnabled()
         map.executionEnabled = hasExecutionEnabled()
         map.nodeFilterEditable = hasNodeFilterEditable()
+        map.excludeFilterUncheck = hasNodesHealthCheckEnabled()
```
gschueler: if the excludeFilterUncheck is default true/false, then it should not be added to the export map.  (also if there is no filterExclude) in that way the XML/yaml serialization for old jobs will not change

miguelantonio: @miguelantonio you mean I just need to remove it or check first if the exists like:
```if(this.filterExclude){
                map.nodefilters.filterExclude = this.filterExclude

                if(this.excludeFilterUncheck){
                    map.excludeFilterUncheck = true
                }
            }```

gschueler: as long as the excludeFilterUncheck defaults to false, that is ok.  if it should be default true it would be the inverse

miguelantonio: @gschueler I sent the fix, the default value is false

Issue #4471, opened 2019-02-08
UUID validation of jobRef breaks bulk import of jobs in clean instance of RunDeck


ajxb:
We often have to build out new instances of RunDeck from scratch from a set of templates that include a number of projects and jobs as part of our development workflow with the tool.  As part of this build process a new, clean instance of RunDeck is spun up, and project / job data is imported from an archive created using the rd cli tool.

The tools first create the projects from the exported data, they then iterate over each project in turn and import the yaml associated with each job for each project.  Prior to v3.0.13 this import did no checking of UUID integrity, so would succeed in importing the jobs no matter what order this is done.  I think that the following changes introduced UUID checking for jobRefs:

- https://github.com/rundeck/rundeck/issues/4337
- https://github.com/rundeck/rundeck/pull/4366

This breaks the job import into a clean system unless we import the jobs in the reverse order to when they are referenced.  We have hundreds of jobs with many references between them over multiple projects so working out the order is going to be difficult, is there any way that we can import the data ignoring the jobRef UUID checks given that the data will have been exported from a known good working system?

**My Rundeck detail**

* Rundeck version: 3.0.13
* install type: rpm
* OS Name/version: CentOS 6.9
* DB Type/version: mysq 5.7

**To Reproduce**

1. Create a new, clean instance of RunDeck (no data, projects, jobs in system)
1. Create a project
1. Using the CLI tool import a job that has a jobRef, without having imported the job that it references

This results in the error : 
`:Workflow has one or more invalid steps: [1: [Job Name does not correspond to the UUID]]
Invalid Option definition: Domain: Workflow has one or more invalid steps: [1: [Job Name does not correspond to the UUID]]
Invalid Option definition: Live_URL: Workflow has one or more invalid steps: [1: [Job Name does not correspond to the UUID]];Invalid Option definition: Domain: Workflow has one or more invalid steps: [1: [Job Name does not correspond to the UUID]]
Invalid Option definition: Next_URL: Workflow has one or more invalid steps: [1: [Job Name does not correspond to the UUID]];Invalid Option definition: Domain: Workflow has one or more invalid steps: [1: [Job Name does not correspond to the UUID]];Invalid Option definition: Live_URL: Workflow has one or more invalid steps: [1: [Job Name does not correspond to the UUID]];Invalid Option definition: Domain: Workflow has one or more invalid steps: [1: [Job Name does not correspond to the UUID]]
`

**Expected behavior**

[...]

PR #4516, opened 2019-02-18
removed UUID validation of jobRef

### LINKED ISSUE #4451: Importing project archive fails if referenced job name/group changed
**Describe the bug**
Projects with jobs referenced by UUID that have had their names or groups updated will cause import failures with "Preserve UUID".

**My Rundeck detail**

* Rundeck version: 3.0.13
* install type: Docker

**To Reproduce**
1. Import the project into one instance
2. Rename `RefMe` or its group
3. Export the project and import into another Rundeck instnace

**Expected behavior**
Rundeck should continue to allow me to export and import projects even if I have renamed referenced jobs or their groups.

**Screenshots**
image

**Additional context**
The error message does not explain the problem or offer a workaround. The logs are the only indicator of the issue:
```
2019-02-04 22:48:45.258 ERROR --- [tp2031149831-30] rundeck.services.ProjectService          : Failed loading (1) jobs from XML at archive path: rundeck-ProjectA/jobs/job-a6d88d66-920b-492b-b7c4-1a22da67333b.xml
2019-02-04 22:48:45.259 ERROR --- [tp2031149831-30] rundeck.services.ProjectService          : Job at index [1] had errors: Validation errors: Workflow has one or more invalid steps: [1: [Job Name does not correspond to the UUID]]; Invalid Option definition: Foo: Workflow has one or more invalid steps: [1: [Job Name does not correspond to the UUID]]
```

[ProjectA-20190204-225828.rdproject.zip](https://github.com/rundeck/rundeck/files/2830123/ProjectA-20190204-225828.rdproject.zip)

jtobard:
Job reference step using UUID and job name pointing to a job that doesn't exist won't fail it will only remove the `UUID` component and leave just the `jobname` reference.

Fix #4451 and 4471

### DIFF
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/WorkflowController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/WorkflowController.groovy
@@ -1139,17 +1139,6 @@ class WorkflowController extends ControllerBase {
             if (!exec.jobName && !exec.uuid) {
                 exec.errors.rejectValue('jobName', 'commandExec.jobName.blank.message')
             }
-            if(exec.uuid && exec.jobName){
-                def refSe = ScheduledExecution.findScheduledExecution(null,null,null,exec.uuid);
-                if(!refSe){
-                    //"Job doesnt exists, using only the uuid to reference the job
-                    exec.uuid = null
-                }else if(refSe?.generateFullName()!=ScheduledExecution.generateFullName(exec.jobGroup, exec.jobName)){
-                    exec.jobName = null
-                    exec.jobGroup = null
-                    //using only the uuid
-                }
-            }
             if(exec.uuid && !exec.jobName){
                 def refSe = ScheduledExecution.findScheduledExecution(null,null,null,exec.uuid);
                 if(refSe){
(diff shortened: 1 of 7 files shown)


ProTip:
@jtobard  are you able to import the archive from #4451 ? I seem to be unable to import it on anything above `3.0.12`, including a build of this branch.. It's odd because the job name and UUID match the referenced job in that archive..

jtobard:
@ProTip my mistake, missing some code on the PR maybe because I moved from another branch by hand, now pushing it.

gschueler:
@jtobard perhaps there needs to be a functional test performing the import

ProTip:
@jtobard Thanks! I have tested this and it does let me import the project, as well as the project if a referenced job has been renamed.

It seems though that the Job names are getting stripped(instead of just ignored for validation purposes). After I import the example project where everything matches, and I view the reference workflow step, only the UUID remains in GUI. Without looking up the name and displaying it users may be surprised that the reference names are gone after an import?

I'm also curious that in the case of #4471 if we strip the UUID when importing jobs out of order, will this leave only the job name? In that case if a job has been renamed or regrouped the UUID may actually be correct and leaving just the name as reference may break the workflows.

PR #4555, opened 2019-02-27
Use master branch of ui trellis

ProTip:
Set ui-trellis dependency to particular commit id

### DIFF
--- a/rundeckapp/grails-spa/package-lock.json
+++ b/rundeckapp/grails-spa/package-lock.json
@@ -1,9 +1,37 @@
 {
-  "name": "project-motd",
+  "name": "rundeck-vue-ui",
   "version": "1.0.0",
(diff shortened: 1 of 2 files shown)


gschueler:
btw, i wonder if it should use a tag/commit instead of branch name ... the branch is likely to change over time

ProTip:
That makes a lot of sense for releases.

ProTip:
Fixes #4593

gschueler:
@ProTip even for non-release code, since ui-trellis can change outside of rundeck code, it makes sense to set the dependency to a particular tag/commit I think

ProTip:
@gschueler I was thinking for some reason it would be good to get updates on master but.. That does not make sense to me ATM. I've pinned the `ui-trellis` version.

Issue #4536, opened 2019-02-23
static tokens file not working


mengesb:
**Describe the bug**
I've defined a static API token in the static tokens file and I'm unable to use that token.

**My Rundeck detail**

* Rundeck version: 3.0.13-20190123
* install type: rpm
* OS Name/version: RHEL 7.x
* DB Type/version: H2

**To Reproduce**
1. Generate `/etc/rundeck/tokens.properties`
2. Update `/etc/rundeck/framework.properties` and set `rundeck.tokens.file=/etc/rundeck/tokens.properties`
3. Restart rundeckd
4. Attempt `curl -vk https://host.domain.tld/api/14/system/info\?authtoken\=__________` using the static token

**Expected behavior**
API call works correctly

**Screenshots**
```
[root@rd ~]# grep tokens /etc/rundeck/framework.properties 
rundeck.tokens.file=/etc/rundeck/tokens.properties
[root@rd ~]# cat /etc/rundeck/tokens.properties
# Generated by Chef for rd
# Local modifications will be overwritten.

admin: BFrVx****

# ---
curl -vk https://host.domain.tld/api/14/system/info\?authtoken\=BFrVx****
# ---
2019-02-23 06:07:44.699 ERROR --- [qtp490625541-26] rundeck.interceptors.SetUserInterceptor  : Invalid API token used: BFrVx****
2019-02-23 06:07:44.701 ERROR --- [qtp490625541-26] r.interceptors.AuthorizationInterceptor  : (Token:BFrVx****) UNAUTHORIZED for api/apiSystemInfo
```

*Desktop (please complete the following information):**
 - OS: N/A
 - Browser N/A
 - Version N/A

**Additional context**
I've sanitized the above.... yes the API token is the correct length. I've also tried with the newer format with the `username: token,role1,role2,rol3` method but this isn't working either

If I use an API token created via the UI, things work as expected. I've attempted several reboots like #971 suggested and #4124 I read but wasn't any help. I still can't use a token inserted in the tokens file.

mengesb:
File permissions should be OK:
```
-rw-r-----.  1 rundeck rundeck 3033 Feb 23 06:00 framework.properties
-rw-r-----.  1 rundeck rundeck  606 Feb 10 21:50 rundeck-config.properties
-rw-r-----.  1 rundeck rundeck  137 Feb 23 06:16 tokens.properties
```

mengesb:
[...]

PR #4567, opened 2019-03-04
Fix and add debugging for static tokens

sjrd218:
Allows users who are trying to use static tokens to gain more insight about the token check.


gschueler:
ref https://github.com/rundeck/rundeck/issues/4536

Issue #4420, opened 2019-01-24
ZipException when deploying 3.x WAR in Tomcat


kristian-lesko:
**Describe the bug**
When deploying a Rundeck 3.x WAR to Tomcat, the following exception happens:
`2019-01-24 14:31:41.577  WARN --- [ost-startStop-1] ationConfigEmbeddedWebApplicationContext : Exception encountered during context initialization - cancelling refresh attempt: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'application': Initialization of bean failed; nested exception is java.security.PrivilegedActionException: java.util.zip.ZipException: zip file is empty`

**My Rundeck detail**

* Rundeck version: 3.0.13 (happens on 3.0.1 & 3.0.9 as well)
* install type: WAR (in Tomcat)
* OS Name/version: CentOS 7
* DB Type/version: Postgres (happens with default DB config as well)

**To Reproduce**
1. Install _tomcat_.
2. Download the installation `.war` file into Tomcat `webapps` directory.
3. (Optional) Unpack the `.war` into a folder; however, Tomcat itself does this
4. Run Tomcat.

**Expected behavior**
Tomcat should start without errors.

**Desktop**
Irrelevant; UI cannot be accessed because application does not start correctly.

**Additional context**
The issue happened for me both when upgrading from Rundeck 2.11 (which worked without issues) and when trying to deploy the new Rundeck 3 from scratch. 
Perhaps the error message is misleading and means something else, but I wasn't able to find any `.zip` file in the installation that would be obviously empty or bad. Also, sometimes the message altered to `java.util.zip.ZipException: error in opening zip file`. I wasn't able to find any mention of filename of the zip that failed to open, though.

sjrd218:
@kristian-lesko  The most probable cause of this issue is that the rdeck.base location is not writable by the tomcat user process. Please make sure that you have launched tomcat with the -Drdeck.base system property set to a location on your file system to which the tomcat user has full access.

bsloeserwij:
I have the same issue. 
My rdeck.base is owned by the tomcat user, has permissions 777 (for testing only) and selinux is disabled. This issue still appears.

kristian-lesko:
@sjrd218 many thanks for your response.

I was actually deploying a tomcat war that was not owned by `tomcat`; however, it seemed to be unpacked to a subdirectory without issues. Even after setting its owner to tomcat, I still get the exact same "zip file is empty" exception.

[...]

PR #4568, opened 2019-03-04
Fix tomcat 7 init issue. #4420.

sjrd218:
Fix first install issue on Tomcat 7 with RPM based installation.

### DIFF
--- a/rundeckapp/grails-app/init/rundeckapp/init/RundeckInitializer.groovy
+++ b/rundeckapp/grails-app/init/rundeckapp/init/RundeckInitializer.groovy
@@ -54,6 +54,8 @@ class RundeckInitializer {
     public static final String SPRING_BOOT_ENABLE_SSL_PROP = "server.ssl.enabled"
     private static final String LINESEP = System.getProperty("line.separator");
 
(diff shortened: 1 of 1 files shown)


rundeckapp/grails-app/init/rundeckapp/init/RundeckInitializer.groovy
```diff
@@ -296,12 +298,22 @@ class RundeckInitializer {
...
             }
         }
 
         try {
+            if(Environment.isWarDeployed()) {
+                String sloc = location.toString()
+                if(sloc.endsWith(".class") && sloc.contains(WAR_BASE)) {
+                    //An unusual case where the location returned by the protection domain code source did not give us a usable location.
+                    //Seen in Tomcat 7 on Centos 7
+                    String webinfclassesdir = sloc.substring(5,sloc.indexOf(WAR_BASE)+WAR_BASE.length())
```
gschueler: what is the magic `5` ?

sjrd218: 5 is the 'file:' prefix for a url with a file protocol. I've updated the code to reflect that.

PR #4576, opened 2019-03-04
Fix stacktrace on menu controller caused by typo. 

sjrd218:
Fix stacktrace on menu controller caused by typo. Made call more efficient.


gschueler:
was there an associated bug filed?

sjrd218:
Not that I know of. I was using the app and saw the Stacktrace when I was doing normal stuff, so I fixed it.

gschueler:
@sjrd218 👍  just making sure there wasn't a missed reference

PR #4580, opened 2019-03-05
Fix french translation for 'kill'

vincsb:
Typo translation fix: "Kill" is used in english and was translate literally with "Tuer", but it's too agressive : we use "kill" to talk about somebody. Instead, we can user "Arrêter" (stop).

Fix two existing translations in french only.


gschueler:
Thank you!

Issue #4419, opened 2019-01-24
Don't invert sort order of multi-valued default values.


NFDWADM:
**Describe the bug**
When creating a multi-valued option the default values are sent to the job (and GUI) in reverse order.

**My Rundeck detail**

* Rundeck version: 3.0.9
* install type: war
* OS Name/version: Red Hat 6.9

**To Reproduce**
1. Create option "ids" in job
2. Set "Multi-valued" to yes, delimiter = ,
3. Set "Default Value" to 1,2,3,4
4. Execute script: `echo @option.ids@`
5. Save job
6. Check the option ids in "Prepare and Run".
7. The order from top to bottom is 4,3,2,1
8. Run the job. Output is 4,3,2,1

**Expected behavior**
Order should be 1,2,3,4 in GUI and output.

NFDWADM:
I also noticed that rundeck picks the items from a multivalued option in a random order. In the beginning I thought it was top down but I have at least one job where it picks the last option first.

NFDWADM:
I found a pattern. I created a job that takes multi-valued option that just echoes the options.

Options (from top to bottom): 0, 3, F

When I run that job for the first time I get: 0,3,F (as expected)
Second run (after clicking on "Run this job again with the same options"): F,3,0
Third: 0,3,F
Fourth: F,3,0
Fifth: 0,3,F.

So it seems that rundeck always switches the first and the last item.

Issue #4389, opened 2019-01-14
Job option (variable) auto-sorting as text the Allowed Values list


Daryes:
**Describe the bug**
When entering a list in the "Allowed Value" text area in a new option (or variable), the list is automatically sorted when the option is saved

**My Rundeck detail**

* Rundeck version: 3.0.11
* install type: deb
* OS Name/version: Ubuntu 18.04

**To Reproduce**
1. Create a new job or edit an existing one
2. add a new option 
3. populate the mandatory informations
4. select the `Allowed Value` text area, and enter : `15,30,60,90,180,360`
5. save the option.
6. edit the saved option, check the `Allowed Value` area, the following information will be present : `15,180,30,360,60,90`
Of course, this sorted order will be kept in the job definition.

**Expected behavior**
expected list : `15,30,60,90,180,360`
The saved value list should have been either not sorted, or numerically sorted.

reinerrdeck:
Confirmed in 3.0.12 WAR.

Issue #1394, opened 2015-08-13
Preserve order of allowed values within a job execution option


danifr:
Hi,

is there a way to sort the list of allowed values for a job option?
I tried adding `preserveOrder="true"` but it doesn't seem to work within a specific option.

``` xml
<option preserveOrder="true" name='execution_mode' values='standard,force' enforcedvalues='true' required='true' />
```

I have some jobs that use different execution modes depending on a flag.

For example in this case, I'd like `standard` to be the option by default (first in the list) instead of `force`.

![selection_004](https://cloud.githubusercontent.com/assets/7135482/9245786/37be1ac0-41a7-11e5-82f6-e04580227685.png)

Am I missing something? Is there a way to do it??
Otherwise I think it'd be nice to have this feature.

Thanks!
Daniel.

gschueler:
I guess you mean preserve the original order of the allowed values? (They are sorted alphabetically automatically).

Currently you would have to set the defaultValue for the option to the one you want selected by default, but the list would appear it sorted order.

danifr:
> I guess you mean preserve the original order of the allowed values? (They are sorted alphabetically automatically).
Yep, exactly that.
Thanks!

rophy:
+1

rophy:
I'm really needing this feature, can we disable the sorting of option values altogether?

Since we're manually specifying allowed options, user can always sort it manually if they wish the values to be sorted.

rootAmeen:
Yea the option to manually sort is important.

steveblackmon:
+1 this is important for my use case as well, because downstream system (amazon emr) cares about the order

chrisabney:
+1 it is important that the alphabetical sorting "feature" not be included in Rundeck

kmcdowell85:
Any news on this? At Least making it a toggleable feature? Needed very much +1

sarathcakurathi:
Any update on this feature?

ahmedchater:
Puppet needs recipes to be sent to it in an explicit order, this would be useful.

matejhasul:
+1

AVShutov:
+1

NFDWADM:
+1

fdevans:
[...]

PR #4599, opened 2019-03-07
new option features ( sort, list delimiter).

ltamaster:
Fix for the following issues:
https://github.com/rundeck/rundeck/issues/1394
https://github.com/rundeck/rundeck/issues/4419
https://github.com/rundeck/rundeck/issues/4389

- Adding a new option field in order to save the option list values as the user set it on the job (not as a list, but a string).
- Also, a new toggle called `sort list` was added in order to sort the option list values. This allows users to preserve the order of the list in case the flag is false.
- Finally, another field called `list delimiter` was added in order to specify the delimiter of the list 

**Additional context**

**New Sort field**
screenshot 2019-03-07 14 37 51

screenshot 2019-03-07 14 38 09

**Sort numeric list**
screenshot 2019-03-07 14 38 32

screenshot 2019-03-07 14 39 11

New delimiter field 

screenshot 2019-03-07 14 40 08
screenshot 2019-03-07 14 40 19


rundeckapp/grails-app/domain/rundeck/Option.groovy
```diff
@@ -52,7 +54,7 @@ public class Option implements Comparable{
     Boolean required
     Boolean isDate
     String dateFormat
-    SortedSet<String> values
+    Set<String> values
```
gschueler: Does changing this affect existing db entries at all? is there a need to change it since we are moving to saving the input string anyway?

rundeckapp/grails-app/domain/rundeck/Option.groovy
```diff
@@ -311,10 +329,12 @@ public class Option implements Comparable{
      */
     void convertValuesList(){
         if(valuesList!=null){
-            def x=new TreeSet()
-            x.addAll(valuesList.split(",").collect{it.trim()}.grep{it} as List)
+            if(valuesListDelimiter == null){
+                valuesListDelimiter=DEFAULT_DELIMITER
+            }
+            def x=new ArrayList()
+            x.addAll(valuesList.split(valuesListDelimiter).collect{it.trim()}.grep{it} as List)
```
gschueler: `valuesList.split(valuesListDelimiter)` if the delimiter is a regex-special char this would not work.  Maybe `valuesList.split(Pattern.quote(valuesListDelimiter))`

rundeckapp/grails-app/domain/rundeck/Option.groovy
```diff
@@ -367,6 +387,27 @@ public class Option implements Comparable{
...
+    }
+
+    void sortValuesList(){
+        def numericValues = values.findAll { it.isNumber() }.collect {it.toDouble()}
+        if(numericValues.size()==values.size()){
+            def stringNumericList= values.findAll { it.isNumber() }.collect {new StringNumericSort(it, it.toDouble())}
+            stringNumericList.sort{ a,b -> a.numericValue <=> b.numericValue }
+            values = stringNumericList.collect{it.strValue}
+        }else{
+            values = values.sort()
```
gschueler: if `Set<String> values` is not a transient, then modifying it is going to persist the value to the DB.  

I think you should not write to the `values` field anymore.  Instead just return the converted values as a List.

When an Option/Job is created/updated, the `values` field should be set to `null` instead

rundeckapp/grails-app/domain/rundeck/Option.groovy
```diff
@@ -367,6 +387,27 @@ public class Option implements Comparable{
...
+            sortValuesList()
+        }
+        values
+    }
+
+    void sortValuesList(){
+        def numericValues = values.findAll { it.isNumber() }.collect {it.toDouble()}
+        if(numericValues.size()==values.size()){
+            def stringNumericList= values.findAll { it.isNumber() }.collect {new StringNumericSort(it, it.toDouble())}
+            stringNumericList.sort{ a,b -> a.numericValue <=> b.numericValue }
```
gschueler: the comparator logic should be moved to StringNumericSort where it could be unit tested

PR #4600, opened 2019-03-07
Add tomcat 7 tests

ProTip:
Adds tomcat 7 API tests and updates Rundeck to achieve passing results.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,10 +76,15 @@ jobs:
     - script_block 'docker.pull' pull_rdtest
     - script_block 'api-test' 'DOCKER_COMPOSE_SPEC=docker-compose-api-mysql.yaml bash test/run-docker-api-tests.sh'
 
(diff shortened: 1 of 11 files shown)


ProTip:
Opened the PR early so we can all see the CI/CD status.

ProTip:
All tests passing, ready for review!

ProTip:
Saw wrong Travis notification and jumped the gun.. Looks like one of the tests is unstable ATM and that `close()` and `flush()` also need to be wrapped in a static compiled method.

ProTip:
@sjrd218 @gschueler All API tests are now passing on Tomcat 7.

I have removed `close` and replaced with `flush` which has the same response commit effect and stops Grails from trying to find a matching view; small change to keep similar behavour. I placed a static `append` and `flush` on the base controller to DRY things up.

Issue #4562, opened 2019-03-01
Upgrade Jsch library to latest (0.1.55)


gschueler:
**Is your feature request related to a problem? Please describe.**

Update Jsch library to latest release (0.1.55)

PR #4609, opened 2019-03-11
updating jsch version to 0.1.55

ltamaster:
Enhancement, updating JSCH lib to the last version


gschueler:
ref https://github.com/rundeck/rundeck/issues/4562

Issue #4520, opened 2019-02-19
project properties key mask feature not working for workflow steps


cwaltherf:
**Describe the bug**
The password mask option in the Edit Project > Edit Project Configuration File is not working for Workflow Node Steps or Node Executors

**Expected behavior**
As masked passwords from resource models are masked with "*******" signs inside the edit view of the project properties from the GUI, they should be also masked for Node Executors and Workflow Steps, also from custom plugins.

PR #4610, opened 2019-03-11
Creating key mask feature for custom plugins on project propeties view #4520

carlosrfranco:
This changes allow mask password properties with pattern "project.plugin.[ServiceName].[providerName].password" for non declared plugin class field.


rundeckapp/grails-app/services/rundeck/services/FrameworkService.groovy
```diff
@@ -57,6 +52,7 @@ import java.util.function.Predicate
 class FrameworkService implements ApplicationContextAware, AuthContextProcessor, ClusterInfoService {
     static transactional = false
     public static final String REMOTE_CHARSET = 'remote.charset.default'
+    public static final List PLUGINS_TYPES = ['Notification','WorkflowNodeStep', 'RemoteScriptNodeStep']
```
gschueler: why is it only listing these plugin types?

rundeckapp/grails-app/controllers/rundeck/controllers/FrameworkController.groovy
```diff
@@ -1060,6 +1028,34 @@ class FrameworkController extends ControllerBase implements ApplicationContextAw
...
+                        ]
+                    }
+
+            try {
+                pluginsPasswordFieldsService.untrack(pluginsConfig, *pluginsDescs)
+                pluginsConfig.each {pconfig ->
+                    def props=pconfig.config.props
+                    def type = pconfig.config.type
+                    frameworkService.getPluginsTypes().each {
+                        String propPassowordName = "project.plugin.${it}.${type}.password"
```
gschueler: it seems this only works if the config entry is named "password"?

rundeckapp/grails-app/services/rundeck/services/PasswordFieldsService.groovy
```diff
@@ -115,6 +116,43 @@ class PasswordFieldsService {
...
+            }
+
+            Description desc = descriptions.find { it.name == config.type }
+            if (desc == null) {
+                configPos++
+                continue
+            }
+
+            for (property in props) {
+                if (isPluginPasswordProperty(property)) {
```
gschueler: since we have the plugin Description, we can look up the property corresponding to the config key, and determine if it is a Password value, by `isPasswordDisplay(Property)`

carlosrfranco: There are some plugins that can use project properties whose is not implemented like a plugin property with display annotations. One example is the Service Now plugin this property:

`project.plugin.Notification.Service-Now-Notification.login=user
project.plugin.Notification.Service-Now-Notification.password=s3cret
project.plugin.Notification.Service-Now-Notification.url=https://server.service-now.com`

So, I wonder if this would be good solution to apply mask if user want even that plugin not have the password display for this property.

carlosrfranco:
> It should handle more plugin types, and should correctly determine if a tracked property is a PASSWORD display type.
> 
> the plugin types could be listed, but instead perhaps just using the input properties `project.plugin.[serviceType].[providerName].*` it could extract only the needed plugin service Types



There are some plugins that can use project properties whose is not implemented like a plugin property with display annotations. One example is the Service Now plugin this property:

project.plugin.Notification.Service-Now-Notification.login=user project.plugin.Notification.Service-Now-Notification.password=s3cret project.plugin.Notification.Service-Now-Notification.url=https://server.service-now.com

So, I wonder if this would be good solution to apply mask if user want even that plugin not have the password display for this property.

rundeckapp/grails-app/services/rundeck/services/FrameworkService.groovy
carlosrfranco: Actually, it is wrong. Should have all groovy and script-based plugin type

rundeckapp/grails-app/services/rundeck/services/PasswordFieldsService.groovy
jtobard: I think it's better to implement it that way and create a github issue for that plugin to be upgraded to annotations.

gschueler:
Note: I am working on some other changes that affect the PasswordFieldService, so I will merge these changes in after that

gschueler:
close in lieu of https://github.com/rundeck/rundeck/pull/4724

Issue #4212, opened 2018-11-15
Unable to pass job.user.email to worflow step


kquilliam:
**Describe the bug**
${job.user.email} not being passed into job resulting in a null value. Im using Rundeck to execute Ansible playbooks and I am trying to pass ${job.user.email} into the playbook to email configlets to the executing user. ${job.user.name} is currently being used without issue.

**My Rundeck detail**
* Rundeck version: 2.11.5
* install type: rpm
* OS Name/version: CentOS Linux release 7.5.1804 (Core)
* DB Type/version: postgres

**To Reproduce**
Create job referencing a playbook with a var assigned to the content of ${job.user.email}.
```
- defaultTab: summary
  description: ''
  executionEnabled: true
  id: 4a24de3e-f576-46e5-90fb-23e4a446faa0
  loglevel: INFO
  name: TESTING
  nodeFilterEditable: false
  scheduleEnabled: true
  sequence:
    commands:
    - configuration:
        ansible-become: 'false'
        ansible-become-method: sudo
        ansible-extra-param: --limit ""
        ansible-extra-vars: |
          rundeck_email: ${job.user.email}
        ansible-playbook: /etc/ansible/new_test/.git/dev/rundeck.yml
      nodeStep: false
      type: com.batix.rundeck.plugins.AnsiblePlaybookWorkflowStep
    keepgoing: false
    strategy: node-first
  uuid: 4a24de3e-f576-46e5-90fb-23e4a446faa0

```
```
---
- name: Test template
  hosts: localhost
  gather_facts: no
  connection: local

  tasks:
    - debug: msg="{{ rundeck_email }}"
```

**Expected behavior**
The email address of the user profile would be passed into the workflow step.

When ran directly from Ansible, passing the email as a var the result is what is expected.
`ansible-playbook rundeck.yml -vvv --extra-vars rundeck_email=email@somemail.com`
```
ok: [localhost] => {
    "msg": "email@somemail.com"
}
```

**Screenshots**

**Desktop (please complete the following information):**

**Additional context**

jorn-k:
Hi folks. This issue has been resolved and merged in [4361](https://github.com/rundeck/rundeck/pull/4631). It will be released very soon.

Thank you, again, Rundeck community for continuing to make Rundeck better every day!

Issue #2607, opened 2017-07-05
Job editor autocomplete: incorrectly suggests ${job.user.email}


gschueler:
Issue type: Bug report

**My Rundeck detail**

* Rundeck version: 
* install type: (rpm,deb,war,launcher?)
* OS Name/version: 
* DB Type/version: mysql? postgres? h2?..

**Expected Behavior**

Autocomplete suggests correct variables.

**Actual Behavior**

Autocomplete suggests non-working variable `${job.user.email}`

See also #947 

**How to reproduce Behavior**

orderedchaosdev:
I am also seeing this behavior and it has caused me problems since I had planned to use this var and can't.

toninis:
Hello,

Any updates on this ? 
Will this feature be available ?

jorn-k:
Hi folks. This issue has been resolved and merged in [4361](https://github.com/rundeck/rundeck/pull/4631). It will be released very soon.

Thank you, again, Rundeck community for continuing to make Rundeck better every day!

Issue #947, opened 2014-09-25
add job.user.email to execution context variables


pwhack:
I was able to pass job.user.email to scripts just fine in v2.1.1, however, the value is empty in v2.2.2. I verified the user's profile contains a valid email address.

dscowan:
It would be helpfull if job.user.email could be used in scripts

zlanyi:
+1

Slm0n87:
+1 - this feature would help me, too. To use it in a inline-script

brandonpal:
+1

david-richer-adsk:
+1

nmamn:
+1 ...

luckytaxi:
What's the status of this?

astapelfeld:
+1

toninis:
Hello,

Any updates on this ?
Will this feature be available ?

kuribas:
+1

ZoeLeah:
+1

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

jorn-k:
Hi folks. This issue has been resolved and merged in [4361](https://github.com/rundeck/rundeck/pull/4631). It will be released very soon.

Thank you, again, Rundeck community for continuing to make Rundeck better every day!

PR #4631, opened 2019-03-15
Make email of user profile accessible on context #4212

carlosrfranco:
This changes allow context access ${job.user.email}, i.e. the email address of the user profile can be passed into the workflow step.


jtobard:
fix #4212 fix #2607 fix #947

PR #4641, opened 2019-03-21
Fix holded connections on referenced jobs

### LINKED ISSUE #4459: Rundeck holds connections on job referenced jobs
**Describe the bug**
Rundeck is holding the connections to backend DB when there are references to other jobs.
These connections will be held as long as the child job is still running and it can be several hours. If there are many such jobs, we will end up in a situation where Rundeck is holding many idle connections for a lot of time with backend DB

**My Rundeck detail**

* Rundeck version: 3.0.9
* install type: rpm
* DB Type/version: Mysql: version: 5.6.22-72

**To Reproduce**
running job referenced jobs

jtobard:
Fix #4459 

`notificationService.triggerJobNotification` on referenced jobs generated new connections when instantiated using the id of the job (in the internal transactional method uses a new transaction in another service and uses the element to check if it has notifications enabled on the trigger) when the complete `scheduledExecution` passes directly from this service, this new connection is avoided and it continues using only one per referenced task.

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
@@ -3476,21 +3476,16 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
         }
         def props = frameworkService.getFrameworkProperties()
         def disableRefStats = (props?.getProperty('rundeck.disable.ref.stats') == 'true')
-        ReferencedExecution refExec
 
 
         Execution exec = Execution.get(execid as Long)
-        refExec = new ReferencedExecution(scheduledExecution: se, execution: exec, status: EXECUTION_RUNNING).save()
-
+        def refId = saveRefExecution(EXECUTION_RUNNING, null, se.id, exec.id)
 
 
         if (!(schedlist[0].successOnEmptyNodeFilter) && newContext.getNodes().getNodeNames().size() < 1) {
             String msg = "No nodes matched for the filters: " + newContext.getNodeSelector()
             executionContext.getExecutionListener().log(0, msg)
-            ReferencedExecution.withTransaction { status ->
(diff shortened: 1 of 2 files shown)


ProTip:
@jtobard I'm still seeing a transaction opened that hangs around for the life of the job run:
```
---TRANSACTION 2559, ACTIVE 50 sec
5 lock struct(s), heap size 1136, 2 row lock(s), undo log entries 1
MySQL thread id 4, OS thread handle 140094787745536, query id 9816 172.22.0.3 rundeck
```

Is that service perhaps opening a transaction and querying some data too soon, or leaving one open where it doesn't need to?

jtobard:
@ProTip I don't see any open connection more than a couple of seconds at the start and finish. But I'm using the command `show full processlist` from the original issue, how are you checking this? maybe there are other places to optimize.

ProTip:
@jtobard I'm running `show engine innodb status` to view open transactions.
After enabling the lock monitor:
```
SET GLOBAL innodb_status_output=ON;
SET GLOBAL innodb_status_output_locks=ON;
```

I'm seeing this from `show engine innodb status`:
```
---TRANSACTION 5660, ACTIVE 10 sec
5 lock struct(s), heap size 1136, 2 row lock(s), undo log entries 1
MySQL thread id 10, OS thread handle 140340071290624, query id 19972 172.22.0.3 rundeck
TABLE LOCK table `rundeck`.`referenced_execution` trx id 5660 lock mode IX
TABLE LOCK table `rundeck`.`scheduled_execution` trx id 5660 lock mode IS
...
TABLE LOCK table `rundeck`.`execution` trx id 5660 lock mode IS
``

I have a job that references a sleep, and this transaction appears to be open until the job completes. It looks like it may have updated `referenced_execution` and read from `scheduled_execution` and `execution` but has not committed.

ProTip:
Superseded by #4654

PR #4654, opened 2019-03-27
Stop holding DB connections during job execution

### LINKED ISSUE #4459: Rundeck holds connections on job referenced jobs

ProTip:
Fixes #4459 

This builds off the work @jtobard completed in #4641.

During testing against MySQL with low `wait_timeout` I was still getting errors about the underlying connection being dead after the job finishes. This was strange as I could not see a running transaction in `show engine innodb status`.

### Test Setup
**MySQL**
```
SET GLOBAL wait_timeout=15
```
**Rundeck**
```
dataSource.properties.validationInterval=5000
dataSource.properties.testOnBorrow=true
dataSource.properties.testWhileIdle=true
dataSource.properties.testOnReturn=true
dataSource.properties.timeBetweenEvictionRunsMillis=5000
```

### Test Results
Running `show full processlist` would display something like the following:
```
mysql> show full processlist;
+-----+---------+------------------+---------+---------+------+----------+-----------------------+
| Id  | User    | Host             | db      | Command | Time | State    | Info                  |
+-----+---------+------------------+---------+---------+------+----------+-----------------------+
|   2 | root    | localhost        | rundeck | Query   |    0 | starting | show full processlist |
| 203 | rundeck | 172.22.0.1:56492 | rundeck | Sleep   |    2 |          | NULL                  |
| 204 | rundeck | 172.22.0.1:45448 | rundeck | Sleep   |    2 |          | NULL                  |
| 205 | rundeck | 172.22.0.1:48274 | rundeck | Sleep   |    2 |          | NULL                  |
| 206 | rundeck | 172.22.0.1:51028 | rundeck | Sleep   |    2 |          | NULL                  |
| 207 | rundeck | 172.22.0.1:53740 | rundeck | Sleep   |    2 |          | NULL                  |
| 208 | rundeck | 172.22.0.1:56334 | rundeck | Sleep   |    2 |          | NULL                  |
| 209 | rundeck | 172.22.0.1:58902 | rundeck | Sleep   |    2 |          | NULL                  |
| 210 | rundeck | 172.22.0.1:33372 | rundeck | Sleep   |    2 |          | NULL                  |
| 211 | rundeck | 172.22.0.1:36322 | rundeck | Sleep   |    2 |          | NULL                  |
| 212 | rundeck | 172.22.0.1:38970 | rundeck | Sleep   |    0 |          | NULL                  |
+-----+---------+------------------+---------+---------+------+----------+-----------------------+
```
The sleep command time would go up to 5 or at most 10, then it would reset as the tomcat JDBC pool runs the validation query. This indicates all open connections are in the pool.

Running a 20 second sleep job ref I could see one connection would go past 10 seconds and right up until the job finishes. This indicated that the connection was checked out of the pool even though there was no command running on it.

### Root Cause
It turns out that the `ExecutionService` class is marked `@Transactional` causing every method to be implicitly wrapped in a transaction and thus Hibernate session. This includes `executeWorkflowStep` which is the entry-point from the `ExecutionServiceImpl`.  Because of this implicit transaction, a hibernate session was being kept open in the background while the step was executed.

### Solution
Logically `runJobRefExecutionItem` is multiple units of work that were all being wrapped up as one:
* [T]Validation and book keeping
* Execute the workflow
* [T]Validation and book keeping

By adding `@Transactional(propagation = Propagation.NOT_SUPPORTED)` to `executeWorkflowStep` we opt out of the implicit transaction. From there sessions are manually opened via `withTransaction`, variables definitions used in the closures hoisted, and some general teasing apart allows for the execution to proceed without holding a database connection open.

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
@@ -3506,7 +3506,6 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
         long startTime = System.currentTimeMillis()
 
         def wresult = metricService.withTimer(this.class.name, 'runJobReference') {
-            println("Made it!!!!")
             WorkflowExecutionService wservice = executionContext.getFramework().getWorkflowExecutionService()
 
             def timeoutms = 1000 * timeout
--- a/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
@@ -3476,21 +3476,16 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
         }
         def props = frameworkService.getFrameworkProperties()
         def disableRefStats = (props?.getProperty('rundeck.disable.ref.stats') == 'true')
-        ReferencedExecution refExec
 
 
         Execution exec = Execution.get(execid as Long)
-        refExec = new ReferencedExecution(scheduledExecution: se, execution: exec, status: EXECUTION_RUNNING).save()
-
+        def refId = saveRefExecution(EXECUTION_RUNNING, null, se.id, exec.id)
 
 
         if (!(schedlist[0].successOnEmptyNodeFilter) && newContext.getNodes().getNodeNames().size() < 1) {
             String msg = "No nodes matched for the filters: " + newContext.getNodeSelector()
             executionContext.getExecutionListener().log(0, msg)
-            ReferencedExecution.withTransaction { status ->
-                refExec.status = EXECUTION_FAILED
-                refExec.save()
-            }
+            saveRefExecution(EXECUTION_FAILED, refId)
 
             throw new StepException(msg, JobReferenceFailureReason.NoMatchedNodes)
         }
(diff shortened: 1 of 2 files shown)


rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -3473,34 +3485,28 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
...
             }
-
-            throw new StepException(msg, JobReferenceFailureReason.NoMatchedNodes)
         }
 
         long startTime = System.currentTimeMillis()
 
-        def wresult = metricService.withTimer(this.class.name,'runJobReference') {
+        def wresult = metricService.withTimer(this.class.name, 'runJobReference') {
+            println("Made it!!!!")
```
sjrd218: Let's kill the println.

PR #4661, opened 2019-03-28
Adds check for disabled flag for Community News

jessemarple:
Adds config check to disable Community News. Link persists. If user clicks on Community News, new tab is opened and page navigates to rundeck.com/community-updates


jessemarple:
I added to the docs in the Docs repo.

PR #4669, opened 2019-03-29
Use Password field type for password encryptor input

### LINKED ISSUE #4624: Make password utilitlity field Value To Encrypt to type password

I am working in a team with about 15 persons, where 3 of them move regulary to new teams. When I need to grant them access to rundeck, I need to open the password utlity tool (https://ecom-buildmaster.lx.hosts.intra/rundeck/passwordUtility/index) and let them enter their password. Before they enter their password I change the type of the password field from type text to type password, so I cannot see the value of the password in plain text during typing.

I see the encrypted password values copy them and start a rundeck job to add these new user data in the files ``server/config/realm.properties``. 

I dislike the password to be visible during typing by default.

1. The minimum idea is just to make the password field to ``<input type="password">`` by default. So you cannot see the typed secret.
2. Even better: use two fields to enter the password, this way you can be sure, there is no typo in the password field.
3. Best for my case, but just nice to have: Let the user select roles on this page as well and add a button "Save local user" to let him save its credentials and roles in the realm.properties just by clicking a button.

If there exists something similar already, please give me a hint. Thanks.

**Additional context**

Rundeck Version 3.0.8-20181029

sjrd218:
Add password encryption utility according to user's request.

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/plugins/ScriptPluginProviderLoader.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/plugins/ScriptPluginProviderLoader.java
@@ -749,7 +749,7 @@ public class ScriptPluginProviderLoader implements ProviderLoader, FileCache.Exp
         try {
             String date = getPluginMeta().getDate();
             return new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssX").parse(date);
-        } catch (IOException | ParseException e) {
+        } catch (IOException | NullPointerException | ParseException e) {
 
         }
         return null;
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/dtolabs/rundeck/core/plugins/ScriptPluginProviderLoader.java
```diff
@@ -749,7 +749,7 @@ public Date getPluginDate() {
         try {
             String date = getPluginMeta().getDate();
             return new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssX").parse(date);
-        } catch (IOException | ParseException e) {
+        } catch (IOException | NullPointerException | ParseException e) {
```
gschueler: I know this works, but it's a code smell... 👃  makes me cringe 😬

Issue #4416, opened 2019-01-23
bulk delete selection for executions issue


cwaltherf:
When performing a bulk delete for executions, clicking on select all in the first page, and then navigating to the second page, on the second page it's not possible to select more executions.

Also, when returning back to the first page, the selection is still active, but when clicking on delete it does not delete the selection anymore.

Tested on Rundeck Pro 3.0.12 any installation any OS.

jessemarple:
Reviewed Jan 23 2019

carlosrfranco:
PR already merged on 18 Apr
https://github.com/rundeck/rundeck/pull/4673

PR #4673, opened 2019-03-29
Allow multi-page selections on job exections history report #4416

carlosrfranco:
This changes allow user to select executions in more than one page for bulk delete.


jtobard:
This works on the activity list on the job page but this should also work on the activity page.

carlosrfranco:
Demonstration of "bulk delete" functionality by selecting multiple pages simultaneously:
multi_page_selection_bulk_delete

Issue #4643, opened 2019-03-22
"Attached as file to Email" should be default when upgrading Rundeck


hs-hub-world:
When introducing new feature (i.e. attach log as file or inline) could you please make one of them a default value (i.e. Attached as file to email should be default) especially when upgrading Rundeck to newer version that means now we have to go back and update all the jobs.  Please see image below.

![image](https://user-images.githubusercontent.com/20629911/54837294-3bd8c700-4c83-11e9-8f58-88e4af71cdc6.png)

Thank you

PR #4688, opened 2019-04-03
compatibility for old versions of email notification (issue #4643)

ltamaster:
Fixing https://github.com/rundeck/rundeck/issues/4643

compatibility for old versions of email notification (log output as attached file)


rundeckapp/grails-app/services/rundeck/services/NotificationService.groovy
```diff
@@ -283,6 +283,11 @@ public class NotificationService implements ApplicationContextAware{
                         if (configAttachLogInFile in ['true',true]) {
                             attachlog = true
                         }
+
+                        //for old versions support
+                        if(!attachlog && !attachlogbody){
```
jtobard: You could add a unit test for this to avoid that by mistake that code goes back in a future change.

Issue #2531, opened 2017-05-26
Configuring an option, impossible to let the Default Value empty, even with default values declared in JSON


fgtl:
Issue type: Bug report

**My Rundeck detail**

* Rundeck version: 2.8.2
* install type: war or launcher?
* OS Name/version: Linux 3.2.0-4-amd64 #1 SMP Debian 3.2.73-2+deb7u3 x86_64 GNU/Linux
* DB Type/version: unknown

**Expected Behavior**

In a scheduled **Job**, configuring an **Option**, declaring **Allowed Values** as a **Remote URL**, referencing a JSON containing default values. (using the selected=true property)
Since the JSON contains default values, I should be able to save my **Option** without a **Default Value** specified. (maybe showing a warning asking to double check that default values are indeed contained in the JSON)

**Actual Behavior**

Either :
- if the task is not scheduled: works as expected
- if the task is scheduled: impossible to empty **Default Value** (error message when saving : “Specify a Default Value for Required options when the Job is scheduled.”)

![default](https://cloud.githubusercontent.com/assets/2150986/26493973/c5c8ca8c-421c-11e7-95f6-fda4cefad57e.png)

**How to reproduce Behavior**

See Expected Behavior, and do a dichotomy between scheduled/unscheduled and filled/empty **Default Value**.

fsampaio:
I got the same error message using 2.8.2 and 2.10 version.

@fgtl did you find a workaround for this issue?

fgtl:
@fsampaio Unfortunately not, I'm sorry to say.

PavloRos:
Hi, 
I have the same problem.  Can I use variables/regex/links in **"default value"**?

maturbet:
+1

PR #4694, opened 2019-04-08
Changes to allow option default value empty with default value declared in JSON (remote options) #2531

carlosrfranco:
With this changes, if a option get values from a remote JSON with a declared default value, so the options do not need to have a default value when it is a required option on scheduled execution


gschueler:
ref #2531

PR #4697, opened 2019-04-09
Add configuration for hiding the user/password login fields when SSO 

sjrd218:
Also allows the local login form to be suppressed until a user logs in via SSO.
This behavior is primarily for Test Drives.


sjrd218:
Docs: https://github.com/rundeck/docs/pull/232

PR #4700, opened 2019-04-09
Flow Control fail flag right behavior

jtobard:
`FlowControl.java` interface defines the Halt method:
`public void Halt(boolean success);`

Currently `FlowControlWorkflowStep` plugin passes `fail` boolean instead of success, to fix this the plugin should call `Halt(!fail)`.

Unit test added to ensure the correct use of `halt` and `fail` properties.


gschueler:
haha 🤦‍♂️

PR #4701, opened 2019-04-09
User group plugin type

sjrd218:
A new plugin type that will add roles to a user upon login.


core/src/main/java/com/dtolabs/rundeck/core/execution/service/ScriptUserGroupSource.java
```diff
@@ -0,0 +1,103 @@
...
+                            groups.add(line);
+                        }
+                    }
+                }
+            } else {
+                LOG.error(new String(errStream.toByteArray()));
+            }
+
+        } catch(Exception ex) {
+            ex.printStackTrace();
```
gschueler: perhaps log the exception via Logger

sjrd218: updated

sjrd218:
Docs here: https://github.com/rundeck/docs/pull/231

PR #4710, opened 2019-04-11
Modularization of "run job later" jobs reschedule routines

ahormazabal:
This PR provides some modularization of code to reschedule ad-hoc scheduled jobs ("Run job later"), in order to be able to validate and reschedule a list of such executions.


ahormazabal:
Added a new test to validate the new method.

rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
```diff
@@ -693,10 +693,27 @@ class ScheduledExecutionService implements ApplicationContextAware, Initializing
...
+        [jobs: succeededJobs, failedJobs: failedJobs, executions: adhocRescheduleResult.executions, failedExecutions: adhocRescheduleResult.failedExecutions]
+    }
+
+    /**
+     * Reschedule the provided ad-hoc scheduled executions. Invalid executions will be cleaned up.
+     * @param serverUUID
+     * @param project
+     * @return
+     */
+    def rescheduleAdHocJobs(List<Execution> executionList) {
```
gschueler: my only concern is that this method name is confusing: `adhoc` is used to refer to Commands/executions not associated with a job. i believe it has also been used by the "run job later" mechanism, as it is essentially the same process to schedule those executions, but this method should be renamed, perhaps to something like `rescheduleOnetimeJobExecutions` , so using "onetime" instead of "adhoc" for jobs executions scheduled at a single time

gschueler:
Thanks @ahormazabal !

PR #4743, opened 2019-04-23
Logout enhancements.

### LINKED ISSUE #4707: Cognito MFA in front of Rundeck - Logout users
After setting up Cognito MFA auth in front of Rundeck using Amazon ALB,  users are unable to log out properly from Rundeck, the user session does not expire and is able to log back in without typing credentials.

According to ALB documentation, "when an application needs to log out an authenticated user, it should set the expiration time of the authentication session cookie to -1 and redirect the client to the IdP logout endpoint (if the IdP supports one)"

All of these things have to be done by the application itself (Rundeck)

Be able to log out from Rundeck and get user's session killed.

sjrd218:
Allow custom cookies to be expired upon logout.
Make the logout redirection url configurable.
Fixes #4707.


sjrd218:
Docs here: https://github.com/rundeck/docs/pull/234

PR #4745, opened 2019-04-23
Fixes StaticHostKeyChecking when using password for git plugin via ssh

### LINKED ISSUE #4620: SCM Git plugins ignore Strict Host Key Checking value
**Describe the bug**
SCM Git Import/Export fails to setup or enable if git host is not in known_hosts.

**My Rundeck detail**

* Rundeck version: 3.0.17
* install type: docker
* OS Name/version: docker @ ubuntu 16.04
* DB Type/version: H2

**To Reproduce**
1. Go to 'Setup SCM' > 'Git Export' > 'Configure'
2. Fill in git ssh url (ssh://git@gitrepo/home/git/my-repo.git)
3. Set 'Authentication' > 'SSH: Strict Host Key Checking' to 'no'
4. Setup

**Error message**
Failed cloning... UnknownHostKey: gitrepo. RSA fingerprint is ...

sjrd218:
Fixes issue with StaticHostKeyChecking when using username and password credentials in git scm plugin.

### DIFF
--- a/plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/BaseGitPlugin.groovy
+++ b/plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/BaseGitPlugin.groovy
@@ -21,6 +21,7 @@ import com.dtolabs.rundeck.core.plugins.configuration.Validator
 import com.dtolabs.rundeck.core.plugins.views.Action
 import com.dtolabs.rundeck.core.storage.ResourceMeta
 import com.dtolabs.rundeck.plugins.scm.*
+import com.jcraft.jsch.Session
 import org.apache.log4j.Logger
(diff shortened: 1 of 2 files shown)


plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/BaseGitPlugin.groovy
```diff
@@ -514,6 +520,7 @@ class BaseGitPlugin {
 
             //test url matches origin
             def config = agit.getRepository().getConfig()
+            println config
```
gschueler:
```suggestion
```

gschueler:
a later todo for this would be to simply use PluginSshSessionFactory after some slight refactoring i think

PR #4747, opened 2019-04-23
Fixes #4622. Multi-valued option now works with option values plugins.

### LINKED ISSUE #4622: "Option Values Plugin" not working when Options --> Multi-valued=yes
**Describe the bug**
Using the "Option Values Plugin" works fine when Options --> Multi-valued="no", but when setting Options --> Multi-valued="yes" the options are not displayed.

**My Rundeck detail**

* Rundeck version: TEAM 3.0.14
* install type: WAR
* OS Name/version: CentOS 7
* DB Type/version: H2 (testing environment)

**To Reproduce**
1.-  Create an Option Values Plugin
2.- Set Options --> Multi-valued = "no"
3.- Before starting job one can see all the options displayed in a drop-down list (as expected)
4.- Edit job and set Options --> Multi-valued = "yes"
5.- Before starting job, one can not see any of the options, therefore you can not select any

**Expected behavior**
When setting Options --> Multi-valued = "yes" one should have the possibility to see and select multiple options

**Screenshots**
Options_Values_1

Options_Values_2

Options_Values_3

 - Browser Chrome Version 72.0.3626.121 (Official Build) (64-bit)

sjrd218:
Fixes #4622. Multi-valued option now works with option values plugins.

### DIFF
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/ScheduledExecutionController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/ScheduledExecutionController.groovy
@@ -2915,6 +2915,7 @@ class ScheduledExecutionController  extends ControllerBase{
             }
             if(opt.optionValuesPluginType) {
                 opt.valuesFromPlugin = optionValuesService.getOptions(scheduledExecution.project,opt.optionValuesPluginType)
+                opt.values = opt.valuesFromPlugin.collect { it.value } as SortedSet
             }
         }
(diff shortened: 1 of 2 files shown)


gschueler:
Using SortedSet to sort the values might change the expected ordering

sjrd218:
True, but since the values variable only takes a sorted set I'm not sure there's an option.

sjrd218:
I moved the values population from the server side to the client side. Values are now rendered in the order they come from the plugin.

Issue #4699, opened 2019-04-09
Exclude Filter Ignored for Scheduled Job but not Adhoc Run


benruset:
**Describe the bug**
I have a job that runs on certain nodes based on the presence of certain AWS tags and a regex on hostnames. When the job runs via the scheduler any of the exclusions from the exclude filter are ignored. When I run the job by hand they are honored. 

**My Rundeck detail**

* Rundeck version:   3.0.19-20190327 
* install type: RPM
* OS Name/version: Amazon Linux 1
* DB Type/version: MySQL

**To Reproduce**
1. Create job with an exclude filter and schedule it.
2. When the job runs via the scheduler it ignores the exclude filter.
3. When the job is run manually it honors it.

**Expected behavior**
Exclude filter should be followed no matter how the job runs.

(Attaching an XML export of my job with the extension changed to .txt.)

[Nightly_Platform_Dev_Shutdown.txt](https://github.com/rundeck/rundeck/files/3059668/Nightly_Platform_Dev_Shutdown.txt)

benruset:
I just made another test job and the exclude filter was ignored for the ad-hoc job as well until I checked the "show excluded nodes" radio button.

reinerrdeck:
Also confirmed on Rundeck 3.0.20.

PR #4750, opened 2019-04-24
Fix: job exclude filter doesn't work for scheduled jobs

ltamaster:
Fix for https://github.com/rundeck/rundeck/issues/4699

Adding the exclude filter when the execution and the execution context are created.

**Additional context**
The exclude filter was just considered when the nodes are selected on the Run Now page.


gschueler:
fixes #4699

Issue #4760, opened 2019-04-26
Wrong list nodes using "Change the Target Nodes" with exclude filter


ltamaster:
**Describe the bug**
On a job with exclude filter, if you change the target nodes on the "Run Now" page, the result node list is wrong.

**My Rundeck detail**

* Rundeck version: 3.0.21
* install type: launcher
* OS Name/version: MacOs 10.14
* DB Type/version: mysql

**To Reproduce**
1. Create a job with exclude filter 
2. Got to run now page
3. Change the original target nodes
4. The job will execute on a wrong node list result

**Expected behavior**
Getting the node list selected on "Change the Target Nodes"

**Screenshots**
<img width="969" alt="Screenshot 2019-04-26 17 12 23" src="https://user-images.githubusercontent.com/6034968/56839640-7bca3580-6851-11e9-8410-5d5d465f313f.png">
<img width="481" alt="Screenshot 2019-04-26 17 12 43" src="https://user-images.githubusercontent.com/6034968/56839641-7bca3580-6851-11e9-9ca2-7ae4b6bd0419.png">

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser chrome
 - Version 74

PR #4761, opened 2019-04-26
Wrong list nodes using "Change the Target Nodes" with exclude filter

ltamaster:
For issue https://github.com/rundeck/rundeck/issues/4760

Cleaning the exclude filter when the target nodes were changed


gschueler:
added to maint-3.0.x

PR #4769, opened 2019-04-30
Enhance ACL policies to enable access rules that apply to users not in a group or not by username.

### LINKED ISSUE #3810: Enhance ACL policies to enable access rules that apply to users not in a group

The ACL policies apply to all groups/users that match the `by:` section. However, multiple roles/groups will apply all matching policies. There is no way to express that the policy rules should apply to anyone *not* in some group.  This is a problem if you want to blanket DENY certain access to everyone *except* some set of groups, you can't express it.

Add new syntax to invert the `by: user:` or `by: group:` selection:

possibly:

```{.yaml}
not-by:
  group:
  - groupA
  - '^someRegex.*$'
```

The ACL rules defined in that policy would only apply if the set of group roles do not include the matching groups.

Using a negative-lookahead regex such as `group: '^(?!someregex).*$'` won't work, because it will apply to any *other* groups the user is a member of, even if they are also a member of "someregex" group.

Currently the only solution is to make sure that ALLOW rules are defined only for certain groups, since any DENY rules will override any ALLOW rules.

jtobard:
A new ACL policy that applies to all users who are not in the specified group or username (including regular expression).

The new syntax replaces `by:` for `notBy:`

acl editor

acl page

As an example:
```
notBy:
  username: dev
description: run-job
for:
  job:
  - allow:
    - run
    equals:
      name: cleanTempFolders
context:
  project: example
```
Everyone that is not the user `dev` can run a job called `cleanTempFolders` in the project `example`

fix #3810

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/authorization/AclRule.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/authorization/AclRule.java
@@ -57,4 +57,6 @@ public interface AclRule {
 //    }
 
     public Set<String> getDenyActions();
+
+    public boolean isBy();
 }
--- a/core/src/main/java/com/dtolabs/rundeck/core/authorization/AclRuleBuilder.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/authorization/AclRuleBuilder.java
@@ -134,6 +134,11 @@ public class AclRuleBuilder {
         return this;
     }
 
+    public AclRuleBuilder by(final boolean isBy){
+        aclRuleImpl.by=isBy;
+        return this;
+    }
+
     public AclRule build() {
         return new AclRuleImpl(aclRuleImpl);
     }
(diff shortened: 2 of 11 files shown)


gschueler:
@jtobard what is the behavior in a `notBy` clause with multiple usernames?

jtobard:
@gschueler if the username does not correspond to the regular expression it takes the permissions.

using 
```
notBy: 
     username: deva|devb
```

The user `deva` is excluded also `devb`, but user `devc` gets the permissions. 
I'm going to include a test for this case.

gschueler:
I wonder if we need some safety rails on this.  the behavior for acls is "reject unless explicitly allowed".  however with the `notBy` clause it might be easy to accidentally grant access, since you can easily do `allow: ..., notBy: username: 'bob'`, which would grant it to everyone else.  thoughts @ahonor ?

ahonor:
True it does go a bit against the deny be default. What safety rails were you considering?

gschueler:
An idea is to make notBy only valid for `deny` rules. that would make it able to only restrict access, and not increase access to anyone. I think that is in the spirit of the original use case in #3810

jtobard:
I will modify the Pull request to reflect that restriction

Issue #4310, opened 2018-12-13
step labels from child jobs not listed in GUI Report view


cwaltherf:
When running a Parent job with job references and step labels, the step labels of the steps of a child job are not displayed in the GUI Report view.

![screenshot from 2018-12-13 16-41-34](https://user-images.githubusercontent.com/36487101/49963400-a34de080-fef6-11e8-96f8-f18cb7762218.png)
![screenshot from 2018-12-13 16-45-37](https://user-images.githubusercontent.com/36487101/49963401-a34de080-fef6-11e8-81c8-85d2f119ffd2.png)

gaetanquentin:
still need child's labels. Output is unreadable without it.

ProTip:
Fixed by #4655

ajxb:
Hi Guys,


I think that there might be a regression in RunDeck 3.0.21 that has re-introduced this bug.

In 3.0.20 I see the following, which is as expected following the introduction of the fix:
![image](https://user-images.githubusercontent.com/1756536/56954210-26827280-6b36-11e9-97c4-a8c8398fe923.png)

In 3.0.21 I see the behaviour that we saw before this fix was introduced:
![image](https://user-images.githubusercontent.com/1756536/56954291-5893d480-6b36-11e9-8b07-8f1f7e359b7b.png)

Cheers

reinerrdeck:
Confirmed.

On 3.0.20:

![3020](https://user-images.githubusercontent.com/43051294/56978152-d9f26380-6b44-11e9-9818-036a143bc2f4.png)

On 3.0.21:

![3021](https://user-images.githubusercontent.com/43051294/56978160-dfe84480-6b44-11e9-955a-510835c36d57.png)

jtobard:
Fixed on 3.1

![image](https://user-images.githubusercontent.com/2894508/61809531-7ad28100-ae0b-11e9-943d-f9017c763cb6.png)

PR #4772, opened 2019-04-30
Merging application properties for production mode. #4310

carlosrfranco:
There were two application config scopes for the production mode and the second overwritten the first one


ProTip:
Fixes #4310

gschueler:
added to maint-3.0.x

PR #4784, opened 2019-05-03
Add config option to enable unsanitized HTML log output Fix #4690

### LINKED ISSUE #4690: Turning on/off the HTML sanitizer / CSS styling for HTML Rendering
Hi,

I'm using 3.0.9-20181127 on Ubuntu 18.04 LTS.

According to:
https://docs.rundeck.com/docs/manual/content-converters/index.html#sanitized-html-output

When rendering formatted data as HTML a sanitizer is "normally" ran over the content to get rid of "tags and certain attributes".

I'm trying to display the output of pytest-html:
https://pypi.org/project/pytest-html/

These are standalone/self-contained HTML files with script elements among other things.

Unsurprisingly the HTML doesn't render properly in Rundeck.

My questions are:

1. How do I turn off/configure/restrict the HTML sanitizer?  
2. How do I get Rundeck to respect CSS declared within the HTML for rendering?

Ideally there would be a per-render setting, with the sanitizer turned on by default, but if it's possible to circumvent sanitization by writing a custom renderer let me know and I'll write one!

I've tried using various combinations of the embed, frame, and link elements to get around this (so the actual HTML I send to the renderer doesn't itself contain a script), but this doesn't seem to work either - I assume because these elements are also removed?

I also had a go of using the junit output and the junit2html tool.  This generates simpler HTML with no scripts, but Rundeck doesn't respect the CSS of the rendered content (I've seen this in previous uses too).

Another alternative would be a junit xml or pytest plugin specifically for Rundeck, but none seems to exist?

I have thought about writing my own Groovy plugin to XSLT transform the raw junit into Rundeck-compatible HTML (i.e. move all the style attributes to each and every element tag to force Rundeck to render the correct style)... I've done this on previous projects including a generic XML Renderer I wrote, but it's laborious when converters already exist.... it seems like the real missing feature is that we can't control things like CSS and Script content given Rundeck already has an HTML renderer.  It really restricts practical use of a potentially great feature IMHO.

Any other advice on how to best achieve this?

sjrd218:
Allow automatic html stripping to be turned off by adding content-meta:no-strip=true to the log entry metadata.


sjrd218:
I've added the requested features for disabling this behavior at both the framework and project level. I've also added the property to the RenderDatatypeFilterPlugin to control the setting. The framework and project settings are enabled by adding entries to the framework.properties and project.properties. Is that sufficient, or is there additional UI that needs to be created to control these settings?

gschueler:
If you want to enable this as a GUI checkbox for project, settings, you can add it here: https://github.com/rundeck/rundeck/blob/d949fc6ffd598879519d9183cefb3ae7fa7a5bfe/rundeckapp/grails-app/services/rundeck/controllers/MenuService.groovy#L11-L51

or in another bean implementing ProjectConfigurable.  

As long as it is documented, GUI presentation can be added later

gschueler:
Merge master again?

Issue #4798, opened 2019-05-08
Scheduled jobs do not respect the cluster remote execution policy.


carlosrfranco:
Customers on cases # 1827 and 1792 are facing the same issue. Scheduled jobs dot respect the cluster remote execution policy

They are using three cluster members, one front end and two workers. No job should run on the frontend node. To achieve that they are using remote execution policies. If they schedule a job, edit it on the front end job, the schedule owner will be the front end node. The job will run on the frontend node no matter what the remote cluster policies say.

PR #4799, opened 2019-05-08
Changes to allow schedule jobs to respect the cluster remote execution policy when it exists #4798

carlosrfranco:
This change makes possible to determine a node to run a scheduled execution on a cluster in accordance with the policy defined.


rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
```diff
@@ -1077,7 +1086,7 @@ class ScheduledExecutionService implements ApplicationContextAware, Initializing
         }
 
         log.info("scheduled job ${se.extid}. next run: " + nextTime.toString())
-        return nextTime
+        return [nextTime, jobDetail?.getJobDataMap()?.get("serverUUID")]
```
jtobard: I think a unit test in `ScheduledExecutionServiceSpec` is missing for this new return value

core/src/main/java/com/dtolabs/rundeck/core/schedule/JobScheduleManager.java
```diff
@@ -67,4 +67,15 @@
      * @return true if the scheduleOWner should change to current node.
      */
     boolean updateScheduleOwner(String name, String group, Map data);
+
+    /**
+     * In cluster mode, return true if the scheduleOWner should change to current node.
```
gschueler: This javadoc description is not correct

PR #4810, opened 2019-05-09
Fix #4703 node detail popup cannot be dismissed

### LINKED ISSUE #4703: GUI: popover will not disapear when selecting a node - BUG
**Describe the bug**

When we select a Node from a project and add it to run a command, the popover will not disappear even when the node it's already selected 

**My Rundeck detail**

* Rundeck version: 3.0.20 
* install type: war

**To Reproduce**
1. Click on 'commands'
2. Click on 'show all nodes'
3. Select a node
4. Popover remains visible even after selecting the node
5. The popover disapears only when changing to another tab menu option

**Expected behavior**

The popover should disappear after selecting the node

**Screenshots**

imagen

**Additional context**
This behavior was not noted in 3.0.19 or earlier versions.

gschueler:
* add knockout dom disposal callback
* constrain popup viewport to main-panel
* add mousedown handler to close popup for event outside of popup
* create unique id for interactive elements in referenced html content when copied to popup

Fix #4703

### DIFF
--- a/rundeckapp/grails-app/assets/javascripts/application.js
+++ b/rundeckapp/grails-app/assets/javascripts/application.js
@@ -821,7 +821,9 @@ function _initPopoverContentRef(parent, options) {
     var opts = {
       html: true,
       content: function () {
-        return jQuery(ref).html();
+        const id = generateId()
+        const html = jQuery(ref).html()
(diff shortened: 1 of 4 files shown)


gschueler:
also in maint-3.0.x

PR #4811, opened 2019-05-09
fix #4809 more/less text not changing for job description in group

### LINKED ISSUE #4809: "more/less" link not correct
Rundeck version: 3.0.21-20190424
The "more/less" link always shows 'Less" instead of changing between More/Less.

image

Thank you,
HS

gschueler:

Fix #4809 

CSS was too lenient


gschueler:
added to maint-3.0.x

PR #4820, opened 2019-05-14
Fix #4765. Add ability to require a role in order for user to log in.

### LINKED ISSUE #4765: Default security role for Active Directory authenticated user?
**Context**
Hello everyone,
We just have deployed 2 Rundeck Pro servers on Azure Cloud and tried to use JAAS module to authenticate with LDAP over Azure Active Directory Domain Services. AADDS is an abstracted layer of Azure Active Directory which permit us to run LDAP query over it (AAD is flat by design and not LDAP requestable). AADDS structure let us to only use one OU (= AADDC Users) to target both userBaseDn and roleBaseDn.

**Describe the bug**
My problem is that anyone from my Users OU can authenticate. Whatever the role assigned to them. They can access the Rundeck page with no permissions since they are in the right OU. Other role-assigned users can authenticate and are granted based on the admin.aclpolicy file as expected.

I presume this is due to a basic secutiy role set somewhere. I know that previous version of Rundeck allowed us to modify the web.xml file to remove basic "user" security role. But now that this is handled diffrently I can't find anything about that.

Thank you in advance for all the help you can give me.

**Expected behavior**
Since we can't use LDAP filters. I would like to know if we can authenticate users based on their assigned roles. 

---

* jaas-ldap.conf

activedirectory {
    com.dtolabs.rundeck.jetty.jaas.JettyCachingLdapLoginModule required
    debug="true"
    contextFactory="com.sun.jndi.ldap.LdapCtxFactory"
    providerUrl="ldap://rlxcom.onmicrosoft.com:389"
    bindDn="myBindUser@myDomain.com"
    bindPassword="myPassword"
    authenticationMethod="simple"
    forceBindingLogin="true"
    userBaseDn="OU=AADDC Users,DC=myDomain,DC=com"
    userRdnAttribute="sAMAccountName"
    userIdAttribute="sAMAccountName"
    userPasswordAttribute="unicodePwd"
    userObjectClass="user"
    roleBaseDn="OU=AADDC Users,DC=myDomain,DC=com"
    roleNameAttribute="cn"
    roleMemberAttribute="member"
    roleObjectClass="group"
    userLastNameAttribute="sn"
    userFirstNameAttribute="givenName"
    userEmailAttribute="mail"
    cacheDurationMillis="300000"
    reportStatistics="true"
    nestedGroups="false";
};

* admin.aclpolicy

description: Rundeck Admin access.
context:
  project: '.*' # all projects
for:
  resource:
    - allow: '*' # allow read/create all kinds
  adhoc:
    - allow: '*' # allow read/running/killing adhoc jobs
  job:
    - allow: '*' # allow read/write/delete/run/kill of all jobs
  node:
    - allow: '*' # allow read/run for all nodes
by:
  group: [mySecurityGroupAdm]

description: Rundeck Admin access.
context:
  application: 'rundeck'
for:
  resource:
    - allow: '*' # allow create of projects
  project:
    - allow: '*' # allow view/admin of all projects
  project_acl:
    - allow: '*' # allow admin of all project-level ACL policies
  storage:
    - allow: '*' # allow read/create/update/delete for all /keys/* storage content
by:
  group: [mySecurityGroupAdm]

---

**My Rundeck detail**

* Rundeck version: 3.0.19-20190327
* install type: rpm
* OS Name/version: RHEL Server release 7.6 (Maipo)
* DB Type/version: mysql

sjrd218:
Fix #4765. Add ability to require a role in order for user to log in.

### DIFF
--- a/rundeckapp/grails-app/controllers/rundeck/interceptors/SetUserInterceptor.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/interceptors/SetUserInterceptor.groovy
@@ -2,9 +2,11 @@ package rundeck.interceptors
 
 import com.dtolabs.rundeck.core.authentication.Group
 import com.dtolabs.rundeck.core.authentication.Username
+import grails.plugin.springsecurity.SpringSecurityUtils
 import org.rundeck.web.infosec.AuthorizationRoleSource
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.context.ApplicationContext
+import org.springframework.security.core.context.SecurityContextHolder
 import rundeck.AuthToken
 import rundeck.User
 import rundeck.services.UserService
@@ -18,7 +20,7 @@ class SetUserInterceptor {
     ApplicationContext applicationContext
 
     UserService userService
-
+    def messageSource
     int order = HIGHEST_PRECEDENCE + 30
 
     SetUserInterceptor() {
@@ -97,6 +99,18 @@ class SetUserInterceptor {
             request.errorCode = 'request.authentication.required'
             render view: '/common/error.gsp'
             return false
+        }
+        def requiredRole = grailsApplication.config.rundeck.security.requiredRole
+        if(!requiredRole.isEmpty()) {
+            if(!request?.subject?.principals?.findAll { it instanceof Group }?.any { it.name == requiredRole }) {
+                log.error("User must have role: ${requiredRole} to log in.")
+                SecurityContextHolder.clearContext()
+                request.logout()
+                response.status = 403
+                flash.loginerror = messageSource.getMessage("user.not.allowed",null,null)
+                render view: '/user/login.gsp'
+                return false
+            }
         }
         return true
     }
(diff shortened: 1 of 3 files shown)


rundeckapp/grails-app/controllers/rundeck/interceptors/SetUserInterceptor.groovy
```diff
@@ -98,6 +100,18 @@ class SetUserInterceptor {
             render view: '/common/error.gsp'
             return false
         }
+        def requiredRole = grailsApplication.config.rundeck.security.requiredRole
+        if(!requiredRole.isEmpty()) {
+            if(!request?.subject?.principals?.findAll { it instanceof Group }?.any { it.name == requiredRole }) {
+                log.error("User must have role: ${requiredRole} to log in.")
```
gschueler: perhaps include the subject name in the error message for the server log

rundeckapp/grails-app/i18n/messages.properties
```diff
@@ -1237,6 +1237,7 @@ set.as.default.filter=Set as Default Filter
 remove.default.filter=Remove Default Filter
 run.a.command.on.count.nodes.ellipsis=Run a command on {0} {1} …
 not.authorized=Not authorized
+user.not.allowed=User not allowed
```
gschueler: Hmm , just a suggestion for rephrasing, but not necessary.
```suggestion
user.not.allowed=Authorization was not within the required role.
```

sjrd218: I would be concerned that this might be giving a user too much information.

ahonor:
Rundeck 2.x would indicate !role to basically mean user did not belong to required role. At a minimum, this scenario has to be logged to a file because it will be a source of Admin setup confusion. I don't mind it being a message though.

sjrd218:
@ahonor The log that appears in the service.log would look like: `User admin must have role: rundeck_required to log in.` Is that sufficient?

PR #4826, opened 2019-05-16
Issue/4825

### LINKED ISSUE #4825: Update Docker image to work with OpenShift
OpenShift runs containers with a random UID for _reasons_. This requires that all files and directories that need to be read and written are in the root group. The user should default to the root group as well so that anything it creates will default to that group.

ProTip:
Update Docker images so they will boot on OpenShift.

Fixes #4825


eniomachado:
Hi @ProTip,
I'm getting an error:

Writing manifest to image destination
Storing signatures
STEP 1: FROM rundeck/ubuntu-base@sha256:4c2bf3da1b01cbfae43a17b9e7967895166ef90133adc02ea2cb177eee1fab6c
STEP 2: COPY --chown=rundeck:root .build .
error: build error: error building at STEP "COPY --chown=rundeck:root .build .": error adding sources [/tmp/build/inputs/.build]: error checking on source /tmp/build/inputs/.build under "/tmp/build/inputs": copier: stat: "/.build": no such file or directory

Issue #4742, opened 2019-04-23
Updating project settings doesn't refresh the settings immediately 


ltamaster:
**Describe the bug**
When you save a change on the project settings menu, the changes are not "shown" immediately. Instead, you need to wait (kind of cache) or perform the update twice.
This happens when you store the project settings on the database. 

It seems this PR, in order to avoid deadlocks on storage table, produce the issue.
https://github.com/rundeck/rundeck/pull/4522

**My Rundeck detail**

* Rundeck version: 3.0.20
* install type: launcher
* OS Name/version: MacOs 10.14
* DB Type/version: mysql

**To Reproduce**
1. Go to Edit Configuration
2. Add any change 
3. Save the settings
4. Go again to the Edit Configuration page and you will see the previous version of the settings.

**Expected behavior**
Changes in project settings should be applied it immediately

**Desktop (please complete the following information):**
 - OS: MacOs
 - Browser Chrome
 - Version 73

fgutierrezz:
this happens in 3.0.19 , 3.0.21 also

DVGeoffrey:
I am experiencing this problem on 3.0.19

gschueler:
needs backport to 3.0 as well

ProTip:
While reviewing the PR for a fix that didn't involve reverting the changes to fix #4269 introduced in #4522, I ended up reviewing those changes as well and discovered the precise reason(I hope!) it caused the issue we are seeing with the stale project settings. I also believe we should revert those changes and tackle the lock timeout issue separately if it re-appears.
 
PR #4522 introduced a new `withNewTransaction` closure [here](https://github.com/rundeck/rundeck/pull/4522/files#diff-7bf7c6a1d5fb256a3a46a7101cf128cbR290). Unfortunately this has two effects:
1. It bypasses the current hibernate session and transaction, instead opening up a transaction on a new connection. The transaction is not "nested" or piggy-backed onto the current transaction.
2. It caused the return [here](https://github.com/rundeck/rundeck/pull/4522/files#diff-7bf7c6a1d5fb256a3a46a7101cf128cbR313) to be the return for the new closure, not the return for the outer closure. Now `saveStorage` always returns false, and the save occurs again inside a `withSession` also bypassing the current session/transaction.

So after these changes _all_ `saveStorage` calls end up outside the existing session/transaction and the storage, if previously loaded, becomes stale in the outer session.

[...]

Issue #4269, opened 2018-11-30
Lock wait timeout exceeded


vvbogdanov87:
From time to time when I save a job, error `Lock wait timeout exceeded; try restarting transaction.` appears. Only restart of Rundeck helps to resolve the issue. A similar problem was fixed by #4169 but the fix was merged to branch release-2.11

**My Rundeck detail**
* Rundeck version: 3.0.8-20181029
* install type: Docker(deb)
* OS Name/version: ubuntu 16.04
* DB Type/version: mysql

**To Reproduce**
1. Go to 'edit job'
2. Click on 'save'

**Desktop**
 - OS: ubuntu 18.10
 - Browser chrome
 - Version 70

<details>
  <summary>Stacktrace</summary>
  2018-11-28 12:13:11.238 ERROR --- [p473344802-1251] o.g.web.errors.GrailsExceptionResolver   : MySQLTransactionRollbackException occurred when processing request: [POST] /project/spark/job/update                                                                                                                                                      [90/1874]
Lock wait timeout exceeded; try restarting transaction. Stacktrace follows:                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                  
org.springframework.dao.CannotAcquireLockException: Hibernate operation: could not execute statement; SQL [n/a]; Lock wait timeout exceeded; try restarting transaction; nested exception is com.mysql.jdbc.exceptions.jdbc4.MySQLTransactionRollbackException: Lock wait timeout exceeded; try restarting transaction                                            
        at org.springframework.jdbc.support.SQLErrorCodeSQLExceptionTranslator.doTranslate(SQLErrorCodeSQLExceptionTranslator.java:259)                                                                                                                                                                                                                           
        at org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:73)                                                                                                                                                                                                                      
        at org.grails.orm.hibernate.GrailsHibernateTemplate.convertJdbcAccessException(GrailsHibernateTemplate.java:731)                                                                                                                                                                                                                                          
        at org.grails.orm.hibernate.GrailsHibernateTemplate.convertHibernateAccessException(GrailsHibernateTemplate.java:719)                                                                                                                                                                                                                                     
        at org.grails.orm.hibernate.GrailsHibernateTemplate.doExecute(GrailsHibernateTemplate.java:303)                                                                                                                                                                                                                                                           
[...]

PR #4871, opened 2019-05-24
fixing updating project settings

ltamaster:
backing to the previous version to fix https://github.com/rundeck/rundeck/issues/4742

the current version of DbStorageService produce the refresh issue


ProTip:
Is this going to re-introduce #4269 ?

ltamaster:
@ProTip I tried something different that worked for me.

PR #4925, opened 2019-06-07
Sets default nodeJS version for bulding.

ahormazabal:
I noticed that the build of grails-spa fails if node is not installed on the building machine. I replicated this behavior on Ubuntu, OSX and Windows 7, even if i clone a fresh copy of master.

Setting a default value for node.install property solves the problem and eliminates the need of installing nodejs on the building machine.

I chose to set it on the general gradle.properties instead of grails-spa gradle.properties to keep al versions settings on the same file.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -28,3 +28,4 @@ groovyVersion = 2.4.15
 jettyVersion=9.4.11.v20180605
 mavenCentralUrl = http://repo1.maven.org/maven2/
 grailsCentralUrl = http://grails.org/plugins
+node.install=10.16.0


ProTip:
@ahormazabal NodeJS is a build time dependency. This was left off by default as it bypasses a developers local nodejs installation.

ahormazabal:
oh i see. Maybe we should we describe this behavior on README.md then? If you don't have node installed on your machine  the build fails and knowing about -Pnode.install parameter would be very useful for the cases on which you don't want to install node or the node version installed breaks the build (for example last time i checked node 12 breaks the rundeckpro build)

gschueler:
i'm kind of in favor of this change, as it makes it easier to build from scratch for new contributors.

gschueler:
@ProTip perhaps we invert the flag behavior, so you have to supply a flag to use local install and not the gradle node install instead.  we would do that in ci

ahormazabal:
From what i see at the gradle nodejs plugin docs, the problem is the behavior of the version setting: "If npmVersion is specified, the plugin installs that version of npm into npmWorkDir by the npmSetup task and use it. If npmVersion is not specified and a locally-installed npm exists, the plugin will use it. Otherwise, the plugin will use the npm bundled with the version of node installation."

To invert the behavior we would have to set an special value. Like node.install=system. And do something like:
```
if node.install == "system"
       version=null
  else if node.install==aVersion
       version=aVersion
  else if node.install == null
       version = defaultVersion.
```

ProTip:
@ahormazabal Yes, that's the gist of it. An issue/PR was opened a bit back that we ended up requesting it be modified for this reason; setting the install option always bypasses the local node/npm installation even if it exists already.

I think this is an odd way to develop personally and typically expect `node_modules` to represent what's installed and in use, and for `npm link`s to work as well. @jessemarple may have some input on this.

I don't believe requiring installing nodesj alongside java is too much of a burden for development; a lot of people may not even have a java sdk installed in their dev environment. However if we do move forward with this it may be good to override this in all of our CI/CD builds so we can take advantage of the nodejs installation already present in the build images.

gschueler:
let's go forward with @ahormazabal's suggestion. the build docs can call out the behavior and inform users to use the `-Pnode.install=system` if they don't want gradle to do the install

PR #4931, opened 2019-06-07
Fix #4930 for ReloadablePropertyFileLoginModule.

### LINKED ISSUE #4930: Jaas login using PropertyFileLoginModule adds username as a role
PropertyFileLoginModule and ReloadablePropertyFileLoginModule add the username to the list of roles when logging in. 

Expected behavior:
The username should not be added to the role list.

sjrd218:
Do not add the username to the role list.


rundeckapp/src/main/groovy/org/rundeck/jaas/jetty/JettyAuthPropertyFileLoginModule.java
```diff
@@ -48,7 +48,8 @@ public void initialize(Subject subject, CallbackHandler callbackHandler, Map sha
             logger.debug("using reloadable realm property file reader");
             module = new ReloadablePropertyFileLoginModule();
         } else {
-            module = new PropertyFileLoginModule();
+            module = new ReloadablePropertyFileLoginModule();
```
gschueler: why not just simplify the whole if clause?

sjrd218: Nice catch.

gschueler:
some conflicts with master @sjrd218

PR #4937, opened 2019-06-11
UI fixes, minor css and html changes, flattened

jessemarple:
A number of minor CSS and HTML issues. 

- Incidents of <div> having a class of "row" creating a card with a width 30px wider than other cards. Selected without the additional "row" class as standard/default UI.
- Sets "primary" buttons to have a 2x border but keeping the 1px external border and adding a 1px inset boxshadow. 
- Removing instances of "btn-simple" to keep with the button style usage.
- Adds property of "word-break":"break-all" where visually seen. (this occured, for example, on the BASE shown in System Report"
- Minor changes to Repository/Plugins UI


jessemarple:
Bypassing Travis, merging.

PR #4940, opened 2019-06-11
[draft] Cleanup Workflow execution

ahormazabal:
This PR allow to specify a second Workflow for a Job, to run as a "Cleanup Workflow".
Currently, if a cleanup workflow is defined for a job or an execution, the system will run the second workflow once the first workflow is completed (successful or not), using the same context.
If the second workflow fails, the succeed status of the first workflow will be preserved.

To create a job with a cleanup workflow, the job import api or the rd cli can be used, specifying a "cleanupSequence" section with the same format as the original "sequence" entry.  Example:
```
- defaultTab: nodes
  description: test description
  executionEnabled: true
  loglevel: INFO
  name: Testjob2
  nodeFilterEditable: false
  options:
  - label: wait
    name: wait
    regex: '[0-9]+'
    value: '15'
  scheduleEnabled: true
  sequence:
    commands:
    - script: |-
        echo "test @option.wait@:" $(date)
        echo "=================================="
        sleep @option.wait@
        echo "done" $(date)
    keepgoing: false
    strategy: node-first
  cleanupSequence:
    commands:
    - exec: echo "test" $(date)
    - exec: echo "testing"
    keepgoing: false
    strategy: node-first

```

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/execution/MultiWorkflowExecutionServiceThread.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/execution/MultiWorkflowExecutionServiceThread.java
@@ -35,7 +35,6 @@ import java.util.function.Supplier;
  * Implementation of {@link WorkflowExecutionServiceThread} which implements the chained execution of
  * multiple workflows after the execution of a primary execution workflow.
  *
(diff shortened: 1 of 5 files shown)


ahormazabal:
I'm closing this PR, as this topic should probably be approached using the new execution plugin.

PR #4965, opened 2019-06-17
Execution mode status API with failing code when status is passive.

ahormazabal:
Issue: https://github.com/rundeckpro/rundeckpro/issues/606

This PR creates a new API endpoing ``/api/V/system/executions/status`` which returns the current
execution mode using the same response format as ``/api/V/system/executions/active`` and ``/api/V/system/executions/passive``.

Additionally, the endpoint will respond with a HTTP status of ``503 - Service Unavailable`` when the current status is **passive**.


ahormazabal:
Docs: https://github.com/rundeck/docs/pull/268

ahormazabal:
Updated api version to 32.

PR #4967, opened 2019-06-17
Added test and extra if conditional to fix #4959

### LINKED ISSUE #4959: Schedule DayOfWeek via terraform/API fails.
**Describe the bug**
When trying to use terraform to create a job schedule with cron of `0 0 0/2 ? * 1-5 *` it comes back with error `Crontab string format was invalid: 0 0 12 ? * ? *` Notice how it changes the day of week from 1-5 to `?`.  The invalid format is actually added to job and can be seen on UI.

**My Rundeck detail**

* Rundeck version: I've tried 2.7.1, 3.0.0, 3.0.22
* install type: docker https://hub.docker.com/r/jordan/rundeck
* OS Name/version: debian stretch
* DB Type/version: mariadb

**To Reproduce**
Go to https://github.com/terraform-providers/terraform-provider-rundeck/blob/master/rundeck/resource_job_test.go#L147 and change that schedule to be the above mentioned cron. Then run `make testacc` providing the mentioned ENV variables

**Expected behavior**
The cron should be set to run SUNDAY-THURSDAY, instead it gives invalid crontab information. 

**Additional context**
If you try to enter `0 0 0/2 ? * 1-5 *` into UI, it succeeds. If you try to enter the invalid `0 0 12 ? * ? *` it shows invalid. However, running it on terraform successfully puts the invalid in the UI.

moosilauke18:
Fixes issue with #4959 where dayofmonth is not null, it is `?`, but it sets dayofweek to also be `?`.

It simply looks if dayofmonth is set to `?`. If it is, then it goes to the next if.

### DIFF
--- a/rundeckapp/grails-app/domain/rundeck/ScheduledExecution.groovy
+++ b/rundeckapp/grails-app/domain/rundeck/ScheduledExecution.groovy
@@ -464,7 +464,7 @@ class ScheduledExecution extends ExecutionContext {
                     se.year = '*'
                 }
                 if(data.schedule.dayofmonth && data.schedule.dayofmonth instanceof Map
-                        && null !=data.schedule.dayofmonth.day){
+                        && null !=data.schedule.dayofmonth.day && '?' !=data.schedule.dayofmonth.day){
                     se.dayOfMonth = data.schedule.dayofmonth.day
                     se.dayOfWeek = '?'
                 }else if(data.schedule.weekday && data.schedule.weekday instanceof Map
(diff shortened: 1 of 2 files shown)


gschueler:
thanks!

gschueler:
I think this has never come up before because in normal usage, the `<day day="?"/>` would not be added to exported xml when `<weekday ..>` is included, and the import would succeed if the `<day .../>` is not present.  the terraform provider always generates that element

PR #4986, opened 2019-06-21
Small improvement to the API forecast

### LINKED ISSUE #4985: Endpoint that show past schedule executions
Actually, Rundeck can check the schedule executions based on a forecast as proactive behavior, specifically that:

https://docs.rundeck.com/docs/api/#get-job-forecast

Maybe is a good idea that forecast shows the past schedule executions to run scheduled executions that were missed.

Based on this:

https://stackoverflow.com/questions/56685061/how-to-run-rundeck-scheduled-executions-that-were-missed

jtobard:
fix #4985 

For those strange cases in which you need to check when a job should have run.

This is a new parameter for the endpoint:
```
GET /api/31/job/[ID]/forecast
```

Adding the parameter `past=true` will return an inverse forecast, that is, considering the current scheduler, when it should have run.
This forecast is only referential, since it will not take into account if the job could have been disabled or not yet been created.

A couple of people in the community channels requested this functionality as a nice to have for administration purpose.

### DIFF
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
@@ -2871,7 +2871,7 @@ class MenuController extends ControllerBase implements ApplicationContextAware{
         //future scheduled executions forecast
         def time = params.time? params.time : '1d'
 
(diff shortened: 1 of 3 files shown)


jtobard:
Documentation: rundeck/docs#271

Issue #4831, opened 2019-05-16
Admin profile page not should show all tokens from DB . 


fgutierrezz:
**Describe the bug**
It will be good to limit when the token table contains many rows. 

**My Rundeck detail**

* Rundeck version: 3.0.21 Cluster
* install type: launcher
* OS Name/version: ubuntu 18.04
* DB Type/version: postgres

**Expected behavior**
No errors expected

**Additional context**
when a high number of tokens its observer high CPU usage

ahonor:
How many is too many rows?

fgutierrezz:
With around 8000 tokens rows rundeck profile page becomes unstable and unresponsive, unable to open any rundeck link . The following screenshot was taken after 1 or 2 minutes, it shows what happen and not responding when clicking on the show token button. 

![Captura de pantalla de 2019-05-16 19-25-25](https://user-images.githubusercontent.com/30966360/57893432-e3502100-7810-11e9-8631-7e5e34595ac2.png)

After some minutes the page is blank

![Captura de pantalla de 2019-05-16 19-33-04](https://user-images.githubusercontent.com/30966360/57893564-7b4e0a80-7811-11e9-92ce-18b71e5e6c32.png)

PR #4990, opened 2019-06-24
Pagination of tokens on user profile page.

ahormazabal:
This PR adds pagination to the token section of the user profile page, in order to support a high amount of tokens.

Tested with 8000+ tokens. Loading time was reduced from 20s to ~300ms

The main changes on this PR are:
- Added a gsp-based paginator which will reload the webpage when a new page is requested.
- Token delete operation now work through http post instead of ajax in order to correctly display the pagination afterwards.
- Token create still works through ajax, but it will reload the page afterwards instead of injecting the new record on the table, for the pagination to display correctly.
- Added a new pagination taglib in order to support gsp pagination using bootstrap markup.


ahormazabal:
Fixes issue: https://github.com/rundeck/rundeck/issues/4831

ahormazabal:
This still needs to be backported to 3.0.x, i'll create a new PR for that.

rundeckapp/grails-app/assets/javascripts/user/profile.js
```diff
@@ -142,7 +142,8 @@ function generateUserToken(login, elem, data) {
         beforeSend: _createAjaxSendTokensHandler('api_req_tokens'),
         success: function (data, status, jqxhr) {
             if (data.result) {
-                addTokenRow(elem, login, data.tokenid);
+              // Reload current page.
+              window.location.reload();
```
sjrd218: Is there a way to detect if the adding a token to the row would push it over the page max and only then do a reload? The user interaction is less pleasant with the reload on every token creation. I'm guessing having over 20 tokens is a corner case anyway, so if we could only reload when absolutely necessary that would be great.

ahormazabal: Ok i adjusted it so it will only will reload if the insertion will overflow the table beyond the max page length.

ahormazabal:
finally the behavior will be the following:
- changed token ordering to newer to older.
- a row will be added dinamically to the table only if we are at the first page and the addition will not change the number of pages.
- a row will be deleted dinamically from the table only if we are at the last page and the deletion will not change the number of pages.
- on all other cases the page will be reloaded.
- The default page size was setted on 50, and added an option to the gui options to override. 

The practical effect of this behavior is that users with a low amount of tokens will only see the dynamic behavior, but as the number of tokens grow the page will switch to the gsp paginator.

PR #5003, opened 2019-06-27
Adds "Cancel" button to UI

jessemarple:
Adds a Cancel button to the PluginConfig UI in Project Nodes Config.

### DIFF
--- a/rundeckapp/grails-app/views/framework/projectNodeSources.gsp
+++ b/rundeckapp/grails-app/views/framework/projectNodeSources.gsp
@@ -133,7 +133,6 @@
 
                               <div class="panel panel-default">
                                 <div class="panel-heading">
-
                                   <g:link controller="framework" action="editProjectNodeSources"
                                           params="[project: params.project ?: request.project]"
                                           class=" btn btn-info btn-sm">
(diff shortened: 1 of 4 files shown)


jessemarple:
@gschueler I can't check in functional from formatting changes. The formatting changes were not made intentionally, but instead with VS Code as it maintains the best practice formatting automatically. I'll get you set up with a formatter such that when you save a file it automatically formats.

gschueler:
@jessemarple the reformatting is not the primary problem, it is the combining the of reformatting with other changes in the same commit.  I've written up my problems with this kind of change here https://github.com/rundeck/rundeck/wiki/Github-Guidelines#reformatting-code-and-non-functional-changes please try to separate the changes in the future.  (If you have to reformat the file, commit that reformatted file before making changes to it.)

jessemarple:
@gschueler I totally understand how problematic this is. The issue we're having isn't that the reformatting changed the code, it's that when the code was originally written, it was formatted incorrectly or not formated at all. Using the formats as defined in the grails-spa .eslintrc.js and .editorconfig files would eliminate the issue. The reason those files are included in the repo is for exactly that reason - to standardize the formatting of Vue and JS files across different editors and developers.

I'll make an update to the Wiki page to include that requirement.

jessemarple:
I've updated the Github Guildlines wiki to include information regarding the StandardJS ruleset and how to use it.

PR #5005, opened 2019-06-27
fix #5004 EnhancedNodeService bean was not loaded

### LINKED ISSUE #5004: Node Enhancer feature is not working
**Describe the bug**
Rundeck was not able to process the enhancement in the node although it is possible to configure on "Edit Node" view.

**My Rundeck detail**
--

**To Reproduce**
- Install a node enhancer plugin on rundeck
- Config the node enhancer on "Edit Node" view
- Note that the nodes were not enhanced

**Expected behavior**
The plugin should make the enhancement if configured

carlosrfranco:
EnhancedNodeService bean was not being loaded then the node enhancement was not processed.


gschueler:
can you create backport for 3.0?

PR #5012, opened 2019-06-28
Migrating the attribute-match-node-enhancer plugin to the rundeck oss core

carlosrfranco:
Migrating a plugin from https://github.com/rundeck-plugins/attribute-match-node-enhancer to Rundeck core. This plugin can fix an issue reported on pro repo: https://github.com/rundeckpro/rundeckpro/issues/607


gschueler:
@carlosrfranco we shouldn't move the code here, instead just add it as a build-time dependency like we do here: https://github.com/rundeck/rundeck/blob/master/rundeckapp/build.gradle#L68-L71

PR #5020, opened 2019-07-02
new Joblifecycle plugin type

ronaveva:
New capability that allows to execute a task before or after a job execution.

Pre job execution criteria:

1.-If the plugin throws an exception, the job execution will not start and the message provided by the exception will be shown in the job execution page
2.-If the plugin returns false, the job execution will not start and the message provided by the the plugin will be shown in the job execution page
3.-If the plugin returns null or true, the execution will continue as usual

Before job execution criteria:

1.-If the plugin throws an exception, the job will be shown as failed and it will stop the execution
2.-If the plugin returns false, the job will be shown as failed and it will stop the execution
3.-If the plugin returns null or true, the execution will continue as usual

After job execution criteria:

1.-If the plugin throws an exception, it will propagate, BUT it will not change the job result
2.-If the plugin returns false, it will throw an exception, BUT it will not change the job result
3.-If the plugin returns null or true, the execution will continue as usual

Plugins can be enable/disable at project level under the project configuration page. This will generate a new property in the project.properties file (2 by each plugin implementation)


gschueler:
A few notes:

* this plugin system only allows plugins to be enabled at project level, and does not seem to support custom plugin configuration properties.
* Plugins should receive all events, it should be up to the plugin to handle them or not.  for example you might want a plugin that needs to see the beforeJob and afterJob events, so it wouldn't make sense to disable one of the events.  If the plugin has customization options for how to behave, that should be part of the plugins custom configuration properties (see below)

I think this should be changed in a few ways:

1. instead of having beforeJob/afterJob toggles at project level for each plugin, the configuration should use the same "list of plugin configurations" pattern that we use for Resource model, Node Enhancer, and Node Healthcheck plugins.  (I can point out locations in code for examples)
    * There is already support in vue UI and project configuration to do this, so only minimal integration work needs to be done.
    * This also enables custom plugin property configuration at the project level
2. "Instance" scoped plugin config properties should be avilable when you edit a job, if any are available due to job Lifecycle plugins. (I can add this as a secondary enhancement.)
3. We might want to rename thse to just "JobPlugins", since the lifecycle part is just a detail of the way they work.

ronaveva:
Plugin type has been changed to "JobPlugin" instead of "JobLifeCyclePlugin".
Plugins can now be enabled/disabled completely under the project configuration page (not by event)

core/src/main/java/com/dtolabs/rundeck/plugins/jobs/JobPlugin.java
```diff
@@ -0,0 +1,34 @@
...
+ * Time: 10:22 AM
+ */
+public interface JobPlugin {
+
+    /**
+     * It triggers before the job starts
+     * @param event event execution data
+     * @return JobEventStatus
+     */
+    default public JobEventStatus beforeJobStarts(JobEvent event) throws JobPluginException{
```
gschueler: the plugin should return a java interface, not a class, perhaps just change JobEventStatus to an interface, and have a default implementation used in the app

core/src/main/java/com/dtolabs/rundeck/core/execution/WorkflowExecutionServiceThread.java
```diff
@@ -79,12 +82,14 @@ public void run() {
 
     public WorkflowExecutionResult runWorkflow() {
         try {
+            iJobPluginService.beforeJobStarts(new JobEventImpl(context));
```
gschueler: the plugin should be able to return a new execution context, and that new context should be used in the `executeWorkflow` on line 87.  When modifying contexts, the pattern is to use the `ExecutionContextImpl.builder(oldContext)` to create a builder from the original context. It can then be modified, and a new one built.  

So in this case, the job plugin service might pass in a builder instead of a StepExecutionContext to the JobPlugin, allowing the plugin to update the context if needed, before returning it to the WorkflowExecutionServiceThread

core/src/main/java/com/dtolabs/rundeck/core/jobs/JobEventStatus.java
```diff
@@ -0,0 +1,30 @@
+package com.dtolabs.rundeck.core.jobs;
+
+public class JobEventStatus {
```
gschueler: should be an interface, and should have a way to return a new `StepExecutionContext`.

PR #5021, opened 2019-07-02
Logged users

ronaveva:
It shows the users logged into rundeck (under the same page as the user summary)

User status can be:
1.-LOGGED IN: it will take as the last date the most recent date between the logged event and the execution of a job event (executed by the user)
2.-LOGGED OUT: if the user has explicitly logged out of rundeck
3.-ABANDONNED: if the user did not logout from rundeck and the last date (between logged in and execution job) has more then 30 minutes (*) compared to the current date.

Also, user summary page has been migrated to VUE.

*It can be modified by adding the property "framework.session.abandonned.minutes" under "framework.properties" file


rundeckapp/grails-app/services/rundeck/services/UserService.groovy
```diff
@@ -22,10 +22,23 @@ import com.dtolabs.rundeck.plugins.user.groups.UserGroupSourcePlugin
...
 
+
 @Transactional
 class UserService {
 
     FrameworkService frameworkService
+    private static final int DEFAULT_TIMEOUT = 30
+
+    enum LogginStatus{
+        LIN('LOGGED IN'),LOUT('LOGGED OUT'),ABND('ABANDONNED'),NL('NOT LOGGED')
```
gschueler:
```suggestion
        LIN('LOGGED IN'),LOUT('LOGGED OUT'),ABND('ABANDONED'),NL('NOT LOGGED')
```

ronaveva:
> There haven't been much activity here. This is stale. Is it still relevant? This is a friendly reminder to please resolve it. :-)

We are waiting for some possible changes to be approved.

PR #5022, opened 2019-07-02
exposing password encrypt utility interface as plugin

robertopaez:
Exposing Password Encrypt Utility as plugin (PasswordUtilityEncrypter). 

In order to have custom encrypt utilities, a new PasswordUtilityEncrypter plugin type was added .
The PasswordUtilityController npw also search the list of plugins implemented (along with ServiceLoader classes) 

**Additional context**


robertopaez:
@gschueler

I decided add a new interface for the PasswordUtilityEncrypterPlugin.
Mainly because if we use the same PasswordUtilityEncrypter interfaz, the CommandLineSetup will be broken.
the controller now will support both implementations. For the plugin based on PasswordUtilityEncrypterPlugin, now the Descriptions  can be used.

core/src/main/java/com/dtolabs/rundeck/core/encrypter/PasswordUtilityEncrypterPlugin.java
```diff
@@ -0,0 +1,8 @@
+package com.dtolabs.rundeck.core.encrypter;
+
+import java.util.Map;
+
+public interface PasswordUtilityEncrypterPlugin {
+
+    Map encrypt(Map config);
```
gschueler: this interface is a bit crude for a plugin type, I would suggest something more like this:

```suggestion
    EncryptorResponse encrypt(Map<String,String> config);
```

where the `EncryptorResponse` is an interface like:

~~~
interface EncryptorResponse{
    boolean isValid()
    String getError()
    Map<String,String> getOutputs()
}
~~~

robertopaez: added @gschueler

robertopaez:
@gschueler I added the docs that you asked

PR #5028, opened 2019-07-03
Audit Events listeners and plugins.

ahormazabal:
This PR implements an auditing listener layer to capture certain system events, focused on auditing user operations.

Current events processed are:
- User successful login 
- User failed login
- User logout
- Access to a project (IE: access to a project's homepage).

Space was left to add future actions and resources to audit, based on the same action and resources  used on ACLs.

The interface implements both listeners that can be used internally, or can be implemented as plugins.

This PR fullfils a requirement from a client to be able to track user logins,  logouts and access to projects.

Example plugin can be found at: https://github.com/ahormazabal/example-audit-plugin


rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
```diff
@@ -258,6 +241,14 @@ class MenuController extends ControllerBase implements ApplicationContextAware{
             return redirect(controller: 'menu', action: 'home')
         }
         def startpage = params.page?: grailsApplication.config.rundeck.gui.startpage ?: 'jobs'
+
+        // Trigger project home access event
+        auditEventsService.eventBuilder()
```
gschueler: I think this trigger should not be put here. I would suggest maybe instead implementing a grails Interceptor that matches certain url paths to observe as project access points. 

(This specific controller action should probably be removed/deprecated. I think it is still used, but is something I would like to remove at some point.)

Also, repeated access to the same project should not trigger another event, correct? Perhaps you will need to have a value stored in the user session indicating if a project has been accessed, and use that to determine if the event needs to fire.

core/src/main/java/com/dtolabs/rundeck/core/audit/AuditEvent.java
```diff
@@ -0,0 +1,84 @@
...
+ *
+ */
+public interface AuditEvent extends Serializable {
+
+
+  /**
+   *
+   * @return The system framework interface..
+   */
+  Framework getFramework();
```
gschueler: If the framework is necessary as part of the event data, use the `IFramework` interface in place of `Framework`

rundeckapp/grails-app/services/rundeck/services/audit/AuditEventsService.groovy
```diff
@@ -0,0 +1,396 @@
...
+
+/**
+ *
+ * Service for capturing and firing auditing events like login, logout or resource access
+ *
+ * @author Alberto Hormazabal
+ *
+ */
+@Transactional
+class AuditEventsService
```
gschueler: This class should not be a grails Service, as it does not need any hibernate session/database access.  It can be moved to the src/groovy directory, and declared in the resources.groovy as a normal spring bean.  As a grails Service, all the method calls to this class are proxied and run within a database transaction

core/src/main/java/com/dtolabs/rundeck/core/audit/ResourceTypes.java
```diff
@@ -0,0 +1,15 @@
+package com.dtolabs.rundeck.core.audit;
+
+/**
+ * Types of resources reported by audit events.
+ */
+public final class ResourceTypes {
```
gschueler: This seems like it should be an enum

ahormazabal: This was indeed an enum. However i changed it to be strings so the interface could support adding more resource and action types dynamically without modiying the actual code, so we could implement generation of dynamic or configurable resource types or actions in the future.

gschueler: Ok, that makes sense

ahormazabal:
Added interceptor for project access tracking.

core/src/main/java/com/dtolabs/rundeck/plugins/audit/AuditEventListener.java
```diff
@@ -0,0 +1,58 @@
+package com.dtolabs.rundeck.plugins.audit;
+
+import com.dtolabs.rundeck.core.audit.AuditEvent;
+
+/**
+ * Interface for implementing an Audit event listener plugin.
+ *
+ * @author alberto
+ */
+public interface AuditEventListener {
```
gschueler: it should be named `AuditEventListenerPlugin`

rundeckapp/grails-app/controllers/rundeck/interceptors/AuditInterceptor.groovy
```diff
@@ -0,0 +1,213 @@
...
+
+/**
+ * This filter tracks user access to trigger auditing events.
+ * For example, track the user access to different projects.
+ *
+ * This filter assumes {@link ProjectSelectInterceptor} has already processed the request.
+ *
+ * @author Alberto Hormazabal
+ */
+class AuditInterceptor {
```
gschueler: perhaps if no audit plugins are configured, this interceptor can short-circuit and do nothing?

rundeckapp/grails-app/controllers/rundeck/interceptors/AuditInterceptor.groovy
```diff
@@ -0,0 +1,213 @@
...
+    private static final int DEFAULT_CACHE_RETENTION_PERIOD = 60 * 30 // default 30 min.
+
+    def frameworkService
+    def auditEventsService
+    final ConfigurationService configurationService
+
+    private Long minProjectNotifyPeriodMillis
+    private Long cacheRetentionPeriodMillis
+
+    private final ConcurrentHashMap<String, ProjectTrackingEntry> projectAccessTrackingMap
```
gschueler: is a Guava cache a better fit here? rather than manually doing cache expiration?

rundeckapp/src/main/groovy/rundeck/services/audit/AuditEventsService.groovy
```diff
@@ -0,0 +1,534 @@
...
+     */
+    void removeListener(AuditEventListener listener) {
+        internalListeners.remove(listener)
+    }
+
+
+    /**
+     * Dispatch the event.
+     */
+    private dispatchEvent(AuditEventBuilder eventBuilder) {
```
gschueler: grails has an internal eventing system, i think perhaps that should be used instead. the Interceptor can send an event that way, and the AuditEventsService can listen for it.  that way future events can be sent from original location without binding to this class

rundeckapp/src/main/groovy/rundeck/services/audit/AuditEventsService.groovy
```diff
@@ -0,0 +1,534 @@
...
+            this.httpRequest = httpRequest
+            return this
+        }
+
+
+        /**
+         * Builds a new immutable event object
+         * @return
+         */
+        protected AuditEvent build() {
```
gschueler: this is a lot of boiler-plate builder code, perhaps use Lombok `@Builder` in a core Java class?

core/src/main/java/com/dtolabs/rundeck/plugins/audit/AuditEventListener.java
ahormazabal: I named it that way to allow both use as plugins or internally from within rundeck code. I could maybe do separete interfaces for both cases.

rundeckapp/grails-app/controllers/rundeck/interceptors/AuditInterceptor.groovy
ahormazabal: Nice idea.

rundeckapp/grails-app/controllers/rundeck/interceptors/AuditInterceptor.groovy
ahormazabal: I did try first with guava but coulnd't make it work right as i expected. I finally implemented like this to have more control about the atomicity of the map entries updating without creating too much contention between concurrent requests.

rundeckapp/src/main/groovy/rundeck/services/audit/AuditEventsService.groovy
ahormazabal: I can refactor the intrface implementation to a separate POJO, but i still need some of that boilerplate to ensure deep data copy between objects and to get correctly the request context.

ahormazabal:
Documentation: https://github.com/rundeck/docs/pull/343

core/src/main/java/com/dtolabs/rundeck/plugins/audit/AuditEventListener.java
ahormazabal: Created dedicated interface for plugins.

rundeckapp/grails-app/controllers/rundeck/interceptors/AuditInterceptor.groovy
ahormazabal: Done

core/src/main/java/com/dtolabs/rundeck/plugins/audit/AuditEventListener.java
ahormazabal: Added a new dedicated interface for plugins.

PR #5033, opened 2019-07-03
reverting changes to allow schedule jobs to respect the remote policy

ltamaster:
Bug:
reverting changes to allow schedule jobs to respect the remote policy execution if exists
https://github.com/rundeck/rundeck/pull/4799

Jobs scheduled can be removed from quartz, so the scheduled job won't be executed on any cluster member

the severNodeUUID cannot be changed, it just can be done when the job is saved.


gschueler:
@ltamaster does this revert all of the behavior of that previous change?

ltamaster:
@gschueler 
​I think it will revert it completely​ because the goal of the previous feature was respected the remote execution policy. We need to double check how to reach that.

PR #5041, opened 2019-07-10
fix #4781 Options from URL not filled when job is scheduled

### LINKED ISSUE #4781: Options from URL not filled when job is schedule
**Describe the bug**
In a job using options data retrieved from a service.
The option is set to the data returned by webservice when the job is ran manually but to the default option value when scheduled.

**My Rundeck detail**
* Rundeck version: [ 3.0.21-20190424 ]
* install type: [war]
* OS Name/version: [CentOS Linux release 7.6.1810 (Core)]
* DB Type/version: [h2]

**To Reproduce**
1. Create a job with option 'servers' whose value is a remote URL.
I use something like that https://rundeckhost/getservers.py?type=http&env=test
2. When the job is ran manually, 
the 'servers' option is filled with webservice data.
3. When the job is scheduled 
the 'servers' option is filled with the default value or not filled if no default value was set.

**Expected behavior**
The option should be populated with webservice data in both case.
At least scheduled and manual job should behave the same way.
 
 - OS: [Windows]
 - Browser [Firefox]
 - Version [67]

** Additional information
Seem to be related with #1675

carlosrfranco:
fixes #4781 changing the behaviour of scheduled job with option from URL

A scheduled job do not get any option values (or get the default value, if defined) when a job has an option from URL with selected values

This changes makes a scheduled job and a job manually executed to have the same behaviour for options with remote URL values, that is, the job will take a default value if defined, or the remote value from URL if default is not defined.


rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -2480,6 +2481,30 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
      * evaluate the options in the input argString, and if any Options defined for the Job have required=true, have a
      * defaultValue, and have null value in the input properties, then append the default option value to the argString
      */
+    def Map addRemoteOptionSelected(ScheduledExecution scheduledExecution, Map optparams, UserAndRolesAuthContext authContext = null) throws ExecutionServiceException {
```
gschueler: the javadoc for the wrong method is now attached to this method, you should make sure the javadoc stays with the correct method, and add new javadoc for this method

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -2480,6 +2481,30 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
...
+        def newmap = new HashMap(optparams)
+
+        final options = scheduledExecution.options
+        if (options) {
+            def defaultoptions=[:]
+            options.each {Option opt ->
+                if(null==optparams[opt.name] && opt.enforced && !opt.optionValues){
+                    Map remoteOptions = scheduledExecutionService.loadOptionsRemoteValues(scheduledExecution, [option: opt.name], authContext?.username)
```
gschueler: I think the `mapConfig` parameter should also include the `optparams` in case the url has references to other options, like this:
```suggestion
                    Map remoteOptions = scheduledExecutionService.loadOptionsRemoteValues(scheduledExecution, [option: opt.name, extra: [option: optparams]], authContext?.username)
```

PR #5058, opened 2019-07-16
Execution cleanup on a cluster

ltamaster:
Enhancement:
On a cluster environment where the serverNodeUUID of the cluster members normally changed (on docker o kubernetes env), the execution cleanup process doesn't work as expected. The main reason is each cluster member will try to remove its own executions (filtering by serverNodeUUID).

1.- A new method to get the list of dead members is added. Each active node will have a different list of dead members in order to not get the same list of executions to delete.
2.- if that list is not empty, then the execution query will search for those uuids (plus the uuid of the server).
3.- the list of dead member can include the "null" option. In this case, it will remove execution with the serverNodeUUID on null (cover cases when rundeck has migrated from not cluster to cluster env).
4.- Some optimizations were added (eg, query just for the execution ID, etc)

**Additional context**

### DIFF
--- a/rundeckapp/grails-app/jobs/rundeck/quartzjobs/ExecutionsCleanUp.groovy
+++ b/rundeckapp/grails-app/jobs/rundeck/quartzjobs/ExecutionsCleanUp.groovy
@@ -39,9 +39,10 @@ class ExecutionsCleanUp implements InterruptableJob {
         ExecutionService executionService = fetchExecutionService(context.jobDetail.jobDataMap)
         FileUploadService fileUploadService = fetchFileUploadService(context.jobDetail.jobDataMap)
         LogFileStorageService logFileStorageService = fetchLogFileStorageService(context.jobDetail.jobDataMap)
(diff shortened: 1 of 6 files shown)


rundeckapp/grails-app/jobs/rundeck/quartzjobs/ExecutionsCleanUp.groovy
```diff
@@ -188,33 +210,60 @@ class ExecutionsCleanUp implements InterruptableJob {
 
     private int totalAllExecutions(ExecutionService executionService, String project){
         ExecutionQuery query = new ExecutionQuery(projFilter: project)
-        Map result = executionService.queryExecutions(query)
-        return null != result ? result.total : 0
+        def criteriaClos = { isCount ->
```
gschueler: this param `isCount` doesn't seem to be used?

ltamaster: your right, I removed it

PR #5060, opened 2019-07-17
fixes #1378 Improve Logs using apache-log4j-extras

### LINKED ISSUE #1378: Feature Request: compress activity log
Activity log may eat up your disk space. 

`.rdlog` files are stored as-is in rundeck server. One option would be to mount an external storage and store activity in there in order to not waste rundeck server disk space. Despite that, activity logs could be compressed. You will lose processing time but just in one use case: when you need to read log files.

I think it worth it.

Romainj1:
fixes #1378
Logs created by rundeck /var/log/rundeck/... are not compressed, and it can cause disk usage issues.
Logs are  handled with log4j, and using the class org.apache.log4j.DailyRollingFileAppender, which allows log rotation. Unfortunately, this last does not support automatic file compression.

Easy solution is to use class org.apache.log4j.rolling.RollingFileAppender, but it cannot as it is part of apache-log4j-extras (and not core log4j), we need to import this library.

Zipping files in a crontab seemed a good solution, but then handles poorly log rotation.
I tried using also org.apache.log4j.RollingFileAppender (different from org.apache.log4j.rolling.RollingFileAppender), but didn't achieve to compress files with this one.

**Additional context**

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -61,6 +61,7 @@ dependencies {
     }
     compile 'org.apache.ant:ant:1.8.3',
         'log4j:log4j:1.2.17',
+        'log4j:apache-log4j-extras:1.2.17',
         'commons-codec:commons-codec:1.10',
         'commons-beanutils:commons-beanutils:1.9.3',
         'commons-collections:commons-collections:3.2.2',


Romainj1:
up

PR #5089, opened 2019-07-25
on cluster environment , the job scheduled should respect the remote exec policies

ltamaster:
Enhancement:
* on cluster environment, the job scheduled should respect the remote execution policies (if it is enabled).
* When the job is created, it will be checked if the job will be scheduled on another cluster member or locally. 
* When a job is updated, that was already considered. But, it was added a flag to avoid registering on the local quartz when the job is owned by another node.
* for takeover API, it will be forced to run locally

A new method was added to JobScheduleManager (scheduleRemoteJob) which will check if the remote policies are available on cluster env (and it will send the takeover message if it is enabled)

**Additional context**

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/schedule/JobScheduleManager.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/schedule/JobScheduleManager.java
@@ -95,5 +95,7 @@ public interface JobScheduleManager {
      *
      * @return true if successful
      */
-    boolean scheduleRemoteJob(Map data);
+    default boolean scheduleRemoteJob(Map data){
+        return false;
+    }
 }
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/dtolabs/rundeck/core/schedule/JobScheduleManager.java
```diff
@@ -87,4 +87,13 @@
...
     List<String> getDeadMembers(String uuid);
+
+    /**
+     * Schedule a job to run later
+     *
+     * @param data  dataRundeckproClusterGrailsPlugin
+     *
+     * @return true if successful
+     */
+    boolean scheduleRemoteJob(Map data);
```
gschueler: a `default` implementation should be provided for the normal behavior

ltamaster: yes, it is here:
https://github.com/rundeck/rundeck/pull/5089/files#diff-c8e0c96ec93bdecafc6f3d719da88211R136

gschueler: i mean use this
```suggestion
    default boolean scheduleRemoteJob(Map data){
        return false;
    }
```

ltamaster: got it

PR #5139, opened 2019-08-08
Allow notification to use workflow exported variables #5061

### LINKED ISSUE #5061: Allow notifications to receive context variables / Ref: #4838
I will be useful to allow notifications to include context variables 

Allow context variables to be passed in notifications. 

I tried using exported global vars (using `export-var` workflow step) and captured data (using `key-value-data` log filter) in notification however is not able to read the variables in the notification

Current output
 headers={Accept=application/json, application/javascript, text/javascript, foo=${export.newval}, trigger=success, fel=${data.value*}

Example output expected 
 headers={Accept=application/json, application/javascript, text/javascript, foo=12345, trigger=success, fel=12345

**Additional context**
example included using http notification call 

[ContextVariables_notifications.yaml.txt](https://github.com/rundeck/rundeck/files/3402415/ContextVariables_notifications.yaml.txt)
.

jtobard:
Notification Service now uses modified global context instead of the original allowing it to read new variables created during workflow.

And export variable workflow plugin creates the variable directly on global  `ContextView`.

Example workflow using notification plugin (Splunk notification plugin):
Workflow

Notification setup

Notification Result

fix #5061

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/NotificationService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/NotificationService.groovy
@@ -16,6 +16,7 @@
 
 package rundeck.services
 
+import com.dtolabs.rundeck.core.dispatcher.ContextView
 import com.dtolabs.rundeck.core.dispatcher.DataContextUtils
 import com.dtolabs.rundeck.core.execution.workflow.WorkflowStrategy
 import com.dtolabs.rundeck.core.logging.LogEvent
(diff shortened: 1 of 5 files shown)


rundeckapp/grails-app/services/rundeck/services/NotificationService.groovy
```diff
@@ -544,7 +545,7 @@ public class NotificationService implements ApplicationContextAware{
             userData['user.lastName'] = user.lastName
         }
         //pass data context
-        def dcontext = content['context']?.dataContext ?: [:]
+        def dcontext = content.context?.getSharedDataContext()?.getData(ContextView.global())?.getData() ?: [:] //usage of modified global context
```
gschueler: instead of adding to the shared data inside the plugin, can it just be consolidated here before getting the data?

jtobard: moved to the workflow operator since it was the original scope of the upvar plugin: https://github.com/rundeck/rundeck/pull/2952/files#diff-a3b56ffe2c1254af9525ee9c1f8ea5efR155

PR #5158, opened 2019-08-15
Dedupe setps on stepctx

### LINKED ISSUE #5157: Execution steps not fully loading
**Describe the bug**
Under certain circumstances the execution steps for a job will not fully load. A "Loading" message with spinner will display forever and the console will contain two errors complaining about accessing properties on `undefined`.

**My Rundeck detail**

* Rundeck version: 2.x - 3.1.1-SNAPSHOT

ProTip:
Fixes #5157

The issue can occur if there are duplicate steps in the state log. KO can't handle duplicate keys. It won't complain out right about them, but `undefined`s will appear in the observable array causing errors in mapping and other parts of the code.

This fix addresses the issue in the front end. After data is loaded the steps are de-duped on `stepctx` before being passed through to get mapped and summarized.

I considered de-duplicating the data in the backend however this information is actually present in the state file. It's unclear to me at this point if it serves a purpose to have multiple entries for the same `stepctx` in the steps array.

### DIFF
--- a/rundeckapp/grails-app/assets/javascripts/executionStateKO.js
+++ b/rundeckapp/grails-app/assets/javascripts/executionStateKO.js
@@ -873,8 +873,25 @@ function RDNode(name, steps,flow){
      * @param steps
      */
     self.loadData=function(data){
+        self.dedupeSteps(data.steps);
         self.updateSummary(data.summary);
         self.updateSteps(data.steps);
+    };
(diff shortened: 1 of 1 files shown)


ProTip:
@gschueler We have the execution logs and project/job that produced them in case `4852` if you want to take a look. It's a massive job though and likely won't run for us, but it may offer some insights.

PR #5182, opened 2019-08-26
Fixes #5147 Allow quotes in the middle of arguments to be interpreted as a single argument

### LINKED ISSUE #5147: Resource model generator with quoted args
**Describe the bug**
When calling a resource model generator script with arguments, quoted arguments are always delimited by spaces, and quoted spaces are ignored.

See amendlik/salt-gen-resource#5

This issue was reported by a user of the [Salt resource model generator](https://github.com/amendlik/salt-gen-resource), but the problem seems to be caused by Rundeck in the way it parses command line arguments before passing them to the resource model generator scripts.

Specifically, the problem is caused by this line: https://github.com/rundeck/rundeck/blob/master/core/src/main/java/com/dtolabs/rundeck/core/resources/ScriptResourceUtil.java#L335

**To Reproduce**
1. Install the [Salt resource model generator](https://github.com/amendlik/salt-gen-resource)
2. Configure it as follows:
```
resources.source.1.config.args=--log-file-level=debug -s -a environment -C 'G@region:SV and G@roles:webapp or G@roles:api or G@roles:job'
resources.source.1.config.file=/rundeck/scripts/SaltGenResource.py
resources.source.1.config.format=resourceyaml
resources.source.1.config.interpreter=sudo
resources.source.1.type=script
```
3. `'G@region:SV and G@roles:webapp or G@roles:api or G@roles:job'` should all be passed as a single argument with embedded spaces.
4. The Salt log file reveals that the argument was space-delimited and passed as multiple arguments.

**Expected behavior**
`'G@region:SV and G@roles:webapp or G@roles:api or G@roles:job'` should all be passed as a single argument with embedded spaces.

**Additional context**

carlosrfranco:
When calling a resource model generator script with arguments, quoted arguments are always delimited by spaces, and quoted spaces are ignored. For example, args like: -C "G@os:ubuntu and G@osarch:amd64"

I created a regex to split method that ignore spaces for quoted args like example above.


carlosrfranco:
> It would be better to change the way the ScriptResourceModelSource allows interpreting the args string. One way would be to add an "argsDelimiter" field to set a delimiter character manually. Another way would be to use the existing mechanism for handling quoted strings used by rundeck when interpreting commands, called OptsUtil. The method `OptsUtil.burst` can split input with quoted arguments. The ScriptResourceModelSource could add a plugin property that uses the OptsUtil method to interpret the arg string as an array.
> 
> However, I think ScriptResourceUtil usage of a single "args" string should probably be deprecated, along with the `.split(" ")`

@gschueler , I made the changes so the ScriptResourceModelSource is using the burst method of OptsUtil now. But I have one question: Do you think that I should remove the single "args" from properties of ScriptResourceModelSource and make it deprecated on ScriptResourceUtil on this same PR?

gschueler:
> Do you think that I should remove the single "args" from properties of ScriptResourceModelSource and make it deprecated on ScriptResourceUtil on this same PR?

No, the config inputs shouldn't change, so that upgrading won't break

PR #5190, opened 2019-08-27
Webhook plugin

sjrd218:
Adds webhook features to Rundeck


sjrd218:
@gschueler Could you give it a once over? I've added the ability to enable and disable webhooks, and to pass request headers to webhooks.

gschueler:
nb: snyk flagged commons-beanutils, but we use version 1.9.4 which is not affected

gschueler:
nb: more false positive from snyk, the c3p0 dep used is 0.9.5.4, which is not affected by the indicated cve

gschueler:
ok to merge?

sjrd218:
Good to go.

PR #5200, opened 2019-09-03
Return validation report as structured data

ProTip:
Also pass save errors through to custom plugin component. Not sure if we should make this consistent with the wider Rundeck API responses?


rundeckapp/grails-spa/src/pages/webhooks/views/WebhooksView.vue
```diff
@@ -189,14 +193,16 @@ export default {
       this.curHook.config = this.selectedPlugin.config
       this.ajax("post", `${rdBase}webhook/admin/save`, this.curHook).then(response => {
         if (response.data.err) {
-          this.setError("Failed to save! " + response.data.err)
+          this.setError("Failed to savez! " + response.data.err)
```
sjrd218: is "savez" on purpose?

ProTip:
```suggestion
          this.setError("Failed to save! " + response.data.err)
```

PR #5208, opened 2019-09-06
Replaces job Status notification email using Pine Email Framework

jessemarple:
Replaces job Status notification email template with a new one crafted with the Pine Email Framework. 

Brings in some of the stats/calculated props see on `_showStats.gsp`

Adds responsive layout.

Tested on iPhone, iPad and Kindle.

New Mobile
Old
New Desktop


jessemarple:
Note that while the file name stayed the same "status.gsp" the entire contents were replaced when moving from one email framework to the next as well as changes to the layout and content.

rundeckapp/grails-app/views/execution/mailNotification/status.gsp
```diff
@@ -1,378 +1,785 @@
...
+                  <td style="padding: 20px;">Log Output</td>
                 </tr>
+              </table>
+            </button>
+          </a>
+          <div class="toggle-content">
+            <table cellpadding="0" cellspacing="0" width="100%" role="presentation" style="border: 1px solid #EEEEEE;">
+              <tr>
+                <td style="padding: 20px;">
+                  ${enc(sanitize:logOutput)}
```
sjrd218: output should go inside a <pre> tag so that line endings from the output show up correctly

rundeckapp/grails-app/views/execution/mailNotification/deprecated_status.gsp
```diff
@@ -0,0 +1,379 @@
+%{--
```
sjrd218: Unless we are planning on providing a configuration option for the user to revert back to this email template, we should delete this file.

sjrd218:
For some reason the email being sent is now being sent with: `Content-Type: text/plain; charset=utf-8`

this is causing the rendering in the email client to look really wonky. This applies wether using the old or the new template.

rundeckapp/grails-app/views/execution/mailNotification/status.gsp
jessemarple: I've added a `<pre>` tag around the log output.

rundeckapp/grails-app/views/execution/mailNotification/deprecated_status.gsp
jessemarple: Because we are feature flagging the new Status email, we've kept "deprecated_status", changed it's name to oldStatus and are flagging the switch in rundeck-config.properties.

PR #5212, opened 2019-09-06
ExecutionLifecyclePlugin and JobLifecyclePlugin

ronaveva:
It adds ExecutionLifecyclePlugin and JobLifecyclePlugin

ExecutionLifecyclePlugin: it trigger events once the execution context already exist.
1.-beforeJobStarts: it can be use to modify the execution context
2.-afterJobEnds

JobLifecyclePlugin: it outside of the execution context.
1.-beforeJobExecution: it can be used to perform validations before executing the job and also change the option values
2.-beforeSaveJob: it can be used to add/delete job options when performing a save/update of a job.


rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -2410,14 +2433,42 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
...
+        def result = checkBeforeJobExecution(scheduledExec, optparams, props, authContext)
+        if(result?.useNewValues()){
+            optparams = result.optionsValues
+            checkSecuredOptions(scheduledExec, props, result.optionsValues)
+        }
         validateOptionValues(scheduledExec, optparams,authContext)
         return optparams
     }
 
+    def checkSecuredOptions(scheduledExecution, props = [:], optionsValues){
```
gschueler: I'm not sure what this method is supposed to do.  I see it is called in `scheduleAdHocJob` and `retryExecuteJob` but i don't know why.

ronaveva: This is to replace values comming from job secure options (by the ones setted up by a JobLifeCyclePlugin)

PR #5216, opened 2019-09-09
User login status tracking

ronaveva:
New fields added:
1.-Session id
2.-Hostname
3.-Event Time
4.-Logged Status

*Added pagination, default 100 (users per page)

Search filters added (uses equals to):
1.-Login (aka username)
2.-Session id
3.-Hostname

*Pagination can be modifed by using the property (server/config/rundeck-config.properties):

rundeck.gui.user.summary.max.per.page=20


ronaveva:
> 
> 
> Storing sessionId in db, and displaying other users' session IDs in the GUI, is a security risk. That feature should be removed. If necessary, a hashed version of the sessionID could be stored perhaps

Added a new property to enable session ID storage and display. False by default.

rundeckapp/grails-app/controllers/rundeck/controllers/UserController.groovy
```diff
@@ -508,13 +508,15 @@ class UserController extends ControllerBase{
             obj.email = it.email
             obj.created = it.dateCreated
             obj.updated = it.lastUpdated
-            def lastExec = includeExec ? Execution.lastExecutionDateByUser(it.login).get() : null
-            if (lastExec) {
-                obj.lastJob = lastExec
+            if(params.includeExec){
```
gschueler: the param value is already used?

gschueler:
@ronaveva it seems like a lot of other changes were added to this branch

PR #5254, opened 2019-09-16
Improve interactivity of page when loading execution log

ProTip:
This improves the interactivity of the page while loading the execution log.

To reduce the amount of time JS is blocking the UI(scrolling primarily) the table rows are being added in chunks. Prior there was an implicit chunking due to the `500` max line limit fetched from the server at one time, however this chunk size proved to feel quite laggy in my testing.

This change decouples the DOM update batching from the data fetching. Until there are over **20k** lines of output the batch size will be 50. Around **20k** the reflow starts to take up a much larger portion of the update time and is constant for the input size. At this point the batch size is raised back to `500` to prevent a much longer loading time for larger logs.

The impact to user experience is that the page will feel much smoother when loading small to medium logs, and will feel smoother for the start of loading larger logs without impacting their overall total loading time.

The use of tables for very large lists is generally considered a poor choice from a performance perspective. Table updates cause long style, layout, and render events to occur. These become progressively worse as more table elements are added:
image

This can be mitigated with very careful, conservative CSS rules along with a fixed table layout however it may not be able to be fully addressed. Moving to elements(divs) and a page layout that does not require long reflows would be ideal however it would also be a much more significant change.

### DIFF
--- a/rundeckapp/grails-app/assets/javascripts/executionControl.js
+++ b/rundeckapp/grails-app/assets/javascripts/executionControl.js
@@ -473,9 +473,91 @@ var FollowControl = function (eid, elem, params) {
         var rowcount= this.countTableRows(this.cmdoutputtbl);
         var compacted = data.compacted;
         var compactedAttr = data.compactedAttr;
+
+        /**Called after table update is finished for housekeeping and fetching next batch if necessary **/
+        var doFinishedLogUpdate = function() {
+
+            this.lineCount+=entries.length;
+
+            if (typeof(this.onAppend) == 'function') {
+                this.onAppend();
+            }
+            if (this.clusterExec && this.showClusterExecWarning) {
+                if (!this.runningcmd.completed) {
+                    //show cluster loading info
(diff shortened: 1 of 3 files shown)


rundeckapp/grails-app/assets/javascripts/executionControl.js
```diff
@@ -473,107 +473,139 @@ var FollowControl = function (eid, elem, params) {
         var rowcount= this.countTableRows(this.cmdoutputtbl);
         var compacted = data.compacted;
         var compactedAttr = data.compactedAttr;
+
+        /**Called after table update is finished for housekeeping and fetching next batch if necessary **/
+        var doFinishedLogUpdate = function() {
```
jessemarple: I wouldn't mind a bit more in the way of commenting on the what this code does. If you think we'll most likely replace this with that Vue list/table component you were showing me the other day, please feel free to disregard this.

gschueler:
let's target this for 3.2

PR #5264, opened 2019-09-17
Fix VUE UI issue on webhook page

sjrd218:
When a new webhook is saved there is no current hook selected from the list so don't try to match.


gschueler:
please assign to appropriate milestone

PR #5265, opened 2019-09-17
Fix manual plugin uninstall

sjrd218:
Fix uninstall of manually installed plugin when only artifact id is sent to API.


gschueler:
please assign to appropriate milestone

PR #5273, opened 2019-09-19
Use containers instead of binaries to run minio

gschueler:

Changes tests for object-store plugin to use docker container for minio via [TestContainers](https://www.testcontainers.org/) instead of launching local binary directly.

**Additional context**

Note: TestContainers spock integration requires groovy 2.5, this changes the test dependencies for object-store plugin to use groovy 2.5, however this required tweaking some of the plugin code to work correctly between groovy 2.4/2.5. I'm not sure if this is a good idea, since there could be issues when running the code in 2.4 that don't appear when running it with 2.5 for the tests. 

For now i'm just putting this up for perusal.  Perhaps a better solution would be to use junit instead of spock and not use groovy 2.5 for the tests.


gschueler:
turns out the testcontainers+spock integration was very minimal, so we don't need it, can go back to 2.4 groovy

plugins/object-store-plugin/src/test/groovy/testhelpers/MinioContainer.groovy
```diff
@@ -0,0 +1,54 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package testhelpers
+
+import io.minio.MinioClient
+import org.testcontainers.containers.GenericContainer
+
+class MinioContainer<SELF extends MinioContainer<SELF>> extends GenericContainer<SELF> {
```
bsideup: FYI you don't **have** to use the self typing here, and can define it as `MinioContainer extends GenericContainer< MinioContainer>`

PR #5105, opened 2019-07-31
Improves readability of options in job executions, answers #5044

jessemarple:
Answers and Fixes #5044

Improves the readability of options, stacking them one over the other, and implements word-breakage.

Screen Shot 2019-07-30 at 8 43 37 PM

Issue #5044, opened 2019-07-11
Improve cluttered options on job list and job info page


jordanhenderson:
Hi, 

I would like to propose a few style changes to improve the display of jobs with a large number of options, or options with moderately long configurable strings. Currently, the options list displayed on job lists and execution information pages quickly become cluttered and cause overflow in tables. It would be good to either hide these, or apply CSS to truncate them - potentially as a configurable option. I used the following CSS on job list and execution pages to substantially improve the options - see the below before and after.

I am using the latest snapshot of rundeck (3.1) in a docker container and cannot currently alter these pages myself without changing the base image, so thought it would be a good idea to post here hoping that someone might consider this for a future release.

```
@media (min-width: 993px) {
    .execution-aux-info.flex-item-1 {
        max-width: calc(50vw - 115px);
    }
}
.argstring-scrollable {
    overflow: hidden;
    word-break: break-all;
    text-overflow: ellipsis;
}
.argstring-scrollable .text-secondary {
    display:block;
}
.argstring-scrollable .optkey {
    display: inline-block;
}
.argstring-scrollable .optvalue {
    overflow: hidden;
    word-break: break-all;
    text-overflow: ellipsis;
    white-space: nowrap;
    display: block;
}
.eventargs .argstring-scrollable {
    display:none;
}
```

**Job Screen - Before**
<img width="800" alt="Screen Shot 2019-07-11 at 12 11 20 pm" src="https://user-images.githubusercontent.com/2040171/61017307-0ee92500-a3d6-11e9-97cb-48468de053f6.png">

**Job Screen - After**
<img width="801" alt="Screen Shot 2019-07-11 at 12 11 57 pm" src="https://user-images.githubusercontent.com/2040171/61017323-1f999b00-a3d6-11e9-9ed6-a232748305a5.png">

**Job List - Before (notice the overflow)**
<img width="891" alt="Screen Shot 2019-07-11 at 12 18 07 pm" src="https://user-images.githubusercontent.com/2040171/61017335-2aecc680-a3d6-11e9-9c44-198045b4a91e.png">

**Job List - After (options moved to Job page, where they can display in full)**
<img width="910" alt="Screen Shot 2019-07-11 at 12 18 27 pm" src="https://user-images.githubusercontent.com/2040171/61017356-3dff9680-a3d6-11e9-9a5c-168a4c733bbd.png">

gaddman:
[...]

PR #5277, opened 2019-09-19
cleanup argstring display in activity list and exec page

gschueler:
activity list: don't wrap, allow horizontal overflow scroll
exec page: show values on separate blocks, max height to scroll on y axis

Update display of argstrings values in activity list and execution page, ref #5044 #5105 

Screen Shot 2019-09-19 at 1 47 07 PM
Screen Shot 2019-09-19 at 1 47 00 PM


gschueler:
note: this would be alternative to #5275

PR #5302, opened 2019-09-26
Fix issue with plugin select value with multiOption

sjrd218:
Java plugins with a property that came from @SelectValues with multOption=true did not send their values to the server correctly. The server wasn't handling the value type correctly. That is now fixed too.

### DIFF
--- a/rundeckapp/grails-app/views/framework/_pluginConfigPropertyFormField.gsp
+++ b/rundeckapp/grails-app/views/framework/_pluginConfigPropertyFormField.gsp
@@ -162,15 +162,16 @@
 
             <g:each in="${propSelectValues}" var="propval" status="n">
                 <div class="optionvaluemulti checkbox">
-                  <g:checkBox name="${fieldname}" checked="${propval.value in defvalset}" value="${propval.value}" id="f_${fieldid}_p_${n}"/>
-                  <label class="grid-row optionvaluemulti" for="f_${fieldid}_p_${n}">
-                    ${propval.label}
-                  </label>
-                </div>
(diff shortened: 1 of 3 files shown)


rundeckapp/grails-app/views/framework/_pluginConfigPropertyFormField.gsp
```diff
@@ -162,15 +162,16 @@
...
-                  <label class="grid-row optionvaluemulti" for="f_${fieldid}_p_${n}">
-                    ${propval.label}
-                  </label>
-                </div>
-            </g:each>
-        </div>
-        </div>
-    </g:elseif>
+              <g:set var="ischecked" value="${propval.value in defvalset ? 'checked="checked"' : ''}"/>
+            <input type="checkbox" id="${fieldname}" name="${fieldname}" ${ischecked} value="${propval.value}"/>
```
gschueler: non gsp tags will need attributes to be encoded
```suggestion
            <input type="checkbox" id="${enc(attr:fieldname)}" name="${enc(attr:fieldname)}" ${ischecked} value="${enc(attr:propval.value)}"/>
```

sjrd218: If you selected multiple options of the checkbox group they would all be sent to the backend as empty strings. eg ["","",""]. Once I made the change the correct values were transferred.

gschueler: ok, thanks

PR #5325, opened 2019-10-01
Fix error on remote opts with enforced values for simple JSON object

### LINKED ISSUE #5314: Option Validation with URLs not working
We are seeing the exact same behavior as #5030 but in a newer version of Rundeck

**My Rundeck detail**

* Rundeck version: 3.1.2.20190927
* install type: rpm
* OS Name/version: CentOS 7
* DB Type/version: postgres

**To Reproduce**
    Create a job with an option using Remote URL
    Restrict using Enforced from Allowed Values
    Required yes
    Add any step.
    Save the job.
    Run the job selecting any value from the list of options.

**Expected behavior**
The job runs successfully

**Screenshots**
See attached: Option option_name value: /mnt/share/NodalSuite/nightly_qa_deployments/QA13.json was not in the allowed values: [] 

 - OS: Fedora
 - Browser Firefox
 - Version 68.0.2

**Context
We saw the same bug in 3.1.0.20190731-3 as well.

ProTip:
Fixes #5314 

Fix

`loadOptionsRemoteValues` can return three(?) different shapes for results:
- `List(Collection)<String>` for `['foo','bar']`
- `List<JSONObject>` for `[{"name": "foo", "value": "bar"}]`
- `List<LinkedHashMap>` for `{"foo":"bar"}`

Changing to `instanceof Map` will match both JSONObject and the Groovy map.

**Additional**
I believe at some point the remote option loading could afford to be refactored out into a utility class. Breaking the logic out across multiple methods and ensuring a uniform(typed?) return schema regardless of the supported input JSON shape could help with test-ability and maintainability.

### DIFF
--- a/rundeckapp/src/integration-test/groovy/rundeck/services/ExecutionServiceIntegrationSpec.groovy
+++ b/rundeckapp/src/integration-test/groovy/rundeck/services/ExecutionServiceIntegrationSpec.groovy
@@ -1,0 +1,4 @@
+package rundeck.services
+
+class ExecutionServiceIntegrationSpec {
+}
--- a/rundeckapp/src/integration-test/groovy/rundeck/services/ExecutionServiceIntegrationSpec.groovy
+++ b/rundeckapp/src/integration-test/groovy/rundeck/services/ExecutionServiceIntegrationSpec.groovy
@@ -1,4 +1,0 @@
-package rundeck.services
-
-class ExecutionServiceIntegrationSpec {
-}
--- a/rundeckapp/src/test/groovy/ExecutionServiceSpec.groovy
+++ b/rundeckapp/src/test/groovy/ExecutionServiceSpec.groovy
@@ -1511,13 +1511,13 @@ class ExecutionServiceSpec extends Specification implements ServiceUnitTest<Exec
         service.scheduledExecutionService = Mock(ScheduledExecutionService)
         when:
 
-        def validation = service.validateOptionValues(se, opts)
+        service.validateOptionValues(se, opts)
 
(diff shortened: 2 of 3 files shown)


rundeckapp/src/test/groovy/ExecutionServiceSpec.groovy
```diff
@@ -1498,28 +1498,31 @@ class ExecutionServiceSpec extends Specification implements ServiceUnitTest<Exec
...
                     srcUrl       : "cleanUrl",
                     err          : null
             ]
         }
 
 
         ExecutionServiceValidationException e = thrown()
+        e.errors.each { k, v ->
+            println("${k} ${v}")
+        }
```
sjrd218: No need to keep the println.

rundeckapp/src/integration-test/groovy/rundeck/services/ExecutionServiceIntegrationSpec.groovy
```diff
@@ -0,0 +1,4 @@
+package rundeck.services
```
sjrd218: This file seems unnecessary. Can we remove it?

ProTip: Hah, yes.. That was abandoned. It seemed like a good idea and turned into a massive PITA.

menicolas89:
any luck with this review. thanks!

sjrd218:
@ProTip Looks like you might have to rebase this to get it to build.

PR #5352, opened 2019-10-07
Travis env fix: use trusty distro

gschueler:

Trying to fix travis builds failing to install oracle jdk 8


gschueler:
close in favor of #5353

PR #5362, opened 2019-10-09
 Move AuthConstants.java from grails app to core 

### LINKED ISSUE #5354: Move AuthConstants class to core
Move `AuthConstants.java` from grails app to core, this will allow its use from java classes like AclTool without repeating the information:

https://github.com/rundeck/rundeck/blob/master/rundeckapp/src/main/groovy/com/dtolabs/rundeck/server/authorization/AuthConstants.java

https://github.com/rundeck/rundeck/blob/master/core/src/main/java/com/dtolabs/rundeck/core/cli/acl/AclTool.java#L1970

jtobard:
Move AuthConstants.java from grails app to core to be used on core classes like AclTool and exposed to other apps.

fix #5354

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/common/AuthConstants.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/common/AuthConstants.java
@@ -1,38 +1,12 @@
-/*
- * Copyright 2016 SimplifyOps, Inc. (http://simplifyops.com)
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
- *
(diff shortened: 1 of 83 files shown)


core/src/main/java/com/dtolabs/rundeck/core/common/AuthConstants.java
```diff
@@ -13,26 +13,15 @@
...
-
-/*
-* AuthConstants.java
-* 
-* User: Greg Schueler <a href="mailto:greg@dtosolutions.com">greg@dtosolutions.com</a>
-* Created: 9/1/11 9:20 AM
-* 
-*/
-package com.dtolabs.rundeck.server.authorization;
+package com.dtolabs.rundeck.core.common;
```
gschueler: i think perhaps it should go in a package like `org.rundeck.core.auth`

PR #5383, opened 2019-10-15
Cleanup test dir in NodeServiceSpec test

sjrd218:
Adds a framework.var.dir to the test so the temp dir that is generated is not in the rundeckapp dir.


rundeckapp/src/test/groovy/rundeck/services/NodeServiceSpec.groovy
```diff
@@ -38,6 +38,7 @@ import spock.lang.Unroll
  */
 @TestFor(NodeService)
 class NodeServiceSpec extends Specification {
+    private static final String RESOURCE_TMP_DIR = '/tmp/rundeckNodeServiceSpec'
```
gschueler: maybe use `File.createTempDir()` instead, so the /tmp isn't hardcoded.

PR #5392, opened 2019-10-17
Addapting OSS so rundeckpro can implement a new interface to save file on storage

carlosrfranco:
This PR is to create an interface so rundeckpro can implement it for a new feature to save file on storage.

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/execution/ExecutionContext.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/execution/ExecutionContext.java
@@ -83,10 +83,6 @@ public interface ExecutionContext {
      */
     public StorageTree getFileStorageTree();
     /**
-     * @return the file storage service
-     */
-    public AuthStorageTree getAuthStorageTree();
-    /**
      * @return the job service
      */
     public JobService getJobService();
--- a/core/src/main/java/com/dtolabs/rundeck/core/execution/ExecutionContext.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/execution/ExecutionContext.java
@@ -39,6 +39,7 @@ import com.dtolabs.rundeck.core.execution.workflow.WorkflowExecutionListener;
 import com.dtolabs.rundeck.core.jobs.JobService;
 import com.dtolabs.rundeck.core.logging.LoggingManager;
 import com.dtolabs.rundeck.core.nodes.ProjectNodeService;
+import com.dtolabs.rundeck.core.storage.AuthStorageTree;
 import com.dtolabs.rundeck.core.storage.StorageTree;
 
 import java.util.Collection;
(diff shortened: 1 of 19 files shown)


core/src/main/java/com/dtolabs/rundeck/core/execution/ExecutionContext.java
```diff
@@ -77,6 +78,14 @@
...
      */
     public StorageTree getStorageTree();
+    /**
+     * @return the file storage service
+     */
+    public StorageTree getFileStorageTree();
+    /**
+     * @return the file storage service
+     */
+    public AuthStorageTree getAuthStorageTree();
```
sjrd218: What is the difference between the file storage tree and the auth storage tree?

core/src/main/java/com/dtolabs/rundeck/core/logging/PluginLoggingContext.java
```diff
@@ -51,4 +52,6 @@
      * @return the data context in the private scope
      */
     public DataContext getPrivateDataContext();
+
+    public StorageTree getFileStorateTree();
```
gschueler:
```suggestion
    public StorageTree getFileStorageTree();
```

sjrd218:
@carlosrfranco I think you also need the changes to `application.yml` from the other PR. I don't think the app will bootstrap without it.

core/src/main/java/com/dtolabs/rundeck/core/storage/StorageTreeFactory.java
```diff
@@ -138,6 +139,10 @@ public StorageTree createTree() throws Exception {
                     ), PathUtil.asPath("/keys")
             );
         }
+        if(null!=fileStorageConverterPluginAdapter && null!=defaultConverters && defaultConverters.contains("FileStorageLayer")){
+            logger.debug("Configuring base converter: FileStorageLayer" );
+            builder=builder.convert(fileStorageConverterPluginAdapter, PathUtil.asPath("/files"));
+        }
```
gschueler: the "defaultConverts" maybe should be refactored completely, allowing the factory resource definition to specify plugin instances/beans to use that way.  It was a small hack to have magic words "KeyStorageLayer", "StorageTimestamperConverter" that added certain converter plugins.  I think we should make a more generic way to do it

rundeckapp/grails-app/services/rundeck/services/AuthStorageManager.groovy
```diff
@@ -0,0 +1,27 @@
...
+import com.dtolabs.rundeck.core.authorization.AuthContext
+import com.dtolabs.rundeck.core.storage.ResourceMeta
+import com.dtolabs.rundeck.core.storage.StorageTree
+import com.dtolabs.rundeck.core.storage.StorageUtil
+import com.dtolabs.rundeck.core.storage.AuthStorageManager
+import grails.gorm.transactions.Transactional
+import org.rundeck.storage.data.MemoryTree
+
+@Transactional
+class AuthStorageTreeService implements AuthStorageManager{
```
gschueler: should avoid adding Service classes to the grails services, which don't need to be grails services.  This class could be removed, and simply inject the `rundeckAuthStorageManager` bean defined here https://github.com/rundeck/rundeck/pull/5392/files#diff-52259838c7d67dfe63914241dcf89443R114

gschueler: i think...

gschueler:
closing as out of date

PR #5394, opened 2019-10-18
Beta Execution output viewer

ProTip:
Part of a larger epic. Ready to merge in for feedback and iteration in snapshot.

## Points of Interest
* Fast
* Vue component
* Gutter is NOT selectable; allow for selecting un-mangled output across lines
* Manipulates DOM directly, but also utilizes Vue components for each log line
* Critical layout mostly scoped and located with elements
* Small-ish theming surface with 'dark' theme example(... and likely light to come)
* BEM class naming
* Props extracted from div data attributes(progressive enhancement)
* Can jump to a line provided through props when it becomes loaded
* Can jump to all the lines any time and on demand
* Fills the parent element completely; wrapper dimensions in `gsp` page setup specifically for the current show page

## Shortcomings
* Doesn't filter on nodes ATM
* No network resiliency right now(will not recover if a request fails while fetching output)
* Existing options are not fully supported(no ANSI color toggle and etc)


ProTip:
Updates:

* Following running executions
* Rendering step labels based on new workflow API endpoint
* Node labels
* Light default theme
* Renamed to "Beta" and tucked into the drop down menu in execution view:
image
* Also available in the follow drop down when running a job:
image

gschueler:
Some issues

1) needs an opt-in flag to enable it. either app-level feature flag, or project config flag or something
2) choosing "follow output beta" loads the output page but then seems to fail to load any output until you reload the page
3) "Display gutter" and "Display timestamps" settings toggles don't seem to do anything until you reload the page
4) some js console warnings from vue `Avoid mutating a prop directly since the value will be overwritten whenever the parent component re-renders. Instead, use a data or computed property based on the prop's value. Prop being mutated: "follow"`

fdevans:
If we can already identify issues or things that aren’t working let’s address those and add the feature flag with default to ON.  We can release as Beta with 3.3.0 and finalize in a subsequent 3.3.x release.

I've moved the milestone on this to 3.3.0

gschueler:
Issue when running a job and choosing "follow Log output beta", if the job does not produce output immediately, no log output is loaded in the page, and it must be reloaded.

Example, make a job with a single command:

    for i in $(seq 30) ; do sleep 10 ; echo hi $i ; done

@ProTip

ProTip:
@gschueler Feature flagged with default to `true` for now.

Resolved the issue where viewer doesn't load if execution output is delayed.

The options settings is expected behavior ATM subject to change.

PR #5399, opened 2019-10-21
[build] Upgrade to gradle 4.8 compatible project structure. 

sjrd218:
This is required for future Grails upgrade work.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -77,7 +77,7 @@ ext.exportedProjects = [
         ":rundeck-storage:rundeck-storage-data",
         ":rundeck-storage:rundeck-storage-filesys",
         ":rundeck-storage:rundeck-storage-conf",
-        ":rundeck"
+        ":rundeckapp"
 ]
 
 project.ext.environment = project.hasProperty('environment') ? environment : 'development'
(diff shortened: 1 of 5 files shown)


gschueler:
i'm curious what the gradle requirement is for this change? the sub project name can't be modified as we did before?

gschueler:
is it breaking to rename the dir? 😨

sjrd218:
@gschueler Here's a link to my gradle issue regarding this. https://github.com/gradle/gradle/issues/8884

gschueler:
I wonder if we can simply modify the rundeckapp build to change the artifact name to "rundeck", instead of having to this kind of thing, and leave the gradle project named "rundeckapp".  the changes here mean any new gradle project under rundeckapp/ has to have its path explicitly defined in the settings.gradle

sjrd218:
I'm in favor of the idea. It's been a long time since I looked into the issue, but I believe there were challenges with renaming the build artifact. I abandoned that route for the one we currently have in place. Since we can't keep what we have, it may be worth looking into again.

sjrd218:
@jessemarple I've checked this again and it looks ok to merge to me.

PR #5405, opened 2019-10-22
Use typed configuration for Rundeck configs.

sjrd218:
This PR is designed to provide a starting place for adding and using Rundeck configs in a type safe way. The underlying mechanism uses Spring's configuration mapping feature.

Please give feedback regarding the technical solution and structuring of the typed configs.

### DIFF
--- a/rundeckapp/build.gradle
+++ b/rundeckapp/build.gradle
@@ -102,10 +102,10 @@ dependencies {
     compile "org.springframework.boot:spring-boot-autoconfigure"
 
     compile "org.grails:grails-core"
-    //compile "org.springframework.boot:spring-boot-starter-actuator"
-    compile ("org.springframework.boot:spring-boot-starter-jetty"){
-        exclude(group:'org.eclipse.jetty.websocket',module:'websocket-server')
-        exclude(group:'org.eclipse.jetty.websocket',module:'javax-websocket-server-impl')
+    compile "org.springframework.boot:spring-boot-starter-actuator"
+    provided("org.springframework.boot:spring-boot-starter-jetty") {
+        exclude(group:"org.eclipse.jetty", module:"jetty-util")
+        exclude(group:"org.eclipse.jetty", module:"jetty-security")
     }
     compile "org.grails:grails-web-boot"
     compile "org.grails:grails-logging"
--- a/rundeckapp/build.gradle
+++ b/rundeckapp/build.gradle
@@ -104,8 +104,8 @@ dependencies {
     compile "org.grails:grails-core"
     compile "org.springframework.boot:spring-boot-starter-actuator"
     compile ("org.springframework.boot:spring-boot-starter-jetty") {
-        exclude(group:"org.eclipse.jetty", module:"jetty-util")
-        exclude(group:"org.eclipse.jetty", module:"jetty-security")
+        exclude(group:'org.eclipse.jetty.websocket',module:'websocket-server')
+        exclude(group:'org.eclipse.jetty.websocket',module:'javax-websocket-server-impl')
     }
     compile "org.grails:grails-web-boot"
     compile "org.grails:grails-logging"
--- a/rundeckapp/build.gradle
+++ b/rundeckapp/build.gradle
@@ -103,7 +103,7 @@ dependencies {
 
     compile "org.grails:grails-core"
     compile "org.springframework.boot:spring-boot-starter-actuator"
-    provided("org.springframework.boot:spring-boot-starter-jetty") {
+    compile ("org.springframework.boot:spring-boot-starter-jetty") {
         exclude(group:"org.eclipse.jetty", module:"jetty-util")
         exclude(group:"org.eclipse.jetty", module:"jetty-security")
     }
@@ -137,7 +137,6 @@ dependencies {
         exclude(group:"org.yaml",module:"snakeyaml")
     }
     profile "org.grails.profiles:web"
-    runtimeOnly "javax.annotation:javax.annotation-api:1.3.2"
 //    runtime "org.springframework:spring-test:5.0.3.RELEASE"
     runtime "com.h2database:h2:1.4.199"
 
--- a/rundeckapp/build.gradle
+++ b/rundeckapp/build.gradle
@@ -137,6 +137,7 @@ dependencies {
         exclude(group:"org.yaml",module:"snakeyaml")
     }
     profile "org.grails.profiles:web"
+    runtimeOnly "javax.annotation:javax.annotation-api:1.3.2"
 //    runtime "org.springframework:spring-test:5.0.3.RELEASE"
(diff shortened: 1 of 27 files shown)


rundeckapp/grails-app/services/rundeck/services/ConfigurationService.groovy
```diff
@@ -212,4 +241,17 @@ class ConfigurationService {
             return defval
         }
     }
+
+    @Override
+    void afterPropertiesSet() throws Exception {
+        appCfg = grailsApplication?.config?.rundeck
+    }
+
+    Map<String,String> deprecatedKeyTrx = [
```
gschueler: maybe this map could be defined put into resources.groovy, or a config of its own, making it easier to extend later if more things come up, unless these are the only values we might need to translate

rundeckapp/src/main/groovy/org/rundeck/app/services/EnhancedNodeService.groovy
```diff
@@ -52,7 +52,7 @@ class EnhancedNodeService
 
     @Override
     void afterPropertiesSet() throws Exception {
-        enabled = featureService.featurePresent('enhanced-nodes', false)
+        enabled = featureService.featurePresent('enhancedNodes', false)
```
gschueler: we might want a Constants class for feature names

rundeckapp/grails-app/services/rundeck/services/ConfigurationService.groovy
sjrd218: As long as future configs are added to the typed configuration it should prevent devs from using config names that can't be mapped to Java property names.

Since these keys already exist, there needed to be some way of falling back from the new name to get the value out of the old if it exists. 

I hope this map doesn't need any additions.

rundeckapp/src/main/groovy/org/rundeck/app/services/EnhancedNodeService.groovy
sjrd218: I guess we could have a Features class that has all of the current features. In the pro side we could extend the class like ProFeatures and add the pro only features. What do you think?

rundeckapp/build.gradle
```diff
@@ -29,6 +29,13 @@ apply plugin:"com.energizedwork.idea-project-components"
 apply plugin:"asset-pipeline"
 apply plugin:"org.grails.grails-gsp"
 
+dependencyManagement {
+    imports {
+        mavenBom "org.grails:grails-bom:$grailsVersion"
+        mavenBom 'org.springframework.cloud:spring-cloud-starter-parent:Edgware.SR6'
```
ahormazabal: Curious that you had to add the grails and spring-cloud boms to the build. This shouldn't be necessary to include spring dependencies. did you run with some issues?

rundeckapp/grails-app/services/rundeck/services/ConfigurationService.groovy
ahormazabal: I would keep this mapping close to where properties are defined, to make easier for developers to find it. Probably this class is fine, or RundeckConfigBase. Also maybe changing this to an enum class could allow you to both define a constant, and map both old an new property keys to keep track of this legacy properties.

rundeckapp/src/main/groovy/org/rundeck/app/services/EnhancedNodeService.groovy
ahormazabal: Yes i agree. probably an enum class would allow us to define both constants and associated mappings when needed.

rundeckapp/build.gradle
sjrd218: I had to add this when I had my dev environment configured with the new enterprise reloadable config plugin. The way our gradle project is set up caused gradle to be unable to find the dependencies until I added this section on the OSS side. We could probably leave it out for the OSS build, but if anyone tries to use the pro config plugin in their dev env it will fail if this section isn't present. Ideas?

rundeckapp/grails-app/views/framework/projectNodeSources.gsp
```diff
@@ -82,7 +82,7 @@
                         </a>
                       </li>
 
-                    <feature:enabled name="enhanced-nodes">
+                    <feature:enabled name="enhancedNodes">
```
gschueler, line 85: maybe the feature:enabled tag lib can accept Feature enum type attribute e.g. `<feature:enabled feature="${Features.ENHANCED_NODES}">`

rundeckapp/grails-app/conf/application.groovy
```diff
@@ -116,6 +113,7 @@ grails.plugin.springsecurity.interceptUrlMap = [
         [pattern: '/user/loggedout', access: ['permitAll']],
         [pattern: '/feed/**',        access: ['permitAll']],
         [pattern: '/api/**',         access: ['permitAll']],
+        [pattern: '/health',         access: ['permitAll']],
```
gschueler, line 116: this seems unrelated to typed configs, why was this added?

rundeckapp/build.gradle
```diff
@@ -102,8 +102,8 @@ dependencies {
     compile "org.springframework.boot:spring-boot-autoconfigure"
 
     compile "org.grails:grails-core"
-    //compile "org.springframework.boot:spring-boot-starter-actuator"
-    compile ("org.springframework.boot:spring-boot-starter-jetty"){
+    compile "org.springframework.boot:spring-boot-starter-actuator"
```
gschueler, line 105: is the actuator necessary?

rundeckapp/grails-app/conf/application.groovy
sjrd218: This is added for the Pro side of this feature. When Rundeck registers itself with Consul, Consul likes to have a health check endpoint so it can test the apps that have registered and make sure they're still alive. The /heath endpoint is the standard endpoint it hits.

rundeckapp/build.gradle
sjrd218: The actuator was added to expose the health endpoint and also to add a refresh endpoint. After further research I'm not sure we really need this. I was using it as a quick way to expose the /health check endpoint, but we can do that ourselves without the other features that actuator also adds.

rundeckapp/grails-app/views/framework/projectNodeSources.gsp
sjrd218: I think we may need to reevaluate the Features enum. It's not a very flexible way to handle referencing features. If we are looking at adding and removing features at runtime I think we need something more flexible.

gschueler: hmm, perhaps a Feature bean type? the specific feature name can be a constant string in the implementation class

PR #5408, opened 2019-10-23
Add API endpoint for getting job workflow tree

ProTip:
* Adds `job/[ID]/workflow` endpoint to public API for retrieving a job's workflow tree

This is preparatory work for moving more functionality out into components in the `grails-spa` project. Specifically it is being utilized by the new execution output viewer component.

**Docs PR**
rundeck/docs#338

### DIFF
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/ScheduledExecutionController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/ScheduledExecutionController.groovy
@@ -184,7 +184,8 @@ class ScheduledExecutionController  extends ControllerBase{
             apiJobDeleteBulk             : ['DELETE', 'POST'],
             apiJobClusterTakeoverSchedule: 'PUT',
             apiJobUpdateSingle           : 'PUT',
-            apiJobRetry                  : 'POST'
+            apiJobRetry                  : 'POST',
+            apiJobWorkflow               : 'GET',
     ]
 
     def cancel (){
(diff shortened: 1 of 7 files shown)


gschueler:
it would "work", but needs some more changes to be a full API endpoint:
* api version check,
* error responses should be use apiService for correct response format
* if only JSON is supported, should respond 405 for other Accept requests.

Ideally, the workflowJson would just be updated to be a full api endpoint, and usage of workflowJson in the client-side js could be changed to use a true API request

ProTip:
@gschueler Converted to API endpoint and the usage I found converted to utilize it via an update `ts-rundeck` client.

Will merge the `ui-trellis` branch into master and update commit ref before merging this one.

rundeckapp/grails-app/controllers/rundeck/controllers/ScheduledExecutionController.groovy
```diff
@@ -627,7 +639,10 @@ class ScheduledExecutionController  extends ControllerBase{
                 ),
                 AuthConstants.ACTION_VIEW, 'Job', params.id
         )) {
-            return
+            return apiService.renderErrorFormat(response, [status: HttpServletResponse.SC_FORBIDDEN,
+                                                           code  : 'api.error.item.unauthorized', args: ['Run', 'Job ' +
```
gschueler:
```suggestion
                                                           code  : 'api.error.item.unauthorized', args: ['View', 'Job ' +
```

ProTip:
I realized when adding the docs that API v33 isn't the pending bump I thought it was.

PR #5440, opened 2019-10-31
Docker - Set uid in passwd for OpenShift support

hille721:
It is both :) 
OpenShift Container Platform runs containers using an arbitrarily assigned user ID, this means the rundeck process will be started with an UID not equal to the rundeck user defined in the passwd. This leads to problems, for example using ssh.
rundeck

Simple add a new rundeck passwd entry with the correct UID during the start of the container. 
Of course the /etc/passwd needs write permissions for the group (->'RUN chmod g=u /etc/passwd' has to be added to the Dockerfile as well)

**Additional context**
Morge background information:
https://docs.openshift.com/container-platform/3.3/creating_images/guidelines.html#openshift-container-platform-specific-guidelines

### DIFF
--- a/docker/official/lib/entry.sh
+++ b/docker/official/lib/entry.sh
@@ -52,7 +52,7 @@ fi
 # Support Arbitrary User IDs on OpenShift
 if ! whoami &> /dev/null; then
   if [ -w /etc/passwd ]; then
-    echo "rundeck:x:$(id -u):0:rundeck user:/home/server:/sbin/nologin" >> /etc/passwd
+    echo "rundeck:x:$(id -u):0:rundeck user:/home/rundeck:/bin/bash" >> /etc/passwd
   fi
 fi
 
--- a/docker/official/lib/entry.sh
+++ b/docker/official/lib/entry.sh
@@ -49,6 +49,13 @@ if [[ ! -z "${RUNDECK_ENVARS_UNSETS:-}" ]] ; then
     unset RUNDECK_ENVARS_UNSETS
 fi
 
+# Support Arbitrary User IDs on OpenShift
+if ! whoami &> /dev/null; then
+  if [ -w /etc/passwd ]; then
+    echo "rundeck:x:$(id -u):0:rundeck user:/home/server:/sbin/nologin" >> /etc/passwd
+  fi
+fi
+
 exec java \
     -XX:+UnlockExperimentalVMOptions \
     -XX:MaxRAMFraction="${JVM_MAX_RAM_FRACTION}" \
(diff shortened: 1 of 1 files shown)


docker/official/lib/entry.sh
```diff
@@ -49,6 +49,13 @@ if [[ ! -z "${RUNDECK_ENVARS_UNSETS:-}" ]] ; then
     unset RUNDECK_ENVARS_UNSETS
 fi
 
+# Support Arbitrary User IDs on OpenShift
+if ! whoami &> /dev/null; then
+  if [ -w /etc/passwd ]; then
+    echo "rundeck:x:$(id -u):0:rundeck user:/home/rundeck:/bin/bash" >> /etc/passwd
```
ProTip: I believe the Rundeck user may already exist in `passwd`. It may be more appropriate to replace the existing line/entry to match the runtime uid?

hille721: yes, that's true. Done with commit c0d4be7:
`sed -i "/rundeck/c\rundeck:x:$(id -u):0:rundeck user:\/home\/rundeck:\/bin\/bash" /etc/passwd`

hille721:
> Hi @hille721
> 
> Thanks for reporting this.. Would you be able to open a PR on the base image to adjust the `passwd` file group so it's modifiable in the entry?
> 
> https://github.com/rundeck/rundeck/blob/master/docker/ubuntu-base/Dockerfile

Done: #5944

docker/official/lib/entry.sh
```diff
@@ -49,6 +49,13 @@ if [[ ! -z "${RUNDECK_ENVARS_UNSETS:-}" ]] ; then
     unset RUNDECK_ENVARS_UNSETS
 fi
 
+# Support Arbitrary User IDs on OpenShift
+if ! whoami &> /dev/null; then
+  if [ -w /etc/passwd ]; then
+    sed -i "/rundeck/c\rundeck:x:$(id -u):0:rundeck user:\/home\/rundeck:\/bin\/bash" /etc/passwd
```
ProTip, line 55:
```suggestion
    sed -i "\#rundeck#c\rundeck:x:$(id -u):0:rundeck user:/home/rundeck:/bin/bash" /etc/passwd
```

docker/official/lib/entry.sh
hille721: With the sed statement Rundeck won't start anymore because the entry script failed with:
sed: couldn't open temporary file /etc/sedF47F9T: Permission denied

I guess sed will generate this as temporary file, but there are no write permissions in /etc. 

Solution would be to use the echo statement as it was at the beginning

PR #5441, opened 2019-10-31
Allow all plugin properties to use groupName and grouping rendering options.

sjrd218:
Allow all plugin properties to use groupName and grouping rendering options.
Also fixes the name of a service in the Plugin api integration test.

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/plugins/configuration/PluginAdapterUtility.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/plugins/configuration/PluginAdapterUtility.java
@@ -249,15 +249,9 @@ public class PluginAdapterUtility {
 
             ReplaceDataVariablesWithBlanks blankReplaceAnnotation = field.getAnnotation(ReplaceDataVariablesWithBlanks.class);
             if(blankReplaceAnnotation != null) pbuild.blankIfUnexpandable(blankReplaceAnnotation.value());
-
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/dtolabs/rundeck/core/plugins/configuration/PluginAdapterUtility.java
```diff
@@ -253,6 +253,11 @@ private static Property propertyFromField(final Field field, final PluginPropert
             for (RenderingOption renderingOption : field.getAnnotationsByType(RenderingOption.class)) {
                 pbuild.renderingOption(renderingOption.key(), renderingOption.value());
             }
+        } else {
+            for (RenderingOption renderingOption : field.getAnnotationsByType(RenderingOption.class)) {
+                if(renderingOption.key().equals(StringRenderingConstants.GROUP_NAME) || renderingOption.key().equals(StringRenderingConstants.GROUPING))
```
gschueler: any reason to restrict this to only "grouping/group_name" keys? i didn't realize non-string properties were skipping this annotation, probably we should just apply it to all properties

gschueler: originally, yes, however i think we should just add them all to the properties.  if they don't make sense for that property type they will be ignored, but better to allow it to be open ended for future options

sjrd218:
Should this go in 3.2 or wait?

gschueler:
ok for 3.2

PR #5445, opened 2019-11-01
Fixes failing test in audit event service.

sjrd218:
Fixes failing test in audit event service.

Provides equivalent data in a test friendly way.


ahormazabal:
Also took the liberty to push a fix to a similar error on another test.

gschueler:
the alternative is to just put the InetAddress call somewhere else that can be stubbed out in the test

ahormazabal:
I like this change because it centralizes the obtention of the hostname on the framework instead of having to call InetAddress independently and eventually stubbing it.
Maybe we could discuss if the result of that method should be the property or the inet call, but personally i like to have it at the framework level.

PR #5453, opened 2019-11-05
Lifecycle plugin updates

gschueler:

Some fixes for lifecycle plugins:

* UI: CODE rendering option in job edit page not loading for exec lifecycle plugins
* Add utility implementation of expected return type for execution lifecycle plugins
* fix: `useSingleComponentOfType` method not working correctly

### DIFF
--- a/rundeckapp/grails-app/views/scheduledExecution/_edit.gsp
+++ b/rundeckapp/grails-app/views/scheduledExecution/_edit.gsp
@@ -1250,6 +1250,9 @@ function getCurSEID(){
 <g:if test="${params.executionLifecyclePluginValidation}">
             jobeditor.addError('plugins');
 </g:if>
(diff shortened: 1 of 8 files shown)


rundeckapp/grails-app/views/scheduledExecution/_edit.gsp
```diff
@@ -1250,6 +1250,9 @@ function getCurSEID(){
 <g:if test="${params.executionLifecyclePluginValidation}">
             jobeditor.addError('plugins');
 </g:if>
+            jQuery('.apply_ace').each(function () {
+                _setupAceTextareaEditor(this, confirm.setNeetsConfirm);
+            });
```
sjrd218: @greg Is this causing the duplicate on the ruleset UI?

PR #5459, opened 2019-11-06
Table word wrap issue, fixes #5384

### LINKED ISSUE #5384: GUI - Unable to Delete User API Tokens with large role list
**Describe the bug**
On the User Profile Page API Tokens section, if a token has a large number of roles or many roles with long names, the Roles column will push/move the Delete button off-screen. The only way to access the delete button is to Scale the Browser window down.

**My Rundeck detail**
* Rundeck version: 3.0.26-20190829 º 3.1.2-20190927 
* install type: rpm
* OS Name/version: Centos 7
* DB Type/version: MySQL

**To Reproduce**
Create a User with many roles, goto Profile page, and create an API token for the user that has all the roles - see attached.

**Expected behavior**
We expect the column to be truncated/wrapped (similar to how it works in the First Name, Last Name, and Groups of the User Profile in the section above) Perhaps the integration of a Horizontal scroll bar would be helpful as well.

**Screenshots**
v.3.0x
**Firefox**
GH-3 0-TokenFirefox
GH-3 0-TokenFirefox67percent
**Chrome**
GH-3 0-TokenChrome
GH-3 0-TokenChrome67percent

v.3.1x
**Firefox**
GH-3 1-TokenFirefox
GH-3 1-TokenFirefox67percent
**Chrome**
GH-3 1-TokenChrome
GH-3 1-TokenChrome67percent

**Desktop:**
 - OS: Ubuntu 18.04 or Windows 10
 - Browser Chrome 77 and Firefox 69

jessemarple:
User token table, when a user has many roles, does not wrap the text therein, obsecuring the delete button off the viewport. This fix adds an inline "word-break" to the containing span and removes "ellipsis" class that is both unnecessary and incorrect.

Screen Shot 2019-11-06 at 11 27 42 AM

### DIFF
--- a/rundeckapp/grails-app/views/user/_token.gsp
+++ b/rundeckapp/grails-app/views/user/_token.gsp
@@ -59,8 +59,8 @@
                 (${token.creator})
             </g:if>
         </td>
-        <td width="30%" class="ellipsis" title="${token.authRoles}">
-            <span>${token.authRoles}</span>
+        <td width="30%" title="${token.authRoles}">
+            <span style="word-break: break-all;">${token.authRoles}</span>
         </td>
         <td width="10%">
           <a style="padding-left:14px; padding-right: 14px; ${wdgt.styleVisible(if: token.token && !(params.showConfirm && params.token==token.token))}"


ProTip:
fixes #5384

niall-munnelly:
Hey, awesome, I was just about to report this in 3.0.x. Will this fix be backported? Thanks.

PR #5460, opened 2019-11-06
Fix duplicate ACE editors being added on job edit page.

sjrd218:
ACE editor setup call is already being done in the jobedit.js. This seems to have the effect of adding it twice.


gschueler:
so it looks like the ace editor for workflow strategy is inited here https://github.com/rundeck/rundeck/blob/master/rundeckapp/grails-app/views/execution/_execDetailsWorkflow.gsp#L95-L100

so the apply_ace block i added (removed in this pr) could be modified to target the  `#tab_execution_plugins` selector to limit it to just that div

gschueler:
```
jQuery('#tab_execution_plugins textarea.apply_ace').each(function () {
```

gschueler:
this is ok to merge, this missing ace editor in exec plugins tab can be fixed later

PR #5469, opened 2019-11-07
fix: max pagination config not working

gschueler:
Fixes issue: max pagination config not working


jtobard:
This works ok on API calls but not on the GUI activity page.

According to the docs:
`Default paging size for the Activity page and results from execution API queries can be changed.`

Using:
```
rundeck.pagination.default.max=40
```

API (set correctly the max to 40):
api-max40

GUI (always show 30 per page):
image


This may be a problem with the documentation if this property doesn't affect the activity page.

MegaDrive68k:
Using `rundeck.pagination.default.max=20` at rundeck-config.properties file, keeps 30 jobs per page.

Screenshot_1

MegaDrive68k:
It works!

`rundeck.pagination.default.max=10` on `rundeck-config.properties` file

Result:

Screen Shot 2021-10-19 at 12 53 39

PR #5500, opened 2019-11-15
WIP: add skipped status for steps

gschueler:

Adds "skipped" status for steps, allowing workflow state to correctly show status instead of "not started" when a step is meant to be skipped.


fdevans:
This came up recently along the lines of Scheduling as well.

PR #5518, opened 2019-11-19
Enable LDAP sync in Docker

kenske:
Allow enabling LDAP Sync using docker environment variables (`RUNDECK_LDAP_SYNC: "true"`)


ltamaster:
Hi @kenske 
This was already added here
https://github.com/rundeck/rundeck/blob/master/docker/official/remco/templates/rundeck-config.properties#L43

you can use `RUNDECK_SECURITY_SYNCLDAPUSER: "true"`

Luis

kenske:
@ltamaster I don't know how I missed that. Thanks, closing.

PR #5532, opened 2019-11-21
variable execution.succeededNodeList to observe real node execution on notification

### LINKED ISSUE #5517: Context Variable ${execution.succeededNodeList} Doesn't Observe Random Subset
**Describe the bug**
I use the Splunk http/json notification frequently. I have some jobs that execute against a random subset of a very large pool of nodes. When the notification fires with context variable "${execution.succeededNodeList}," that expands to _all_ nodes picked up by the filter, rather than the, say, twenty or so selected by the random subset orchestrator. There's no easy way to track executions against nodes in jobs using the random subset orchestrator. This is admittedly an outlier in my environment, but it's not ideal.

**My Rundeck detail**

* Rundeck version: 3.0.16
* install type: launcher-war
* OS Name/version: RHEL 7
* DB Type/version: postgres

**To Reproduce**
1. Create a job with an inclusive node filter, but use the random subset orchestrator to limit nodes to a small number
2. Set up a notification that would typically support context variable ${execution.succeededNodeList}. In the Splunk notification, I add it to the "events" field.
3. Execute job
4. View notification

**Expected behavior**
If random subset is set for 20 nodes, then I would expect 20 nodes to be in the list for ${execution.succeededNodeList}. Instead, I see everything included by the node filter (and, in this case, that's quite a lot of nodes).

**Screenshots**
Sorry.

 - OS: Mac
 - Browser Chrome, Safari
 - Version Latest

**Additional context**
Thanks.

jtobard:
Modified `${execution.succeededNodeList}`  on notifications based on real node list not on the target node list.
Uses the `workflowService` to obtain the list of nodes and add only the nodes that really run the job, in case the workflow executor execute only in a portion of the nodes, as example, the random subset Orchestrator.

Fix #5517

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/NotificationService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/NotificationService.groovy
@@ -70,6 +70,7 @@ public class NotificationService implements ApplicationContextAware{
     def LoggingService loggingService
     def apiService
     def executionService
+    def workflowService
     OrchestratorPluginService orchestratorPluginService
 
     def ValidatedPlugin validatePluginConfig(String project, String name, Map config) {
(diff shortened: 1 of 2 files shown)


MegaDrive68k:
Tested and works OK.

Without Random Subset.

Screenshot_1

With Random Subset = 1

Screenshot_2

PR #5534, opened 2019-11-21
Task/5504/add support type

jessemarple:
Implements a support type notification that the OSS version of Rundeck is unsupported and provides no warranty.

To better make use of the space provided in the sidebar and allow for the placement of another instance of the support type indicator; the build date has been moved into a tooltip that is visible on hover of the version number.

Screen Shot 2019-11-21 at 6 20 39 PM

Screen Shot 2019-11-21 at 6 21 19 PM


jessemarple:
The Licenses notification has been removed from the footer on the Login/Logout pages.

PR #5542, opened 2019-11-26
fix #602 Changes to preserve quotes on arguments commands so rundeck will interpret it as unique argument

### LINKED ISSUE #602: preserve quotes used in commands
If a command has a quoted string e.g. `echo " hi there"`, the quotes are interpreted when read by rundeck, but not preserved literally on output.  Current rules require double-quoting of input strings to emit literal quotes: http://rundeck.org/docs/manual/job-workflows.html#quoting-arguments-to-steps

Suggested enhancement to make this more intuitive:

if an argument string has spaces and is not quoted (and is does not contain variable expansion like `${option.value}`, it should be emitted quotes added

carlosrfranco:
fixes #602 

If a command has a quoted strings like `command -m 'arg1 arg2'`, this arguments was not preserved by rundeck and the argument `'arg1 arg2'`  was interpreted as two separated args.

Quotes will be preserved and is interpreted as unique argument


rundeckapp/grails-app/views/execution/_wfitemEdit.gsp
```diff
@@ -475,6 +475,21 @@
...
                     </div>
+                    <div class="form-group" >
+                        <label class="col-sm-2 control-label"></label>
+                        <div class="col-sm-10">
+                            <div class="checkbox">
+                                <g:checkBox name="preserveQuotesOnArguments"
+                                            checked="${item?.preserveQuotesOnArguments}"
+                                            id="preserveQuotesOnArgumentsCheck" value="true"
+                                />
+                                <label for="preserveQuotesOnArguments">
```
gschueler: incorrect value prevents label from working
```suggestion
                                <label for="preserveQuotesOnArgumentsCheck">
```

carlosrfranco: Fixed

carlosrfranco:
> I think we need some additional clarity on the behavior:
> 
>     * preserving quotes should be the default behavior when creating new commands.  however it should should be off for existing commands for compatibility, and the checkbox to toggle it should also explain a bit more about why you might want to disable it for compatibility.
> 
I added a tooltip with additional explaining for the behavior.

Captura de ecrã de 2020-01-03 11-34-37

> 
> Commands run from the "Commands" section work in the old (non-preserving) way. Perhaps another toggle needs to be added to Commands UI to change this as well

Added on last commit

ahonor:
Is there a feature flag that can toggle this feature off/on?

carlosrfranco:
> Is there a feature flag that can toggle this feature off/on?

In this current state does not have the feature flag. I understand that the idea is to keep the previous behavior for existing jobs so even without the feature flag, the jobs will not be affected. But of course it wouldn't be a problem to add the flag as well

gschueler:
@carlosrfranco  @ahonor yes let's add a feature toggle. if turned off, the old behavior is used and the checkbox for command steps and commands tab is not shown.  if turned on, the new behavior is default, but checkboxes are shown.  I dislike the checkboxes, but not sure of a better way to allow the new behavior to be default and still have compatibility for old job definitions

rundeckapp/src/main/groovy/com/dtolabs/rundeck/app/api/ApiRunAdhocRequest.groovy
```diff
@@ -37,10 +37,12 @@ class ApiRunAdhocRequest implements Validateable{
     String fileExtension
     Boolean nodeKeepgoing
     Integer nodeThreadcount
+    Boolean preserveQuotesOnArgumentsCheck
```
gschueler: suggest renaming this to `preserveQuotes`

gschueler: also, as this now affects the API, the `apiRunCommand` in ScheduledExecutionController should support this parameter only in newer API version.

carlosrfranco:
@gschueler Will the feature flag be off by default?

rundeckapp/src/main/groovy/com/dtolabs/rundeck/app/api/ApiRunAdhocRequest.groovy
gschueler: I think also, given newer API version, the run command API should use preserveQuotes=true by default if it is not specified, otherwise set it to false.

ahonor: @carlosrfranco yes, it should be toggled off by default.

carlosrfranco:
PR updated and moved to https://github.com/rundeck/rundeck/pull/6887

PR #5555, opened 2019-11-28
Fix unexpected behavior when killing a job that includes Job Reference of workflow type. 5478

jtobard:
Abort referenced execution before it has time to start if the execution has been aborted. 

Now the aborted job reference step works in job references that run as a workflow step.
image

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
@@ -3697,15 +3697,19 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
             )
 
             thread.start()
+            if(exec.abortedby){
+                thread.abort()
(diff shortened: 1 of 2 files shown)


rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -3697,15 +3697,19 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
             )
 
             thread.start()
```
gschueler: if the exec.abortby is set already, why start the thread at all?

jtobard: I had problem not starting the thread (the steps stays forever in Waiting status) but now I tested again and it looks like those problems were caused by Intellij using debug mode and some stuck threads, now it works ok. I'm going to push the fix without starting it.

PR #5564, opened 2019-12-03
Add uuid to webhook

sjrd218:
Replace uses of auth token to identify a webhook with a neutral uuid. This allows recreation of auth token upon import and export and doesn't leak the token anywhere.


rundeckapp/webhooks/grails-app/services/webhooks/WebhookService.groovy
```diff
@@ -179,6 +179,13 @@ class WebhookService {
         }
     }
 
+    boolean importIsAllowed(Webhook hook, Map hookData) {
+        if(hook.authToken == hookData.authToken) return true
+        if(!hook.authToken && Webhook.countByAuthToken(hookData.authToken) == 0) return true
```
gschueler: an additional check needed is something like `AuthToken.countByToken(hookData.authToken)==0`

PR #5566, opened 2019-12-03
Webhook validation fixes

gschueler:

Fix save and validation problems with webhooks
- [x] during "save", if validation error for a plugin, the changes still get saved
- [x] gui should show plugin validation problems
- [x] switching plugin types should clear any previous plugin config values
- [x] auth token should not be created if new webhook validation fails
- [x] auth token should be deleted if gorm save fails for new webhook


gschueler:
@sjrd218 thanks i'll fix that

PR #5577, opened 2019-12-06
ssh export rd variables

ronaveva:
Enhancement:
You can now inyect RD_ variables as part of the ssh commands, so no modification to the ssh configuration on the client side is needed

- This can be enabled/setup at node level making it able to run with different sintaxis depending on the node config.

EG:
1.-Linux node
ssh-variable-export-pattern="export {key}='{value}'"
2.-Windows node
ssh-variable-export-pattern="set {key}='{value}'"

- Custom separator can also be set with the following node attribute:

ssh-variable-export-separator

EG:
1.-ssh-variable-export-separator=";"
2.-ssh-variable-export-separator="\n"

- Node attributes can now be excluded from the export, using the following node attribute:

ssh-variable-export-exclude-nodes="true"

- Example job to print the environment variables

[simpleCommand.zip](https://github.com/rundeck/rundeck/files/4010135/simpleCommand.zip)

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/execution/ExecutionServiceImpl.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/execution/ExecutionServiceImpl.java
@@ -413,7 +413,13 @@ class ExecutionServiceImpl implements ExecutionService {
         );
 
         NodeExecutorResult result = null;
+
         String[] commandArray = commandList.toArray(new String[commandList.size()]);
+
+        if(nodeExecutor.supportVariableInjection()){
+            commandArray = NodeExecutorUtils.getExportedVariablesForNode(node, nodeContext, commandList);
+        }
+
         try {
             result = nodeExecutor.executeCommand(nodeContext, commandArray, node);
         } finally {
@@ -427,6 +433,5 @@ class ExecutionServiceImpl implements ExecutionService {
     public String getName() {
         return SERVICE_NAME;
     }
-
 
 }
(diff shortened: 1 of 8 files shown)


ltamaster:
Hi @ronaveva 
My comments:
1) I think the pattern variables shouldn't be on `JschNodeExecutor`class, maybe it should be part of the INodeEntry interface

2) maybe you could add a way to filter the environment variables. For example, if the node has 300 attributes the command could fail, or maybe I could not need the node variables but the context ones.

ronaveva:
Modified according to requested changes

ahonor:
Execution Service does not seem the right place to inject Unix shell syntax

ronaveva:
@ahonor @jtobard but those attributes are more related to the execution it self and the node than the actual OS where the command will be executed.

gschueler:
not every executor will allow that kind of variable setting.  If we want multiple executors to allow this, we can move it to a utility class that can be used from any executor. but it should be up to the NodeExecutor to do it, so, perhaps a better place is a new class like `NodeExecutorUtils`

core/src/main/java/com/dtolabs/rundeck/core/execution/ExecutionServiceImpl.java
```diff
@@ -33,6 +33,7 @@
 import com.dtolabs.rundeck.core.execution.dispatch.DispatcherException;
 import com.dtolabs.rundeck.core.execution.dispatch.DispatcherResult;
 import com.dtolabs.rundeck.core.execution.dispatch.NodeDispatcher;
+import com.dtolabs.rundeck.core.execution.impl.jsch.JschNodeExecutor;
```
ahonor: Is this import necessary?

PR #5579, opened 2019-12-06
Fix #5001 - Adapting query so dashboard can show one time scheduled job (run job later) as scheduled and not as running job

### LINKED ISSUE #5001: one time scheduled job seen as running in Rundeck Dashboard
When scheduling a job to run one time, using the "Schedule Job later" option on the top right side of the jobs page, the job will be shown as running in the Rundeck Dashboard, instead of its schedule.https://files.slack.com/files-pri/T0UG2B0HK-FKL3GHRK5/image.png

schedule_once

schedule_once_dashboard

But when it's scheduled by editing the job with the "schedule to run it repeatedly" option, it appears correctly scheduled in the Dashboard graph.

schedule_to_run_repeatedly

schedule_to_run_rep_dashboard

carlosrfranco:
Fix #5001 
A customer requested to change dashboard behaviour so jobs that will run later should be shown as scheduled and not as running

The query was adapted to filter jobs with date start setted to a date future

### DIFF
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
@@ -412,6 +412,14 @@ class MenuController extends ControllerBase implements ApplicationContextAware{
                                 serverOwner   : se.serverNodeUUID == serverNodeUUID
                         ]
                     }
+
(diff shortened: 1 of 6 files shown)


carlosrfranco:
> Can you add test to this

Tests was added and the bugfix was validated by support.

PR #5585, opened 2019-12-09
new version pywinrm plugin 2.0.5

ltamaster:
plugin bug fix:
https://github.com/rundeck-plugins/py-winrm-plugin/issues/39

**Additional context**


ltamaster:
@jtobard 
I would say is not necessary a backport PR, because the workaround is just installing the new version of the plugin.

PR #5615, opened 2019-12-18
Support toggling features with case sensitive names

ProTip:


ProTip:
@Mergifyio backport maint-3.2.x

PR #5623, opened 2019-12-19
 Fix incorrect job schedule display in cluster mode

sjrd218:
Fixes an issue where job next execution times may display incorrectly in a cluster.


sjrd218:
See https://github.com/rundeckpro/rundeckpro/issues/834 for a description of the problem being fixed.

PR #5658, opened 2019-12-30
update jgit to 5.6.0.201912101111-r

lswith:
A clear and concise description of the bug you are fixing (reference issue number), or enhancement you are implementing.
This is a simple dependency update for the jgit library. Quite a few bugs have been fixed in the last 4 years by this library.

I've updated the library, and fixed 1 api call. The `release()` has now been changed to `close()`.

### DIFF
--- a/plugins/git-plugin/build.gradle
+++ b/plugins/git-plugin/build.gradle
@@ -37,11 +37,11 @@ repositories {
 
 dependencies {
     compile project(":core")
-    pluginLibs( 'org.eclipse.jgit:org.eclipse.jgit:3.7.1.201504261725-r') {
-        exclude module: 'slf4j-api'
+    pluginLibs( 'org.eclipse.jgit:org.eclipse.jgit:5.6.0.201912101111-r') {
         exclude module: 'jsch'
-        exclude module: 'commons-logging'
+        exclude module: 'slf4j-api'
     }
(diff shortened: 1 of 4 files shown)


plugins/git-plugin/build.gradle
```diff
@@ -37,11 +37,11 @@ repositories {
...
 dependencies {
     compile project(":core")
-    pluginLibs( 'org.eclipse.jgit:org.eclipse.jgit:3.7.1.201504261725-r') {
-        exclude module: 'slf4j-api'
+    pluginLibs( 'org.eclipse.jgit:org.eclipse.jgit:5.6.0.201912101111-r') {
         exclude module: 'jsch'
-        exclude module: 'commons-logging'
+        exclude module: 'slf4j-api'
     }
+    compile 'org.slf4j:slf4j-api:1.7.29'
```
lswith: this is the same version as what is required from the springboot library.

jtobard:
Tested and it works ok to me, just one question. There's any documentation about why `release()` has now been changed to `close()`?
Maybe there are other enhancement or modifications we can make using this new version.

lswith:
@sjrd218 fixed.
@jtobard I think there were a few things broken in this plugin which is why I was hoping to upgrade. I'm hoping someone can continue to support the GitPlugin in the future and use those key enhancements/modifications.

PR #5695, opened 2020-01-09
Bugfix: close logread in tailExecutionOutput

dbt:
When the browser is polling for logging updates on an active job,
tailExecutionOutput can be called many times in succession. Each
call leaks a file descriptor until the garbage collector cleans up
the underlying FileInputStream.

This can be easily verified by writing a long running noisy job,
trying to watch the logs in the browser, then monitoring the list
of open files using something like lsof or, on linux,
/proc/$rundeckPid/fd/.


dbt:
Thanks for testing. What's left to be able to merge?

PR #5703, opened 2020-01-13
Upgrade snakeyaml

### LINKED ISSUE #5702: Upload plugin artifact to private repository fails
**Describe the bug**
Attempting to upload a plugin artifact to a private repo configured in Rundeck results in an error. The stacktrace shows a mismatched yaml library as the cause.

**My Rundeck detail**

* Rundeck version: any version greater than 3.1.0

**To Reproduce**
Configure a private repository.
Try to add a plugin to the repo using the api.

**Expected behavior**
Plugin upload succeeds.

sjrd218:
Fixes #5702 by upgrading snakeyaml to the same version used by the repository library.

### DIFF
--- a/core/src/test/java/com/dtolabs/rundeck/core/common/TestNodesYamlParser.java
+++ b/core/src/test/java/com/dtolabs/rundeck/core/common/TestNodesYamlParser.java
@@ -307,25 +307,25 @@ public class TestNodesYamlParser extends TestCase {
         Assert.assertEquals(null, recv.nodes.get("bill").getHostname());
 
     }
-// AS OF SNAKEYAML 1.21 scalars can now contain colons, so this test will no longer fail
-//    public void testParseInvalid_unexpecteddatatype() throws Exception {
-//            //unexpected data type for string field
-//            testReceiver recv = new testReceiver();
-//            ByteArrayInputStream is = new ByteArrayInputStream(
-//                ("test: \n"
-//                 + "  nodename: bill\n"
-//                 + "  hostname: test\n"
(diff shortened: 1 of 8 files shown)


core/src/test/java/com/dtolabs/rundeck/core/common/TestNodesYamlParser.java
```diff
@@ -307,24 +307,25 @@ public void testParseInvalid_allow_missing_hostname() throws Exception {
...
-        }
+// AS OF SNAKEYAML 1.21 scalars can now contain colons, so this test will no longer fail
+//    public void testParseInvalid_unexpecteddatatype() throws Exception {
+//            //unexpected data type for string field
+//            testReceiver recv = new testReceiver();
+//            ByteArrayInputStream is = new ByteArrayInputStream(
+//                ("test: \n"
+//                 + "  nodename: bill\n"
+//                 + "  hostname: test\n"
+//                 + "  username: {test:value, a:b}\n").getBytes());
```
gschueler: can this be changed to valid yaml array syntax, so that the test still tries to parse invalid datatype for this field?

gschueler: otherwise, could this just be rewritten instead of removed?  i'd have to review exactly what it is testing, but it appears to try to send a non-string attribute value to the node yaml parser, expecting an error. if the current input evaluates to a valid string, perhaps just changing it to result in a map/array would work

sjrd218: Added the test back

PR #5707, opened 2020-01-14
Fix ordering violation in Gradle scripts

theosotr:
Hello

This pull request fixes an ordering issue appearing in Gradle script.
Specifically, the task `core:compileJava` creates the `src/main/resources/META-INF/com/dtolabs/rundeck/core/application.properties` file from this [template](https://github.com/rundeck/rundeck/blob/master/core/src/main/meta/com/dtolabs/rundeck/core/application.properties).

This generated file is treated as resource; thus, the task `processResources` must run after `compileJava` so that `application.properties` exists in the file system. When `processResources` runs before `compileJava`, the build process produces an in correct jar file, i.e., the directory `META-INF` does not contain `application.properties`.

That's the diff of the contents of the `core:jar` file

````
< META-INF/com/
< META-INF/com/dtolabs/
< META-INF/com/dtolabs/rundeck/
< META-INF/com/dtolabs/rundeck/core/
< META-INF/com/dtolabs/rundeck/core/application.properties
````

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -150,9 +150,12 @@ def expandTemplate = {
 }
 
 
+compileJava.outputs.file "$projectDir/src/main/resources/META-INF/com/dtolabs/rundeck/core/application.properties"
+compileJava.inputs.file "$projectDir/src/main/meta/com/dtolabs/rundeck/core/application.properties"
 compileJava.doFirst {
         expandTemplate()
 }
+processResources.mustRunAfter compileJava
 
 test{
     systemProperties 'rdeck.base': "$projectDir/build/rdeck_base"


ProTip:
Hi @theosotr ,

Thanks for finding this and proposing a fix! We hit this a while back attempting to use `--parallel` in the build and had to back out.

Lately we've been on a bit of a tear fixing up our Gradle build configurations. I ended up reviewing this and didn't like the way we had originally implemented it; didn't seem very idiomatic Gradle. So in PR #5832 I converted the code to a task and made `processResources` depend on that which seems to order correctly.

Could you check to see if that resolves the issue for you as well? If so do you want to convert this to an issue so we can link to it and get it in the release notes?

Thanks again!

theosotr:
Hi @ProTip 
Indeed https://github.com/rundeck/rundeck/pull/5832 fixes this issue.
Thank you so much for your reply.

PR #5737, opened 2020-01-28
 Changing the mechanism that loads the plugins into memory to consider the type of the plugin in addition to its name

carlosrfranco:
This changes is a bugfix for https://github.com/rundeckpro/rundeckpro/issues/849

This changes fixes a wrong behaviour that did not allow plugins of different types to have the same provider name

Was changed the way that the plugins is registered into registryMap and loaded in memory. So now it is using the type in addition to name provider on the registryMap


core/src/main/java/com/dtolabs/rundeck/core/plugins/PluginRegistry.java
```diff
@@ -210,4 +210,12 @@ ValidatedPlugin validatePluginByName(String name, PluggableProviderService servi
...
      */
     public PluginMetadata getPluginMetadata(String service, String provider) throws ProviderLoaderException;
+
+    /**
+     * Register a plugin into map using type and name as key to load it when requested
+     * @param type
+     * @param name
+     * @param beanName
+     */
+    void registryPlugin(String type, String name, String beanName);
```
gschueler: suggested rename of this method
```suggestion
    void registerPlugin(String type, String name, String beanName);
```

PR #5746, opened 2020-01-30
Update asset pipeline plugin for gradle and grails.

sjrd218:
Update asset pipeline plugin for gradle and grails.

### DIFF
--- a/rundeckapp/grails-app/assets/less/library/markdeep.less
+++ b/rundeckapp/grails-app/assets/less/library/markdeep.less
@@ -26,7 +26,7 @@
   OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
   THE POSSIBILITY OF SUCH DAMAGE.
 */
-
+@import "bootswatch";
 .md code, pre {
   font-family: Menlo, 'Lucida Console', monospace;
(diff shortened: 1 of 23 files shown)


gschueler:
were all of the scss/less changes required for the new plugin version?

sjrd218:
Apparently the newer libraries used by the latest version of the plugin are more strict than the previous ones. The changes made were necessary for the plugin to process the files.

rundeckapp/grails-app/assets/less/library/markdeep.less
```diff
@@ -156,9 +156,9 @@
 
 .md table.table {
   margin: auto;
-  .table-bordered();
-  .table-condensed();
-  .table-hover();
+ // .table-bordered();
```
jessemarple: Have you seen what this looks like in a Markdeep/markdown UI somewhere in the app? 
table-bordered, -condensed, -hovered - all contribute to the way a table looks in markdown output. What we'd see here is a table w/o borders, cell padding, and no onHover effect (row highlighting).

Can you confirm that there's no UI regression please.

sjrd218: There appears to be a regression since the 2.11.x days. The current behavior is the same as the 3.0.x version, so the regression was introduced in that transition.

PR #5749, opened 2020-01-31
Fixing a n+1 issue removing a code that accessed the list of executions and was making a query for each one. 

### LINKED ISSUE #5735: JobsAjax causing Rundeck to become unresponsive
**Describe the bug**
When there are many historical executions for jobs in a project, the `jobsAjax` call can result in `n+1` queries against the database. In my case the following query runs ~100k times(~1 time per entry in the execution table):

<details>
<summary>Query</summary>

```
select logfilesto0_.id as id1_5_0_, logfilesto0_.version as version2_5_0_, logfilesto0_.plugin_name as plugin_n3_5_0_, logfilesto0_.date_created as date_cre4_5_0_, logfilesto0_.filetype as filetype5_5_0_, logfilesto0_.last_updated as last_upd6_5_0_, logfilesto0_.completed as complete7_5_0_, logfilesto0_.execution_id as executio8_5_0_ from log_file_storage_request logfilesto0_ where logfilesto0_.execution_id=68260;
```
</details>

This can lead to the following sequence of events on the Dashboard page with scheduling:
1. The first query runs and takes ~30 seconds to complete
2. Subsequent queries start before the first finishes
3. Queries continue to stack up until the heap usage approaches the limit, at which point Rundeck becomes largely unresponsive

This can occur with one dashboard page open, eventually, or much more quickly with multiple dashboard pages open in separate tabs/browsers.

**Impact**
Rundeck has a good chance of becoming unresponsive under the following conditions:
* ~4GB max heap
* Remote execution storage setup(perhaps without as well)
* Large (>~80k) numbers of historical executions
* One or more users on a page that hits `jobsAjax`, such as the Enterprise dashboards page with job scheduling displayed

Larger heap sizes may be able to mitigate unresponsiveness, however more executions and concurrent `jobsAjax` requests will eventually fill the heap.

**My Rundeck detail**

* Rundeck version: 3.2.1
* install type: docker
* OS Name/version: n/a
* DB Type/version: mysql

**Screenshots**

image

**Additional Context**
VisualVM profile during request:
[dashboard_profile.nps.gz](https://github.com/rundeck/rundeck/files/4120382/dashboard_profile.nps.gz)

carlosrfranco:
Fixes #5735 
Fixes rundeckpro/rundeckpro#873

Mistakenly, was inserted a code which made a query for each execution of a job. It caused a n+1 issue and that strongly impacted performance

Was changed the code so instead of a query for each execution, now should run only one query for the executions.

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
@@ -4374,9 +4374,9 @@ class ScheduledExecutionService implements ApplicationContextAware, Initializing
     /**
      * Return a map with date start (timestamp) of one time scheduled executions (Job Run Later)
      * @param se
-     * @return Map with timestamp scheduled of a ScheduledExecutions
+     * @return Map<Long, Date> with timestamp scheduled of a ScheduledExecutions
      */
-    Map nextOneTimeScheduledExecutions(List<ScheduledExecution> se) {
+    Map<Long, Date> nextOneTimeScheduledExecutions(List<ScheduledExecution> se) {
         Date now = new Date()
         Map item = [:]
         Execution.createCriteria().list() {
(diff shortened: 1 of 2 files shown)


rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
```diff
@@ -355,7 +356,7 @@ class ScheduledExecutionService implements ApplicationContextAware, Initializing
                 }
 
                 add(Restrictions.disjunction().add(Subqueries.exists(subquery)).add(restr))
-            }
+            } else if(boolfilters.any {key, val -> null!=query["${key}Filter"]}){ add(restr) }
```
gschueler: should not be a single line, it becomes unreadable

ProTip:
@Mergifyio refresh

Issue #1351, opened 2015-07-21
Slow main page load


RomeyRome206:
We have 500~ jobs
18 projects
9000~ daily executions
MySQL back-end (no slow queries in log)

/menu/home takes about 30-45 seconds to load. It has progressively gotten worse as we add jobs.

ahonor:
My guess is the database needs indexing. Any DBA there that can check?

rgazaryants:
That was my first guess too, but the longest query in the slow log is 2~ seconds.

mathieuchateau:
Hello,

If no big/slow query, it may be there numbers (if sequential).
Can you do a tcpdump on rundeck node (filtering to MySQL IP) ?

So we may see how long time it is doing SQL query out of the 30-45 total
seconds.

Cordialement,
Mathieu CHATEAU
http://www.lotp.fr

2015-07-21 4:43 GMT+02:00 rgazaryants <notifications@github.com>:

gschueler:
this might be slow/inefficient page rendering of the jobs page.  for comparison, how long does it take to execute "rd-jobs" to list the jobs for the same project (or use the API directly)?  If there is a significant difference this would point to page rendering being the cause of the delay

rgazaryants:
Our biggest project:

```
# time rd-jobs list -p DataWarehouse
real	1m12.425s
user	0m1.504s
sys 	0m0.136s

# time curl "http://localhost/api/12/jobs?project=DataWarehouse&authtoken=***
real	0m1.632s
user	0m0.000s
sys     0m0.004s
```

gschueler:
I'd like to reproduce this...you say you have 500 jobs in this project.  what kind of folder/group structure do you have? (number of levels, number of jobs per folder)

rgazaryants:
Actually, it's 500 jobs across 18 projects. Nothing more than 2 levels deep - so **/folder_1/folder_2/job**. 71 jobs in 1 folder at the extreme. Typical is 10~.

Just to add, if I import all the same jobs to a clean server with no execution history, I don't see the loading delay. Seems like the delay is in counting up executions+users in the last day.

rgazaryants:
I set up a retention job to delete executions older than 30 days - it needed to be done anyway.

It matched about 280k executions. We only started migrating to Rundeck about 3 months ago. It's going pretty slow(via API), but that's expected since it's the first run. We'll see how it performs once it's done.

rgazaryants:
[...]

PR #5779, opened 2020-02-11
Some improvements to job list page speed with many jobs

gschueler:
* job list page with many jobs is slow

* reduce inefficient resource access and curtail unnecessary code paths
* fix UI issue when job list paging was enabled

### DIFF
--- a/rundeckapp/grails-app/views/menu/jobs.gsp
+++ b/rundeckapp/grails-app/views/menu/jobs.gsp
@@ -24,7 +24,7 @@
     <meta name="skipPrototypeJs" content="true"/>
     <g:set var="projectName" value="${params.project ?: request.project}"/>
     <g:set var="projectLabel" value="${session.frameworkLabels?session.frameworkLabels[projectName]:projectName}"/>
-    <g:set var="paginateJobs" value="${grailsApplication.config.rundeck.gui.paginatejobs}" />
+    <g:set var="paginateJobs" value="${grailsApplication.config.rundeck.gui.paginatejobs.enabled}" />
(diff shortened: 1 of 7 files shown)


jessemarple:
> tested local. Looks good.

@jtobard - here's 2K jobs https://github.com/rundeckpro/rundeckpro/files/3984649/many_jobs_2000-20191219-135237.rdproject.jar.zip

Can you add that job and give this a test.

rundeckapp/grails-app/views/menu/jobs.gsp
```diff
@@ -24,7 +24,7 @@
     <meta name="skipPrototypeJs" content="true"/>
     <g:set var="projectName" value="${params.project ?: request.project}"/>
     <g:set var="projectLabel" value="${session.frameworkLabels?session.frameworkLabels[projectName]:projectName}"/>
-    <g:set var="paginateJobs" value="${grailsApplication.config.rundeck.gui.paginatejobs}" />
+    <g:set var="paginateJobs" value="${grailsApplication.config.rundeck.gui.paginatejobs.enabled}" />
```
sjrd218: If this property is set to false in the rundeck-config.properties it seems to still show the pagination controls.

ltamaster:
I tested it, 1200 jobs in 9s more less (like 2 min without the fix)

rundeckapp/grails-app/views/menu/jobs.gsp
gschueler: ok i'll check that

gschueler: should be fixed now

gschueler:
ref #1351

PR #5813, opened 2020-02-21
adds ProjectComponent for archive import/export, refactor Webhooks data to use it

gschueler:

Adds component type for project archive import/export, updates webhooks project data import/export to use it

ProjectComponent type allows defining import/export aspects of project archive in a modular way.


ronaveva:
It only displays import options if the implementation has config properties.

imagen

Without that property the whole block would not show

rundeckapp/src/main/groovy/com/dtolabs/rundeck/app/support/ProjectArchiveParams.groovy
```diff
@@ -44,8 +42,10 @@ class ProjectArchiveParams implements ProjectArchiveImportRequest, Validateable{
     Boolean exportReadmes
     Boolean exportAcls
     Boolean exportScm
-    Boolean exportWebhooks
-    Boolean whkIncludeAuthTokens
+    Map<String, Boolean> importComponents
+    Map<String, Map<String, String>> importOpts
```
sjrd218: When trying to import a project with the api I'm not sure how to represent these values as query parameters. Since the project archive is obtained by reading the request input stream, there doesn't seem to be another way to post these values. Do you know how this should work?

gschueler: I believe you can use `?importOpts.component.option=value` and `?importComponents.component=true`  as URL parameters. But perhaps a nicer pattern can be put in place for API calls.

Also, for webhooks, we should honor the old query params used , i will make that change.

sjrd218: I was looking at the rundeck-cli to figure out how to pass the params. Do you think retrofit can handle that structure?

gschueler: Yes, I think the `@QueryMap` annotation can be used

gschueler: We may want to add an api that lists components and their inputs (similar to scm plugins api)

gschueler: added api v34 webhook params support

gschueler:
@sjrd218 thanks, investigating

gschueler:
@sjrd218 api export should work now

PR #5896, opened 2020-03-18
job query enhancement

ronaveva:
Enhancement: job query components can now show the search parameters in jobs list

**Additional context**
Currently: When an implementation of job query is used, the aditional parameters used by the implentation are not shown in the jobs list page.
Proposed: extendCriteria method can now return a map with the values to be shown in the jobs list page

imagen

imagen

### DIFF
--- a/rundeckapp/grails-app/views/menu/jobs.gsp
+++ b/rundeckapp/grails-app/views/menu/jobs.gsp
@@ -407,7 +407,6 @@ search
      <a href="#">
 
         <g:if test="${wasfiltered}">
-            ${filterName}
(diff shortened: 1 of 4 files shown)


rundeckapp/grails-app/views/menu/jobs.gsp
```diff
@@ -406,6 +406,7 @@ search
      <a href="#">
 
         <g:if test="${wasfiltered}">
+            ${filterName}
```
gschueler: accidental change?
```suggestion
```

gschueler:
@Mergifyio rebase

PR #5902, opened 2020-03-20
Verirfying list of listeners to avoid an exception thrown before handle result

### LINKED ISSUE #5901: checkpoint.time.X not working under Rundeck 3.2.X com.rundeck.rundeckpro.amazon-s3
I'm not able to see the "live log" between the cluster members; it can be seen when the job is finished.
Issue reported on: https://github.com/rundeckpro/rundeckpro/issues/886

carlosrfranco:
Fix #5901 and https://github.com/rundeckpro/rundeckpro/issues/886
the "live log" between the cluster members was not refreshed; it can be seen when the job is finished.

at some point, the list of listeners is empty and an exception is thrown and log processing is interrupted. This changes is to avoid the process to be interrupted.

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/LogFileStorageService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/LogFileStorageService.groovy
@@ -186,7 +186,7 @@ class LogFileStorageService
      * @return
      */
     def runRetrievalRequestTask(Map task){
-        def listener = logFileRetrievalListeners.remove(task.ruuid)
+        def listener = !logFileRetrievalListeners?.isEmpty() && logFileRetrievalListeners[task.ruuid] ? logFileRetrievalListeners.remove(task.ruuid) : null
         def future = runRetrievalRequest(task)
         future.handle(
                 { result, throwable ->
--- a/rundeckapp/grails-app/services/rundeck/services/LogFileStorageService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/LogFileStorageService.groovy
@@ -186,7 +186,7 @@ class LogFileStorageService
      * @return
      */
     def runRetrievalRequestTask(Map task){
-        def listener = !logFileRetrievalListeners?.isEmpty() && logFileRetrievalListeners[task.ruuid] ? logFileRetrievalListeners.remove(task.ruuid) : null
+        def listener = task.ruuid?logFileRetrievalListeners.remove(task.ruuid):null
         def future = runRetrievalRequest(task)
         future.handle(
                 { result, throwable ->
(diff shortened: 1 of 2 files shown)


rundeckapp/grails-app/services/rundeck/services/LogFileStorageService.groovy
```diff
@@ -186,7 +186,7 @@ class LogFileStorageService
      * @return
      */
     def runRetrievalRequestTask(Map task){
-        def listener = logFileRetrievalListeners.remove(task.ruuid)
+        def listener = !logFileRetrievalListeners?.isEmpty() && logFileRetrievalListeners[task.ruuid] ? logFileRetrievalListeners.remove(task.ruuid) : null
```
gschueler: based on the stacktrace, it's not the `logFileRetrievalListeners` which is null, it is the `task.ruuid`.  I think the problem is in the `requestLogFileRetrievalPartial` doesn't use a ruuid for a listener, so this method just needs to check

```suggestion
        def listener = task.ruuid?logFileRetrievalListeners.remove(task.ruuid):null
```

carlosrfranco: @gschueler when I debugged the whole process, I realized that the listener was empty and the "ruuid" is not null. The map content is something like this:
`[state:PENDING_LOCAL, file:/var/lib/rundeck/logs/rundeck/PROJECT_TEST/job/85a9de20-5226-4fb9-831c-2e91759a5f03/logs/30.state.json, filetype:state.json, ruuid:2c2d971b-7e24-4dc1-a458-5279e8854ff6, storage:com.rundeck.rundeckpro.plugins.S3LogFileStoragePlugin@4ea47792, id:30:state.json, name:com.rundeck.rundeckpro.amazon-s3, count:0]`

gschueler: ruuid is not null for a "full" storage request, but for a "partial" (checkpoint) storage request, it is null because there is no listener-based callback method (async) for partial storage requests. I think the addition of the async storage requests caused this bug by failing to check for ruuid. the listener may be null for full storage request if the caller didn't use the async mechanism

you should debug the partial storage retrieval request.  the stacktrace is pretty clear that the ruuid is null

carlosrfranco: I took a bad example. But, anyway, the listener is also empty on partial:
`task: [state:PENDING_LOCAL, file:/var/lib/rundeck/logs/rundeck/PROJECT_TEST/job/85a9de20-5226-4fb9-831c-2e91759a5f03/logs/30.state.json.part, filetype:state.json, storage:com.rundeck.rundeckpro.plugins.S3LogFileStoragePlugin@5f43b838, id:30:state.json:part, name:com.rundeck.rundeckpro.amazon-s3, count:0, date:Thu Mar 19 05:26:12 UTC 2020, partial:true]`
and
`logFileRetrievalListeners: [:]`

so my solution also worked. Should the listener be empty at this point? Or is it an issue?

PR #5912, opened 2020-03-25
Remove tabulation and line break from the content of text area

### LINKED ISSUE #5911: Problem with WebHook Notification URL when the values is > 100
**Describe the bug:**
Exported YAML contains blank spaces when WebHook Notification URL is longer than 100 characters (multi-line)

**My Rundeck details:**
RUNDECK ENTERPRISE 3.2.2-20200204
Google Chrome Version 80.0.3987.149 (Official Build) (64-bit)

**Steps to reproduce the behavior:**
    1. import attached job
    2. edit the job definition
    3. view on failure notification
    4. you will be able to see that spaces have been added to the start and end of the url
    5. save the job
    6. export yaml job definition
    7. exported yaml will contain the spaces in the url

**Screenshots**:
Original Job definition:
before

View from GUI: 
GUI

After saving/export:
exported

**Additional Content:**
WebHooks notifications work fine even when blank spaces are wrongly added to the URL.

jtobard:
fix #5911
This was caused by the line break and tabulation inside the `<textarea>` tag
This happens only on notifications that has a length of more than 100 characters.
In that case, the input type in the GUI changes to a text area and adds empty spaces and line break characters.  If the job is saved without removing the empty spaces is saved with the new characters every time.

Fix validated by @runwaldo see original issue: rundeckpro/rundeckpro#931

### DIFF
--- a/rundeckapp/grails-app/views/scheduledExecution/_editNotificationsTriggerForm.gsp
+++ b/rundeckapp/grails-app/views/scheduledExecution/_editNotificationsTriggerForm.gsp
@@ -127,9 +127,8 @@ implied. - See the License for the specific language governing permissions and -
 
                 <g:set var="notifurlcontent" value="${params[triggerUrlFieldName] ?: defUrl?.content}"/>
                 <g:if test="${notifurlcontent && notifurlcontent.size() > 100}">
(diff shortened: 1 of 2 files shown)


gschueler:
travis results were successful but not reported

PR #5919, opened 2020-03-28
Rd authorization libs

gschueler:

Refactor some authorization code out of core/ dir into separate libs:

migrate and refactor code into 3 java libraries:

* rundeck-authz-api: acl evaluation and interface types
* rundeck-authz-core: app level resource and action definitions
* rundeck-authz-yaml: acl policy format validation and parsing

**Additional context**

* end goal is to remove rd-acl (AclTool) and put it in the rundeck-cli project or an extension for rundeck-cli tool


rundeck-authz/rundeck-authz-core/src/main/java/org/rundeck/core/auth/AuthResources.java
```diff
@@ -0,0 +1,232 @@
+package org.rundeck.core.auth;
+
+import java.util.*;
+
+public class AuthResources {
```
sjrd218, line 5: It would be great if we could come up with a way to add type, kinds, actions etc... with a service loader or beans. That way plugins could add that info rather than having to hand edit this file every time we need something new. SCM and webhooks are prime candidates for that type of behavior. Don't know if we could work it into this refactor. What do you think?

gschueler: that would be a great enhancement.  i think for this phase i just want to deal with current usage

PR #5923, opened 2020-03-30
Fix job scheduling transaction errors

ProTip:
Bug fix. The use of `REQUIRES_NEW` to create executions has impacted existing code paths that were in the same unit of work. This has caused a number of scheduling issues to pop up related to transaction handling.

Jobs added to the Quartz scheduler through the existing code-path have their triggers placed into a `PENDING` trigger group, which is _paused_.

The `QuartzScheduleManager` subscribes to Grails GORM events in order to "activate" the job in the Quartz schedule after the transaction creating the `Execution` commits. This happens by rescheduling the job with a trigger set to the appropriate 

**Additional context**
First introduced in #4498

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/JobSchedulerService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/JobSchedulerService.groovy
@@ -6,6 +6,7 @@ import grails.events.annotation.Subscriber
 import groovy.transform.CompileStatic
 import groovy.util.logging.Log4j
 import org.grails.datastore.mapping.engine.event.PostInsertEvent
+import org.grails.datastore.mapping.engine.event.PostUpdateEvent
 import org.quartz.JobBuilder
 import org.quartz.JobDataMap
 import org.quartz.JobKey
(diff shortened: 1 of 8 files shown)


rundeckapp/grails-app/services/rundeck/services/JobSchedulerService.groovy
```diff
@@ -137,4 +214,80 @@ class QuartzJobScheduleManager implements JobScheduleManager {
...
+     * If an execution has been re-assigned to the local node we reschedule the PENDING job
+     */
+    @Subscriber
+    void afterUpdate(PostUpdateEvent event) {
+        if(!(event.entityObject instanceof Execution))
+            return
+
+        Execution execution = event.entityObject as Execution
+
+        if (isExecutionReassignedToLocal(execution)) {
```
gschueler, line 245: should this also check `isExecutionTypeUser` and `isExecutionStatusScheduled` just like the afterInsert method does?

gschueler: @ProTip @sjrd218

PR #5968, opened 2020-04-15
Fix #5081 log file storage request should be unique for execution

### LINKED ISSUE #5081: 500 error: More than one row with the given identifier was found: ..., for class: rundeck.LogFileStorageRequest
**Describe the bug**

500 error occurs on activity page, with message 

> More than one row with the given identifier was found: XXX, for class: rundeck.LogFileStorageRequest

(XXX is an ID number)

**My Rundeck detail**

* Rundeck version: 3.x

**To Reproduce**

Inconsistent, possibly aborting an execution right as it also finishes normally, or doing so from another cluster member.

**Expected behavior**

No error

**Additional context**

Execution declare `hasOne` for LogFileStorageRequest, however there is no  `unique` constraint applied to the associate.  Gorm 6.1 declares [`it's a good idea to add a unique constraint on one side of the one-to-one relationship`](http://gorm.grails.org/6.1.x/hibernate/manual/index.html#manyToOneAndOneToOne) but does not mention any consequence of not doing so.  The unique constraint should be applied, and the LogFileStorageService should handle failure to create the request if it already exists for an execution.

gschueler:

Fix #5081 

Add unique constraint for LogFileStorageService.execution

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/LogFileStorageService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/LogFileStorageService.groovy
@@ -636,6 +636,11 @@ class LogFileStorageService
             return
         }
         //multi storage available
+        //avoid multiple storage requests for the same Execution
+        def orig = LogFileStorageRequest.findByExecution(e)
+        if(orig){
+            return
+        }
         LogFileStorageRequest request = createStorageRequest(e, '*')
         request.discard()
         def reqid = request.execution.id.toString() + ":" + request.filetype
--- a/rundeckapp/grails-app/domain/rundeck/LogFileStorageRequest.groovy
+++ b/rundeckapp/grails-app/domain/rundeck/LogFileStorageRequest.groovy
@@ -28,7 +28,7 @@ class LogFileStorageRequest {
     static belongsTo = Execution
 
     static constraints = {
-        execution nullable: false
+        execution nullable: false, unique: true
         pluginName maxSize: 255
         filetype nullable: true
     }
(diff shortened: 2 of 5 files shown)


ProTip:
@gschueler I'm guessing we don't but do we ever do multi-part storage updates for partial logs or etc that would result in multiple storage requests for the same exec?

rundeckapp/grails-app/services/rundeck/services/LogFileStorageService.groovy
```diff
@@ -636,6 +636,11 @@ class LogFileStorageService
             return
         }
         //multi storage available
+        //avoid multiple storage requests for the same Execution
+        def orig = LogFileStorageRequest.findByExecution(e)
+        if(orig){
+            return
+        }
         LogFileStorageRequest request = createStorageRequest(e, '*')
```
ProTip, line 644: I believe the flush in `createStorageRequest` would raise an error if the unique index constraint were violated? Might be able to avoid an extra db call.

In either case I believe that because the transaction is started in `ExecutionEventsService` two could being going at the same time and not "see" each other, so the request could be queued and the commit would fail when `ExecutionEventsService.executionComplete` returns. I'm not sure if that's a problem per-say, just an observation.

gschueler:
@ProTip partial logs don't create a persistent storage request object, only the final logs. I don't think there will be any similar contention there

rundeckapp/grails-app/services/rundeck/services/LogFileStorageService.groovy
gschueler: yes, i added that findBy check earlier in my test, it's maybe not useful

gschueler: these are both kind of an experimental fix since we haven't reproduced the conditions that cause the eventual problem

gschueler:
pending docs update

PR #5969, opened 2020-04-15
Fix #2301 - Changing to keep empty option value (if no value is passed) even if the option has a default value

### LINKED ISSUE #2301: Empty option with default value
# Bug report

**My Rundeck detail**

* Rundeck version: 2.6.9-1
* install type: rpm
* OS Name/version: Red Hat Enterprise Linux Server release 6.8

**Expected Behavior**
A job has an "inline script" step with only one line: echo "value:@option.opt@". This job has an option with a default value. If on Prepare and Run tab the default value is deleted (before running the job), the option should appear as an empty string during the job run.

**Actual Behavior**
Before running the job I delete the default value on the Prepare and Run tab, but the option will use the default value, instead of an empty string.

**How to reproduce Behavior**
1. _Import_ the next yml file as a job:  
[default_test.txt](https://github.com/rundeck/rundeck/files/739200/default_test.txt) (you should modify the extension back to yml, txt is just because of the uploading.)
2. Click the job, leave the default value of the _opt_ option.
3. Run the job. It will print _value:defval_
4. Click on _Run again_
5. Remove the _defval_ string from _opt_ option.
6. Run the job again. It will pring _value:defval_ again, instead of the _value:_ .

carlosrfranco:
#2301 

The option value is replacing the empty value with default value instead of keeping the value as empty (if a empty value is passed)

The option params should be selected even if the value empty.


gschueler:
i will need to review this more for possible side-effects

carlosrfranco:
> i will need to review this more for possible side-effects

@gschueler ok but, just to mention, ExecutionService:2083 implements that behaviour in the same way as this change is... So, I think that is the expected way

https://github.com/rundeck/rundeck/blob/aef24d1d855ac91447e52ca0f85dd344d2cfade6/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy#L2083

gschueler (approved):

gschueler (approved):

PR #5970, opened 2020-04-15
Remove remaining SCM calls prior to job list on GUI

### LINKED ISSUE #5967: Remove scm calls prior to job list.
Load times on job page can be improved by removing the few SCM calls that still exists in the using the `_gui_min_scm` parameter and moving them to the existent Ajax call.

remove the usage of `scmExportActions` and `scmImportActions` from `jobsFragment` and use only the one on `listExport`. Moving those to the fragment of code that runs only if `minScm` is false, doesn't affect any other functionality (like API calls).

jtobard:
Enhancement to improve the load time of the job list page using SCM.

`scmExportActions` and `scmImportActions` are no longer used on` jobsFragment` method if is called from the GUI (parameter `_gui_min_scm=true`).

Now on the job page the actions section looks like this:
SharedScreenshot01

Until the ajax call of scm return with the info (that info was already on that call, we were just using the one from the method to load that menu as son as the page load):
SharedScreenshot02

As usual this only affect if the SCM plugin is enabled.

fix #5967

### DIFF
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
@@ -488,14 +488,12 @@ class MenuController extends ControllerBase implements ApplicationContextAware{
                                                                  ),
                                                                  [AuthConstants.ACTION_ADMIN, AuthConstants.ACTION_EXPORT, AuthConstants.ACTION_SCM_EXPORT]
             )) {
-                if(frameworkService.isClusterModeEnabled()){
+                if((!minScm) && frameworkService.isClusterModeEnabled()){
                     if (!scmService.projectHasConfiguredExportPlugin(params.project)) {
                         //initialize if in another node
                         scmService.initProject(params.project, 'export')
                     }
-                    if(minScm){
-                        scmService.fixExportStatus(authContext, params.project, results.nextScheduled)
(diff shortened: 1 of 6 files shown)


jtobard:
Tested by @drodriguez-rundeck

rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
```diff
@@ -3286,6 +3283,8 @@ class MenuController extends ControllerBase implements ApplicationContextAware{
                 if (frameworkService.isClusterModeEnabled()) {
                     //initialize if in another node
                     scmService.initProject(params.project, 'import')
+                    scmService.fixImportStatus(authContext, params.project, result.nextScheduled)
+                    scmService.importPluginStatus(authContext, params.project)
```
gschueler, line 3287: the result of this method is never used, is this a redundant call?

jtobard: The response is not needed just to check the status for the future method to check individual file, this must be refactored on next enhancement to SCMon cluster too.

gschueler: can you file an issue about any refactoring that should happen?

PR #5971, opened 2020-04-15
Use pessimistic locking on job stats update

### LINKED ISSUE #5963: Contention updating job statistics causing thread pile-up
**Describe the bug**
After a job execution completes certain bookkeeping actions take place against the database. Among these is updating the Job's execution statistics entry. Under certain circumstances many executions for the same job finishing around the same time can cause contention on the row being updated in the database.

The update utilizes hibernate optimistic concurrency controls along with a retry loop and a back-off. The initial back-off is quite high and can cause the workflow threads to pause for ~5-10s each retry. It has been observed that a "thundering herd" type of scenario can occur where the updates keep resulting in concurrency errors for a large number of the threads, and then threads and work accumulate increasing system resource usage.

In an installation with high usage and high execution-churn this can cause a "pile-up" of jobs/threads as the blocked threads start waiting instead of exiting.

**My Rundeck detail**

* Rundeck version: 3.2.5
* install type: n/a
* OS Name/version: n/a
* DB Type/version: MySQL 5.7

**To Reproduce**
`TODO`

**Expected behavior**
A few hundred executions finishing at the same time should not be contended for so long updating a database row. Parameterized jobs encourage reuse and running many dozens or hundreds of executions of the same job is natural scenario.

**Additional context**
This is impacting at least one Enterprise user.

There are a few solutions that may mitigate this issue:
**SELECT FOR UPDATE**
The code path could select the job statistic row `FOR UPDATE`. This may be an ideal option if we continue to update after every execution. The required lock acquisition in the database will "line up" the updates and they will complete based on the commit and network latency without stomping on each others toes.

**Reduce the back-off time**
The current default is quite high. While this may allow the "queue" to clear faster the optimistic concurrency controls will cause a lot of unnecessary retries to occur as each thread "races" to update the row.

**Update asynchronously**
We could update the job statistics asynchronously after the execution information has been saved. In this process we could "de-bounce" the updates to reduce the number of updates that are necessary.

**Write new statistic entries**
We could write new statistic entries to the database. This would eliminate contention on single rows. The individual entries could then be aggregated at read time, and "rolled up" occasionally to clean house. This is a somewhat common pattern for ingesting and processing data at a high frequency.

ProTip:
Fixes #5963

Disable optimistic concurrency control on the job stats domain object. This will allow a "last writer" wins scenario but should reduce/remove update contention. If we don't necessarily care about the update order this may be a viable solution.

Solutions outlined in #5963 

**Additional context**
https://gorm.grails.org/latest/hibernate/manual/#locking
https://gorm.grails.org/latest/hibernate/manual/#ormdsl

### DIFF
--- a/rundeckapp/grails-app/domain/rundeck/ScheduledExecution.groovy
+++ b/rundeckapp/grails-app/domain/rundeck/ScheduledExecution.groovy
@@ -1153,10 +1153,10 @@ class ScheduledExecution extends ExecutionContext implements EmbeddedJsonData {
         }
     }
 
-    ScheduledExecutionStats getStats() {
+    ScheduledExecutionStats getStats(lock = false) {
         def stats
         if(this.id) {
-            stats = ScheduledExecutionStats.findBySe(this)
+            stats = ScheduledExecutionStats.findBySe(this, [lock: lock])
             if (!stats) {
                 def content = [execCount   : this.execCount,
                                totalTime   : this.totalTime,
--- a/rundeckapp/grails-app/domain/rundeck/ScheduledExecutionStats.groovy
+++ b/rundeckapp/grails-app/domain/rundeck/ScheduledExecutionStats.groovy
@@ -8,10 +8,14 @@ import org.hibernate.annotations.OptimisticLocking
 class ScheduledExecutionStats {
     String content
 
+    long _version = 0
+
     static belongsTo=[se:ScheduledExecution]
     static transients = ['contentMap']
 
     static mapping = {
+        version false
+        _version column: 'version'
         content type: 'text'
     }
 
--- a/rundeckapp/grails-app/domain/rundeck/ScheduledExecutionStats.groovy
+++ b/rundeckapp/grails-app/domain/rundeck/ScheduledExecutionStats.groovy
@@ -11,6 +11,7 @@ class ScheduledExecutionStats {
     static transients = ['contentMap']
 
     static mapping = {
+        version false
         content type: 'text'
     }
 
--- a/rundeckapp/grails-app/domain/rundeck/ScheduledExecutionStats.groovy
+++ b/rundeckapp/grails-app/domain/rundeck/ScheduledExecutionStats.groovy
@@ -2,16 +2,16 @@ package rundeck
 
 import com.fasterxml.jackson.core.JsonParseException
 import com.fasterxml.jackson.databind.ObjectMapper
+import org.hibernate.annotations.OptimisticLockType
+import org.hibernate.annotations.OptimisticLocking
 
 class ScheduledExecutionStats {
-
     String content
 
     static belongsTo=[se:ScheduledExecution]
     static transients = ['contentMap']
 
     static mapping = {
-        version false
         content type: 'text'
     }
 
--- a/rundeckapp/grails-app/domain/rundeck/ScheduledExecutionStats.groovy
+++ b/rundeckapp/grails-app/domain/rundeck/ScheduledExecutionStats.groovy
@@ -2,8 +2,6 @@ package rundeck
 
 import com.fasterxml.jackson.core.JsonParseException
 import com.fasterxml.jackson.databind.ObjectMapper
-import org.hibernate.annotations.OptimisticLockType
-import org.hibernate.annotations.OptimisticLocking
 
 class ScheduledExecutionStats {
     String content
--- a/rundeckapp/grails-app/jobs/rundeck/quartzjobs/ExecutionJob.groovy
+++ b/rundeckapp/grails-app/jobs/rundeck/quartzjobs/ExecutionJob.groovy
@@ -46,7 +46,7 @@ import java.util.function.Consumer
 class ExecutionJob implements InterruptableJob {
 
     public static final int DEFAULT_STATS_RETRY_MAX = 5
-    public static final long DEFAULT_STATS_RETRY_DELAY = 5000
+    public static final long DEFAULT_STATS_RETRY_DELAY = 1000
     public static final int DEFAULT_FINALIZE_RETRY_MAX = 10
     public static final long DEFAULT_FINALIZE_RETRY_DELAY = 5000
 
--- a/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
@@ -2983,6 +2983,7 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
      * @param execution
      * @return
      */
+    @NotTransactional
     def updateScheduledExecStatistics(Long schedId, eId, long time) {
         def success = false
         try {
@@ -3022,8 +3023,6 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
                     }
                     success = true
                 }
-
-
             }
         } catch (org.springframework.dao.ConcurrencyFailureException e) {
             log.warn("Caught ConcurrencyFailureException, will retry updateScheduledExecStatistics for ${eId}")
--- a/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
@@ -2988,8 +2988,7 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
         try {
             ScheduledExecutionStats.withTransaction {
                 def scheduledExecution = ScheduledExecution.get(schedId)
-                def seStats = scheduledExecution.getStats()
-
+                def seStats = scheduledExecution.getStats(true)
 
                 if (scheduledExecution.scheduled) {
                     scheduledExecution.nextExecution = scheduledExecutionService.nextExecutionTime(scheduledExecution)


ProTip:
This would result in inaccurate execution count numbers without utilizing something like `executeUpdate` to do a `+1` on it. We could do that or `lock()` if that's a concern..

ProTip:
> This would result in inaccurate execution count numbers without utilizing something like `executeUpdate` to do a `+1` on it. We could do that or `lock()` if that's a concern..

Ah, we are storing them in a blob so we can't really do those sort of DB-side updates. I pushed a change for what locking when we update could look like.

gschueler:
the execution count is only used for the average duration calculation, so wouldn't matter if it misses an execution

gschueler:
I got this error creating a job.  we can't just remove the version from the domain class, existing schema for the table will have the version field as "not null"

```
Caused by: org.springframework.dao.DataIntegrityViolationException: Hibernate operation: could not execute statement; SQL [n/a]; NULL not allowed for column "VERSION"; SQL statement:
insert into scheduled_execution_stats (id, se_id, content) values (null, ?, ?) [23502-199]; nested exception is org.h2.jdbc.JdbcSQLIntegrityConstraintViolationException: NULL not allowed for column "VERSION"; SQL statement:
insert into scheduled_execution_stats (id, se_id, content) values (null, ?, ?) [23502-199]
		at org.springframework.jdbc.support.SQLErrorCodeSQLExceptionTranslator.doTranslate(SQLErrorCodeSQLExceptionTranslator.java:242)
		at org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:73)
		at org.grails.orm.hibernate.GrailsHibernateTemplate.convertJdbcAccessException(GrailsHibernateTemplate.java:731)
		at org.grails.orm.hibernate.GrailsHibernateTemplate.convertHibernateAccessException(GrailsHibernateTemplate.java:719)
		at org.grails.orm.hibernate.GrailsHibernateTemplate.doExecute(GrailsHibernateTemplate.java:303)
		at org.grails.orm.hibernate.GrailsHibernateTemplate.execute(GrailsHibernateTemplate.java:243)
		at org.grails.orm.hibernate.GrailsHibernateTemplate.execute(GrailsHibernateTemplate.java:117)
		at org.grails.orm.hibernate.AbstractHibernateGormInstanceApi.performSave(AbstractHibernateGormInstanceApi.groovy:251)
		at org.grails.orm.hibernate.AbstractHibernateGormInstanceApi.save(AbstractHibernateGormInstanceApi.groovy:168)
		at org.grails.datastore.gorm.GormInstanceApi.save(GormInstanceApi.groovy:119)
		at org.grails.datastore.gorm.GormEntity$Trait$Helper.save(GormEntity.groovy:98)
		at rundeck.ScheduledExecution.getStats(ScheduledExecution.groovy:1165)
		at rundeck.ScheduledExecution.getRefExecCountStats(ScheduledExecution.groovy:1172)
		at rundeck.controllers.ScheduledExecutionController.show(ScheduledExecutionController.groovy:461)
		... 75 common frames omitted
	Caused by: org.h2.jdbc.JdbcSQLIntegrityConstraintViolationException: NULL not allowed for column "VERSION"; SQL statement:
insert into scheduled_execution_stats (id, se_id, content) values (null, ?, ?) [23502-199]
		at org.h2.message.DbException.getJdbcSQLException(DbException.java:457)
		at org.h2.message.DbException.getJdbcSQLException(DbException.java:427)
		at org.h2.message.DbException.get(DbException.java:205)
		at org.h2.message.DbException.get(DbException.java:181)
		at org.h2.table.Column.validateConvertUpdateSequence(Column.java:418)
		at org.h2.table.Table.validateConvertUpdateSequence(Table.java:824)
		at org.h2.command.dml.Insert.insertRows(Insert.java:175)
		at org.h2.command.dml.Insert.update(Insert.java:132)
		at org.h2.command.CommandContainer.update(CommandContainer.java:133)
		at org.h2.command.Command.executeUpdate(Command.java:267)
		at org.h2.jdbc.JdbcPreparedStatement.executeUpdateInternal(JdbcPreparedStatement.java:200)
		at org.h2.jdbc.JdbcPreparedStatement.executeUpdate(JdbcPreparedStatement.java:154)
		at org.hibernate.engine.jdbc.internal.ResultSetReturnImpl.executeUpdate(ResultSetReturnImpl.java:204)
		at org.hibernate.dialect.identity.GetGeneratedKeysDelegate.executeAndExtract(GetGeneratedKeysDelegate.java:57)
		at org.hibernate.id.insert.AbstractReturningDelegate.performInsert(AbstractReturningDelegate.java:42)
		at org.hibernate.persister.entity.AbstractEntityPersister.insert(AbstractEntityPersister.java:2919)
		at org.hibernate.persister.entity.AbstractEntityPersister.insert(AbstractEntityPersister.java:3490)
		at org.hibernate.action.internal.EntityIdentityInsertAction.execute(EntityIdentityInsertAction.java:81)
		at org.hibernate.engine.spi.ActionQueue.execute(ActionQueue.java:626)
		at org.hibernate.engine.spi.ActionQueue.addResolvedEntityInsertAction(ActionQueue.java:280)
		at org.hibernate.engine.spi.ActionQueue.addInsertAction(ActionQueue.java:261)
		at org.hibernate.engine.spi.ActionQueue.addAction(ActionQueue.java:306)
		at org.hibernate.event.internal.AbstractSaveEventListener.addInsertAction(AbstractSaveEventListener.java:318)
		at org.hibernate.event.internal.AbstractSaveEventListener.performSaveOrReplicate(AbstractSaveEventListener.java:275)
		at org.hibernate.event.internal.AbstractSaveEventListener.performSave(AbstractSaveEventListener.java:182)
		at org.hibernate.event.internal.AbstractSaveEventListener.saveWithGeneratedId(AbstractSaveEventListener.java:113)
		at org.hibernate.event.internal.DefaultSaveOrUpdateEventListener.saveWithGeneratedOrRequestedId(DefaultSaveOrUpdateEventListener.java:192)
		at org.hibernate.event.internal.DefaultSaveOrUpdateEventListener.entityIsTransient(DefaultSaveOrUpdateEventListener.java:177)
		at org.hibernate.event.internal.DefaultSaveOrUpdateEventListener.performSaveOrUpdate(DefaultSaveOrUpdateEventListener.java:97)
		at org.hibernate.event.internal.DefaultSaveOrUpdateEventListener.onSaveOrUpdate(DefaultSaveOrUpdateEventListener.java:73)
		at org.grails.orm.hibernate.support.ClosureEventTriggeringInterceptor.onSaveOrUpdate(ClosureEventTriggeringInterceptor.java:130)
		at org.hibernate.internal.SessionImpl.fireSaveOrUpdate(SessionImpl.java:651)
		at org.hibernate.internal.SessionImpl.saveOrUpdate(SessionImpl.java:643)
		at org.hibernate.internal.SessionImpl.saveOrUpdate(SessionImpl.java:638)
		at org.grails.orm.hibernate.AbstractHibernateGormInstanceApi$_performSave_closure3.doCall(AbstractHibernateGormInstanceApi.groovy:252)
		at org.grails.orm.hibernate.GrailsHibernateTemplate.doExecute(GrailsHibernateTemplate.java:299)
		... 84 common frames omitted
```

ProTip:
I'll take a look. I would have though it would drop the column. Looks like it's just not supplying a value anymore.

ProTip:
I'm guessing it may not drop columns as a protective measure. Supplying a default may do the trick.

ahormazabal:
I'm not sure if enabling pessimistic locking will improve the contention situation. Probably will just move the contention to the database.

I'm not very familiar on how job stats works but looking at the code my opinion is that the most effective solution we have on hand is to enqueue "stats update requests" and process them asynchronously with one thread. This way you eliminate all contention on the table, and process the updates queue with just one thread ensuring that you will not lose data. The drawbacks i see with this are:
- All places where stats are updates would have to be converted
- On peak periods, updates would pile up so would be temporarily outdated while the updater catches up.

On another note, the code from ScheduledExecution.getStats() is not thread safe when multiple threads call it and the first stats object has not been saved to db. If is the case that the method is called by multiple threads a safer version would be:

```
    ScheduledExecutionStats getStats(lock = false) {
        def stats
        if (this.id) {
            stats = ScheduledExecutionStats.findBySe(this, [lock: lock])
            if (!stats) {
                synchronized (this) {
                    stats = ScheduledExecutionStats.findBySe(this, [lock: lock])
                    if (!stats) {

                        def content = [execCount   : this.execCount,
                                       totalTime   : this.totalTime,
                                       refExecCount: this.refExecCount]

                        stats = new ScheduledExecutionStats(se: this, contentMap: content).save()
                    }
                }
            }
        }
        stats
    }

```

gschueler:
@ahormazabal not sure that's any safer, the ScheduledExecution object will not be the same object if two threads did a ScheduledExecution.get() nor if they are on different cluster members.  Also it appears the ScheduledExecutionStats class doesn't have a `unique: true` constraint on the ScheduledExecution, so it's perfectly possible to create two ScheduledExecutionStats objects for the same ScheduledExecution.  ScheduledExecution should declare `hasOne` and ScheduledExecutionStats should add a constraint

ProTip:
> @ahormazabal so it's perfectly possible to create two ScheduledExecutionStats objects for the same ScheduledExecution. ScheduledExecution should declare `hasOne` and ScheduledExecutionStats should add a constraint

To your point @gschueler there is a place in the code, that I can't locate ATM, where it checks to see if multiple stats are returned and deletes them :)

> I'm not sure if enabling pessimistic locking will improve the contention situation. Probably will just move the contention to the database.

I probably shouldn't have spoken so loosely.. The contention for the row is already in the database, it's just in the form of a `WHERE` clause on the version field :) As long as we want read/write consistency with efficient updates we would need to find a way to order the atomic read/write units? A lock queue sorts that out either in the app or in the DB(which `SELECT .. FOR UPDATE` essentially provides). Even if it's in the app it's still probably appropriate to select the rows for update in case multiple servers processing updates against the same row unless a lock-free strategy is adopted.

I agree that async processing in a queue would free up jobs to exit in the case there is so much update demand it needs to be spread over time. I believe all the book-keeping would need to be offloaded otherwise it will just end up on another row? For example, this code also updates the next scheduled time on the scheduled execution. I haven't modified it, but it's likely to cause the same type of problem for jobs that happen to be scheduled.

Given a constraint of read/write consistency the optimistic concurrency strategy is terribly inefficient when a row is being heavily contended for. I believe the latest update I push optimizes this via the pessimistic locking; the work units all queue up nicely and don't smash their heads together. The transaction is tightly confined to this method, so hopefully impact of the extra locking will be contained..

gschueler:
> For example, this code also updates the next scheduled time on the scheduled execution.

This should be removed, IMO. I think it used to be used as a way to sort jobs by next execution time, but is no longer used.

ProTip:
@gschueler For posterity:

Looks like next scheduled is part of the public API as of ver `18`.

I did some load testing trying to create concurrency failures(deadlocks or optimistic concurrency) and it looks like at least under a heavy adhoc(API) load because the sched_ed does not become dirty it is not updated and their is no concurrency conflict.  A heavy scheduled load or mix might be a different story.

PR #5989, opened 2020-04-20
Plugin resource loader: remove unnecessary WARN log

gschueler:

Remove unnecessary warning log output in plugin loader checking

### DIFF
--- a/rundeckapp/src/main/groovy/com/dtolabs/rundeck/server/plugins/RundeckPluginRegistry.groovy
+++ b/rundeckapp/src/main/groovy/com/dtolabs/rundeck/server/plugins/RundeckPluginRegistry.groovy
@@ -551,7 +551,6 @@ class RundeckPluginRegistry implements ApplicationContextAware, PluginRegistry, 
                 if (bean instanceof PluginResourceLoader) {
                     return bean
                 }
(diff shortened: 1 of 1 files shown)


rundeckapp/src/main/groovy/com/dtolabs/rundeck/server/plugins/RundeckPluginRegistry.groovy
```diff
@@ -551,7 +551,6 @@ class RundeckPluginRegistry implements ApplicationContextAware, PluginRegistry,
                 if (bean instanceof PluginResourceLoader) {
                     return bean
                 }
-                log.warn("No resource loader for bean plugin: ${provider}. Provided bean: ${beanName}")
```
ahormazabal, line 554: Maybe we could change it to debug or trace instead of just deleting it? I'm not a fan of deleting log messages.

gschueler: it could be debug but i don't think it's very useful. most bean plugins don't implement that interface, so it's going to be just a lot of noise.

ahormazabal: yea i think you're right.

PR #5990, opened 2020-04-21
Fix #5938 error when logging in with empty credentials.

### LINKED ISSUE #5938: If entering empty credentials rundeck shows a blank screen
**Describe the bug**
If you enter empty credentials Rundeck shows a blank screen, that happens if you launch Rundeck directly with `java -jar rundeck-3.2.4-20200318.war`, the launch needs `-Drundeck.jaaslogin=true` flag to work correctly.

If you launch rundeck 3.1.6 directly with `java -jar rundeck-3.1.6-20200210` works well.

**My Rundeck detail**

* Rundeck version: 3.2.0, 3.2.1, 3.2.2, 3.2.3, 3.2.4 and 3.2.5-SNAPSHOT.
* Install type: WAR
* OS Name/version: Centos 7
* DB Type/version: H2 

**To Reproduce**

1. Start Rundeck directly using `java -jar rundeck-3.2.4-20200318.war`
2. Click on "Login Button" without any credentials.
3. See error.

**Expected behavior**
"Invalid username and password" red warning must appear.

**Screenshots**

With `java -jar rundeck-3.2.4-20200318.war`

Screenshot from 2020-04-03 16-54-22

With `java -Drundeck.jaaslogin=true -jar rundeck-3.2.4-20200318.war`:

Screenshot from 2020-04-03 17-13-09

 - OS: Ubuntu 19.10
 - Browser: Firefox 
 - Version: 74.0

**Aditional context**
No errors on service.log, nor GUI.

sjrd218:
Disallows submitting the login form with an empty username.
Login toggles a spinner to replace the login button so multiple form submissions cannot be attempted.

### DIFF
--- a/rundeckapp/grails-app/views/user/login.gsp
+++ b/rundeckapp/grails-app/views/user/login.gsp
@@ -92,7 +92,7 @@
               </div>
             </g:if>
             <div class="col-md-4 col-sm-6 col-md-offset-4 col-sm-offset-3">
-              <form action="${g.createLink(uri:"/j_security_check")}" method="post" class="form " role="form">
+              <form action="${g.createLink(uri:"/j_security_check")}" method="post" class="form " role="form" onsubmit="return onLoginClicked()">
                 <div class="card" data-background="color" data-color="blue">
                   <div class="card-header">
                     <h3 class="card-title">
@@ -154,6 +154,7 @@
                     </div>
                         <div class="card-footer text-center">
                             <button type="submit" id="btn-login" class="btn btn-fill btn-wd "><g:message code="user.login.login.button"/></button>
+                            <span id="login-spinner" style="display: none;"><i class="fas fa-spinner fa-pulse"></i></span>
                         </div>
                     </g:showLocalLogin>
                   </div>
@@ -190,6 +191,16 @@
 
       </div>
     </div>
+      <script type="text/javascript">
+          function onLoginClicked() {
+            let lbtn = jQuery("#btn-login")
+            if(jQuery('#login').val() === '') return false
+            let spinner = jQuery("#login-spinner")
+            lbtn.hide()
+            spinner.show()
+            return true
+          }
+      </script>
     <g:render template="/common/footer"/>
   </div>
 </div>
--- a/rundeckapp/grails-app/views/user/login.gsp
+++ b/rundeckapp/grails-app/views/user/login.gsp
@@ -164,6 +164,9 @@
                           <span><g:enc>${flash.loginerror}</g:enc></span>
                       </div>
                     </g:if>
+                  <div class="alert alert-danger" style="display:none;" id="empty-username-msg">
+                      <span><g:message code="user.login.empty.username"/></span>
+                  </div>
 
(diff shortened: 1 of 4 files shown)


rundeckapp/grails-app/views/user/login.gsp
```diff
@@ -190,6 +191,16 @@
 
       </div>
     </div>
+      <script type="text/javascript">
+          function onLoginClicked() {
+            let lbtn = jQuery("#btn-login")
+            if(jQuery('#login').val() === '') return false
```
ahormazabal, line 197: This will give no feedback if the user press the Log In button without entering a username. I know is a very border case but i think that maybe this case should be handled just like a bad password or no pasword situation (give the "Invalid username and password."message).
I'm not even sure if its necessary to block the submit given the fix below.

rundeckapp/grails-app/views/user/login.gsp
```diff
@@ -190,6 +191,16 @@
...
+            let spinner = jQuery("#login-spinner")
+            lbtn.hide()
```
ahormazabal, line 199: Wouln't be easier to just disable the button instead of hiding it and placing another element on top?
jQuery("#btn-login").prop( "disabled", true );

rundeckapp/src/main/groovy/rundeck/services/audit/AuditEventsService.groovy
```diff
@@ -280,11 +281,12 @@ class AuditEventsService
     /**
      * Extract the username from an authentication object.
      */
-    private static String extractUsername(Authentication authentication) {
-        if (!authentication) {
+    @PackageScope
+    static String extractUsername(Authentication authentication) {
+        if (!authentication || authentication.principal == "") {
```
ahormazabal, line 286: I don't remember the exact reason, but i remebmer there was a reason for which i try to evaluate first authentication.name before authentication.principal.name, which has to do with the principal not always available or usable. This addition to the if would return null if principal is null even if we have auth.name available.

rundeckapp/src/main/groovy/rundeck/services/audit/AuditEventsService.groovy
```diff
@@ -280,11 +281,12 @@ class AuditEventsService
...
             return null
         }
-        return authentication.name ?: authentication.principal.name ?: null
+        return authentication.name ?: authentication.principal?.name ?: null
```
ahormazabal, line 289: This fix should be enough and cover the same case of the above line.

sjrd218:
> Playing with it, the spinner looks quite cool white waiting for login.
> 
> I found a small bug. If i supply a wrong password (rendering the "Invalid username and password.") and then supply an empty username, i get two error messages:
> Screen Shot 2020-04-21 at 16 04 51
> 
> I also tested the password auth using curl and could not replicate the original problem, so that part should be ok too.

I'm not sure I agree that the screenshot shows a bug. The user is attempting to submit the form with an empty username so they get a message telling them that. The other message is the normal static message a user receives when they have not entered correct credentials. That one doesn't disappear until the user logs in successfully. The new message is merely additive.

ahormazabal:
Fair enough.

PR #6007, opened 2020-04-29
Fix #6001 - useName property of referenced jobs should be true when it is null and there is name but there is not uuid on map

### LINKED ISSUE #6001: Jobs migrated from Rundeck2 to Rundeck 3 have "search by UUID" option set by default.
**Describe the bug**
When a referenced Jobs using the name but not UUID in Parent Job Rundeck version 2.X (specifically 2.4.11) when the job is migrated/imported into a Rundeck version 3.2.5 on the **Select job by** section its appears selected by UUID and it's not available.

**My Rundeck detail**

* Rundeck version: 3.2.5
* install type: rpm, war
* OS Name/version:  ubuntu 18.04
* DB Type/version: MySQL - H2

**Expected behavior**
The Jobs remains referenced by Name.

**Screenshots**
image_004

image_005

image_006

image_008

image_007

image_009

image_010

**Additional context**
Reviewing the yaml file with the job definition for the Rundeck version 3.x uses a **useName: 'true'** argument when a child job is referenced by Name that is not used by Rundeck 2.x

carlosrfranco:
fix #6001 

When a job with referenced jobs is exported on Rundeck 2.x and imported on Rundeck 3.2.x, the option "Search job by" have the value 'uuid' marked instead of 'name'. The file exported have only the name of referenced jobs

If there is a name of referenced job and there is not uuid, the property useName should  be true even the useName property is null

### DIFF
--- a/rundeckapp/src/test/groovy/rundeck/JobExecSpec.groovy
+++ b/rundeckapp/src/test/groovy/rundeck/JobExecSpec.groovy
@@ -136,7 +136,7 @@ class JobExecSpec extends Specification {
 
     }
 
-    def "from map use name without nodeName property"() {
+    def "from map use name without useName property"() {
         given:
         def map = [
                 jobref     : [
--- a/rundeckapp/src/test/groovy/rundeck/JobExecSpec.groovy
+++ b/rundeckapp/src/test/groovy/rundeck/JobExecSpec.groovy
@@ -135,6 +135,32 @@ class JobExecSpec extends Specification {
         false         | _
 
     }
+
+    def "from map use name without nodeName property"() {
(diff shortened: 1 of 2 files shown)


rundeckapp/src/test/groovy/rundeck/JobExecSpec.groovy
```diff
@@ -134,6 +134,32 @@ class JobExecSpec extends Specification {
         true          | _
         false         | _
 
+    }
+
+    def "from map use name without nodeName property"() {
```
jtobard, line 139: small typo is not `nodeName` is `useName`

PR #6010, opened 2020-04-29
Fixing the job import process where the ruleset strategy was not saved correctly

carlosrfranco:
fix https://github.com/rundeckpro/rundeckpro/issues/973

When importing a job definition, the ruleset description was lost

When handling the data map of the strategy plugin an error caused the map to become inconsistent and thus it was not possible to interpret it when loading the job.

### DIFF
--- a/rundeckapp/grails-app/domain/rundeck/Workflow.groovy
+++ b/rundeckapp/grails-app/domain/rundeck/Workflow.groovy
@@ -1,6 +1,7 @@
 package rundeck
 
 import com.dtolabs.rundeck.app.support.DomainIndexHelper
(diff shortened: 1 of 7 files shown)


fdevans:
Have we done comprehensive testing on this to make sure it doesn't impact working rulesets, and fixes broken ones?

carlosrfranco:
> Have we done comprehensive testing on this to make sure it doesn't impact working rulesets, and fixes broken ones?

I'm sending to Jonathan (support engineer) so he can validate with more cases.

jtobard:
This was tested by the support engineer that detect it @G3NSVRV, something else we need to do to merge this before the code freeze 0f 3.2.7?

PR #6051, opened 2020-05-13
Grails 4 Upgrade

gschueler:
Upgrade Rundeck to Grails 4, Gradle 5 🎊 

Many thanks to @niravassar for the bulk of the upgrade work!

Ref #3413


gschueler:
note: the travis failure is due to old Tomcat 7 api tests in master travis config, can ignore

PR #6085, opened 2020-05-27
Fix the Japanese translation of "retry"

mnogu:
This patch fixes a typo in the Japanese translation The correct translation of "retry" is not "ライ" (\u30e9\u30a4) but "リトライ" (\u30ea\u30c8\u30e9\u30a4).


fdevans:
Thank you @mnogu

PR #6091, opened 2020-05-29
Avoiding loading all plugins whenever a method needs to search for a plugin description

carlosrfranco:
Fix https://github.com/rundeckpro/rundeckpro/issues/924

The first time the Job edit screen is loaded, it takes a long time to load. In some environments, it took more than 1 minute to open

These changes allow you to find a description of the plugin without having to load all plugins of a given type. Before, all plugins were loaded for each search for a description

**Additional context**
I couldn't find the reason why this case is more critical in Windows environments


pwonson:
This fix reduces the time to 10 seconds is there another issue open for this?

PR #6100, opened 2020-06-04
Fix issue/1019 where running jobs is aways zero even if there are jobs running.

carlosrfranco:
Fix https://github.com/rundeckpro/rundeckpro/issues/1019

Some changes affected the query that returns info to Dashboard page where the running jobs counter was showing zero jobs even if there are jobs running and, in addition, the schedules that was created by run job later option is not showing too.

The criteria to query jobs was fixed to get jobs running even the status of the execution is null and fixed `JobQueryService` to consider jobs executed with `run job later` option as scheduled job.

**Additional context**
If the addon enterprise-schedules is enabled the jobs scheduled by `run job later` will not be shown on dashboard: https://github.com/rundeckpro/rundeckpro/issues/1036

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
@@ -428,7 +428,10 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
                             isNull('dateCompleted')
                             if(!query.considerPostponedRunsAsRunningFilter){
                                 le('dateStarted', now)
-                                ne('status', EXECUTION_SCHEDULED)
+                                or {
+                                    ne('status', EXECUTION_SCHEDULED)
+                                    isNull('status')
+                                }
                             }
                         }
                     } else {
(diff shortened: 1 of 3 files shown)


rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -428,7 +428,10 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
                             isNull('dateCompleted')
                             if(!query.considerPostponedRunsAsRunningFilter){
                                 le('dateStarted', now)
-                                ne('status', EXECUTION_SCHEDULED)
+                                or {
+                                    ne('status', EXECUTION_SCHEDULED)
```
gschueler, line 432: should this be changed to `eq`?
```suggestion
                                    eq('status', EXECUTION_SCHEDULED)
```

The original query (status!='scheduled') meant to show only currently running jobs, but missed the case where a scheduled execution had started.  
the new query should include (status==null) which is normal executions, as well as (status=='scheduled') which is scheduled executions that have started.

carlosrfranco: @gschueler When jobs is scheduled with `Run job later` option, the status is not defined as 'scheduled'. Previously, the query was only checking if the date started was before then 'now' and status different of 'scheduled' to avoid that `Run job later` was considered as running job. Now I just include the `isNull` clause on query.

gschueler: yes, status is set to `scheduled` when you use run job later.

later if the execution starts, it seems it is still status=scheduled.  so you can have a running execution with status 'scheduled'

PR #6103, opened 2020-06-04
Docker - Optional bindPassword variable for jass-loginmodule

ltamaster:
optional bindPassword environment variable for jass-loginmodule.

If you try to use the enterprise password encryption this value must be empty.
https://docs.rundeck.com/docs/administration/configuration/encryptable-properties.html#rundeck-enterprise-config-property-encryption

Enable the attribute  if the variable is set

**Additional context**

### DIFF
--- a/docker/official/remco/templates/jaas-loginmodule.conf
+++ b/docker/official/remco/templates/jaas-loginmodule.conf
@@ -6,7 +6,7 @@
         contextFactory="{{ getv("/rundeck/jaas/ldap/contextfactory", "com.sun.jndi.ldap.LdapCtxFactory") }}"
         providerUrl="{{ getv("/rundeck/jaas/ldap/providerurl") }}"
         bindDn="{{ getv("/rundeck/jaas/ldap/binddn") }}"
-        {% if exists("/rundeck/jaas/ldap/bindpassword") -%}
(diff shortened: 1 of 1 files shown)


docker/official/remco/templates/jaas-loginmodule.conf
```diff
@@ -6,7 +6,9 @@
         contextFactory="{{ getv("/rundeck/jaas/ldap/contextfactory", "com.sun.jndi.ldap.LdapCtxFactory") }}"
         providerUrl="{{ getv("/rundeck/jaas/ldap/providerurl") }}"
         bindDn="{{ getv("/rundeck/jaas/ldap/binddn") }}"
+        {% if exists("/rundeck/jaas/ldap/bindpassword") -%}
```
ProTip, line 9:
```suggestion
    {% if exists("/rundeck/jaas/ldap/bindpassword") -%}
```

docker/official/remco/templates/jaas-loginmodule.conf
```diff
@@ -6,7 +6,9 @@
...
         bindPassword="{{ getv("/rundeck/jaas/ldap/bindpassword") }}"
+        {% endif %}
```
ProTip, line 11:
```suggestion
    {% endif %}
```

PR #6108, opened 2020-06-05
Accept Environment variables optional for jsch executor

### LINKED ISSUE #4137: Failed: SSHProtocolFailure: channel is not opened .... JSCH default timeout
**Describe the bug**
When connecting to a node using JSCH, connection timeouts with the following error: `Failed: SSHProtocolFailure: channel is not opened.` This does not correlate to parameters set up in `sshd_config` and/or `framework.properties` .

**My Rundeck detail**

* Rundeck version: tested  2.11.7 ( related code is no different in 3.0.x )
* install type: 
* OS Name/version: 

**To Reproduce**

1. Direct SHH to remote node  (takes around 15-20 sec to complete successfully due to network/protocols/etc ). Password auth used.

2. In Rundeck , run a simple command against the same node ( configured with JSCH as node executor) . Connection takes takes the same time and finally timeouts with error: `Failed: SSHProtocolFailure: channel is not opened.` ( password auth  )

3. Change Node executor to, e.g., OpenSSH, run same command against same node. Connection takes the same time but succeeds.

**Expected behavior**
Timeout should honor values configured in `sshd_config` and/or `framework.properties` .

**Additional context**
The defaut timeout for SSH Channel opening seems to be low and cannot be changed

ref:
[https://github.com/rundeck/rundeck/blob/48b0c71381225096ac67943c1d81a94dfe3436d1/core/src/main/java/com/dtolabs/rundeck/core/tasks/net/ExtSSHExec.java#L459](url)

jtobard:
fix #4137

Jsch library has a problem with infinite timeout on channel connection, it has a hardcoded retry of 20000ms.

The solution is to use a number different to 0 on the timeout, but that has the collateral impact to throw an error previously ignored from Jsch: If the remote server is not configured to accept environment variables throws an exception.

The solution proposed is to make Environment variables optional, so you enable only if you are sure you already configured that on the server:
https://docs.rundeck.com/docs/administration/projects/node-execution/ssh.html#passing-environment-variables-through-remote-command

Another collateral effect of this change is connections with high latency that doesn't pass the variable environment has a better time on the execution of the step since it doesn't waste time passing multiple environment variables and waiting for the response.

The current configuration of jsch executor will have accept-env-var on false and need to be enabled manually
 
image

Solution tested in an environment with very high latency by @g3nsvrv and @cwaltherf

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/execution/impl/jsch/JschNodeExecutor.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/execution/impl/jsch/JschNodeExecutor.java
@@ -165,6 +165,10 @@ public class JschNodeExecutor implements NodeExecutor, Describable {
     public static final String FWK_SSH_CONFIG_PREFIX = FWK_PROP_PREFIX + SSH_CONFIG_PREFIX;
     public static final String PROJ_SSH_CONFIG_PREFIX = PROJ_PROP_PREFIX + SSH_CONFIG_PREFIX;
 
+    public static final String NODE_ATTR_PASS_ENV = "ssh-accept-env";
+    public static final String FWK_PROP_PASS_ENV = FWK_PROP_PREFIX + NODE_ATTR_PASS_ENV;
+    public static final String PROJ_PROP_PASS_ENV = PROJ_PROP_PREFIX + NODE_ATTR_PASS_ENV;
+
     private Framework framework;
 
     public JschNodeExecutor(final Framework framework) {
@@ -181,6 +185,7 @@ public class JschNodeExecutor implements NodeExecutor, Describable {
     public static final String CONFIG_CON_TIMEOUT = "ssh-connection-timeout";
     public static final String CONFIG_COMMAND_TIMEOUT = "ssh-command-timeout";
     public static final String CONFIG_BIND_ADDRESS = "ssh-bind-address";
+    public static final String CONFIG_PASS_ENV = "ssh-accept-env";
 
     static final Description DESC ;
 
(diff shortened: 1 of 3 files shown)


ProTip:
It looks like this is a breaking change since the default used to be, effectively, `true` on passing the envars through?

Is a week of timeout on connections and commands a bit excessive compared to the old default of 20 seconds?

jtobard:
> It looks like this is a breaking change since the default used to be, effectively, `true` on passing the envars through?

It is a breaking change but the opposite, leave the default `true` means every server not using env vars and not accepting them in the server is going to fail. A major release with a note on upgrade looks like a good time to do it. 
And that should have been the case since the beginning, is just a Jsch library fail to only thorw that error on the connection when you have the timeout on other value than zero, the error happens anyway is just ignored now.

> Is a week of timeout on connections and commands a bit excessive compared to the old default of 20 seconds?

It is just a placeholder since it is replaced with the user configuration and if the user put `0`, it supposed to be infinite and that's the close we can do. 
Anyway, every connection with a connection latency so high that it lasts more than an hour waiting for some response from the server ends anyway with another error like connection refused or host unreachable from the server side, this behavior is replicated using ssh command directly with high latency there's no configuration possible and it's on the server admin to fix it.

ltamaster:
maybe we can use this feature by default to avoid breaking stuff for the lack of environment variables
https://github.com/rundeck/rundeck/pull/5577

G3NSVRV:
Just for the record, I tested this implementation using the following delay on the network card:

```
g3nsvrv@LAPTOP-S60HCV51:~$ ssh root@node01.rundeck.local -t 'tc qdisc add dev ens33 root netem delay 1200ms 1800ms 25%'
Connection to node01.rundeck.local closed.

g3nsvrv@LAPTOP-S60HCV51:~$ ssh root@node02.rundeck.local -t 'tc qdisc add dev ens33 root netem delay 1800ms 3800ms 25%'
Connection to node02.rundeck.local closed.

g3nsvrv@LAPTOP-S60HCV51:~$ ssh root@node03.rundeck.local -t 'tc qdisc add dev ens33 root netem delay 3800ms 6800ms 25%'
Connection to node03.rundeck.local closed.
```

My Rundeck has these three nodes, each one with a different delay config; the three of them ran successfully after apply this config.

gschueler:
we need to mention the potential breakage in the release/upgrading notes

jtobard:
closing to refactor for missing code and make a new pr

PR #6130, opened 2020-06-12
Accept Environment variables optional for jsch executor. fix #4137

### LINKED ISSUE #4137: Failed: SSHProtocolFailure: channel is not opened .... JSCH default timeout

jtobard:
fix #4137

Jsch library has a problem with infinite timeout on channel connection, it has a hardcoded retry of 20000ms.

The solution is to use a number different to 0 on the timeout, but that has the collateral impact to throw an error previously ignored from Jsch: If the remote server is not configured to accept environment variables throws an exception.

The solution proposed is to make Environment variables optional, so you enable only if you are sure you already configured that on the server:
https://docs.rundeck.com/docs/administration/projects/node-execution/ssh.html#passing-environment-variables-through-remote-command

Another collateral effect of this change is connections with high latency that doesn't pass the variable environment has a better time on the execution of the step since it doesn't waste time passing multiple environment variables and waiting for the response.

The current configuration of JSCH executor will have `accept-env-var` on true and need to be disabled manually (changed from original PR since it looks like most rundeck users use `RD_` variables).

Solution tested in an environment with very high latency by @g3nsvrv and @cwaltherf

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/execution/impl/jsch/JschNodeExecutor.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/execution/impl/jsch/JschNodeExecutor.java
@@ -165,6 +165,10 @@ public class JschNodeExecutor implements NodeExecutor, Describable {
     public static final String FWK_SSH_CONFIG_PREFIX = FWK_PROP_PREFIX + SSH_CONFIG_PREFIX;
     public static final String PROJ_SSH_CONFIG_PREFIX = PROJ_PROP_PREFIX + SSH_CONFIG_PREFIX;
 
+    public static final String NODE_ATTR_PASS_ENV = "ssh-accept-env";
+    public static final String FWK_PROP_PASS_ENV = FWK_PROP_PREFIX + NODE_ATTR_PASS_ENV;
+    public static final String PROJ_PROP_PASS_ENV = PROJ_PROP_PREFIX + NODE_ATTR_PASS_ENV;
+
     private Framework framework;
 
     public JschNodeExecutor(final Framework framework) {
@@ -181,6 +185,7 @@ public class JschNodeExecutor implements NodeExecutor, Describable {
     public static final String CONFIG_CON_TIMEOUT = "ssh-connection-timeout";
     public static final String CONFIG_COMMAND_TIMEOUT = "ssh-command-timeout";
     public static final String CONFIG_BIND_ADDRESS = "ssh-bind-address";
+    public static final String CONFIG_PASS_ENV = "ssh-accept-env";
 
     static final Description DESC ;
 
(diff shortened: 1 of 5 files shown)


jtobard:
Copying comment from @gschueler on the previous PR: 
> we need to mention the potential breakage in the release/upgrading notes

gschueler:
is this a summary of the potentially breaking changes?

* old behavior used a timeout config of 0, causing an internal jsch timeout of 20s, which caused some connections to fail
* a new default timeout config is set for jsch (instead of 0), this fixes the timeout issue in cases where network is unstable
* however by setting the timeout config, jsch now responds differently to the case when remote sshd does not accept remote env vars,  causing connections to fail when env vars are sent (which we do by default)
* we added a new config (`send-env-var`?)which is set to true.  If set to false, we do not send env vars

However, it seems like we are now defaulting the configs to be less compatible:

1. only some users were encountering the timeout error in the old behavior due to network problems
2. since remote sshd config needs to be sent to accept env vars, it's likely that it is not set in most cases

Should we change the new defaults to be less likely to cause issues?  e.g. set the "send-env-vars" to false by default? Since it requires active reconfigure of remote sshd, perhaps we can make the send env vars require additional config to enable.

core/src/main/java/com/dtolabs/rundeck/core/execution/impl/jsch/JschNodeExecutor.java
```diff
@@ -165,6 +165,10 @@
     public static final String FWK_SSH_CONFIG_PREFIX = FWK_PROP_PREFIX + SSH_CONFIG_PREFIX;
     public static final String PROJ_SSH_CONFIG_PREFIX = PROJ_PROP_PREFIX + SSH_CONFIG_PREFIX;
 
+    public static final String NODE_ATTR_PASS_ENV = "ssh-accept-env";
```
gschueler, line 168: suggest changing to ssh-send-env
```suggestion
    public static final String NODE_ATTR_PASS_ENV = "ssh-send-env";
```

core/src/main/java/com/dtolabs/rundeck/core/execution/impl/jsch/JschNodeExecutor.java
```diff
@@ -181,6 +185,7 @@ public JschNodeExecutor(final Framework framework) {
     public static final String CONFIG_CON_TIMEOUT = "ssh-connection-timeout";
     public static final String CONFIG_COMMAND_TIMEOUT = "ssh-command-timeout";
     public static final String CONFIG_BIND_ADDRESS = "ssh-bind-address";
+    public static final String CONFIG_PASS_ENV = "ssh-accept-env";
```
gschueler, line 188:
```suggestion
    public static final String CONFIG_PASS_ENV = NODE_ATTR_PASS_ENV;
```

jtobard:
@gschueler Yes, that is the summary of the changes. 
I can change to default false and modify the tests if that is a better option

gschueler:
@jtobard please add upgrading notes here https://docs.rundeck.com/3.3.x/upgrading/upgrading-to-rundeck-3.3.html

jtobard:
Upgrading notes: https://github.com/rundeck/docs/pull/555

gschueler:
@mergifyio rebase

Issue #6119, opened 2020-06-09
"Unable to parse posted JSON data"  


wrogrammer:
**Describe the bug**
When I'm requesting the Webhook generated by Webhook Management I'm receiving: 
```
{
    "err": "Unable to parse posted JSON data"
}
```

**My Rundeck detail**

* The installed version of Rundeck is 3.3.0.
* Docker

**To Reproduce**
Request POST to Webhook without data parameter.

**Expected behavior**
200 Status Code

sjrd218:
Try posting an empty json document to the endpoint `{}`

wrogrammer:
I was trying and it is working. I'm using the external system and I'm not able to send additional data. 

I can use only the URL format: eg. 
/api/35/webhook/<ID>#test

wrogrammer:
It's working now.
Thanks @sjrd218 !

PR #6133, opened 2020-06-12
Add webhook plugin ability to write to the response stream

sjrd218:
Add the ability to write custom responses to the response output stream from webhook plugins.

Also fixes issue #6119 by relaxing the requirement to send an empty json document to a webhook that uses the JobRunWebhookEventPlugin.


core/src/main/java/com/dtolabs/rundeck/plugins/webhook/WebhookEventPlugin.java
```diff
@@ -25,5 +25,5 @@
  */
 public interface WebhookEventPlugin {
     default List<String> getRequestHeadersToCopy() {return new ArrayList<>(); };
-    void onEvent(WebhookEventContext context, WebhookData data) throws WebhookEventException;
+    WebhookResponder onEvent(WebhookEventContext context, WebhookData data) throws WebhookEventException;
```
gschueler, line 28: i think to preserve compatibility, this interface method shouldn't change.  perhaps add another method that returns WebhookResponder type that has a default implementation so older implementations still work e.g. `default WebhookResponder handleEvent(...){ return null; }`. to differentiate which method to use, there could be another method `default boolean handlesEvent(){ return false; }`

rundeckapp/webhooks/src/main/groovy/webhooks/responder/DefaultJsonWebhookResponder.groovy
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+package webhooks.responder
+
+import com.dtolabs.rundeck.plugins.webhook.WebhookResponder
+
+import javax.servlet.http.HttpServletResponse
+
+
+class DefaultJsonWebhookResponder implements WebhookResponder {
```
gschueler, line 23: maybe a default Json responder implementation would be useful in core, so plugin authors don't have to serialize json manually, just pass a Map/Collection to the default responder that uses jackson serialization

core/src/main/java/com/dtolabs/rundeck/plugins/webhook/WebhookEventPlugin.java
sjrd218: I went back and forth on maintaining backwards compatibility. 

I think it would be better to take advantage of the 3.3.x version bump and document the breaking change, rather than increase the complexity of the API and pollute it with deprecated methods that should not be used. My thought is that since the webhook API is relatively new, the need to maintain backward compatibility is less urgent.

gschueler: ok, if we add the WebhookResponder return type to the method, then it would be nice if existing implementations could just update their methods to return `null`, and have the same default response be sent as it was before.  I think the WebhookService or controller would have to change to handle that

gschueler:
please add upgrading notes to https://github.com/rundeck/docs/blob/3.3.x/docs/upgrading/upgrading-to-rundeck-3.3.md

PR #6152, opened 2020-06-18
Webhook save must flush the session

sjrd218:
Rundeck running on postgres requires the webhook save to flush in order to save the webhook.


ProTip:
Seems like an odd thing to need to add but it shouldn't hurt anything.

Issue #5730, opened 2020-01-24
api metrics "500 server error"


jplassnibatt:
**Describe the bug**
When metrics api call is used, it gives a `500 Server Error`.

**My Rundeck detail**

* Rundeck version: 3.2
* install type: war
* OS Name/version: ubuntu 18.04
* DB Type/version: mysql and h2, **NOT HAPPENING WITH POSTGRESQL**

**To Reproduce**
1. rd-cli command: rd executions metrics
2. see output: Request failed: 500 Server Error

**Additional context**
**Mysql output**

```
[2020-01-24 17:52:57.340]  WARN SqlExceptionHelper --- [qtp461803248-23] SQL Error: 0, SQLState: S1009
[2020-01-24 17:52:57.340] ERROR SqlExceptionHelper --- [qtp461803248-23] Bad format for Time '-40896541' in column 2
[2020-01-24 17:52:57.342] ERROR StackTrace --- [qtp461803248-23] Full Stack Trace:

org.hibernate.exception.GenericJDBCException: could not execute query
	at org.hibernate.exception.internal.StandardSQLExceptionConverter.convert(StandardSQLExceptionConverter.java:47)
	at org.hibernate.engine.jdbc.spi.SqlExceptionHelper.convert(SqlExceptionHelper.java:111)
	at org.hibernate.loader.Loader.doList(Loader.java:2625)
	at org.hibernate.loader.Loader.doList(Loader.java:2605)
	at org.hibernate.loader.Loader.listIgnoreQueryCache(Loader.java:2434)
	at org.hibernate.loader.Loader.list(Loader.java:2429)
	at org.hibernate.loader.criteria.CriteriaLoader.list(CriteriaLoader.java:109)
	at org.hibernate.internal.SessionImpl.list(SessionImpl.java:1787)
	at org.hibernate.internal.CriteriaImpl.list(CriteriaImpl.java:363)
	at org.hibernate.internal.CriteriaImpl.uniqueResult(CriteriaImpl.java:385)
	at grails.orm.HibernateCriteriaBuilder.executeUniqueResultWithProxyUnwrap(HibernateCriteriaBuilder.java:144)
	at org.grails.orm.hibernate.query.AbstractHibernateCriteriaBuilder.invokeMethod(AbstractHibernateCriteriaBuilder.java:1704)
	at org.grails.orm.hibernate.query.AbstractHibernateCriteriaBuilder.get(AbstractHibernateCriteriaBuilder.java:1586)
[...]

PR #6155, opened 2020-06-19
fix issue 5730: It's a workaround to allow calc metrics when database does not support sql contained within the criteria.

carlosrfranco:
Fix https://github.com/rundeck/rundeck/issues/5730

The criteria to query metrics have a sqlProjection that is not compatible with some databases and was returning an error 500 when this API was called

This changes is a workaround to calculate the metrics loading the data on memory avoiding using native sql on criteria. It will be executed only for databases that not support the original criteria with sqlProjection

First, I tried to change the native SQL on criteria that could be supported by any type of database but it was not possible since H2 and MySql has different behaviours on this scenario. After I tried to create a native sql to test the syntax query (and maybe faster) before the execution but that was also not possible since postgresql was returning error while the criteria is executed with success. So, for now, the workaround is to use a simple criteria to test compatibility before the query.

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
@@ -3950,7 +3950,7 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
         if(isSqlCompatible()){
             return queryExecutionMetricsByCriteria(query)
         } else {
-            log.warn("Error processing the query via criteria. Trying to perform in-memory processing")
+            log.debug("Execution metrics query using local calculation")
             return queryExecutionMetricsOnMemory(query)
         }
     }
--- a/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
@@ -62,6 +62,7 @@ import grails.web.mapping.LinkGenerator
 import groovy.transform.ToString
 import org.apache.commons.io.FileUtils
 import org.grails.web.json.JSONObject
+import org.hibernate.JDBCException
 import org.hibernate.StaleObjectStateException
 import org.hibernate.criterion.CriteriaSpecification
 import org.hibernate.type.StandardBasicTypes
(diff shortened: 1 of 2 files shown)


rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -3946,8 +3947,43 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
      * @return result map [total: long, duration: Map[average: double, max: long, min: long]]
      */
     def queryExecutionMetrics(ExecutionQuery query) {
+        if(isSqlCompatible()){
+            return queryExecutionMetricsByCriteria(query)
+        } else {
+            log.warn("Error processing the query via criteria. Trying to perform in-memory processing")
```
gschueler, line 3953: should not use `warn` here, maybe `info`

gschueler, line 3953:
```suggestion
            log.debug("Execution metrics query using local calculation")
```

rdoering:
We are running rundeck 3.3.4 and this issue does still exists :-(.

```
GET /api/29/project/<..>/executions/metrics HTTP/1.1
X-Rundeck-Auth-Token: 5q<..>
Accept: application/json
User-Agent: PostmanRuntime/7.26.5
Cache-Control: no-cache
Postman-Token: cdd7f27d-1d0c-4c73-9acd-419339a7b1dc
Host: task-mgr.<..>
Accept-Encoding: gzip, deflate, br
Connection: keep-alive
Cookie: JSESSIONID=node01kb080xwirhr364twp4j6d25u1042192.node0
HTTP/1.1 500 Server Error
Date: Tue, 03 Nov 2020 20:56:14 GMT
Server: Apache/2.4.18 (Ubuntu)
X-Frame-Options: deny
X-XSS-Protection: 1
X-Content-Type-Options: nosniff
Content-Security-Policy: default-src 'none' ; script-src 'self' 'unsafe-inline' 'unsafe-eval' ; style-src 'self' 'unsafe-inline' ; img-src * data: ; font-src 'self' data: ; connect-src 'self' https://api.rundeck.com ; form-action 'self' ;
Content-Type: application/json
Connection: close
Transfer-Encoding: chunked
{"error":"A server error occurred"}
```

If I call this request for another project on the same server it works maybe.

PR #6156, opened 2020-06-19
Quell a common grails/spring bean warning

sjrd218:
Quell a spring bean warning log message that is common for grails applications on startup. The grails guys don't appear to be in a hurry to fix it, so this log4j entry bumps the threshold to error for that particular class.

### DIFF
--- a/docker/official/remco/templates/log4j2.properties
+++ b/docker/official/remco/templates/log4j2.properties
@@ -99,6 +99,12 @@ logger.spring-core.level = warn
 logger.spring-core.additivity = false
 logger.spring-core.appenderRef.stdout.ref = STDOUT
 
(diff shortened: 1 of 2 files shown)


gschueler:
what is the warning message?

sjrd218:
`WARN  beans.GenericTypeAwarePropertyDescriptor - Invalid JavaBean property 'exceptionMappings' being accessed! Ambiguous write methods found next to actually used [public void grails.plugin.springsecurity.web.authentication.AjaxAwareAuthenticationFailureHandler.setExceptionMappings(java.util.List)]: [public void org.springframework.security.web.authentication.ExceptionMappingAuthenticationFailureHandler.setExceptionMappings(java.util.Map)]`

PR #6163, opened 2020-06-22
BUG: with #5440 Rundeck does not start anymore on OpenShift

hille721:
BUG: With #5440 Rundeck does not start anymore in OpenShift because the entry script failed with:
sed: couldn't open temporary file /etc/sedF47F9T: Permission denied

I guess sed will generate this as temporary file, but there are no write permissions in /etc.

Solution would be to use the echo statement (it does not matter if there is the Rundeck user entry twice)

another possibility would be to make the /etc directory group writable. But from security perspective I like the other way more.

**Additional context**
In a first version of #5440 there was also this echo statement instead of the sed. After short discussion with @ProTip I changed it to sed statement. Sorry, I hadn't considered this behavior.

### DIFF
--- a/docker/official/lib/entry.sh
+++ b/docker/official/lib/entry.sh
@@ -52,7 +52,7 @@ fi
 # Support Arbitrary User IDs on OpenShift
 if ! whoami &> /dev/null; then
   if [ -w /etc/passwd ]; then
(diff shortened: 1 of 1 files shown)


ProTip:
@hille721 Thanks for pointing this out! Checkout #6168 , I believe this should finally be resolved!

hille721:
@ProTip great solution in #6168. Already checked and verified it this afternoon. Works as aspected, thanks!

PR #6164, opened 2020-06-22
Docker - Environment Variable $HOME not set

hille721:
BUG: When starting the container with a different UID as 1000 (e.g. in OpenShift), the HOME environment variable is set to /.
This can lead to problems, e.g. Ansible is using ~/.ansible/tmp as local temorary directory.

simply declare it

### DIFF
--- a/docker/official/lib/entry.sh
+++ b/docker/official/lib/entry.sh
@@ -10,6 +10,7 @@ done
 export HOSTNAME=$(hostname)
 
 export RUNDECK_HOME=/home/rundeck
+export HOME=$RUNDECK_HOME
 
 export REMCO_HOME=/etc/remco
 export REMCO_RESOURCE_DIR=${REMCO_HOME}/resources.d


ProTip:
@hille721 Do the changes in #6167 resolve this for you? With those changes `$HOME` appears to be set correctly when the uid is set randomly.

hille721:
> @hille721 Do the changes in #6167 resolve this for you? With those changes `$HOME` appears to be set correctly when the uid is set randomly.

@ProTip : I just tested on our Openshift Environment. Checking on the terminal looks fine:
```
sh-4.4$ id
uid=1000370000(rundeck) gid=0(root) groups=0(root),27(sudo),1000370000
sh-4.4$ echo $HOME
/home/rundeck
```

But in the Rundeck session, `$HOME` is still wrong:
image

Tested with Rundeck Version Rundeck 3.3.0 RC1-20200623 and SNAPSHOT.

ProTip:
Ah, yes.. Because it's not running via a login shell. Thanks!

PR #6186, opened 2020-06-25
Fix: schedule takeover results in 0 jobs when it should match

gschueler:
* query was not correctly selecting jobs

Fix: schedule takeover results in 0 jobs when it should match

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
@@ -24,6 +24,7 @@ import com.dtolabs.rundeck.core.plugins.DescribedPlugin
 import com.dtolabs.rundeck.core.plugins.JobLifecyclePluginException
 import com.dtolabs.rundeck.core.plugins.ValidatedPlugin
 import grails.gorm.transactions.NotTransactional
+import grails.orm.HibernateCriteriaBuilder
 import groovy.transform.CompileStatic
 import org.hibernate.criterion.DetachedCriteria
 import org.hibernate.criterion.Projections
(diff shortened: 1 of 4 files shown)


rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
```diff
@@ -4221,17 +4222,64 @@ class ScheduledExecutionService implements ApplicationContextAware, Initializing
...
-        def scheduledExecutions = c.listDistinct {
+        return ScheduledExecution.createCriteria().listDistinct {
             or {
-                if(!ignoreInnerScheduled){
-                    eq('scheduled', true)
-                }
-                executions(CriteriaSpecification.LEFT_JOIN) {
-                    eq('status', ExecutionService.EXECUTION_SCHEDULED)
-                    isNull('dateCompleted')
+                applyAdhocScheduledExecutionsCriteria(delegate, selectAll, fromServerUUID, toServerUUID, projectFilter)
```
sjrd218, line 4278: This line makes the query slower by orders of magnitude. We might want to find a way to optimize this.

gschueler: hmm, i just moved the `executions` join construction to a separate method call, that shouldn't make it any slower.  do you mean the restructuring of the query made it slower? this is possible but unfortunately the previous query was not correct for the results we are looking for.

gschueler:
@Mergifyio rebase

rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
sjrd218: Maybe it was slow to begin with. If I comment out this line and run the api takeover the query time goes from 80 seconds to about 1 second.

gschueler:
@Mergifyio rebase

gschueler:
we may be able to improve the query times by splitting the job vs executions queries into two separate queries not needing a join

Issue #5723, opened 2020-01-22
Multiline in-line script on YAML export displayed in single line


sinister3vil:
**Describe the bug**
When exporting the job definition to YAML in-line scripts display in a single line, with special characters like \t, \n and line wraps.
Previously (3.0.6) in-line scripts would use YAMLs block style and chomp indcators : |-  and the script would be included in human readable format.

This affects exporting from the web interface but also via rd job list command.

currently
```
commands:
    - script: "#!/bin/bash\n\n#change these based on job----------------------------------------------\n\
       \t\texport PLAYBOOK=test #do not add extension\n\t\texport BASEDIR=/tmp/dryrun\n\
       rest of script redacted"

```

previously
```
commands:
    - script: |-
        #!/bin/bash

        #change these based on job----------------------------------------------
        		export PLAYBOOK=test #do not add extension
                export BASEDIR=/tmp/dryrun
                rest of script redacted
```

Issue seems to affect jobs with larger in-line scripts or possible edited.

**My Rundeck detail**

* Rundeck version: 3.1.3, 3.2.1
* install type: official docker image

**To Reproduce**
Export job definition in YAML format

**Expected behavior**
Expecting in-line scripts to properly use multi-line YAML syntax

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser Chrome
 - Version 79.0.3945.130

MegaDrive68k:
Hi @sinister3vil 

I tested your issue since our first contact at IRC channel and I found a way to reproduce the issue exactly.

1. Import this job:

```
- defaultTab: nodes
  description: ''
  executionEnabled: true
  id: c5e431e9-ac55-4c80-8409-f4dbb84fbf44
  loglevel: INFO
  name: HelloWorld
  nodeFilterEditable: false
  scheduleEnabled: true
  sequence:
    commands:
    - script: |-
        echo "hi"
        ls -la
    keepgoing: false
    strategy: node-first
  uuid: c5e431e9-ac55-4c80-8409-f4dbb84fbf44
```

2. Modify the inline-script leaving all in one line and Save the job.

![Screenshot_1](https://user-images.githubusercontent.com/51376003/72925052-cde0ae80-3d30-11ea-8213-46327d2f7b88.png)


3. Modify the inline-script again like the original job def and Save again:

![Screenshot_2](https://user-images.githubusercontent.com/51376003/72925078-d933da00-3d30-11ea-892e-421493bca471.png)


[...]

PR #6200, opened 2020-06-29
Enhance YAML multiline export functionality

sjrd218:
Allows a user to direct the yaml exporter to trim lines before exporting in order to get a nice literal text dump of their code blocks.

The user who reported their job exporting in an unexpected format #5723 was apparently depending on the older snakeyaml lib behavior which was more lax about string space representation. The newer snakeyaml lib tries to be more accurate in its string spacing representation and defaults to the YAML double quote representation which provides the most accurate rendering of complex multiline output.

Stripping all trailing spaces on all lines seems to allow the YAML exporter to use the literal representation more deterministically.

Default for the trim spaces is off.

New Rundeck config:
`rundeck.job.export.yaml.trimSpaces=true`

### DIFF
--- a/rundeckapp/src/main/groovy/com/dtolabs/rundeck/app/support/BuilderUtil.groovy
+++ b/rundeckapp/src/main/groovy/com/dtolabs/rundeck/app/support/BuilderUtil.groovy
@@ -230,6 +230,17 @@ class BuilderUtil {
     }
 
     /**
+     * Trim all empty space from each line in the input string and
+     * replace all line endings with the given string
+     * @param os input string
+     * @param lineEnding line ending string to use
+     * @return new string
+     */
+    static String trimAllLinesAndReplaceLineEndings(String os, String lineEnding) {
+        os.readLines().collect { it.trim() }.join(lineEnding)
+    }
+
+    /**
      * Add entry to the map for the given key, converting the key into an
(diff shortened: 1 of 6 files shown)


rundeckapp/src/main/groovy/com/dtolabs/rundeck/app/support/BuilderUtil.groovy
```diff
@@ -229,6 +229,17 @@ class BuilderUtil {
...
 
+    /**
+     * Trim all empty space from each line in the input string and
+     * replace all line endings with the given string
+     * @param os input string
+     * @param lineEnding line ending string to use
+     * @return new string
+     */
+    static String trimAllLinesAndReplaceLineEndings(String os, String lineEnding) {
+        os.readLines().collect { it.trim() }.join(lineEnding)
```
gschueler, line 240: String.trim removes both *leading* and *trailing* whitespace.  I think it should not remove leading whitespace, so `.trim()` is probably not the right method to use.  Test cases don't have any cases with leading whitespace.

sjrd218: I've changed it to trim the end of the line and added tests.

PR #6254, opened 2020-07-10
fix #6239 acls not imported during fs conversion upgrading from 3.2 to 3.3

### LINKED ISSUE #6239: Project ACL's are gone after upgrading to v3.3
All of our **project ACL** setting files are no longer visible from the UI. But the ACL from the settings are still in place. 

RUNDECK
APIVERSION : 35
BASE : /var/lib/rundeck
BUILD : 3.3.0-20200701
BUILDGIT : v3.3.0-0-g66e3ef7
SERVERUUID :
VERSION : 3.3.0-20200701

image

Thanks,
HS

gschueler:

Fix #6239 

Converting from project storage type FS to DB now imports all config resources

**Additional context**

Using WAR deployment in rundeck 3.2 and earlier, the default for the `rundeck.projectsStorageType` was `file` not `db` based on internal app default, whereas other install types (deb,rpm,docker) use `db` by default.

On upgrade from 3.2.x to 3.3.0, a war install using the default config value would convert from file to db, however project level ACLs were not getting converted.

**mitigation**

If you upgrade to Rundeck 3.3.0 using a war install, and did not set `rundeck.projectsStorageType` explicitly in your config file, any filesystem-stored projects would be converted to DB storage, but ACLs would not be copied.  You will need to manually install the aclpolicy files from the filesystem by using the Rundeck UI or API to add them to the project.

If you upgrade to Rundeck 3.3.1(TBR) from 3.2.x, the ACLs will be imported.  

If you upgrade to Rundeck 3.3.1(TBR) from 3.3.0 you can reimport the missing files by adding this config: `rundeck.projectsStorageImportResources=always` and renaming any `project.properties.imported` back to `project.properties` before starting up 3.3.1.  This will reimport any files (acls, readme) from the filesystem to existing projects.  You can rename any file to end with ".imported" before startup to have it be skipped by this process.

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/ProjectManagerService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ProjectManagerService.groovy
@@ -877,15 +877,22 @@ class ProjectManagerService implements ProjectManager, ApplicationContextAware, 
                 projectProps.putAll(other.getProjectProperties())
                 def newProj=createFrameworkProject(other.name,projectProps)
                 //import resources
-                ["readme.md","motd.md"].each{ fpath ->
-                    if(other.existsFileResource(fpath)){
-                        log.warn("Importing ${fpath} for project ${other.name}...")
+                List paths=other.listDirPaths('')
+                while(paths.size()>0) {
+                    String path = paths.remove(0)
+                    if(path=="/etc/project.properties"){
+                        continue
+                    }
+                    if(path.endsWith('/')){
+                        paths.addAll(other.listDirPaths(path))
+                    }else{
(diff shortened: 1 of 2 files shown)


MegaDrive68k:
From 3.2.8 -> 3.3.0 -> 3.3.1-SNAPSHOT (branch issue/6239) i saw same behavior:

Screenshot_1

But the ACL keeps on Filesystem:

Screenshot_2

From 3.2.8 -> 3.3.1-SNAPSHOT (branch issue/6239):

Screenshot_3

Screenshot_4

PR #6255, opened 2020-07-10
Docker - Add / on rundeck/jaas/file/required

cpoule23:
Add / at the begining of rundeck/jaas/file/required. Without "/", the env variable is not taken in consideration


ProTip:
Thanks!

PR #6283, opened 2020-07-17
Reduce workflow engine meta logging in log output by marking as wfmeta

### LINKED ISSUE #5290: Workflow execution debug information appearing in log
**Describe the bug**
Debug information about workflow step execution is appearing in the execution log.

**My Rundeck detail**

* Rundeck version: 3.1.1

IMG_20190923_195259

gschueler:
* hide warn/debug in log view by default unless `?wfmeta=true` param is sent
* don't include workflowID in exec output log data

Reduces DEBUG/WARN log output when workflow engine details are not necessary.

Fixes #5290

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/execution/workflow/EngineWorkflowExecutor.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/execution/workflow/EngineWorkflowExecutor.java
@@ -325,7 +325,9 @@ public class EngineWorkflowExecutor extends BaseWorkflowExecutor {
         return (String message) -> {
             String logMessage = String.format("[wf:%s] %s", workflowId, message);
             logger.debug(logMessage);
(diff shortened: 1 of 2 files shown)


ProTip:
Looks like this should do the trick. Do you think we should have it default to `true` to preserve the current behavior?

Something I hadn't thought much about before is that the inclusion/exclusion of output will affect the "line numbers" and line number linking. This would also not be a live updating preference since the data would need to be re-fetched from the server.

gschueler:
I think the workflow meta log data should be hidden by default, but have a way to reveal it if needed via UI.  Another option would be introduce a new "metalog" entry type, so it is completely independent of normal log output.  either way, a UI toggle would have to reload from server to get the data

fdevans:
Update from today: @gschueler to review and determine if this should continue or close and revisit later.

PR #6298, opened 2020-07-22
Issue 1137 [WIP]: Creating a cache to store the list of project names for which the current user is authorized to create a job.

carlosrfranco:
Fix: https://github.com/rundeckpro/rundeckpro/issues/1137

This changes avoid the authorization checks for every project whenever it's necessary causing some pages to take a long time to load.

Was created a cache to store the project names in a list for which the current user is authorized to create a job. This list should be updated if the project list or ACLs changes.


jtobard:
- [ ] Use Google Guava for caching system
- [ ] Compare times from every page from user perspective using this proposed solution (@nvalenzuela20).
- [ ] Once this is approved include unit test and selenium ( qa framework) test for every page to check response times scaling.

PR #6302, opened 2020-07-22
Fix node view output not rendering

### LINKED ISSUE #6292: Logs in Node View are broken since 3.3.0
**Describe the bug**
Running Jobs in parallel Mode. Trying to view logs from one step. In rundeck versions previous to 3.3.0 you could open the logoutput from one Step( see screenshot). But now the loading spinner won't stop and the logs arren't loaded.

I prefere view logs in the node view to the logoutput view, as the logoutput view can be confusing in parallel mode.

I assume it has something to do with the new logviewer introduced in Version 3.3.0. https://docs.rundeck.com/docs/history/3_3_x/version-3.3.0.html#new-log-viewer

**My Rundeck detail**

* Rundeck version: 3.3.0
* install type: rpm
* OS Name/version: CentOS Linux release 7.8

**To Reproduce**
1. Run a Job.
2. Switch to Node View, if needed.
3. Select a host and click on the arrow button in front of it, to show the steps.
4. Select a step and click on the arrow button int front of it, to show it's logs.
5. The loading spinner won't stop and the logs aren't loaded.

**Expected behavior**
Logs of a specified step are loaded.

**Screenshots**
grafik

 - OS: macOS 10.14.6
 - Browser Firefox (78.0.2), Chrome (84.0.4147.89)
 - But im quite sure it has nothing to do with the browser or its version.

ProTip:
Fixes #6292
Fixes rundeckpro/rundeckpro#1150

This fixes execution output not rendering in the "node view". More likely to occur with job reference steps.

The KO code uses a _cleaned_(stripped of parameters like node) step context when creating an `RDNodeStep`. The execution output store was grouping on the raw `stepctx` causing a lookup failure if it contained encoded parameters.

This change groups the output entries by the cleaned `stepctx` so the lookups will always be successful.

N/A

### DIFF
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/components/execution-log/logBuilder.ts
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/components/execution-log/logBuilder.ts
@@ -48,15 +48,9 @@ export class LogBuilder {
 
     const {node, stepCtx} = this.opts
 
-    if (node && !stepCtx) {
-      const query = `${this.opts.node}`
+    if (node) {
       this.observeFiltered(() => {
-        return executionOutput.entriesByNode.get(query)
-      })
-    } else if (node && stepCtx) {
-      const query = `${node} ${stepCtx}`
-      this.observeFiltered(() => {
-        return executionOutput.entriesbyNodeCtx.get(query)
+        return executionOutput.getEntriesByNodeCtx(node, stepCtx)
       })
     } else {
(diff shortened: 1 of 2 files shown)


ProTip:
@sjrd218 I'll put the project from the Enterprise bug here if you want to check it out further: 
[project_test.zip](https://github.com/rundeck/rundeck/files/4962272/project_test.zip)

Issue #6315, opened 2020-07-25
feature/enhancement : optimization for Dockerfiles using debian package manager tweaks


Rajpratik71:
**Is your feature request related to a problem? Please describe.**
No . It is not related to problem . It is optimization tweak for docker images build using dockerfiles

**Describe the solution you'd like**

By default, Ubuntu or Debian based "apt" or "apt-get" system installs recommended but not suggested packages .

By passing "--no-install-recommends" option, the user lets apt-get know not to consider recommended packages as a dependency to install.

This results in smaller downloads and installation of packages .

Refer to blog at [Ubuntu Blog](https://ubuntu.com/blog/we-reduced-our-docker-images-by-60-with-no-install-recommends) .

ProTip:
Hi @Rajpratik71 ,

Thanks for this.. We are aware of that option however without an analysis of size savings, as well as a list of packages that would get left out and what the impact would be on users if they "went missing" there wouldn't be much appetite for this change ATM.

A few of these are testing containers internal to project development as well.

PR #6316, opened 2020-07-25
optimization debian package manager tweaks

Rajpratik71:
Pull for issue #6315 

By default, Ubuntu or Debian based "apt" or "apt-get" system installs recommended but not suggested packages .

By passing "--no-install-recommends" option, the user lets apt-get know not to consider recommended packages as a dependency to install.

This results in smaller downloads and installation of packages .

Refer to blog at [Ubuntu Blog](https://ubuntu.com/blog/we-reduced-our-docker-images-by-60-with-no-install-recommends) .

Signed-off-by: Pratik Raj <rajpratik71@gmail.com>

### DIFF
--- a/docker/debinstall/Dockerfile
+++ b/docker/debinstall/Dockerfile
@@ -2,7 +2,7 @@ FROM rdubuntu16.04-util:latest
 
 ## General package configuration
 RUN apt-get -y update && \
-    apt-get -y install \
+    apt-get --no-install-recommends -y install \
         sudo \
         software-properties-common \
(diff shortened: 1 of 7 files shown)


ProTip:
Hi @Rajpratik71 ,

Thanks for this.. We are aware of that option however without an analysis of size savings, as well as a list of packages that would get left out and what the impact would be on users if they "went missing" there wouldn't be much appetite for this change ATM.

A few of these are testing containers internal to project development as well.

PR #6331, opened 2020-07-31
Issue 1137 [WIP]: ACL evaluation cache system

carlosrfranco:
Fix: https://github.com/rundeckpro/rundeckpro/issues/1137

This changes avoid the authorization checks for every project whenever it's necessary causing some pages to take a long time to load.

Was created a cache to store the project names in a list for which the current user is authorized to create a job. This list should be updated if the project list or ACLs changes.


rundeckapp/src/main/groovy/com/dtolabs/rundeck/server/projects/AuthContextEvaluatorCacheManager.groovy
```diff
@@ -0,0 +1,136 @@
+package com.dtolabs.rundeck.server.projects
```
jtobard, line 1: Move up to `rundeck.server` package since is not related to projects anymore

rundeckapp/grails-app/controllers/rundeck/controllers/ExecutionController.groovy
```diff
@@ -69,6 +70,7 @@ class ExecutionController extends ControllerBase{
     FileUploadService fileUploadService
     PluginService pluginService
     ConfigurationService configurationService
+    AuthContextEvaluatorCacheManager authContextEvaluatorCacheManager
```
jtobard, line 73: unused import

plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/BaseGitPlugin.groovy
```diff
@@ -631,7 +631,7 @@ class BaseGitPlugin {
         try {
             push = pushb.call()
         } catch (Exception e) {
-            plugin.logger.debug("Failed push to remote: ${e.message}", e)
+            logger.debug("Failed push to remote: ${e.message}", e)
```
jtobard, line 634: this change came from rebase, we can ignore by now, later we can create a new clean PR

gschueler:
the authContextEvaluatorCacheManager doesn't seem to work correctly for me.  It caches auth decisions when it should have invalidated them, I'm not sure why.  
I created a test user and some test system acls to grant access to several projects.  At some point i could view the projects in the project list but when attempting to access them I get authorization denied error.  I can update the system acl policy live, and see some authorizations change for the user, such as api token creation.  But for some reason accessing different projects seems to get cached beyond the normal cache

rundeckapp/grails-app/services/rundeck/services/AuthorizationService.groovy
```diff
@@ -319,10 +331,12 @@ class AuthorizationService implements InitializingBean{
                 resourceModified:{String path->
                     log.debug("resourceModified ${path}, invalidating")
                     sourceCache.invalidate(path)
+                    storedPolicyPathsCache.invalidateAll()
                 },
                 resourceDeleted:{String path->
                     log.debug("resourceDeleted ${path}, invalidating")
                     sourceCache.invalidate(path)
+                    storedPolicyPathsCache.invalidateAll()
```
gschueler, line 339:
```suggestion
                    if(path.startsWith(ACL_STORAGE_PATH_BASE) && path.endsWith('.aclpolicy')) {
                        storedPolicyPathsCache.invalidateAll()
                    }
```

rundeckapp/grails-app/services/rundeck/services/AuthorizationService.groovy
```diff
@@ -319,10 +331,12 @@ class AuthorizationService implements InitializingBean{
                 resourceModified:{String path->
                     log.debug("resourceModified ${path}, invalidating")
                     sourceCache.invalidate(path)
+                    storedPolicyPathsCache.invalidateAll()
```
gschueler, line 334:
```suggestion
                    if(path.startsWith(ACL_STORAGE_PATH_BASE) && path.endsWith('.aclpolicy')) {
                        storedPolicyPathsCache.invalidateAll()
                    }
```

gschueler: This invalideAll should also be called inside `resourceCreated` handler

gschueler:
actually i think it's not a problem with authContextEvaluatorCacheManager, but the user session caches the "SubjectAuthContext" object, which holds a reference to the system Authorization object when it was created.  If the user logs out and in again it resets to the new data...

PR #6350, opened 2020-08-05
Fix #6349: failure to serialize jobs for git due to mkdirs

### LINKED ISSUE #6349: git plugin: Serialization failure updating multiple jobs due to mkdirs failing
**Describe the bug**

Sometimes updating multiple jobs (via API or rd jobs load) can fail to serialize the jobs for git, and shows an error in the log about failing at `mkdirs`.

**My Rundeck detail**

* Rundeck version: 3.2,3.3

**To Reproduce**
1. Use api or rd to import multiple jobs in the same group path, using `${job.group}` in the file pattern of git-export plugin config
2. See errors in system log
3. See missing files in local git workspace

**Expected behavior**
Files are serialized correctly

gschueler:

Fix: #6349 
when serializing multiple jobs in the same dir at the same time a race
condition around the parent dir creation can cause a failure

this synchronizes creation of parent dirs


gschueler:
@mergifyio rebase

PR #6353, opened 2020-08-06
SCM and Git plugin fixes

gschueler:

Fix several issues with SCM/git serialization.

- [x] renaming a job does not correctly show as "Modified" state, instead shows as "Created" state
- [x] deleting a job and later recreating it or adding another job with same scm file path, can fail to serialize properly
- [x] updating multiple jobs via API/rd can fail to serialize some jobs


jtobard:
Tested on local and is working. We have some SCM API tests, it is possible to include tests for this? (In a future iteration I mean)

PR #6370, opened 2020-08-14
Set webpack public path dynamically

ProTip:
Fixes rundeckpro/rundeckpro#1200

The public path was not set correctly to load assets referenced from source files. This also needs to be set dynamically in the entry in order to account for non-bare base url paths(ie running in tomcat under `/rundeck`).

It may be cleaner to have a single webpack runtime file loaded and then set the public path once in `central/main`. This is a big runtime change to be made so close to release though.


MegaDrive68k:
Tested and works.

Screenshot_13

PR #6394, opened 2020-08-21
Forcing a new attempt when to execute script and the script file is busy because of chmod command executed just before

carlosrfranco:
Fix https://github.com/rundeckpro/rundeckpro/issues/1080

Sometimes, if there is a lot of scripts running in parallel and this scripts produce output and output erros, some lines should be printed on a single line (i.e. merging more than one line).

A newLocalExecutor had already been implemented and fixed part of the problem. However, the "chmod" command executed before the script was causing the file to be busy at the time it was called to execute. A change was made to try again after a short wait

Before was implemented a solution to capture the stdout and stderr in separate streams before this lines is printed and breaks this line to print them correctly. https://github.com/rundeck/rundeck/pull/6273

**Additional context**
This problem appears at random

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/execution/impl/local/NewLocalNodeExecutor.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/execution/impl/local/NewLocalNodeExecutor.java
@@ -35,13 +35,19 @@ import java.util.Map;
  */
 public class NewLocalNodeExecutor implements NodeExecutor {
     private static final String MESSAGE_ERROR_FILE_BUSY_PATTERN = "Cannot run program.+: error=26.*";
-    private static final int TIME_TO_WAIT_BEFORE_TRY_AGAIN = 1000;
+    private static final int MAX_TIME_TO_WAIT_BEFORE_TRY_AGAIN = 3000;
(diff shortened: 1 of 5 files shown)


carlosrfranco:
> I don't think this change should be made here, if the issue is with chmod on the script, then the Script Node executor should have the retry action.

@gschueler I see the error when using the newLocalNodeExecutor. I did not see this error happening in another executor or in any other situation except here and in the other solution previously implemented https://github.com/rundeck/rundeck/pull/6273 (both are local executors).

carlosrfranco:
@gschueler The "chmod" command is called here: https://github.com/rundeck/rundeck/blob/5e26a975687c61ba694dfb7391b07d86bda30942/core/src/main/java/com/dtolabs/rundeck/core/execution/workflow/steps/node/impl/DefaultScriptFileNodeStepUtils.java#L291

maybe the change could be here but as this is a very specific case, I thought it best to leave it in the newLocalNodeExecutor class

drodriguez-rundeck:
I did some tests with the snapshot 3.3.3 given by @carlosrfranco . More than 400.000 executions without errors

image (11)

ProTip:
@gschueler @carlosrfranco Running the `sync` command after `chmod` may also resolve this issue.

Issue #6401, opened 2020-08-25
When a job is marked as "Incomplete", the global variables are not being expanded in the email notification 


jtobard:
When a job is marked as "Incomplete", the global variables from framework.properties are not being expanded in the email notification.

The easiest way to reproduce the issue is to setup email notification in a job, add an inline script step and run a long running script (for instance, a loop counter), wait a few moments and stop rundeck service while the job is still running. After restart an email is received but the global variables are not expanded. Below there are some screenshots for reference.

framework.properties
![framework_properties](https://user-images.githubusercontent.com/48293178/82250172-f95b5a80-9918-11ea-8bc7-b6932b071089.jpg)

Email Notification Configuration
![notification_config](https://user-images.githubusercontent.com/48293178/82250231-1132de80-9919-11ea-97e7-3e2226fb7fb8.jpg)

Activity Section
![activity](https://user-images.githubusercontent.com/48293178/82250276-2576db80-9919-11ea-9f72-7511b060a150.jpg)

Received Email Notification
![received_email](https://user-images.githubusercontent.com/48293178/82250317-3c1d3280-9919-11ea-97b5-0473438e1734.jpg)

Service.log after restart
![service_log](https://user-images.githubusercontent.com/48293178/82250367-4fc89900-9919-11ea-8e38-a3457e85b9ba.jpg)

PR #6407, opened 2020-08-26
Fixes when send an email notification when a job is marked as "Incomplete" #6401

nvalenzuela20:
Fix https://github.com/rundeck/rundeck/issues/6401

This is a bug that appears when a job is marked as "Incomplete" and the system sends an email notification, the problem causes the string "${globals.environment}" to appear in the subject.

The method "frameworkService.getProjectGlobals()" has been added in the code that generates the email's subject. If the method has already been called it will be not called again.


nvalenzuela20:
Create a new PR

PR #6414, opened 2020-08-27
Rundeck initialization refactoring

sjrd218:
Refactor server uuid assignment and loading.
Cleanup initialization code, and refactor it to work the same in container and non-container installations.
Add an unsecured health check endpoint.
Add cleanup code for metrics and log storage service upon server restart or shutdown.
Add plumbing for hot restart capability.

The initialization method in the FrameworkService has been removed and the serverUuid now comes from a system property that is set in the prebootstrap phase of the Rundeck startup. This helps avoid some of the issues we've had in the past with the system triggering the initialization at the wrong time.

~~The server uuid now comes from a file called `serverId` located in the server config directory. The id from framework.properties is migrated into the file if it exists. If the environment variable `RUNDECK_SERVER_UUID` or the `rundeck.server.uuid` system property is set by the user, that value will be used for the server id and will be installed in the serverId file.~~

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -26,6 +26,8 @@ mavenCentralUrl=https://repo1.maven.org/maven2/
 grailsCentralUrl=https://grails.org/plugins
 jacksonDatabindVersion=2.10.1
 snakeyamlVersion=1.26
+gsonVersion=2.8.6
+httpClientVersion=4.5.12
 #
 # Override of spring-boot dependencies versions.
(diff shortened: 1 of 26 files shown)


sjrd218:
@gschueler @ProTip I have update this PR as discussed. For some reason Travis won't let me restart the failed test run, otherwise it'd be passing.

gschueler (approved):

PR #6426, opened 2020-09-01
internal refactor: don't throw runtime exception on missing property lookup

gschueler:

* change behavior of PropertyLookup.getProperty - don't throw PropertyLookupException on missing property, return null


gschueler:
@Mergifyio rebase

Issue #6436, opened 2020-09-08
Minor product issue: If a node executor plugin is missing, the project's "Default Node Executor" section is blank


jtobard:
After importing a project from another Rundeck that had a "Node Executor Plugin" that does not exist on the server where the project was added, the project's "Default Node Executor" section appears blank (or if the plugin is installed via upload on the plugin manager page in one of the cluster member)

**Steps to reproduce:**
Rundeck server 1, install a plugin like sudo local executor 
Create a project and select the executor plugin you added.
Login to server 2 in the same cluster and edit the project settings.

![Screenshot (1036)](https://user-images.githubusercontent.com/48293178/84104475-c2420b80-a9e3-11ea-9fee-29231ca7d8b1.png)

![Screenshot (1038)](https://user-images.githubusercontent.com/48293178/84104494-cb32dd00-a9e3-11ea-82d6-6e1179a92333.png)

Obviously, to avoid this, the plugin should be installed in all cluster members, but if not, it would be a good idea to show a "Missing Plugin" message instead of appearing blank.

PR #6437, opened 2020-09-08
Node Executor or File Copier Plugin are missing #6436

nvalenzuela20:
Fix https://github.com/rundeck/rundeck/issues/6436

If a node executor or file copier plugin are missing, the project's "Default Node Executor" or "Default File Copier" section will appear empty. 
This error is generated when importing a project that has a node executor or file copier plugins selected from a cluster to another cluster that does not have these plugins previously installed

Validation messages are added so that the user knows that the plugin is not available

image


nvalenzuela20:
Create a new PR #6454

Issue #6442, opened 2020-09-10
Rundeck is not importing jobs from 2.4.x correctly


nvalenzuela20:
### Bug Description:
A customer is exporting jobs from **Rundeck OSS 2.11.14**, those jobs have into a Job Reference Step with the "_Import Options?_" Checkbox enabled.
When the job definition is imported in **Rundeck Enterprise 3.2.6**, that Checkbox into Job reference Stop is unchecked. This happens when the job is exported as XML or YAML. It also happens even if the entire project is exported and imported into the new instance as a jar archive. 

### Rundeck Details:
Tested on:
Rundeck 3.0.9
Rundeck 3.0.27
Rundeck 3.2.6
Rundeck 3.2.7
Rundeck 3.2.9
Rundeck 3.3.0

### Steps to Reproduce the Bug:

- Load the attached Job Definition "run-inventory-listing-smoke-test-oldrundeck.yaml".
- Go to Job Workflow.
- Open the Job Reference Step.
- Notice the "_Import options?_" is unchecked.

### Screenshot:
![Screenshot (5 1)](https://user-images.githubusercontent.com/53315711/88235373-be431380-cc48-11ea-8fb6-3a618ac8ec20.png)

### File:
[run-inventory-listing-smoke-test-oldrundeck.zip](https://github.com/rundeckpro/rundeckpro/files/4963008/run-inventory-listing-smoke-test-oldrundeck.zip)

PR #6444, opened 2020-09-10
ImportOptios flag is not importing correctly when upload a job definition #6442

nvalenzuela20:
Fix https://github.com/rundeck/rundeck/issues/6442

When the user imports a job definition with the "_importOptions_" flag set to "_true_", Rundeck fails to save it on the database.

Fixed the job importing logic by adding an added an alternative block to the code that obtains  "_importOptions_" from “_commands_” when there are no options on “_jobref_”


gschueler (approved):

nvalenzuela20:
Tested by @nsuarezdev

jtobard (approved):

PR #6456, opened 2020-09-17
Fix #6449 - Sync command disabled as default and without arguments

### LINKED ISSUE #6449: "sync: ignoring all arguments" warning message on script step (only on 3.3.3)
**Describe the bug**

`sync: ignoring all arguments` always appears when using script-step on Rundeck 3.3.3, it seems only a warning message, the script works OK.

**My Rundeck detail**

* Rundeck version: 3.3.3
* install type: WAR/RPM/Docker
* OS Name/version: Ubuntu 20.04.1
* DB Type/version: H2

**To Reproduce**

1. Create a job with a script-step, the job must be dispatched to a remote SSH node.
2. Run the job.
3. Check the output, always appears `sync: ignoring all arguments` warning at the beginning.

**Expected behavior**

No warnings on output.

**Screenshots**

Screenshot_3

 - OS: Ubuntu 20.04.1
 - Browser: Firefox
 - Version: 80.0.1

**Additional context**

This warning doesn't appear on Rundeck 3.3.2. 
The message disappears adding  `enable-sync="false"` attribute on node definition (resource.xml).

carlosrfranco:
Fix #6449 

The message "sync: ignoring all arguments" appears when using script-step on Rundeck 3.3.3 for some OS despite the script is running ok.

Some OS were returning a warning message because the command does not use any arguments so it was dropped. I also changed the default behavior so that sync is disabled by default and being enabled via node attribute "enable-sync = true"

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/execution/workflow/steps/node/impl/DefaultScriptFileNodeStepUtils.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/execution/workflow/steps/node/impl/DefaultScriptFileNodeStepUtils.java
@@ -304,10 +304,10 @@ public class DefaultScriptFileNodeStepUtils implements ScriptFileNodeStepUtils {
             }
 
             Map<String, String> nodeAttribute = node.getAttributes();
-            if(!"false".equals(nodeAttribute.get(NODE_ATTR_ENABLE_SYNC_COMMAND))) {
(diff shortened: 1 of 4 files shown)


jtobard:
This log is confusing some users since it wasn't included in the release notes and affecting one customer with a custom plugin: https://github.com/rundeck/rundeck/issues/6449#issuecomment-694846356 maybe we need to include this on documentation too?

gschueler (approved):
@carlosrfranco please add documentation of this behavior to rundeck/docs

PR #6457, opened 2020-09-17
Notification plugin key storage

ltamaster:
enhancement: 
* adding access to key storage for notification plugins. So, authentication parameters can be loaded from key storage
* enabling key storage selector on VUE plugin form 

Notification plugins must AcceptsServices to get access to key storage services.
```
class TestNotificationPlugin implements NotificationPlugin, AcceptsServices{

        private Services services

        @Override
        void setServices(Services services) {
            this.services = services
        }

        @Override
        boolean postNotification(String trigger, Map executionData, Map config) {
            StorageTree storageTree = this.services.getService(KeyStorageTree)
            def resource = storageTree.getPassword("keys/path/sth")
            .....
        }
    }
```

**Additional context**


gschueler (approved):
@ltamaster the `AcceptsServices` should be added to the Notification plugin documentation in rundeck/docs

PR #6473, opened 2020-09-22
Fix sidebar active indicator on webhooks page

### LINKED ISSUE #6472: Nav indicator when on Webhooks points at Dashboards
Nav indicator when on Webhooks points at Dashboards:
image

ProTip:
Fixes #6472


sjrd218 (commented):
It no longer points to the wrong location, but it doesn't point to the webhook menu item. I think additional logic would need to be added to the code that lists dynamic menu items in order for this to point to the webhooks location.

If all we care about is not pointing to the wrong location, then this is sufficient.

fdevans:
It's fine if it doesn't point to Webhooks for now.  The others outside main nav don't work either.

sjrd218 (approved):

PR #6478, opened 2020-09-23
Fix ace collision

ProTip:
Fix. The brace fix previously added seems to cause issues when ace is not already included loaded via the grails app gsp pages.

Ditched brace for [ace-builds](https://github.com/ajaxorg/ace-builds), and official project from the maintainers of Ace. Ace builds is has better compatibility with our webpack setup, where as brace was designed for use with browserify.

Attempt to hack up the `requires` more..


ProTip:
Superseded by #6479

PR #6490, opened 2020-09-24
Fix #6304 where changing only the crontab would not update the scheduled execution.

### LINKED ISSUE #6304: Updating a JOB Crontab Schedule doesn't save new value
**Description:**
Cannot update schedule on the Jobs/deployments, editing an existing job after the update and changing only the year on the crontab doesn't keep the new time.

**Rundeck Details:**

* Rundeck version: 3.3.0
* install type: war
* OS Name/version: SUSE Linux Enterprise Server 12 SP5
* DB Type/version: postgres

**To Reproduce**
1. Go to Jobs -> Deployments
2. Click on the Job to edit and pick the Schedule tab
3. Change the Crontab value (in this case just the year) and save
4. New value is only kept the first time, if you edit a second time it does not keep the value.

**Expected behavior**
The new schedule should be always kept on the crontab field.

**Screenshots**
image

 - OS: Windows 10
 - Browser Chrome
 - Version 84.0.4147.89 (Official Build) (64-bit)

sjrd218:
GORM dirty checking wasn't picking up the changes unless the setter was used explicitly.

### DIFF
--- a/rundeckapp/grails-app/domain/rundeck/ScheduledExecution.groovy
+++ b/rundeckapp/grails-app/domain/rundeck/ScheduledExecution.groovy
@@ -730,7 +730,7 @@ class ScheduledExecution extends ExecutionContext implements EmbeddedJsonData {
      * of the ScheduledExecution appropriately.  Returns false if the size
      * is wrong
      */
-    def boolean parseCrontabString(String crontabString){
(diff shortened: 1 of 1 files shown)


gschueler (approved):
i was unable to verify this as a fix since I could not reproduce the issue. however these changes should be fine

PR #6494, opened 2020-09-25
Notifications on its own thread

ronaveva:
Fix for https://github.com/rundeckpro/rundeckpro/issues/1248

Option to trigger notifications on a separated thread and set a time out for it

**Qa framework with tests to check for the notification reception**
https://github.com/rundeckpro/rundeck-tester/pull/33

**Documentation for the new properties**
https://github.com/rundeck/docs/pull/670


gschueler (changes requested):
please add a feature flag to disable this behavior

ronaveva:
> 
> 
> please add a feature flag to disable this behavior

This new behavior is already behind a flag "rundeck.notifications.own.thread", want me to change this to a feature one? @gschueler

gschueler:
@ronaveva yes, it should conform to the feature flag pattern we have been using. Add an entry in Features class, and add feature flag, and any config props to the RundeckConfigBase class.

gschueler:
@ronaveva see example here https://github.com/rundeck/rundeck/pull/6527/files

gschueler (approved):
using 2 new threads per notification will work, but ideally this should use a thread pool.  you could use the TaskService perhaps although it might need some tweaking.  I think we will have to refactor the Notification task management in the future

PR #6506, opened 2020-09-29
Changing to consider logged user roles for ACL cache

carlosrfranco:
Fixes https://github.com/rundeckpro/rundeckpro/issues/1275

When using `ReloadablePropertyFileLoginModule`, if the user properties is changed on `realm.properties` file the changes have no effect without restart Rundeck because the ACL rules validation will already stored in cache.

Now, if roles changes the cache value will not be used.

**Additional context**
 Also removed some unused lines of codes


gschueler (changes requested):
no tests were added to validate the behavior.

Note: https://github.com/rundeckpro/rundeckpro/wiki/Coding-Best-Practices-Guide#testing

gschueler (approved):

PR #6515, opened 2020-10-01
Removes Rundeck preboot initialization from application.groovy 

sjrd218:
In some cases having imports from application code in the application.groovy can prevent startup. I have added a spring initializer to the test suite to do the preboot initialization instead, which fixes this issue.


gschueler (approved):

gschueler:
@Mergifyio rebase

PR #6518, opened 2020-10-02
Fix the display of Enterprise plugins when listing plugins

sjrd218:
Fix the display of Enterprise plugins when going to the "Find Plugins" page, then choosing to filter by installed and uninstalled.


ProTip:
@sjrd218 If I'm reading this correctly:
* `showWhich..` determines the display of all/installed/uninstalled plugins
* On an Enterprise install, Enterprise plugins will now only show up when viewing installed plugins

Does that sound accurate?

sjrd218:
`showWhich` is the filter: null is show all, true is show installed, false is show uninstalled. Enterprise plugins will always show up if showWhich is null.

ProTip (approved):

PR #6513, opened 2020-10-01
add feature toggle for sidebar project list 

gschueler:
**Is this a bugfix, or an enhancement? Please describe.**

Add feature toggle to disable sidebar project listing, this can speed up UI when many projects are present.

    rundeck.feature.sidebarProjectListing.enabled=false

Also adds feature toggle to enable session project list caching. The user session projects cache is enabled by default, but can be disabled:

    rundeck.feature.userSessionProjectsCache.enabled=false

cache delay can be configured, default 5 minutes:

    rundeck.userSessionProjectsCache.refreshDelay=300000

cache can be refreshed if `?refresh=true` is added to home page request URL

PR #6523, opened 2020-10-06
Project list updates

gschueler:

Extends https://github.com/rundeck/rundeck/pull/6513

more improvements to project list (home) page speed

- [x] project stats loading deferred to ajax request unless already cached
- [x] projectNamesAjax uses session cache from #6513 
- [x] update home ui to work better when everything is loaded async, and to support pro plugin behavior


ProTip:
Are we moving away from the `rundeck.gui` namespace for stuff like the paged job list and the project list? I noticed they were going into `rundeck.feature`.. The legacy execution output also went rogue..

gschueler:
@ProTip the convention we are using for feature flags uses `rundeck.feature.X.enabled`. Config values can be at a different path, but the feature itself needs to use that convention currently

gschueler:
@mergifyio rebase

ProTip (approved):

PR #6525, opened 2020-10-06
Fix the display of Enterprise plugins when listing plugins

ProTip:
Original PR from @sjrd218 :

Fix the display of Enterprise plugins when going to the "Find Plugins" page, then choosing to filter by installed and uninstalled.


ProTip:
Accidentally merged this and went through the process of backing that out to see what it was like..

1/10, would not recommend.

sjrd218 (approved):

PR #6526, opened 2020-10-06
Async startup of execution history cleaner tasks

gschueler:
Speedup app startup with many projects


rundeckapp/src/main/groovy/com/dtolabs/rundeck/app/support/ExecutionCleanerConfigImpl.groovy
```diff
@@ -0,0 +1,62 @@
+package com.dtolabs.rundeck.app.support
+
+import groovy.transform.CompileStatic
+
+/**
+ * Implementation of ExecutionCleanerConfig
+ */
+@CompileStatic
```
ProTip, line 8: Yum

ProTip (approved):

PR #6550, opened 2020-10-12
Fix loading zero projects

ProTip:
`projectNames` in `home.gsp` is not `null` with zero projects when running a  _build_.

Check if `null` **or** if `projectnames.Size() == 0`.

Checking in the javascript for it. Passing around empty lists is preferable to `null` though as it composes better, and appears to be the original intention.

**Additional context**
I could not reproduce this running in development mode; the `null` check was enough. I'm not sure why this is.. I created a build to verify the issue and then created a build with the fix to verify it was resolved..

### DIFF
--- a/rundeckapp/grails-app/views/menu/home.gsp
+++ b/rundeckapp/grails-app/views/menu/home.gsp
@@ -30,7 +30,7 @@
     <meta name="layout" content="base"/>
     <meta name="tabpage" content="home"/>
     <title><g:appTitle/></title>
-    <g:if test="${projectNames==null || projectNames.size() == 0}">
+    <g:if test="${!projectNames}">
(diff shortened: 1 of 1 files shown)


rundeckapp/grails-app/views/menu/home.gsp
```diff
@@ -30,7 +30,7 @@
     <meta name="layout" content="base"/>
     <meta name="tabpage" content="home"/>
     <title><g:appTitle/></title>
-    <g:if test="${projectNames==null}">
+    <g:if test="${projectNames==null || projectNames.size() == 0}">
```
gschueler, line 33: or using groovy "truthiness" test: 
```suggestion
    <g:if test="${!projectNames}">
```

gschueler (approved):

PR #6556, opened 2020-10-13
Update nowrunning ajax query to support wildcard value

gschueler:
Fixes https://github.com/rundeckpro/rundeckpro/issues/1298


jtobard (approved):
I tested this and forget to approve last week

Issue #5125, opened 2019-08-03
Incorrect time on job activity page


paulholden:
**Describe the bug**
See following screenshot (the job was executed at 00:15 BST):

![Rundeck time](https://user-images.githubusercontent.com/1126541/62409414-b66d0800-b5ce-11e9-8924-986654b1fbdd.png)

The _Start time_ column shows the time as 11:15:20 pm, presumably because it hasn't applied the timezone offset.

The correct time is listed three times on the page, although using three wildly different formats.

**My Rundeck detail**

* Rundeck version: 3.1.0.20190731-2
* install type: deb
* OS Name/version: Ubuntu 16.04
* DB Type/version: MySQL 5.7.27

**To Reproduce**
1. Execute a job
2. View the execution log
3. Confirm that the time(s) listed are each correct

MegaDrive68k:
Confirmed.

On 3.1.0:

![Screenshot_1](https://user-images.githubusercontent.com/51376003/62489567-0dbcd500-b795-11e9-8081-e3e56f8e42f3.png)

On 3.0.24:

![Screenshot_2](https://user-images.githubusercontent.com/51376003/62489925-11049080-b796-11e9-835f-e20ea1d9bbee.png)

ahormazabal:
Fixed timezone offset at: https://github.com/rundeck/rundeck/pull/5213

paulholden:
Thanks all, looks better in 3.1.1.20190923-1 👍

cwaltherf:
Seing different Time Formats with Rundeck 3.3.2

![image](https://user-images.githubusercontent.com/36487101/94069485-10c6ba80-fdc7-11ea-8419-bb467ac37ef6.png)

nvalenzuela20:
Tested @cwaltherf

PR #6559, opened 2020-10-14
Show date formatted correctly in the GUI #5125

nvalenzuela20:
Fix: https://github.com/rundeck/rundeck/issues/6564

The start date, complete date and tooltip were being printed on the page without the standard format
image
image

Page rendering was modified to print the timestamp formatted correctly

image
image

### DIFF
--- a/rundeckapp/grails-app/views/execution/_wfstateSummaryLine.gsp
+++ b/rundeckapp/grails-app/views/execution/_wfstateSummaryLine.gsp
@@ -69,7 +69,7 @@
                 <span class="timerel execution-completed text-secondary" data-bind="visible: completed()">
 
                     <g:message code="at"/>
-                    <span data-bind="text: formatTimeAtDate(endTime()), attr: {title: endTime() }">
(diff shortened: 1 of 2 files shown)


ProTip:
Seems like we are dropping quite a bit of information(date, seconds) with the format change on start/end. The tooltip change to match the same format seems to reduce its usefulness.

jtobard (changes requested):
I agree with @ProTip, customers have had problems with the reduction of information in the dates, the seconds is an important part in some workflows

ProTip:
@jtobard It feels like the linked issue may have been re-opened under a different problem that is more of a feature enhancement. The original issue was for the time being incorrect, and the poster flagged these times as being correct, but it was re-opened for a consistency problem; the times are correct but one is ISO 8601 in UTC and the other shorter local time.

jtobard:
@ProTip ya I agree it should be another issue. 
But is still an issue for the customer point of view since he only see different times for the same execution, it may need more information ont he same gui on why one is using UTC, even a `(UTC)` at the end of the tooltip would help

PR #6567, opened 2020-10-15
Sending an event to cleanup acl caches If cluster mode is active

carlosrfranco:
Fixes https://github.com/rundeckpro/rundeckpro/issues/1283

If the acl policy is changed on a instance, the others instances on cluster will keep the old policy on cache and the changes will not affect them.

A new event will be triggered if acl policy changes so that all instances on a cluster have their caches cleaned

**Additional context**
The DbStorageService was changed to refresh the resource after it is obtained from the database because the object returned did not contain the changes that had just been made in the db.

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/AuthorizationService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/AuthorizationService.groovy
@@ -27,12 +27,14 @@ import com.dtolabs.rundeck.core.authorization.AuthorizationUtil
 import com.dtolabs.rundeck.core.authorization.ValidationSet
 import com.dtolabs.rundeck.core.authorization.providers.*
 import com.dtolabs.rundeck.core.config.Features
+import com.dtolabs.rundeck.server.AuthContextEvaluatorCacheManager
 import com.google.common.cache.CacheBuilder
 import com.google.common.cache.CacheLoader
 import com.google.common.cache.LoadingCache
 import com.google.common.util.concurrent.Futures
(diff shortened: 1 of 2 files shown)


rundeckapp/grails-app/services/rundeck/services/AuthorizationService.groovy
```diff
@@ -369,4 +378,20 @@ class AuthorizationService implements InitializingBean{
...
+        if (frameworkService.isClusterModeEnabled()) {
+            sendAndReceive(
+                    'cluster.clearAclCache',
+                    [
+                            uuidSource: frameworkService.getServerUUID(),
+                            path: path
+                    ]
+            ) { resp ->
+                log.debug("Cleaning the cache in the cluster is ${resp.clearCacheState}: ${resp.reason}")
+            }
```
gschueler, line 392: it seems it should still call `cleanCaches(path)` in cluster mode

gschueler (changes requested):

rundeckapp/grails-app/services/rundeck/services/AuthorizationService.groovy
carlosrfranco: @gschueler cleanCaches method will be called by all active instance on cluster so I didn't think it necessary to include this call there as I would call it twice. perhaps the sender should not receive the message to execute this method via messaging service?

gschueler: core would still need it to be called

gschueler (approved):

PR #6568, opened 2020-10-16
performance: Update ACL list GUI pages to have asynch behavior

gschueler:

* [x] Project and System ACL GUI pages  load ACL policy metadata asynchronously on demand instead of ahead of time
* [x] Fix a bug: no pagination shown in Project ACL page

* only load list of ACL names ahead of time
* defer metadata loading about ACLs (validation, structure) to on-demand when paged into view

**Additional context**

Fix very slow loading of system ACLs page with many thousands of ACLs.


sjrd218 (approved):
Very fast loading with 2000 system acls. Pagination seems to work.

PR #6573, opened 2020-10-20
Adds an option to run referenced jobs using child's node set

### LINKED ISSUE #6572: Request to run referenced Jobs on child nodes
Request to add an option to execute referenced jobs against their own node set instead of the parent's node set.

See: https://github.com/rundeckpro/rundeckpro/issues/1189

jtobard:
fix #6572 

Edit Jobref

With this new option you can select to use the child node set instead of the one from the parent job.
If the parent  job is not from another project has no  effect.


rundeckapp/src/test/groovy/rundeck/ExecutionServiceSpec.groovy
```diff
@@ -5518,4 +5532,47 @@ class ExecutionServiceSpec extends Specification implements ServiceUnitTest<Exec
...
+
+    def "use referenced job project if enabled"() {
+        given:
+
+        def orgProject = 'prgProj'
+        def jobProj = 'testproj'
+        service.frameworkService = Mock(FrameworkService) {
+            0 * filterNodeSet(null, orgProject)
+            1 * filterNodeSet(null, jobProj)
+            1 * filterAuthorizedNodes(*_)
```
gschueler, line 5544: could verify correct project is used as parameter to filterAuthorizedNodes

gschueler (approved):

gschueler:
@jtobard one conflict to merge

PR #6574, opened 2020-10-20
Fixing GUI to allow select a node clicking at arrow link on matched nodes popup

carlosrfranco:
Fixes https://github.com/rundeckpro/rundeckpro/issues/1176

An error in the GUI was preventing the selection of a node to apply to the filter from the popup that is displayed in the matched nodes field 

Some attributes was on wrong places and it was avoinding the string values to be applied on filter inputs


jtobard (approved):
Tested LGTM.
Let's add a QA framework task to add a test for this

PR #6577, opened 2020-10-21
DB migration plugin to streamline upgrade procedures

imad6639:
Implement usage of the grails database-migration plugin.  When rundeck is started, the gorm entities are diffed against the database schema, creates a diff file and runs the diff file to bring the database in sync with the entities.  
The property variable:

> dataSource.dbCreate = none

 should be set in order not to run the gorm db update.
The property variable:

> grails.plugin.migration.updateOnStart = true

should be set to use the db migration scheme.


imad6639:
changed approach and opened up a new pullrequest

PR #6584, opened 2020-10-26
EventStore and webhooks debugging

ProTip:
## Feature
The bulk of this PR adds a general purpose event storage service that can be used by various internal subsystems to store and retrieve JSON serializable objects.

* Supports Jackson annotations; uses Jackson to serialize objects
* Efficient lookup and bulk deleting via criteria builder
  * Date ranges for lookup and efficient cleanup!
* Added `SimpleServiceProvider` to core so each addon does not need to re-implement
* Added feature service interface to core

## Service interface

The interface to the service is currently very slim:
* `void storeEvent(Event)`
* `void storeEventBatch(List<Event>)`
* `EventQueryResult queryEvent(EventQuery)`
* `EventStoreService scope(...)`

For queries the result will contain the total number of records and list of events depending on the query type(COUNT, SELECT, DELETE):

```
@Data
public class EventQueryImpl implements EventQuery {
    String projectName;
    String subsystem;
    String topic;
    String objectId;
    Date dateFrom;
    Date dateTo;
    Integer maxResults = 20;
    Integer offset = 0;
    EventQueryType queryType = EventQueryType.SELECT;
}
```

```
EventQueryResultImpl implements EventQueryResult {
    Long totalCount;
    List<? extends Event> events;
}
```

### COUNT
Runs a select query for count, response object contains count and empty event list.

### Deleting
Submit a query of type `DELETE` and the response object will have rows deleted as `totalCount` and an empty event list.

### Scoping
A call to `scope` can be made with event and query **templates**. This will return a new service in which calls to `store` and `query` will have their object properties overwritten with the template properties. This allows a "scoped" service to be passed to a consumer limiting its access to the store.

This could be expanded in the future to support more advanced `contains` and topic path scopes.

## Service Impl
Currently there is one implementation and backend; a grails service that persists the events to the db via GORM: `GormEventStoreService`.

### JPA
Indexing is not finalized however a composite index may be the most efficient "primary" index. Grails GORM has had an issue forever(as far as I can tell) that will only allow creating composite indexes in alphabetical column order :| GORM added JPA and bean validation annotation support a while back. The domain object currently utilizes these annotations instead of the GORM mapping and constraint builders.

### GOJOd POJOs
Java classes are extended in Groovy in a number of places with empty subclasses: `class Wee exends Impl {}`. This pattern is used to capture the Groovy and Grails pixie dust(map constructors, data binding for command objects, etc).

## Webhooks changes
Webhooks have been updated to:
* Add a scoped event store to the webhook context if the `eventStore` feature is enabled
* Attach the webhook UUID to the webhook data passed into plugins
* Pass the full webhook plugin UI data into the plugin UI so it can access the UUID and other general information


ProTip:
@gschueler I could use some feedback on the implementation in the commit to support scoped access to the event store.

rundeckapp/grails-app/services/rundeck/services/GormEventStoreService.groovy
```diff
@@ -65,13 +64,15 @@ class EventStoreService implements com.dtolabs.rundeck.core.event.EventStoreServ
...
                     order('lastUpdated', 'desc')
+                    order('sequence', 'desc')
                 }.list(max: query.maxResults, offset: query.offset)
 
                 return new EvtQueryResult(
                     totalCount: count,
                     events: events
                 )
             case EventQueryType.DELETE:
+                println(query.maxResults)
```
gschueler, line 75:
```suggestion
```

gschueler (approved):

rundeckapp/grails-app/conf/spring/resources.groovy
```diff
@@ -197,7 +198,8 @@ beans={
             (ClusterInfoService)         : ref('clusterInfoService'),
             (ApiInfo)                    : ref('rundeckApiInfoService'),
             (ExecutionFileManagerService): ref('logFileStorageService'),
-            (ResourceFormats)            : ref('pluginService')
+            (ResourceFormats)            : ref('pluginService'),
+//            (EventStoreService)          : ref('eventStoreService')
```
gschueler, line 202: should clean up

PR #6586, opened 2020-10-27
Add ability to mark a schedule as missed.

sjrd218:
Add the ability to create an entry in the execution and exec report tables for a job that did not execute at its scheduled time.

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
@@ -1406,6 +1406,7 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
     public static String EXECUTION_FAILED_WITH_RETRY = "failed-with-retry"
     public static String EXECUTION_STATE_OTHER = "other"
     public static String EXECUTION_SCHEDULED = "scheduled"
+    public static String EXECUTION_MISSED = "missed"
 
     public static String ABORT_PENDING = "pending"
     public static String ABORT_ABORTED = "aborted"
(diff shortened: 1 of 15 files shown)


rundeckapp/grails-app/views/execution/mailNotification/_newStatus.gsp
```diff
@@ -784,6 +797,7 @@
     </td>
     </tr>
   </table>
+        </g:else>
```
gschueler, line 800: was this a missing close tag?

gschueler (approved):

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -766,6 +766,9 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
                 isNull('status')
                 ne('status', EXECUTION_SCHEDULED)
             }
+            and {
```
gschueler, line 769: `cleanupRunningJobsAsync`, and `findRunningExecutions` both do the same query, the code should probably be unified to build criteria builder in a single method.  Also, the findRunningExecutions query should be tested

gschueler (changes requested):

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
sjrd218: I've removed the change to that query. When I first added missed executions I didn't set the dateCompleted and so it would cause the cleanup to pick up the missed records and change them. 

As an aside I'm wondering if we could remove the duplication by changing the `cleanupRunningJobsAsync` to be:
```
def cleanupRunningJobsAsync(String serverUUID = null, String status = null, Date before = new Date()) {
    executorService.submit {
            cleanupRunningJobs(findRunningExecutions(serverUUID, before), status)
    }
}
```
Seems odd to look up the ids on the main thread then use the ids to lookup the objects in a different thread.

gschueler: yes that seems better

gschueler (approved):

Issue #6582, opened 2020-10-26
Request to use server time on ISO 8601 dates on activity page


nvalenzuela20:
Since the upgrade to 3.3.2 (unsure if it was a problem before this version) the rundeck UI displays certain UI elements in the UTC while others are displayed in the local timezone of the server.

Expected behavior is to have all elements using the local timezone of the server

![image](https://user-images.githubusercontent.com/2894508/96730325-6a051800-138c-11eb-99fe-7285e27d8a77.png)

jbanda15:
I have the same request with another customer.
They need to see the local timezone too because they have users in different regions and timezone.

- Status Time
![Status Time](https://user-images.githubusercontent.com/45765283/118048056-b4121300-b349-11eb-93b9-2c71d0c7ab0b.png)

- Start and End Time
![Start-End Time](https://user-images.githubusercontent.com/45765283/118048106-cb510080-b349-11eb-82b7-0cd351f9d3b9.png)

-Target Node Time
![Target Node Time](https://user-images.githubusercontent.com/45765283/118048139-d73cc280-b349-11eb-8d20-810e1110d626.png)

They want to have all elements using the local timezone of the server

Ticket: https://rundeck.lightning.force.com/lightning/r/Case/5006e00001Ro1uXAAR/view

tazou:
Hello,
I have the Rundeck 3.3.11.20210507-1 version on Debian 9, and I have the same problem.
On the web interface, I see "Zulu" timezone for jobs.

Example : 
![image](https://user-images.githubusercontent.com/6482957/118820943-0a85d100-b8b7-11eb-81d1-de676f0abfd3.png)

I tried to put this option on the java command line : 
-Duser.timezone=""
And a second test with : 
-Duser.timezone=Europe/Paris

Each time, I restarted the rundeckd service to try, but the problem is still here.

My Debian system is in CEST like this : output of "date" command : Wed May 19 15:32:07 CEST 2021
And I have "Europe/Paris" in the file /etc/timezone 

Thanks.

lcanavosio:
Another client requesting the same feature
Case: 83536
Client: Transunion
Version: Enterprise 3.4.10

lcanavosio:
Another client requesting the same feature
Case: 83536
Client: Transunion
Version: Enterprise 3.4.10
Currently it displays the time of the Browser's host, where the user is logged in.
[...]

PR #6587, opened 2020-10-27
RUN-18: Show date in ISO 8601 format in the GUI #6582

nvalenzuela20:
Fix: https://github.com/rundeck/rundeck/issues/6582

The start date, complete date and tooltip were being printed on the page without the date in ISO 8601 format

image
image

Page rendering was modified to print the timestamp the date in ISO 8601 format correctly

image
image


ProTip (commented):
With `Z` being dropped how would the user know what the absolute time is? ie:
* The server is US Central TZ but the user is in California(PST); would they be able to figure out the absolute(UTC) time?

Changing these times from UTC to a server local time may be very surprising to many users(particularly in such a subtle way as the `Z` being dropped). At a minimum it may be need to be behind a feature flag or configuration; @gschueler ?

nvalenzuela20:
@ProTip 

 I have been assuming the following about timestamps:
- that the timestamps on the database are stored in local server time
- Rundeck on the server is correctly reading this and obtaining a UTC value

Using that, the proposed fix changes the formatting on the page so that timestamps appear in the timezone of the browser that is rendering the page. This is done by passing a UTC timestamp from the server and then parsing and formatting to local time on the client browser. Am I understanding correctly how are timestamps handled by Rundeck?

Yes, I agree that the formatting is confusing, would it be ok then if I change all the timestamp formatting on this page so that its clear what is the timezone that was used to format the time? It could look like this:

2020-10-12T18:37:47 -03:00

ProTip (commented):
I believe there may be some confusion over the requirements and behavior. The issue states that there is a need for the time to be displayed in the _servers_ local time.

The taglib addition `formatTimeInLocalTimeZone` will actually render the time in the servers TZ, however this value gets replaced by the knockout binding.

The knockout code will format the time in the _users_ local time, not the servers local time.

I believe we may need to shape up the following:
* What sort of configuration this should go behind
* What local time the time should be formatted in per the original ask(the servers TZ or the users TZ)
* The format for the time in the UI

fdevans:
Based on the open questions for this I'm going to defer it for 3.3.7

fdevans:
Since this PR does not fix the issue and the original PR was done on version 3.3.x I'm going to close this and we will revisit #6582 in the future.

PR #6590, opened 2020-10-27
Issue/1321 Activity lists not showing on Job page when using oracle database

imad6639:
Fixed issue of jobs not showing due to oracle returning empty activity lists.  Oracle treat not equal '' as null, and when used in createCriteria alway returns no results.  Made this criteria conditional if Oracle is used.

The alternatives I tried to use were the isNotEmpty, and the sizeGt methods in org.grails.datastore.mapping.query.api.Criteria interface. isNotEmpty required a list, and sizeGt would throw an exception


rundeckapp/grails-app/services/rundeck/services/ReportService.groovy
```diff
@@ -560,4 +563,9 @@ class ReportService  {
 
         return frameworkService.authorizeProjectResources(authContext,resHS, constraints, project)
     }
+
+    private boolean isOracleDatasource(){
```
sjrd218, line 567: Did you check if this technique works with JNDI data sources? Some of our Oracle clients use JNDI.

sjrd218 (approved):

PR #6594, opened 2020-10-29
new sshj plugin

ltamaster:
enhancement: adding the new SSHJ plugin 
https://github.com/rundeck-plugins/sshj-plugin

Node Executor/ File Copier (based on SSHJ lib) as an alternative of JSCH plugin 

**Additional context**
* It has the same configuration name attributes as the JSCH plugin, so this [documentation](https://docs.rundeck.com/docs/administration/projects/node-execution/ssh.html) works for this plugin 

* sudo authentication is experimental


gschueler (approved):
tested private key file and storage

PR #6608, opened 2020-11-04
Fixing multivalued options validation with no errors but that appear as invalid when another option (not multivalued) is invalid

carlosrfranco:
Fixes https://github.com/rundeckpro/rundeckpro/issues/1324

When a job has options and multivalued options with validations, if the validation of the non-multivalued options fails, all validations for the other multivalued options also fail.

was made a change on validation so If the option validation fail, it should not impact to the other options validations

**Additional context**
When multivalued option validation fail, the GUI it was breaking duplicating the options fields. A fix was made on javascript


imad6639 (approved):

carlosrfranco:
Moving this PR to this one: https://github.com/rundeck/rundeck/pull/6641

Issue #5925, opened 2020-03-31
RUNDECK_MAIL_PROPS option don't work as expectation


bienkma:
I'm using docker-compose for launching a rundeck instance. Also, I wanna setup notification by Gmail but when I set environment RUNDECK_MAIL_PROPS don't work as document (It's ready worked in groovy config file - worked same my account Gmail. But I don't want to use Groovy file config for production). My docker-compose.yml is below:
```
---
version: '3.1'
services:
  rundeck:
    restart: always
    image: rundeck/rundeck:3.2.4
    ports:
      - 4440:4440
    depends_on:
      - db
    container_name: rundeck
    environment:
      - TS="Asia/Ho_Chi_Minh"
      - RUNDECK_SERVER_FORWARDED=true
      - RUNDECK_SERVER_ADDRESS=0.0.0.0
      - RUNDECK_DATABASE_URL=jdbc:mysql://db/rundeck?autoReconnect=true&useSSL=false
      - RUNDECK_DATABASE_USERNAME=root
      - RUNDECK_DATABASE_PASSWORD=change_me_now
      - RUNDECK_GRAILS_URL=http://10.110.64.20:4440
      - SERVER_URL=http://10.110.64.20:4440
      - RUNDECK_SECURITY_SYNCLDAPUSER=true
      - SETTING_RUNDECK_FORWARDED=true
      - RUNDECK_DATABASE_DRIVER=org.mariadb.jdbc.Driver
      - RUNDECK_LOGGING_STRATEGY=CONSOLE
      - RUNDECK_MAIL_SMTP_HOST=smtp.gmail.com
      - RUNDECK_MAIL_SMTP_PORT=587
      - RUNDECK_MAIL_SMTP_USERNAME=rundeck@lab.vn
      - RUNDECK_MAIL_SMTP_PASSWORD=change_me_now
      - RUNDECK_MAIL_FROM=rundeck@lab.vn
      - RUNDECK_MAIL_PROPS=["mail.smtp.auth":"true","mail.smtp.port":"587","mail.smtp.starttls.enable":"true","mail.smtp.socketFactory.port":"587","mail.smtp.socketFactory.fallback":"false"]
    volumes:
      - /home/bienkma/config/jaas-ldap.conf:/home/rundeck/servers/config/jaas-ldap.conf:ro
      - ./etc/admin.aclpolicy:/home/rundeck/etc/admin.aclpolicy:ro
[...]

PR #6613, opened 2020-11-06
Emit RUNDECK_MAIL_PROPS as groovy to fix SMA email auth

craigminihan:
A fix for issue #5925. As noted by the issue author and other contributors RUNDECK_MAIL_PROPS does not function as described.

Emit the value of RUNDECK_MAIL_PROPS in `server/config/rundeck-config.groovy` rather than the properties file.

This is an implementation of a suggested solution in the comment stream of #5925 and https://stackoverflow.com/questions/30601574/rundeck-gmail-smtp-not-working.

**Additional context**
RUNDECK_MAIL_* with the exception of PROPS are still emitted to the properties file. Only PROPS is emitted to the groovy file.

### DIFF
--- a/docker/official/lib/entry.sh
+++ b/docker/official/lib/entry.sh
@@ -35,9 +35,11 @@ echo "rundeck.server.uuid = ${RUNDECK_SERVER_UUID}" > ${REMCO_TMP_DIR}/framework
 
 # Combine partial config files
 cat ${REMCO_TMP_DIR}/framework/* >> etc/framework.properties
-cat ${REMCO_TMP_DIR}/rundeck-config/* >> server/config/rundeck-config.properties
+cat ${REMCO_TMP_DIR}/rundeck-config/*.properties >> server/config/rundeck-config.properties
 cat ${REMCO_TMP_DIR}/artifact-repositories/* >> server/config/artifact-repositories.yaml
 
+# Copy the groovy config files
+cp ${REMCO_TMP_DIR}/rundeck-config/*.groovy server/config/
 
 # Store settings that may be unset in script variables
 SETTING_RUNDECK_FORWARDED="${RUNDECK_SERVER_FORWARDED:-false}"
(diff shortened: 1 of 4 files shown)


docker/official/lib/entry.sh
```diff
@@ -35,9 +35,11 @@ echo "rundeck.server.uuid = ${RUNDECK_SERVER_UUID}" > ${REMCO_TMP_DIR}/framework
 
 # Combine partial config files
 cat ${REMCO_TMP_DIR}/framework/* >> etc/framework.properties
-cat ${REMCO_TMP_DIR}/rundeck-config/* >> server/config/rundeck-config.properties
```
craigminihan, line 38: Limit the concatenation to just the properties files

docker/official/lib/entry.sh
```diff
@@ -35,9 +35,11 @@ echo "rundeck.server.uuid = ${RUNDECK_SERVER_UUID}" > ${REMCO_TMP_DIR}/framework
...
+cat ${REMCO_TMP_DIR}/rundeck-config/*.properties >> server/config/rundeck-config.properties
 cat ${REMCO_TMP_DIR}/artifact-repositories/* >> server/config/artifact-repositories.yaml
 
+# Copy the groovy config files
+cp ${REMCO_TMP_DIR}/rundeck-config/*.groovy server/config/
```
craigminihan, line 42: Copy the new groovy config file

docker/official/remco/templates/rundeck-config.groovy
```diff
@@ -0,0 +1,7 @@
+{%- if exists("/rundeck/mail/props") %}
```
craigminihan, line 1: Conditionally emit `/rundeck/mail/props`

ProTip:
Hello,

Please see the comment in the attached issue https://github.com/rundeck/rundeck/issues/5925#issuecomment-729843044 . Going to close this for now as it's not clear this is necessary to resolve the issue based on our past experience running STARTTLS through the `RUNDECK_MAIL_PROPS` and have not heard back on the issue.

PR #6628, opened 2020-11-17
Enable Project based ACLs for Key Storage under keys/project/<NAME>/ paths

gschueler:

first WIP for adding project ACLs for key storage under `keys/project/NAME/` subpaths.


core/src/main/java/com/dtolabs/rundeck/core/storage/AuthRundeckStorageTree.java
```diff
@@ -58,13 +58,23 @@ public AuthRundeckStorageTree(StorageTree storageTree, AuthStorageContextProvide
...
+        boolean authorized = false;
+        Set<Attribute> environments = contextProvider.environmentForPath(path);
+        for (Attribute env : environments)
+        {
+            Decision evaluate = context.evaluate(
+                    resourceForPath(path),
+                    action, Collections.singleton(env)
+
+            );
+            if (evaluate.isAuthorized()){
```
gschueler, line 72: The behavior should also check if any of the decision.explain().getCode() is a REJECTED_DENIED state. in that case, the access should be rejected even if the other decision is authorized

gschueler (commented):
auth evaluation should handle DENIED result

gschueler:
1. the feature flag should also disable the way the authorization checks work for project, i think perhaps you could use a different implementation of AuthStorageContextProvider injected in the spring context if the flag is disabled
2. the sidebar menu has changed, i'm not sure how you will want to add the new link (@ProTip )

imad6639:
> 1. the feature flag should also disable the way the authorization checks work for project, i think perhaps you could use a different implementation of AuthStorageContextProvider injected in the spring context if the flag is disabled
> 2. the sidebar menu has changed, i'm not sure how you will want to add the new link (@ProTip )

Should be addressed

core/src/main/java/com/dtolabs/rundeck/core/storage/KeyStorageContextProvider.java
```diff
@@ -0,0 +1,56 @@
...
+     * @param path path
+     *
+     * @return authorization environment: a project environment if the path matches /project/name/*, otherwise the
+     *         application environment
+     */
+    @Override
+    public Set<Attribute> environmentForPath(Path path) {
+        String[] paths = path.getPath().split("/");
+        if (paths != null && paths.length > 2 && paths[0].equals(PROJECT_PATH_COMPONENT)) {
+            return AuthorizationUtil.projectContext(paths[1]);
```
gschueler, line 50: shouldn't the environment always be the RUNDECK_APP_ENV for KeyStorageContextProvider ?

gschueler:
Does the Vue implementation of the key storage browser (http://github.com/rundeck/rundeck/blob/main/rundeckapp/grails-spa/packages/ui-trellis/src/components/plugins/KeyStorageSelector.vue) needs similar updates?

PR #6632, opened 2020-11-19
Enhance missing schedule UI and info

sjrd218:
Enhances the missing schedule UI.

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
@@ -70,6 +70,7 @@ import org.hibernate.JDBCException
 import org.hibernate.StaleObjectStateException
 import org.hibernate.criterion.CriteriaSpecification
 import org.hibernate.type.StandardBasicTypes
+import org.quartz.CronExpression
 import org.rundeck.app.components.jobs.JobQuery
 import org.rundeck.core.auth.AuthConstants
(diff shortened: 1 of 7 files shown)


rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -4246,4 +4252,33 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
             )
         }
     }
+
+    String createMisfireReport(ScheduledExecution scheduledExecution, Date misfireTime) {
```
sjrd218, line 4256: @gschueler should the report be a string or structured data that is rendered at view time? It seems the ExecReport message indicates a string, but if you have a better way I'll see if I can make it work.

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -4246,4 +4252,33 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
...
+        String crontabExp = scheduledExecution.generateCrontabExression()
```
gschueler, line 4257: I think instead of doing this we need to ask the jobSchedulesService.  there is currently a `nextExecutionTime` method, but we could add a `executionTimesBetween` method or something.  The reason is the schedule may not be defined on the job itself

gschueler (approved):

PR #6658, opened 2020-12-04
Fix/job life cycle node list

ronaveva:
This is a bug fix for issue https://github.com/rundeckpro/rundeckpro/issues/1409

Issue: Previously when using job life cycle plugins, the node list was only passed to the plugin implementation only if the user had read and run permissions on the nodes.

Solution: It no longer requires user node permissions to send the matches nodes to the method "beforeSaveJob" from JobLifecyclePlugins.


gschueler (changes requested):
this code is relying on the behavior of FrameworkService.filterAuthorizedNodes returning the original list if the actions is null. It would make more sense to simply call frameworkService.filterNodeSet to get the node list if you do not want to filter the nodes by authorization.

gschueler (approved):

PR #6662, opened 2020-12-07
fix an issue with notification plugin + dynamic properties.

ltamaster:
bug fix

fix an issue with notification plugin + dynamic properties.
When the dynamic property plugin has an exception (for example a key is not found in the storage), the transaction is rolling back and the references fields from scheduledExecution object are lost, which produces an error in the edit form.

call `listNotificationPluginsDynamicProperties` with a new session

**Additional context**


rundeckapp/src/test/groovy/rundeck/controllers/ScheduledExecutionControllerSpec.groovy
```diff
@@ -2944,4 +2944,79 @@ class ScheduledExecutionControllerSpec extends HibernateSpec implements Controll
...
+        controller.rundeckAuthorizedServicesProvider = Mock(AuthorizedServicesProvider)
+        controller.optionValuesService = Mock(OptionValuesService){
+            getOptions('AProject', 'test',_) >> {throw new Exception ("error calling plugin")}
+        }
+
+        def notificationPlugins = ["notification-plugin": ["value1": "label1"]]
+
+        controller.notificationService = Mock(NotificationService) {
+            1 * listNotificationPlugins() >> [:]
+            1 * listNotificationPluginsDynamicProperties("AProject", _) >> notificationPlugins
```
gschueler, line 2991: was the error caused when an exception was thrown in the listNotificationPluginsDynamicProperties method? should the test reflect that?

ltamaster: yes you right, but the problem didn't happen because listNotificationPluginsDynamicProperties returned an exception, it happened when an exception was triggered inside that method (but handled ). Can I reproduce that in the test?

gschueler (approved):
I was not able to reproduced the issue, so I could not confirm the fix. however the code looks good to me

PR #6685, opened 2020-12-17
Updating mode-yaml.js to later version and fixing to handle a variable when undefined.

carlosrfranco:
Fixes https://github.com/rundeckpro/rundeckpro/issues/1432

The javascript that handle yaml format on editor has a bug when typed a character like a pipe ` | ` is causing an error when calling a function from an undefined variable.
 
In addition to updating the mode-yaml.js script to the latest version, I made a small change to handle the variable if it is undefined. `r = r.replace` to `r = r? .replace`


ltamaster:
@carlosrfranco 
I tested it and it worked, I am wondering if we need to upgrade the full ace library

carlosrfranco:
@ltamaster 
do you mean the ace-builds on package.json?

ltamaster:
I mean the files on https://github.com/rundeck/rundeck/blob/main/rundeckapp/grails-app/assets/javascripts/ace
for example ace.js

what do you think @ProTip ?

ProTip:
@carlosrfranco @ltamaster Is there a way to resolve the issue without modifying the Ace code? It's really hard to verify the source and integrity of these vendored files even if they haven't been changed :) It's also hard to review and reason about exactly what the function does and what the implications of modifying it are..

Also, I believe that optional chaining operator is not the best choice; support was recently added to browsers and I know some users are on extended support versions of Chrome and Firefox that it won't work with.

ProTip:
RE: Upgrade: It might be a good idea to upgrade it all. We could do it from an ace-builds tag, note the tag in the README, and that way would have a way to verify the source of the files and their integrity.

carlosrfranco:
@ProTip Yes, I agree, modifying a third party file is not the best way but I don't know if it is possible to solve it in another way. It is worth investigating. If the only possible fix for this issue is in the code of vendored files, perhaps the other option would be to create a PR in their repository and wait for a new version?

ProTip:
UPDATE: I think there may be something else going on here as it works fine with their online demo:

image

ProTip:
@ltamaster @carlosrfranco Upgrading `ace.js` and `ext-language_tools.js` appears to work without modifying the code. I think maybe we should go ahead and upgrade..

ltamaster:
sounds good to me, I tested the upgrade too and worked.

carlosrfranco:
@ltamaster @ProTip done. Can you try it now?

ProTip:
Can we upgrade all the modes/files?

carlosrfranco:
> Can we upgrade all the modes/files? I think that may be best to avoid any incompatibilities.

It's true... I'm going to upgrade all this files

carlosrfranco:
@ProTip It's done. Must be well tested to see if there was an impact elsewhere on the Rundeck

ProTip:
@ltamaster Looks like the indentation was off a bit causing the shasum not to match. I've updated them, verified the shasums, and updated the README to include the Ace Build version. Could you approve if it all looks good to you?

ltamaster (approved):
tested

PR #6692, opened 2020-12-22
Migrate auth call from framworkService to new acl classes in repository controller.

sjrd218:
Fixes auth call using frameworkService.


gschueler:
You can use AuthContextProcessor interface for a single bean “rundeckAuthContextProcessor”

gschueler (approved):

PR #6694, opened 2020-12-23
Pro property loader wasn't decrypting property value

sjrd218:
Some property file loaders require the get method to be called to transform the value data.

This fixes an issue where the property value was not being decrypted by the encrypted property loader in pro.
https://github.com/rundeckpro/rundeckpro/issues/1456


jtobard (approved):
tested and LGTM

PR #6701, opened 2021-01-04
Changes to allow to impor nodes only separately from Configurarion

carlosrfranco:
Fixes https://github.com/rundeckpro/rundeckpro/issues/1379

It can be considered an enhancement to allow the import (or not) of the node sources can be defined by a separate option from the Project Configurations.

A new option(parameter) was added to import project to define if node sources will be imported (on previous version the current behaviour should be kept).


rundeckapp/grails-app/services/rundeck/services/ProjectService.groovy
```diff
@@ -1077,9 +1079,18 @@ class ProjectService implements InitializingBean, ExecutionFileProducer, EventPu
 
                 'files/' {
                     'etc/' {
-                        if (importConfig) {
+                        if (importConfig || importNodes) {
                             'project.properties' { path, name, inputs ->
                                 configtemp = copyToTemp()
+                                String resourceConfig = ""
+                                configtemp?.text?.eachLine {
+                                    if(it.contains("resources.source")){
```
gschueler, line 1087: I think it would be better to exclude properties starting with `resources.source` after parsing it.  This check is not exact

gschueler (changes requested):

rundeckapp/grails-app/services/rundeck/services/ProjectService.groovy
```diff
@@ -1442,9 +1432,16 @@ class ProjectService implements InitializingBean, ExecutionFileProducer, EventPu
         def newprops = new Properties()
         newprops.putAll(map)
 
-        if(propertiesToMerge){
-            project.mergeProjectProperties(newprops, propertiesToMerge)
-        } else {
+        Properties propResourcesSource = newprops.findAll {String key, String value ->
+                key.contains("resources.source.")
```
gschueler, line 1436:
```suggestion
                key.startsWith("resources.source.")
```

gschueler (changes requested):

gschueler:
need documentation pr for rundeck/docs to document api version update

gschueler (approved):

PR #6723, opened 2021-01-12
Fix to get schedules both via addon (enterprise-schedules) and via job scheduling (job edit page)

carlosrfranco:
Fixes https://github.com/rundeckpro/rundeckpro/issues/1442

The API to get scheduled jobs was returning only jobs scheduled from job edit page but not the ones scheduled by enterprise-scheduled addon.

And the API to get job forecast, the field `futureScheduledExecutions` was being filled only with schedules via addon (if there was one where the job was linked)

The method of the controller was changed to get all scheduled jobs from jobSchedulesService and was created another PR (https://github.com/rundeckpro/enterprise-schedules/pull/63) to fix the addon so that the API to get job forecast returns the future scheduled executions correctly

### DIFF
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
@@ -3163,7 +3163,7 @@ class MenuController extends ControllerBase implements ApplicationContextAware{
         }
 
 
-        def list = jobSchedulesService.getAllScheduled(uuid, null)
+        def list = jobSchedulesService.getAllScheduled(uuid)
         //filter authorized jobs
         Map<String, UserAndRolesAuthContext> projectAuths = [:]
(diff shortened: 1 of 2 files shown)


rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
```diff
@@ -3163,7 +3163,7 @@ class MenuController extends ControllerBase implements ApplicationContextAware{
         }
 
 
-        def list = ScheduledExecution.findAllByServerNodeUUIDAndScheduled(uuid,true)
+        def list = jobSchedulesService.getAllScheduled(uuid, null)
```
imad6639, line 3166: I don't think you need to add in null as the second parameter since the stub shows
getAllScheduled(String serverUUID = null, String project = null)

imad6639 (approved):

Issue #6693, opened 2020-12-23
Slow jobs list/job show page loading with many projects


gschueler:
**Describe the bug**
Very slow loading of (grouped) jobs list page and job show page when there are many (2000+) projects.

It is caused by evaluating authorization for every project to see if Job Create action is allowed, so that the "Copy Job to Other Project..." action modal can list available projects.

fix: do not check authorizations until the Copy Job to Other Project action is selected

**My Rundeck detail**

* Rundeck version: 3.3 and earlier

**To Reproduce**
1. create many projects (2000+)
2. Click on Project to go to grouped jobs list by default OR click a job name to view a job
3. Wait a long time if no authorizations have been cached

**Expected behavior**
page loads faster

PR #6725, opened 2021-01-13
Issues/6693 slow job list show page

imad6639:
Fixes issue #6693 slow job list very slow jobs list page and job show page when there are many (2000+) projects.  This is an enhancement and due to the project names which the user has create authorization for to be returned to the page.  This job name list is only needed when Duplicate this job to another project selection is clicked.  Solution is to use an ajax call when this selection is chosen.


rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
```diff
@@ -2469,6 +2450,29 @@ class MenuController extends ControllerBase implements ApplicationContextAware{
         )
     }
 
+    def authProjectsToCreateAjax() {
+        AuthContext authContext = rundeckAuthContextProcessor.getAuthContextForSubjectAndProject(
```
gschueler, line 2454: I think there is no need for "getAuthContextForSubjectAndProject", probably just "getAuthContextForSubject", since this endpoint does not require a project parameter

imad6639: changed it to only retrieve auth context by subject

gschueler (approved):

Issue #2993, opened 2017-12-12
Update of project configuration overwrites cache settings.


bzlowrance:
Issue type: Bug report

Rundeck version: Rundeck 2.10.0-1
install type: rpm
OS Name/version: RHEL 7.4
DB Type/Version: h2

When updating the project configuration, via "Edit Configuration..."the following cache settings will be overwritten with the defaults of.

project.nodeCache.delay=30
project.nodeCache.enabled=false

We use the following configuration with Ansible, and when this happens it kicks off a Gather Facts job which takes 67 minutes to complete. During this time the inventory disappears, & jobs will fail with a no nodes found error.

project.nodeCache.delay=1209600
project.nodeCache.enabled=true

With the above configuration the inventory is cached for 14 days. If an update is needed before that time, we'll kick off a node refresh job at a time that will not impact existing jobs.

I can think of a few improvements.

1 ) Do not overwrite the existing configuration settings.
2 ) Definition of the default values, for project.nodeCache.delay and project.nodeCache.enabled

If an update is made to the node configuration via "Edit Nodes..."
Only update then inventory for a new resource, or changed resource.

PR #6730, opened 2021-01-19
edit framework controller to accept resourceModelSources config when creating a new project

chrismcg14:
Enhancement

By default, Node Health checks and the Health Status enhancer will now be turned on in new projects for Enterprise customers. 
Updated the category predicate to include the resourceModelSources so they are included. Also updated the default value for the rundeckProHealthChecker and added a check for if the inputMap value at a key is null. If it is, set the value to its default value.

Added a iterator on valid props of the string type. If it is null, set to the defaultValue. If it is still null, remove it from the map to avoid a null pointer exception.


gschueler (changes requested):
tests should be added to verify the changes
also, the use of the predicate in the validateProjectConfigurableInput was added to FrameworkController with this commit message "Fix #2993 node cache settings reset when saving project config", so that change should also be verified that it still works

gschueler (approved):

PR #6734, opened 2021-01-20
fixes rundeckpro/rundeckpro#1454, add in event bus notification when …

imad6639:
Enhancement.  rundeckpro/rundeckpro#1454
Added in the ability to notify/send a message to other cluster members via grails eventBus when the project properties project.disable.executions or project.disable.schedule are changed via editProject in the user interface or the put api method of the project property config.  When either of those properties are changed, the project jobs ned to be scheduled or unscheduled.


rundeckapp/grails-app/controllers/rundeck/controllers/ProjectController.groovy
```diff
@@ -1459,7 +1461,35 @@ class ProjectController extends ControllerBase{
...
+                    isScheduleDisabledNow,
+                    active
+            )
+            if (active) {
+                scheduledExecutionService.rescheduleJobs(frameworkService.isClusterModeEnabled() ? frameworkService.getServerUUID() : null, project)
+            } else {
+                scheduledExecutionService.unscheduleJobsForProject(project, frameworkService.isClusterModeEnabled() ? frameworkService.getServerUUID() : null)
+            }
+        }
+
```
gschueler, line 1492: this duplicates the existing code in FrameworkController, i'd suggest refactoring both to have a new method in FrameworkService

gschueler (approved):

PR #6751, opened 2021-01-27
Update tabs style

ProTip:
image


ProTip:
@mishingo Updated so baseline is full-width:

image

mishingo (approved):

PR #6768, opened 2021-02-04
Duplicate options

chrismcg14:
A clear and concise description of the bug you are fixing (reference issue number), or enhancement you are implementing.

This is a bugfix. When duplicating a job option, most of the properties in the new option did not match that of the original option. 

Updated the toMap() Option method to properly set the properties on the returned map. Updated the _setOptionFromParams() function to properly set the properties on the new option based off returned map from the toMap method. Added inputType property to Options class.

### DIFF
--- a/rundeckapp/grails-app/domain/rundeck/Option.groovy
+++ b/rundeckapp/grails-app/domain/rundeck/Option.groovy
@@ -80,6 +80,7 @@ public class Option implements Comparable{
     Boolean hidden
     Boolean sortValues
     List<String> optionValues
+    String inputType
 
 
     static belongsTo=[scheduledExecution:ScheduledExecution]
@@ -113,6 +114,7 @@ public class Option implements Comparable{
         valuesList(nullable: true)
         valuesListDelimiter(nullable: true)
         sortValues(nullable: true)
+        inputType(nullable: true)
     }
 
 
@@ -166,9 +168,6 @@ public class Option implements Comparable{
         if (null!=sortIndex) {
             map.sortIndex = sortIndex
         }
-        if(enforced){
-            map.enforced=enforced
-        }
         if(required){
             map.required=required
         }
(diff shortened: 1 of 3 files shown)


chrismcg14:
Related PRO issue: https://github.com/rundeckpro/rundeckpro/issues/1369

gschueler (changes requested):
this shouldn't avoid adding a new field to the domain class.

step 1 should be create a new test case that fails with the existing behavior, and then proceed from there for implementing the solution. 

I see changes that don't seem related to the problem (updating the _validateOption method), and no test cases showing what that is meant to accomplish.

any changes to the Option or EditOptsController class should have accompanying test cases.  The tests should be the way to answer the question: "what is this supposed to do?" the tests should fail with existing code, and pass with the new code, showing that the behavior was broken and is now fixed.

rundeckapp/grails-app/domain/rundeck/Option.groovy
```diff
@@ -205,12 +208,25 @@ public class Option implements Comparable{
...
+            map.inputType = inputType
+        }
         if(secureInput){
             map.secure=secureInput
         }
+        if(enforced == true) {
+            map.put("enforcedType", "enforced")
+            map.enforced = true
+        }
+        if(enforced == "regex"){
```
gschueler, line 221: this doesn't make sense, the `enforced` field is a Boolean, so can't be compared to a String

gschueler:
I think i found the true source of the bug: the _duplicateOption method passes the value of `option.toMap()` into the `params` when calling _applyOptionAction here http://github.com/rundeck/rundeck/blob/main/rundeckapp/grails-app/controllers/rundeck/controllers/EditOptsController.groovy#L1001-L1001

However, the input "params" map from (form fields sent in the http request) is different than the toMap() map.  normally the requests that modify the options list parse the "params" map into an Option value, however with the Duplicate request, there is not a set of input params, so the toMap() was used instead.

gschueler:
the _applyOptionAction call can instead pass `params: _getParamsFromOption(newOption)`.

gschueler:
different fix https://github.com/rundeck/rundeck/pull/6784

Issue #1464, opened 2015-10-06
tmp directory


ocuil:
Hi there, first of all I'm so happy with that software !!! is awesome and I love it :-)

I have a little question, I try to run a simple bash script using a virtual ip (balance mode in round robin) and dont works because, when the rundeck copy the script in the target use the /tmp directory and when made a new connection to exec that the load balance send to another node of the cluster and dont find the script to exec.

I have a nfs share folder and I would like to setup (to this job) the folder where upload the script to that shared folder.

it is possible ? or have you any other idea to do that ?

thanks a lot and have great week !!!
Alek

gschueler:
There are a couple of options: you can either use actual nodes for each non-virtual ip, so that the target is actually correct for both the file-copy and command-execution phases.  You can also use the Orchestrator plugin system to for example only pick one node out of the pool of those that are represented by the VIP.

Another option, if you want to use the shared NFS as the way to copy the file, is to use another File-Copier plugin.  You can create this kind of plugin using shell scripts if you want, see [File Copier Plugin](http://rundeck.org/docs/developer/file-copier-plugin.html#script-plugin-type) and [Script Plugin Development](http://rundeck.org/docs/developer/plugin-development.html#script-plugin-development).  You would create your plugin to copy the file to the NFS share in a known location, and the command-execution step on the node would be able to read it from there.

ocuil:
Thanks a lot, tomorrow I'll try that options !!!

ahonor:
Please reopen if you have any other concerns

PR #6771, opened 2021-02-09
Fix: "Show all" button is not clickable #1464

nvalenzuela20:
FIX: https://github.com/rundeckpro/rundeckpro/issues/1464

is this a bug fix, or an enhancement? Please describe.
the job screen's "Show all" button is not clickable when If too many of the same tags are used in the nodes. there is a JavaScript error show on the console.

image

Loaded prototype.min.js script on the page


jtobard:
LGTM but I'll add @ProTip to review since it includes a js library

PR #6816, opened 2021-02-24
Deprecated API: version 10 and lower

chrismcg14:
Enhancement: removed functionality for API versions < 11. Updated Spock and shell tests.

Require the minimum version to be 11, update functionality for versions requiring versions less than 11.


gschueler (changes requested):

rundeckapp/grails-app/controllers/rundeck/controllers/ExecutionController.groovy
```diff
@@ -2133,7 +2129,7 @@ setTimeout(function(){
      * API: /api/5/executions query interface, deprecated since v14
```
gschueler, line 2129: I think maybe this entire API endpoint mapping in UrlMappings should be removed, but I will have to review that

rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
```diff
@@ -3282,7 +3276,7 @@ class MenuController extends ControllerBase implements ApplicationContextAware{
                                                            code  : 'api.error.parameter.required', args: ['project']])
         }
         //allow project='*' to indicate all projects
-        def allProjects = request.api_version >= ApiVersions.V9 && params.project == '*'
+        def allProjects = request.api_version >= ApiVersions.API_EARLIEST_VERSION && params.project == '*'
```
gschueler, line 3279: the api version should always be >= 11 here, so can just take out the condition
```suggestion
        def allProjects = params.project == '*'
```

gschueler (approved):
approved for the api version checking updates.  there are some other changes we need to make, can go in a separate PR

Issue #5179, opened 2019-08-23
Add a button to delete the SCM configuration completely


jtobard:
**Describe the solution you'd like**
A new button on the SCM configuration screen to delete completely the scm-import.properties/scm-export.properties configuration files.

**Describe alternatives you've considered**
You can disable the SCM configuration and it has the same effect but it would be nice to have the capability to completely remove it.

**Additional context**
When you are testing a project or setting up a wrong project, it is desirable to be able to completely eliminate the information so that there are no further doubts as to why this configuration is unused

A new button beside the Clean button, to perform the same cleaning as the clean button and delete the configuration of the storage or the local files depending on the server configuration.
This button should only appears if the configuration is disabled and the user has permission to modify the configuration.

![image](https://user-images.githubusercontent.com/2894508/63600009-bb4f2700-c590-11e9-9ee0-b1c56dff5beb.png)

gmolkvk:
I need this as well. It seems the SCM plugin is blocking project deletes. I could work-around this by deleting the SCM plugin data first.

<details><summary>My problem</summary>

```
$ rd projects delete -p myproject --confirm -y
Request failed: 500 Server Error

$ tail .../log/rundeck.log
2020-09-23 17:27:44,224 [qtp1989780873-1791] ERROR org.codehaus.groovy.grails.web.errors.GrailsExceptionResolver - StaleObjectStateException occurred when processing request: [DELETE] /api/24/project/myproject
Row was updated or deleted by another transaction (or unsaved-value mapping was incorrect) : [rundeck.Storage#670110]. Stacktrace follows:
org.hibernate.StaleObjectStateException: Row was updated or deleted by another transaction (or unsaved-value mapping was incorrect) : [rundeck.Storage#670110]
        at rundeck.services.DbStorageService$_deleteResource_closure13.doCall(DbStorageService.groovy:261)
        at org.grails.datastore.gorm.GormStaticApi.withTransaction(GormStaticApi.groovy:815)
        at org.grails.datastore.gorm.GormStaticApi.withTransaction(GormStaticApi.groovy:767)
        [...]
```

I have traced this to a row in the storage table. I'm using PostgreSQL. After deleting the record id 670110 I can delete the project "myproject" without issue

```
rundeck => SELECT id, date_created, last_updated, dir, name FROM storage WHERE dir LIKE 'projects/myproject/etc';
   id   |      date_created       |      last_updated       |        dir         |         name
[...]

PR #6822, opened 2021-02-26
Scm configuration deletion

ronaveva:
This one is an enhacement and a bug fix for SCM
Enhacement: use is now able to delete the scm configuration ones disabled
ER: https://github.com/rundeck/rundeck/issues/5179
imagen
imagen
imagen

Bug: Storage was getting retrieved from cache, this was causing inconsistency issues when in cluster mode
BUG: https://github.com/rundeckpro/rundeckpro/issues/1559


ltamaster:
moved to https://github.com/rundeck/rundeck/pull/6839

PR #6823, opened 2021-03-01
Fix: Add support for unquoted options (rundeck will supply the value as-is)

unicolet:
Fixes: #6903

IMPORTANT: Before submitting your Pull Request, please review the following instructions:

1. :heavy_check_mark: Please follow the [Developer Guidelines](https://github.com/rundeck/rundeck/wiki/Developer-Guidelines) document. 
2. :heavy_check_mark: Are you implementing a feature or enhancement? Please search the existing Issues and look 
   at the [Rundeck Roadmap Trello Board](https://trello.com/b/sn3g9nOr/rundeck-development) for your idea before posting.
   If your enhancement is not listed, it is better to 
   [post a new enhancement request](https://github.com/rundeck/rundeck/issues/new?template=feature_request.md)
   and get feedback from maintainers and the community *before* submitting an Pull request to implement it.
3. :heavy_check_mark: Please be sure the issue you are addressing is referenced in a commit, or the body of your PR,
   using [github keywords](https://help.github.com/articles/closing-issues-using-keywords/), e.g. "fixes #XXX".

The changes in this PR provide a way to obtain the behavior that, to the best of my understanding, was lost due to issue #602. The new `unquotedoption.<option>` prefix
instructs rundeck to not quote an option, even in presence of characters that would require quoting, i.e. a whitespace.

I just ran all tests with `./gradlew test` and they passed:

```
BUILD SUCCESSFUL in 9m 45s
116 actionable tasks: 116 executed
```

In addition to `option.<option>` to reference a job option the user can now use `unquotedoption.<option>`. The latter instructs rundeck to not apply any quoting
to the option value, but to use the value as-is instead.

As shown in the unit tests, the option value `test with blank` would normally be rendered by rundeck as `'test with blank'` which can be undesirable.
When using `unquotedoption.<option>` instead, the option will be rendered as-is, or as `test with blank`.

I should probably supply a PR for the docs, but I guess that makes sense only this has passed initial review and accepted as potential solution :-)

I have considered adding a flag to `option.<option>` instead but that seemed to complicate the logic of option handling even more, so I went for this approach.

**Additional context**

n.a.


unicolet:
mmh, the travis failure seems unrelated?

```
Reading package lists...
W: The repository 'https://dl.bintray.com/rundeck/rundeck-deb  Release' does not have a Release file.
E: Failed to fetch https://dl.bintray.com/rundeck/rundeck-deb/Packages  403  Forbidden
```

carlosrfranco:
@unicolet Hi! I'm trying to test your code but I'm getting this error:

unquotedJobError

Can you check it, please? I'm using this job definition:
[quotedCommandJob.zip](https://github.com/rundeck/rundeck/files/6156257/quotedCommandJob.zip)

Also, would be good to include the unquoted.\<option\> on suggestions options as following:
suggestion

unicolet:
@carlosrfranco checking

unicolet:
@carlosrfranco looks like the var is not replaced. I can reproduce by firing up `/bin/sh` and then:

```sh
$ ls ${unquotedoption.argsText}
sh: 1: Bad substitution
```

I wonder if the replacement code that I added applies to the usage directly in a command rather than in a script :thinking: I'll give it a look, perhaps you could point me to a guide for running rundeck from sources?

As for the autocomplete I'll look into that too :-)

Thanks for checking my PR :+1:

unicolet:
FYI I was able to reproduce locally, I'll keep digging. For the record I built and started the docker image using:

```sh
./gradlew :docker:officialBuild
docker run -it --rm -v $HOME/rundeck:/home/rundeck/server/data -v rundeck_ssh:/home/rundeck/.ssh --name rundeck --net=host rundeck/rundeck:SNAPSHOT
```

the dirs and volumes are from my previous experiments, YMMV

unicolet:
@carlosrfranco wow, it was quite the hunt, but I finally got it working :sweat_smile: Autocompletion works too in both args and script body

image

let me check why CI is failing, maybe I need to merge with master/main :thinking:

unicolet:
not sure why mergeable fails and I don't seem to have access :thinking:

@carlosrfranco ready for review, AFAICT

fdevans:
Mergable requires a milestone. We'll handle that assignment. Thanks for all the hard work on this. Let us know if we can send you some stickers.

carlosrfranco:
@unicolet LGTM. Can you just add a test for the changes in DataContextUtil class?

unicolet:
> Can you just add a test for the changes in DataContextUtil class?

@carlosrfranco I think I can and I will :-)

unicolet:
@carlosrfranco added test. I am wondering if [this change](https://github.com/rundeck/rundeck/pull/6823/files#diff-f0df9937be6fbaea280c6113e1a9723edb4170a7f329e2e8d61f79c3434cbc9dR144-R147) in `SharedDataContextUtils` is needed, or if I should just remove it :thinking: 

Let me try to run it locally, AFAIU it has no effect on the use case I'm trying to address but I'm wondering if it plays a role in others that we still want to address?

unicolet:
added tests for `SharedDataContextUtils` to document its behavior. Good for another review, thanks!

carlosrfranco (approved):
@unicolet I think it would be better to create a new issue for this case describing it specifically. The issue #602 does not seem to me to be related to this fix since the problem is the quotes that are not preserved when executing a command.

gschueler:
There are some issues with this, the `unquotedoption.name` is auto-suggested in Script step content, but is not actually expanded in them.  However, I think it doesn't make sense to expand the value in Script context, since the `@option.name@` syntax in scripts is already essentially unquoted in that context. 

Second: enabling `${unquotedoption.name}` in Command steps, is potentially opening up a security issue, which is the reason for enforcing quoting in the first place.  If you have a command like `echo ${unquotedoption.myopt}` and the value of myopt is `something ; rm /`, that value will be replaced unquoted resulting in a command `echo something ; rm /` which will execute two commands.  In some cases you may want to allow this unsanitized user input, so I agree that having the flexibility is something we want to add.  However, I'm not sure what the best way to allow that is while making sure the Job author is aware of the risks. Currently the `@option.name@` syntax in scripts will give you this flexibility within a script if you need it. (although that could also be considered a security issue). 

Perhaps both Command and Script steps need a checkbox added to enable the unquoted option behavior.

unicolet:
@gschueler thanks for taking the time to review this PR :-)

> There are some issues with this, the unquotedoption.name is auto-suggested in Script step content, but is not actually expanded in them

@gschueler I didn't notice that, I'll try to check. Any pointers on where in the codebase that expansion is handled would help a ton (searching the rundeck codabse is not easy, even with an IDE :-) )

> enabling `${unquotedoption.name}` in Command steps, is potentially opening up a security issue, which is the reason for enforcing quoting in the first place

I understand, to mitigate this I could add a big bold warning in https://github.com/rundeck/rundeck/pull/6823/files#diff-12984d852fc24a00ce13832c1c9474df65c8e8e099f1c28ba432b9f5d2c682f9R684 . Do you think we could consider this sufficient? If you'd prefer the checkbox along the lines of _yes I am aware of the risks, I want to proceed_ I'll look into it, but I suppose that entails a major design change that I don't think I can work out on my own. For example where would we store the answer?

unicolet:
@gschueler I've tried to address all items raised in https://github.com/rundeck/rundeck/pull/6823#issuecomment-839136654

I did some limited testing locally, and all appears to work.

Here's how the warning looks like:

image

thanks!

gschueler (approved):

gschueler:
Merging this for 3.4.1, thanks @unicolet ! There is a "needs-documentation" tag applied, however. We will need a pr to the https://github.com/rundeck/docs repo to add appropriate documentation.

unicolet:
@gschueler thank you! I'll get to it asap. Happy that I was able to contribute a teeny tiny bit ❤️

PR #6826, opened 2021-03-02
Handling case when executing a job through the API without Node selected by default

carlosrfranco:
Fixes https://github.com/rundeckpro/rundeckpro/issues/1467

If the user should explicitly select target nodes when running a job and no filters for Node Selection is passed through API request, the job was not failing. 

If a job does not have nodes selected by default, so the node filters should be replaced by the filter passed through API request. If none is passed, the job will fail.


gschueler (changes requested):
should add unit test for ScheduledExecutionController change

gschueler (approved):

PR #6829, opened 2021-03-02
fix: 500 error for workflow API response if no read or view auth

gschueler:

Fixes a 500 error calling the Job Workflow API, if the user doesn't have the `read` or `view` authorization for the job. It should return appropriate unauthorized response.

Fix a typo.

### DIFF
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/ScheduledExecutionController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/ScheduledExecutionController.groovy
@@ -657,9 +657,13 @@ class ScheduledExecutionController  extends ControllerBase{
                 ),
                 AuthConstants.ACTION_VIEW, 'Job', params.id
         )) {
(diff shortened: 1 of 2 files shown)


rundeckapp/grails-app/controllers/rundeck/controllers/ScheduledExecutionController.groovy
```diff
@@ -657,9 +657,13 @@ class ScheduledExecutionController  extends ControllerBase{
                 ),
                 AuthConstants.ACTION_VIEW, 'Job', params.id
         )) {
-            return apiService.renderErrorFormat(response, [status: HttpServletResponse.SC_FORBIDDEN,
-                                                           code  : 'api.error.item.unauthorized', args: ['View', 'Job ' +
-                    'ID', jobid]]
```
gschueler, line 662: note: `jobid` is not a valid variable name here

ProTip (approved):

PR #6830, opened 2021-03-03
Initial utility bar implementation

ProTip:
* Adds utility bar to bottom of view
* Establishes initial API through the root store for dynamically adding/modifying entries
* Moves tours trigger to utility bar
* Adds a "Rundeck Info" widget skeleton

image


gschueler:
Issues:

1. I am using a context path `/rundeckpro`, 
    * when I click the "Rundeck 3.4.0" I see the popup with release and cluster member.  
    * (?) It says "Latest releases" but that section is empty.  
   * ❌ When I click the "Rundeck Community 3.4.0" link, I get a 404 page on the `/menu/welcome` path (it doesn't have the context `/rundeckpro` part)
   * screenshot: Screen Shot 2021-03-04 at 9 12 31 AM

2. I'm not sure if this is a problem with the new toolbar, or the welcome Project or the Tours:
   * I have a copy of the welcome project called "Demo2"
   * √ I see the Tours link in the utility bar
   * √ I can click the Tours link and see a list of tours
   * ❌ I can't click any of the Tours, nothing happens
   * ❌ If I click the "Tour Manager" it redirects to `/rundeckpro/tourManager/index?project=welcome_project` instead of `demo2` (This may just be a problem with the tours somehow)

3. Help link
   * √ Help link takes me to the rundeck docs
   * ❌ Help link should use correct url parameters, this was previously generated with the `UtilityTagLib.helpLinkUrl` tag

rundeckapp/grails-spa/packages/ui-trellis/src/components/widgets/rundeck-info/RundeckInfo.vue
```diff
@@ -0,0 +1,96 @@
...
+
+    @Prop()
+    server!: ServerInfo
+
+    async mounted() {
+
+    }
+
+    welcomeUrl() {
+        return url('/menu/welcome').toString()
```
gschueler, line 65: this will need to extend the rd base url

rundeckapp/grails-spa/packages/ui/src/components/navbar/main.ts
```diff
@@ -19,4 +24,48 @@ function init() {
...
+    },
+    {
+        "type": "action",
+        "id": "utility-support",
+        "container": "root",
+        "group": "right",
+        "class": "fas fa-question-circle fas-xs",
+        "label": "Help",
+        "visible": true,
+        "action": () => {window.open('https://docs.rundeck.com/docs/', "_blank")}
```
gschueler, line 56: perhaps the gsp can inject the help url into window.rundeck, and pass that here instead of the hardcoded URL

gschueler (changes requested):

rundeckapp/grails-spa/packages/ui-trellis/src/components/widgets/rundeck-info/RundeckInfo.vue
ProTip: That url helper should join it with rdbase, is it not working? Could be because I used a leading `/`

fdevans:
**RE:** redirects to `/rundeckpro/tourManager/index?project=welcome_project`

This is a URL in the Tour content. Currently tour URLs don't support variables so it's a hard coded URL for that first step which is the issue.

ProTip:
This PR only changes how the tours modal is triggered. There were some manager interactivity issues related to the grid and z-index that have already been resolved in the tours addon.

I believe everything should be addressed now.

gschueler (approved):
* √ help link seems correct now
* √ welcome page url works correctly

The only weird thing is the Rundeck info popup does not have a close button, and the utility bar item that opens it does not stay in an active state.  Clicking outside of the popup doesn't close it, and it stays open unless you click on the Rundeck 3.4.0 text again, which is not necessarily obvious

PR #6833, opened 2021-03-06
Fix: use correct project auth context to eval some project access

gschueler:

Issue: some project-level access is checked using non project-scoped AuthContextProvider, which means that depending on how the auth context is generated, the incorrect authorization result would be cached.  Later, when checking using the correct AuthContextProvider via the cache, the incorrect result is still returned.

This fix addresses two specific places this occurs: checking "event read" and "job create" access for projects when loading the Project List metadata ("homeAjax").


gschueler:
To reproduce the issue:
1. use enterpriseacl feature
2. add acl to enable "read" to "events" within a project for a user
3. log in with that user and view home page
4. choose the project
5. see the "Activity" link is present in the sidebar for the project
6. click Activity and see an Unauthorized message

PR #6846, opened 2021-03-10
Remove File Base Project storage capability

gschueler:

Removes the `filesystem` storage option for project definitions.  This was previously deprecated and `db` made default, this change enforces the `db` behavior.

* Logs a warning if the old config value is used
* Performs import of file-system defined projects to DB automatically
* Refactors the import step so that it occurs prior to other Addons, allowing them to access any ACLs imported to the Project Storage table if necessary (e.g. Enterprise ACL conversion)
* 
Remove adaptor beans and code for filesystem based project definitions.

**Additional context**

Allows more internal refactoring later to remove other code that relates to filesystem based projects.


gschueler:
I may have to revert part of this, seems like it doesn't work correctly in a new install perhaps

gschueler:
(docs content:)

## Update Notes

**Removal of "filesystem" based Project Storage**

* The configuration value `rundeck.projectsStorageType=filesystem` or `file` is no longer supported.  This mode was a legacy mode where Project definitions were stored on the filesystem in a directory with a `project.properties` file. The Rundeck projects storage type `db` has been the default since Rundeck 3.0, and is required for any clustering behavior. This change may only affect you if you are upgrading a legacy configuration of a solo Rundeck server.
* If you are using this mode, you will see a warning at startup, and projects will be converted to DB storage automatically.
* Known file resources stored in each project's directory, such as readme/motd and aclpolicy files, will be imported to the DB storage automatically.
* You can remove the `rundeck.projectsStorageType` configuration key from your rundeck-configuration.properties file

sjrd218 (approved):

Issue #6817, opened 2021-02-24
Password utility only suggests broken algorithms for property file authentication provider


NattyNarwhal:
**Describe the bug**

The documentation for the property file implies that it only supports some weak, broken algorithms like crypt or md5, as this is what the password utility generates. I don't know if this is an intrinsic limitation of the authentication provider, or the documentation and tooling neglecting to mention better and more appropriate algorithms.

I have the impression the property file provider is only intended for small dev/test installations, but that doesn't mean they should have compromised security.

**My Rundeck detail**

* Rundeck version: 3.3.9
* install type: war
* OS Name/version: i 7.4
* DB Type/version: h2

**To Reproduce**

Attempt to create a new password hash (in password utility or `--encryptpwd`)

**Expected behavior**

A better hashing algorithm (i.e bcrypt)

**Screenshots**

The [relevant page in the docs](https://docs.rundeck.com/docs/administration/security/authentication.html#propertyfileloginmodule) if it changes;

![Screenshot from 2021-02-24 18-18-11](https://user-images.githubusercontent.com/3161292/109073653-1190af00-76cd-11eb-8a99-4b91d5c2b710.png)

**Additional context**

[crypt](https://en.wikipedia.org/wiki/Crypt_%28Unix%29#Breaking_crypt(1)_encryption) and [MD5](https://en.wikipedia.org/wiki/MD5#Security) have been known to have been broken for years. Casual use of them hurts my confidence in the software.

ProTip:
Hi @NattyNarwhal ,

In general we do not recommend using `realm.properties` in production. Within the context of `realm.properties` though, do you have any specific concerns utilizing `md5` to hash a secure, random password with high entropy?

NattyNarwhal:
Sorry for not replying earlier. I'm certainly not a crypto expert, but I would think anything I say about this isn't radical.

I do understand your recommendations on auth providers. We've been testing mostly in development environments, but we've experimented with a small production deployment; in those cases, the platform we're deploying to doesn't support PAM, and few run LDAP (or if they do, don't expose it to that part of infra). So the realm file should be OK in that scenario, but we're open to getting a better replacement. Regardless of where it's deployed (dev/prod), I would hope that the password hashing is better than MD5. Using it today makes me raise my eyebrows a little.

[...]

PR #6847, opened 2021-03-10
Add bcrypt credential type for PropertyFileLoginModule

ProTip:
Addresses #6817 

* Adds a bcrypt credential provider and type
* Requires jaas login to be enabled
* Spring security version we use apparently does not support salt revisions so `$2a$` but not `$2y$` works

### DIFF
--- a/rundeckapp/build.gradle
+++ b/rundeckapp/build.gradle
@@ -230,6 +230,7 @@ bootRun {
             '-Xmx2048m')
     systemProperties(System.properties)
     String springProfilesActive = 'spring.profiles.active'
+    systemProperty "rundeck.jaaslogin", "true"
     systemProperty springProfilesActive, System.getProperty(springProfilesActive)
     systemProperty "build.ident", project.version
     sourceResources sourceSets.main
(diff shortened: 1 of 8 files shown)


rundeckapp/build.gradle
```diff
@@ -230,6 +230,7 @@ bootRun {
             '-Xmx2048m')
     systemProperties(System.properties)
     String springProfilesActive = 'spring.profiles.active'
+    systemProperty "rundeck.jaaslogin", "true"
```
sjrd218, line 233: Was this put in for testing? I'm not sure we want jaaslogin to be default.

ProTip: I need to double check but I believe it wouldn't pick up the service provider at all without jaas login engabled(I'm not sure why this is or how it works at all without it enabled).

JAAS login is on by default in our Docker, deb, and rpm distributions .

sjrd218 (approved):
Worked for me. I added the bcrypt encryption to the password encryption utilities and made sure it also worked in the spring security mode.

gschueler (approved):

PR #6848, opened 2021-03-10
Update help text to use quotes on demarcation text

carlosrfranco:
fixes https://github.com/rundeckpro/rundeckpro/issues/1433

The global filter to format the output data was not working well if handle a section of the output (and not the full output)

The help text of the plugin was updated to instruct the user to use quotes in the demarcation text


carlosrfranco:
Correct use can be verified in the documentation: https://docs.rundeck.com/docs/manual/log-filters/render-formatted-data.html#usage

Issue #6674, opened 2020-12-14
Execution History Clean feature doesn't work as expected since Rundeck 3.3.4


MegaDrive68k:
**Describe the bug**

Execution History Clean feature doesn't work since Rundeck 3.3.4, testing deleting execution every 20 seconds, the executions keep on Activity page (Work as expected on Rundeck 3.2.9 until 3.3.3).

**My Rundeck detail**

* Rundeck version: 3.3.4, 3.3.5, 3.3.6, 3.3.7
* install type: WAR
* OS Name/version: Ubuntu 20.10
* DB Type/version: H2

**To Reproduce**

1. Configure a project with the following Execution History Clean config:

![Screenshot_9](https://user-images.githubusercontent.com/51376003/102141216-3b149b00-3e3f-11eb-9d24-10dc2c1f3692.png)

2. Use a job that executes every 5 seconds (to generate executions):

```xml
<joblist>
  <job>
    <defaultTab>nodes</defaultTab>
    <description></description>
    <executionEnabled>true</executionEnabled>
    <id>d97d8e6d-361f-4b0c-b966-5506ed621fa0</id>
    <loglevel>INFO</loglevel>
    <name>HelloWorld</name>
    <nodeFilterEditable>false</nodeFilterEditable>
    <plugins />
    <schedule>
      <dayofmonth />
      <month day='*' month='*' />
      <time hour='*' minute='*' seconds='0/5' />
      <year year='*' />
    </schedule>
    <scheduleEnabled>true</scheduleEnabled>
    <sequence keepgoing='false' strategy='node-first'>
      <command>
        <exec>whoami</exec>
      </command>
    </sequence>
    <uuid>d97d8e6d-361f-4b0c-b966-5506ed621fa0</uuid>
  </job>
</joblist>
```
3. Test on Rundeck 3.3.3 (also tested on 3.2.9), works as expected:

![Screenshot_10](https://user-images.githubusercontent.com/51376003/102141279-52538880-3e3f-11eb-8ed6-48cd13885f87.png)

4. Now, do the same on Rundeck 3.3.4/3.3.5/3.3.6/3.3.7):

![Screenshot_11](https://user-images.githubusercontent.com/51376003/102141571-ca21b300-3e3f-11eb-8b3b-336b25a52985.png)

![Screenshot_8](https://user-images.githubusercontent.com/51376003/102141399-86c74480-3e3f-11eb-9582-d7547bd5b0f9.png)

**Expected behavior**

[...]

PR #6859, opened 2021-03-11
cleanup execution in non-cluster mode 

ltamaster:
bug fix: execution cleanup doesn't work as expected when cluster mode is off
cleanup execution in non-cluster mode shouldn't filter by serverNodeUuid.

remove serverNodeUuid filter on execution table when cluster mode is off

**Additional context**
 issue https://github.com/rundeck/rundeck/issues/6674


fdevans:
We will try to review and merge for 3.4.0.  If we can't get through it we'll push to 3.4.1

gschueler (approved):

PR #6865, opened 2021-03-12
Update AWS source model plugin version.

sjrd218:
Move jackson dataformat cbor dependency into Rundeck for coordinated lib versioning.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -2,7 +2,7 @@
 rundeck:
   plugins: # Extra plugins to bundle
   - "com.github.Batix:rundeck-ansible-plugin:3.1.0"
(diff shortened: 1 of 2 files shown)


ProTip:
@sjrd218 The AWS SDK in use by the s3 model source uses an older line of Jackson(2.6.x): https://mvnrepository.com/artifact/com.amazonaws/aws-java-sdk-core/1.11.973 .

Do you think this will cause any issues?

sjrd218:
I'm not sure. The previous version exported a 2.6 version of the dataformat cbor lib, which was mismatched against the 2.10.1 version of jackson core we use, which I would've expected to be problematic, but it seemed to work. This just syncs the version of the jackson libraries we use in Rundeck.

ProTip (approved):

PR #6871, opened 2021-03-12
Replacing code that fetches the list of ScheduledExecutions parents to use criteria and improve query time

carlosrfranco:
fixes https://github.com/rundeckpro/rundeckpro/issues/1519

In the scenario where a job has many referenced executions, the method that returns the of parent jobs becomes quite slow and this slows down the loading of the job page

The method was changed to use criteria to improve query time.


gschueler (changes requested):
looks good, but needs some tests

gschueler (approved):

PR #6887, opened 2021-03-16
RUN-152: Fix: Preserve quotes on arguments commands so rundeck will interpret it as unique argument

### LINKED ISSUE #602: preserve quotes used in commands

carlosrfranco:
fixes #602

If a command has a quoted strings like command -m 'arg1 arg2', this arguments was not preserved by rundeck and the argument 'arg1 arg2' was interpreted as two separated args.

Was implemented a new feature that allow to define if quotes should be preserved or not on command step.

This PR was migrated/updated from https://github.com/rundeck/rundeck/pull/5542


gschueler:
This works well, on the whole.  I think it makes command steps and adhoc commands behave more like what users expect.

However, there are some cases where the quoting doesn't come through exactly as you might expect.

Here are some command strings, showing differences in how Bash interprets them and how Rundeck will interpret them:

A) Command: `TEST=test && echo $TEST`
| bash | preserveQuotes=false | preserveQuotes=true |
|------|---|--------------------|
|`test` | `test` | `test`|

B**) Command: `TEST=test && echo '$TEST'`
| bash | preserveQuotes=false | preserveQuotes=true |
|------|---|--------------------|
**`$TEST`** | `test` | `test`|

C) Command: `TEST=test && echo "$TEST"`
| bash | preserveQuotes=false | preserveQuotes=true |
|------|---|--------------------|
`test` | `test` | `test`|


D) Command: `TEST=test && echo $TEST test2`
| bash | preserveQuotes=false | preserveQuotes=true |
|------|---|--------------------|
|`test test2` | `test test2` | `test test2`|

E**) Command: `TEST=test && echo '$TEST test2'`
| bash | preserveQuotes=false | preserveQuotes=true |
|------|---|--------------------|
`$TEST test2` | **`test test2`** | `$TEST test2`|

F**) Command: `TEST=test && echo "$TEST test2"`
| bash | preserveQuotes=false | preserveQuotes=true |
|------|---|--------------------|
`test test2` | `test test2` | **`$TEST test2`**|

These differences are for two reasons: 1) because Rundeck treats a single word argument without quotes, special chars, or spaces, as not needing quoting, even when preserveQuotes is enabled (case B), and 2) because Bash treats double-quotes and single-quotes differently when interpolating values inside them (case E and F).

This PR detects when an argument is quoted, then enforces a re-quoting when creating the command array to execute, however it re-quotes the value with a single-quote in all cases.

To update this to fix cases E and F: perhaps we need to detect a double-quoted argument and have it re-quoted with double-quotes. 

For case B, we would also probably need to enforce quoting even when the argument is a single word and would not otherwise need quoting.

gschueler:
I force-pushed changes: 
* rebase on main
* add db migrations
* some i18n text updates, fix text encoding for html

gschueler (changes requested):
Changes needed to address cases B & F

PR #6909, opened 2021-03-22
scm delete nows perfoms a clean before deleting the configuration

ronaveva:
It performs a clean before deleting the configuration.


fdevans:
If this is part of the SCM fixes please add the backport-3.3 tag

Issue #6910, opened 2021-03-23
Scm clean fails when repo is down


ronaveva:
When the scm configuration is failing to start correctly and you try to perform a clean, you will get an error 500

* Rundeck version: 3.3.9
* install type: war
* OS Name/version: ubuntu 20.04
* DB Type/version: h2

**To Reproduce**
1. Create a project
2. Add scm configuration (import or export) whit issues so it cannot be enabled
3. Go back to scm home screen
4. Try to perform a clean

**Expected behavior**
It should perform the clean anyway

![imagen](https://user-images.githubusercontent.com/50217398/112162478-da79c480-8bca-11eb-8594-2c9614eb4d39.png)

PR #6911, opened 2021-03-23
Fix/scm clean

ronaveva:
fix https://github.com/rundeck/rundeck/issues/6910

New method on the scm plugin factory with initializacion flag.


fdevans:
If this is part of the SCM fixes please add the backport-3.3 tag.

jtobard (approved):
Tested and it works

PR #6917, opened 2021-03-23
DB Migration fixup

ProTip:
* Adds the `grails.plugin.databasemigration.updateOnStart` option to Docker
* Defaults the above and `dbCreate` settings to the Rundeck default if no override is provided
* Fixes failing tests due to connection abandon timeout
* Adjusts the `node_filter` filter column
* Adds loggers for the Grails db migration plugin and Liquibase set to `INFO`

This would supersede #6923

### DIFF
--- a/rundeckapp/grails-app/migrations/core/NodeFilter.groovy
+++ b/rundeckapp/grails-app/migrations/core/NodeFilter.groovy
@@ -58,4 +58,23 @@ databaseChangeLog = {
             column(name: "project", type: '${varchar255.type}')
         }
     }
+
+    changeSet(author: "rundeckuser (generated)", failOnError:"false", id: "3.4.0-1616620097", dbms: "h2") {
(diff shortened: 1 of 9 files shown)


rundeckapp/grails-app/migrations/core/NodeFilter.groovy
```diff
@@ -58,4 +58,23 @@ databaseChangeLog = {
             column(name: "project", type: '${varchar255.type}')
         }
     }
+
+    changeSet(author: "rundeckuser (generated)", failOnError:"false", id: "3.4.0-1616620097", dbms: "h2") {
```
ProTip, line 62: @imad6639 Is it okay to use timestamps in these IDs so the latest doesn't need to be hunted down through the files/db/etc? This is a pretty common db migration pattern for odering changes(timestamps in the names).

imad6639 (approved):

PR #6918, opened 2021-03-24
Refreshing healthCheck status cache before job execution to avoi job failing when it has node health status filter

carlosrfranco:
fixes https://github.com/rundeckpro/rundeckpro/issues/1539

if a Node becomes unhealthy after the last health check and the cache is not updated before a job starts, the job that has a filter that excludes unhealthy Nodes will fail

The core was updated to allow a node enhancer plugin refreshing node status when a job is preparing to start the execution.

**Additional context**
Related to https://github.com/rundeckpro/rundeckpro/pull/1633


carlosrfranco:
@Mergifyio rebase

gschueler:
note: closed because design was re-evaluated

PR #6922, opened 2021-03-24
Cleanup: jobs import api endpoint requires apiv14

gschueler:

Some code cleanup for Jobs Import API
adds some tests


sjrd218 (approved):

sjrd218:
Are the API tests failing because the error returns json by default instead of xml?

gschueler:
hmm, maybe i broke something

PR #6938, opened 2021-03-30
Handles job pagination (if enabled) on referenced job picker modal

carlosrfranco:
fixes https://github.com/rundeckpro/rundeckpro/issues/1571

If job pagination is enabled, there is no way to handle pagination on modal to choose job reference

Was included pagination mechanism to handle pagination calling controller without reloading page


mishingo (commented):
UI needs adjustment, PR is created with updated styles

mishingo (approved):

PR #6940, opened 2021-03-30
retrieving scheduled executions without using cache

ronaveva:
Fixes https://github.com/rundeckpro/rundeckpro/issues/1657

Modified to not use cache when searching for scheduled executions.


jtobard (approved):
LGTM I don't think it is possible to have a unit test for this isn't? @ronave

PR #6945, opened 2021-04-02
Adjust SCM UI and Installed Plugins Style

mishingo:
- Fix orientation of row content
- Adjust spacing between rows
- Add card container

- Installed Plugins adjust style


gschueler:
this seems to include some changes to `ProviderCard.vue` that are unrelated to the SCM UI

PR #6948, opened 2021-04-05
fix:  use aclFileManager to read project acls for export

gschueler:
fix:  use aclFileManager to read project acls for export

Fixes https://github.com/rundeckpro/rundeckpro/issues/1668


chrismcg14 (commented):
Tested the branch and exported a project with acls at project and system level and couldn't see them when I expanded the jar. Also let Greg know on Slack.

chrismcg14 (approved):
Confirmed it is working!

PR #6950, opened 2021-04-06
Revert "Adjust SCM UI and Installed Plugins Style"

gschueler:
Reverts rundeck/rundeck#6945


gschueler:
reverted pr with extra changes

PR #6952, opened 2021-04-06
Enhance SCM UI 

mishingo:
- Add styles to match current UI
- Fix Spacing issues
- Inline Diff file export path


gschueler:
I fixed a few things, and restored multi-line view for renamed job path. it was intended to be multi line so that the two paths can be more easily compared.  we can revisit the design on that (I think it would be better to just show the new path, and change the color + show an icon indicating it was modified, show the original path on hover)

gschueler (approved):

PR #6953, opened 2021-04-06
Enh: Adding access to Key Storage for Webhook Plugins

chrismcg14:
Add a the keyStorage service provider to the context services being passed to onevent, so web hook processing can use key storage


grails-webhooks/src/test/groovy/webhooks/WebhookServiceSpec.groovy
```diff
@@ -88,6 +89,10 @@ class WebhookServiceSpec extends Specification implements ServiceUnitTest<Webhoo
                 webhookProviderService
             }
         }
+        service.storageService = Mock(MockStorageService){
+            storageTreeWithContext(_) >> {Mock(KeyStorageTree)}
```
ltamaster, line 93: maybe you can move this to the `then` clause and mark the number of times this should be called (like 1* storageTreeWithContext(_) ...)

ltamaster (changes requested):

ltamaster (approved):
LGTM

gschueler (approved):

PR #6963, opened 2021-04-07
Sec: Remove apache httpclient 3.1

sjrd218:
Replace  apache http client 3.1 with newer apache http client version. Created new http client interface and replaced the locations where the 3.1 code was used with the new interface.


fdevans:
Is the able to be backported to 3.3.x series?

sjrd218:
> Is the able to be backported to 3.3.x series?

Yes, we can backport.

gschueler (approved):

PR #6965, opened 2021-04-08
Adjust style of modal pagination

mishingo:
UI Enhancement

Adjust the styling of the modal pagination and spacing of pagination context


rundeckapp/grails-app/views/menu/_jobPickerModalPagination.gsp
```diff
@@ -1,21 +1,31 @@
 <g:if test="${max && max < total && jobsListSize < total}">
-    <div id="modal-pagination" style="display: none">
-        <span>
```
carlosrfranco, line 3: This can be removed. In my PR I had included the offset, total and max attributes but that shouldn't be here.

carlosrfranco (changes requested):

rundeckapp/grails-app/views/menu/_jobPickerModalPagination.gsp
mishingo: Offset values span has  been removed

PR #6967, opened 2021-04-08
Enh: Change default server.address to localhost instead of hostname to generate default rundeck-config.properties

ahormazabal:
By default, rundeck uses the machine hostname to generate both `server.address` and `grails.serverURL` properties when building the default rundeck-config.properties at first run. However, this assumes that the hostname is resolvable through DNS or /etc/hosts, which is not always the case, causing two side effects:

- The application binds to just one network interface, which can be any of the available.
- The grails.serverURL points to an unresolvable URL.

This causes the application to be unreachable after the first run, requiring the installer to run it, take it down to adjust the generated file, and run it again.

This PR makes the following changes to the default properties file to address those issues:
- server.address is now assigned to 0.0.0.0 so the application binds to all available network interfaces.
- default grails.serverURL is now http://localhost:$port/$context so at least the app is always reachable at localhost after the first run.
- adds server.port property explicitly so installers can change the listen port without having to google/review the documentation.

### DIFF
--- a/rundeckapp/templates/config/rundeck-config.properties.template
+++ b/rundeckapp/templates/config/rundeck-config.properties.template
@@ -4,7 +4,7 @@ rdeck.base=${rdeck.base}
 
 #rss.enabled if set to true enables RSS feeds that are public (non-authenticated)
 rss.enabled=false
-server.address=127.0.0.1
+server.address=localhost
 server.port=${server.http.port}
 grails.serverURL=http://localhost:${server.http.port}${server.web.context}
 dataSource.dbCreate = none
--- a/rundeckapp/templates/config/rundeck-config.properties.template
+++ b/rundeckapp/templates/config/rundeck-config.properties.template
@@ -4,7 +4,8 @@ rdeck.base=${rdeck.base}
 
 #rss.enabled if set to true enables RSS feeds that are public (non-authenticated)
 rss.enabled=false
-server.address=${server.hostname}
+server.address=127.0.0.1
+server.port=${server.http.port}
 grails.serverURL=http://${server.hostname}:${server.http.port}${server.web.context}
 dataSource.dbCreate = none
 dataSource.url = jdbc:h2:file:${server.datastore.path};DB_CLOSE_ON_EXIT=FALSE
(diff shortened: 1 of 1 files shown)


sjrd218 (approved):
I'm in favor of these changes.

ProTip:
I'm trying to figure out what scenario this would be _more correct_ than the status quo?

**Installing on a remote server**
The final domain name is almost impossible to guess. How would the installer access Rundeck in a browser via localhost? That's almost certainly going to need to change and require a restart.

**Installing locally**
Localhost or hostname seem to work equally well. However, binding a remote execution service with the credentials `admin/admin` to `0.0.0.0` on install seems like a step backwards security-wise; for local and remote.

fdevans:
let's stay with `localhost` for server.address and look at documentation enhancements that can better guide customers on setting these properly.

ahormazabal:
As agreed, the grails.serverURL setting was set back to the original value, and server.address was changed to 127.0.0.1 as default.

gschueler:
with the change to localhost, it seems there needs to be some new "first run" documentation about what to change, as by default now a new war install will not be listening on an external interface

fdevans:
Added "needs documentation" label.  We can be sure that the docs get updated along with the Learning resource updates.

gschueler (approved):

gschueler:
I updated the `grails.serverURL` to use localhost by default now

sjrd218 (approved):

imad6639 (approved):

gschueler (approved):

imad6639 (commented):
When I run this, the console shows me `Grails application running at http://127.0.0.1:4440 in environment: production`.   When I go to that url, and try to log in, the developer console shows me Refused to send form data to 'http://localhost:4440/' because it violates the following Content Security Policy directive: "form-action 'self'".  But if I go to http://localhost:4440, then it's fine.

ProTip:
That console output shows the address it's bound to instead of the URL :|

gschueler:
what if we set the bind address to `localhost`, that should resolve to (a) right bind address, and then the url would be correct, right?

PR #6970, opened 2021-04-09
Fix/scm status sync

ronaveva:
fixes https://github.com/rundeckpro/rundeckpro/issues/1681

It now checks for the scm configuration in DB instead of going directly to memory.


gschueler (approved):

ltamaster (approved):
LGTM

PR #6979, opened 2021-04-14
adjust margin for icon in card component

mishingo:
Enhancement: Adjusts margin on icon in cards

Adds margin on X axis of icons in card componente


gschueler:
I think we should remove that margin completely, i'm not sure why it was added so broadly.

PR #6986, opened 2021-04-15
Including job list type on parameters when triggering job search

carlosrfranco:
Fixes https://github.com/rundeckpro/rundeckpro/issues/1615

Job search was broken when using grouped job list and with paged job list enabled

It was included job list type on parameters for searching jobs


gschueler:
added a change to include the jobListType param in the link used for group folders

gschueler (approved):

Issue #5848, opened 2020-03-02
Too many deadlocks in MSSQL deleting old executions


G3NSVRV:
**Describe the bug**
It appears that the project settings for deleting old execution logs were causing a lot of deadlocks in the database. 

**My Rundeck detail**

* Rundeck version: 3.2.1
* install type: WAR
* OS Name/version: Windows Server (as a service)
* DB Type/version: MSSQL

**To Reproduce**
Steps to reproduce the behaviour:
1. Setup a project that made many execution logs (e.g. 8k+)
2. Go to project settings > edit configuration > execution history clean
3. Enable old executions cleanup that deletes almost everything

**Expected behavior**
No database lock, normal and quick behaviour

**Screenshots**
![RundeckDeadlocks](https://user-images.githubusercontent.com/11720535/75707592-e46a1600-5c9d-11ea-84be-717dc1c31b27.jpg)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome
 - Version: 80.0.3987.122	

**Additional context**
Our DBAs looked at the database for a few days and were able to create an index that caused the locks to disappear almost immediately. Here is the code they ran:

```
CREATE NONCLUSTERED INDEX [NCIX_error_handler_id] ON [dbo].[workflow_step]
(
[error_handler_id] ASC
)WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, SORT_IN_TEMPDB = OFF, DROP_EXISTING = OFF, ONLINE = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON)

GO
```

as you can see in the attached picture, the deadlocks disappeared almost instantly. Maybe this will be helpful for other customers using MS SQL, or can be an enhancement added to a future version of Rundeck?

fdevans:
Closing due to inactivity

jgarces-pd:
@fdevans do we have these indexes on our latest version?

jplassnibatt:
As this has been reported in 3.3.3 I'm reopening it.

jgarces-pd:
hi @jplassnibatt it happens too on the latest version?

jplassnibatt:
> hi @jplassnibatt it happens too on the latest version?

don't know, 3.3.3 is still supported though.

jgarces-pd:
@fdevans do we have these indexes on our latest version?

gschueler:
[...]

PR #6987, opened 2021-04-15
Fix: Create index for errorHandler on WorkflowStep table to help prevent deadlocks in MSSQL

carlosrfranco:
Fixes https://github.com/rundeck/rundeck/issues/5848

Deleting old execution logs were causing a lot of deadlocks in the database.

Creating index for errorHandler on WorkflowStep table

### DIFF
--- a/rundeckapp/grails-app/migrations/changelog.groovy
+++ b/rundeckapp/grails-app/migrations/changelog.groovy
@@ -103,7 +103,7 @@ databaseChangeLog = {
         include file: 'core/RDOPTION.groovy'
         include file: 'core/Webhook.groovy'
         include file: 'core/ConstraintsIndexesKeys.groovy'
-        include file: 'core/Tag-3.4.0.groovy'
         include file: 'core/WorkflowStep.groovy'
+        include file: 'core/Tag-3.4.0.groovy'
 
 }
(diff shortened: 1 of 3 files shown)


rundeckapp/grails-app/migrations/changelog.groovy
```diff
@@ -104,5 +104,6 @@ databaseChangeLog = {
         include file: 'core/Webhook.groovy'
         include file: 'core/ConstraintsIndexesKeys.groovy'
         include file: 'core/Tag-3.4.0.groovy'
+        include file: 'core/WorkflowStep.groovy'
```
gschueler, line 107: I think this should be included before the 3.4.0 tag, right @imad6639 ?

imad6639: Depends on the intended release.  If this is for 3.4.0 then before the 3.4.0 tag.  If it's for the 3.4.1 release, then a 3.4.1 tag should be created and it should go between the 3.4.0 tag and the 3.4.1 tag

carlosrfranco: FIxed considering that it will be included in 3.4.0. Please, let me know if it should be only in 3.4.1

rundeckapp/grails-app/migrations/core/WorkflowStep.groovy
```diff
@@ -0,0 +1,12 @@
+databaseChangeLog = {
+    changeSet(author: "rundeckuser (generated)", id: "workflow_step") {
```
imad6639, line 2: I think we are changing id's to be epoch time moving forward since they are easier to debug/analyze the order in the changelog:  1619464667662 would be an example of the id right now

carlosrfranco: Fixed

imad6639 (approved):

gschueler (approved):

PR #6993, opened 2021-04-19
Enh: Add button to Commit SCM changes in jobs list page

ronaveva:
Fixes https://github.com/rundeckpro/rundeckpro/issues/1710

The button was added in jobs list page instead of job show because when the user performs a push it cannot choose the commit to push, so if this would be in jobs show page, the user could think that he is pushing just that single job instead of all the pending commits.

The modification in scm controller was to avoid showing the pending jobs to be commited when the requested action is to push and also to know the push action id a new method was added to the export interface.

imagen


ltamaster (approved):
LGTM

PR #6995, opened 2021-04-21
Sec: Resolve Vulnerabilities in various UI components

mishingo:
Provides a way to manage JS modules and update to latest releases and resolve security vulnerabilities

Implement Mix to manage modules from npm and build to designated dir

### Vendoring
The Gradle build has been updated to _vendor_ the files during the build process.

The vendor location is `grails-app/assets/javascripts/vendor`.

JS `//require`s and GSP imports have been updated accordingly.


ProTip (changes requested):
Currently failing CI/CD. The previous attempt to update to Bootstrap `3.4.x` failed.

The package management infrastructure should be excluded from the Grails asset pipeline.

ProTip (commented):
We should probably add instructions on how to update the dependencies in the future.

Is there a good way to verify the source of the deps being checked in? There doesn't seem to be a good way to review these blobs of minified code for security or other issues. Perhaps we could pin the deps and wire the mix process into the Gradle build so the deps are fetched from npm during build?

mishingo:
Agreed tying this into the gradle build is more secure

ProTip:
After discussing with @mishingo I will wire this into the Gradle build and update the PR.

ProTip (approved):

PR #7007, opened 2021-04-26
Kill process when thread for local command is interrupted

carlosrfranco:
Fixes https://github.com/rundeckpro/rundeckpro/issues/1518

When using an script executor plugin to run a job and then kill this job, the subprocess wasn't being killed by rundeck

when the Thread is interrupted when killing the job, the process should also be destroyed


gschueler (approved):

NapoSky:
same in 3.4.2. I need to escape every semicolon and "&" in a commandline workflow. No needs to do it in script workflow.

PR #7008, opened 2021-04-26
Fix: deleted jobs not reflected in project synch state

gschueler:
* happens if jobs deleted out of band of job events, i.e. other cluster member

fixes https://github.com/rundeckpro/rundeckpro/issues/1688

cache job id and file path within the plugin, track list of jobs refreshed during bg process. If some job ids are missing during bg refresh, and were previously cached in the plugin, remove the local files to reflect the deleted status.

**Additional context**


ltamaster (approved):
tested

ronaveva:
deletion working correctly

ronaveva (approved):
LGTM

ronaveva:
@gschueler i've found a small thing. When a deletion is made in cluster
1.-Cluster member where the job was deleted: shows job name deleted
2.-Other members: instead of showing the job name, only shows "(Deleted job)"

imagen

gschueler:
@ronaveva yes, I saw that. I think it is ok for this pr. We could enhance it in the future to fix that I think, we would need to cache job details (name/group) along with the path when the bg process scans jobs

gschueler:
i force pushed a change, it removes the change to the ScmExportPlugin, and allows the job name to show up correctly in the second cluster member

PR #7010, opened 2021-04-28
fix: SCM job import action when both import/export are enabled can show incorrect status

gschueler:
* if git import/export use "preserve" uuid feature, and both are enabled, after git import is performed, jobs can still show modified status, because a random UUID was being saved in job metadata as "srcId".

fixes https://github.com/rundeckpro/rundeckpro/issues/1700

* Removed the behavior that stored random UUID as srcId on job create from export listener
* refactor export listener into separate class
* add tests for export listener
* also alleviates a race condition: both job import and export would create new job metadata entry in the DB in different threads unaware of the other entry.


ronaveva (approved):
LGTM

PR #7022, opened 2021-04-30
SCM: Fix job sourceId not appearing in exported file path

gschueler:

Fixes https://github.com/rundeckpro/rundeckpro/issues/1729

Fixes two issues:
1. corrects earlier change that removed metadata loading for jobs when getting export paths, and which prevented `${job.sourceId}` from working in template paths
2. Fixes an issue where imported jobs would be serialized with wrong file path when scm import and export were both enabled; the job metadata was stored after the event was fired triggering the export.  Fixed this to store the metadata before firing the job change event.


ronaveva (approved):
LGTM

ltamaster:
@gschueler I created an issue for it : https://github.com/rundeckpro/rundeckpro/issues/1729

ltamaster (approved):
tested

PR #7030, opened 2021-05-04
RUN-151: Fix: Properly detect when a job was renamed in SCM Import

ltamaster:
BigFix: when a job is renamed, SCM Import deletes and creates the job again

for issue https://github.com/rundeckpro/rundeckpro/issues/1731

- using jgit, check if a path was added and deleted (gitdiff)
- if that is the case save the renamed job in the import tracker
- when the job is imported, save the current UUID and the sourceId. Force to preserve original UUID when the job is re-imported.
- clean import trackers of the old path
- split job plugin metadata per integration

**Additional context**

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/plugins/scm/JobImporter.java
+++ b/core/src/main/java/com/dtolabs/rundeck/plugins/scm/JobImporter.java
@@ -66,7 +66,7 @@ public interface JobImporter {
      *
      * @return result
      */
-    ImportResult importFromStream(String format, InputStream input, Map importMetadata, boolean preserveUuid, Map renamedJob);
+    ImportResult importFromStream(String format, InputStream input, Map importMetadata, boolean preserveUuid, JobRenamed renamedJob);
 
     /**
      * Deleted a job deleted remotely
--- a/core/src/main/java/com/dtolabs/rundeck/plugins/scm/JobImporter.java
+++ b/core/src/main/java/com/dtolabs/rundeck/plugins/scm/JobImporter.java
@@ -62,10 +62,11 @@ public interface JobImporter {
      * @param input          input stream
      * @param importMetadata metadata to attach to the job
      * @param preserveUuid   if true, preserve any UUID on import, otherwise remove it
+     * @param renamedJob     pass original ID/UUID if the job is renamed
      *
      * @return result
      */
-    ImportResult importFromStream(String format, InputStream input, Map importMetadata, boolean preserveUuid);
+    ImportResult importFromStream(String format, InputStream input, Map importMetadata, boolean preserveUuid, Map renamedJob);
 
     /**
      * Deleted a job deleted remotely
--- a/plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/GitImportPlugin.groovy
+++ b/plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/GitImportPlugin.groovy
@@ -436,6 +436,18 @@ class GitImportPlugin extends BaseGitPlugin implements ScmImportPlugin {
                         "${entry.changeType} ${entry.oldPath}->${entry.newPath}"
                     }.join("\n")
                     )
+
+                    def findNewPath = changes.find { it.changeType == DiffEntry.ChangeType.ADD && it.newPath != null }
+                    def findOldPath = changes.find { it.changeType == DiffEntry.ChangeType.DELETE && it.oldPath != null }
+
+                    if(findNewPath && findOldPath){
+                        def oldPath = findOldPath.oldPath
+                        def newPath = findNewPath.newPath
+                        log.error("Rename detected from ${oldPath} to ${newPath}")
+                        this.importTracker.jobRenamed(job, oldPath, newPath)
+                        return ImportSynchState.IMPORT_NEEDED
+                    }
+
                     def found = pathChanges.find { it.oldPath == path }
                     if (found && found.changeType == DiffEntry.ChangeType.DELETE) {
                         return ImportSynchState.DELETE_NEEDED
@@ -514,7 +526,14 @@ class GitImportPlugin extends BaseGitPlugin implements ScmImportPlugin {
                 break;
 
             case JobChangeEvent.JobChangeEventType.MODIFY_RENAME:
-                importTracker.jobRenamed(reference, path, newpath)
+                if(importTracker.originalValue(newpath)){
+                    //if a job was imported with a new name traker needs to be cleaned
+                    def oldPath=importTracker.originalValue(newpath)
+                    importTracker.renamedTrackedItems.trackItem(oldPath, oldPath)
+                    importTracker.untrackPath(oldPath)
+                }else{
+                    importTracker.jobRenamed(reference, path, newpath)
+                }
         //TODO
 //            case JobChangeEvent.JobChangeEventType.CREATE:
 //            case JobChangeEvent.JobChangeEventType.MODIFY:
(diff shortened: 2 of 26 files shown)


ltamaster:
@Mergifyio rebase

rundeckapp/grails-app/services/rundeck/services/scm/ScmLoaderService.groovy
```diff
@@ -43,6 +42,7 @@ class ScmLoaderService implements EventBusAware {
     final Map<String, ScheduledFuture> scmProjectLoaderProcess = Collections.synchronizedMap([:])
     final Map<String, Boolean> scmProjectInitLoaded = Collections.synchronizedMap([:])
     final Map<String, ScmPluginConfigData> scmPluginMeta = Collections.synchronizedMap([:])
+    final Map<String, ScmPluginConfigData> scmFailedProjectInit = Collections.synchronizedMap([:])
```
gschueler, line 45: I think these changes to the ScmLoaderService should not be part of this PR, it doesn't have to do with the renamed job import, right?

rundeckapp/grails-app/services/rundeck/services/ScmService.groovy
```diff
@@ -1455,7 +1459,7 @@ class ScmService {
                 def newmeta = [name: job.getJobName(),groupPath: job.getGroupPath(),version: job.version, pluginMeta: jobStatus.commit.asMap()]
                 jobMetadataService.setJobPluginMeta(
                         job,
-                        STORAGE_NAME_IMPORT,
+                        STORAGE_NAME_EXPORT,
```
gschueler, line 1462: if the metadata now uses the integration type to store separate metadata for each job, how does it work when you have both import and export configured, and you e.g. rename a job?  previously the metadata was used to track the sourceId when you import a job, and later when you export it. is the sourceId still passed from import to export in the same way?

core/src/main/java/com/dtolabs/rundeck/plugins/scm/JobImporter.java
```diff
@@ -62,10 +62,11 @@
      * @param input          input stream
      * @param importMetadata metadata to attach to the job
      * @param preserveUuid   if true, preserve any UUID on import, otherwise remove it
+     * @param renamedJob     pass original ID/UUID if the job is renamed
      *
      * @return result
      */
-    ImportResult importFromStream(String format, InputStream input, Map importMetadata, boolean preserveUuid);
+    ImportResult importFromStream(String format, InputStream input, Map importMetadata, boolean preserveUuid, Map renamedJob);
```
gschueler, line 69: looking at where this `renamedJob` is used, it would be better if it was not a Map, eg. add a new interface that has `getUuid()` and `getSourceId()`

gschueler (changes requested):

ltamaster:
@Mergifyio rebase

plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/GitImportPlugin.groovy
```diff
@@ -436,6 +436,20 @@ class GitImportPlugin extends BaseGitPlugin implements ScmImportPlugin {
                         "${entry.changeType} ${entry.oldPath}->${entry.newPath}"
                     }.join("\n")
                     )
+
+                    String jobSerializerId = job?.sourceId !=null ? job.sourceId: job.id
+
+                    def findNewPath = changes.find { it.changeType == DiffEntry.ChangeType.ADD && it.newPath && it.newPath.contains(jobSerializerId) }
+                    def findOldPath = changes.find { it.changeType == DiffEntry.ChangeType.DELETE && it.oldPath && it.oldPath.contains(jobSerializerId) }
```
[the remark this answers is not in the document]
ltamaster: @gschueler 
when a job is renamed externally (it is an external event basically), the plugin cannot find the relation between the old path and the new path.
What I tried to do is to find  - according to the git changes - if the job was renamed

Screen Shot 2021-06-17 at 14 28 15

ltamaster:
@Mergifyio rebase

rundeckapp/grails-app/services/rundeck/services/ScmService.groovy
```diff
@@ -965,18 +941,24 @@ class ScmService {
      * @param serializer predefined serializer, or null to create lazy serializer
      * @return JobScmReference
      */
-    JobScmReference scmJobRef(ScheduledExecution job, JobSerializer serializer = null) {
-        def metadata = getJobPluginMeta(job)
-        scmJobRef(job, serializer, metadata)
+    JobScmReference scmImportJobRef(ScheduledExecution job, Map metadata = null) {
```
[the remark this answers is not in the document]
ltamaster: working on that

rundeckapp/grails-app/services/rundeck/services/scm/ScmLoaderService.groovy
ltamaster: on

rundeckapp/grails-app/services/rundeck/services/ScmService.groovy
ltamaster: @gschueler 
with this code
https://github.com/rundeck/rundeck/pull/7030/files#diff-8af69461e09e7c99b0829830e738be41d7d6809725ea047db37711357c3381acR440-R443

using the git changes plus the sourceID (or ID) I am trying to detect that the job was renamed.

plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/GitImportPlugin.groovy
```diff
@@ -514,7 +528,14 @@ class GitImportPlugin extends BaseGitPlugin implements ScmImportPlugin {
...
             case JobChangeEvent.JobChangeEventType.MODIFY_RENAME:
-                importTracker.jobRenamed(reference, path, newpath)
+                if(importTracker.originalValue(newpath)){
+                    //if a job was imported with a new name traker needs to be cleaned
+                    def oldPath=importTracker.originalValue(newpath)
+                    importTracker.renamedTrackedItems.trackItem(oldPath, oldPath)
+                    importTracker.untrackPath(oldPath)
+                }else{
+                    importTracker.jobRenamed(reference, path, newpath)
+                }
```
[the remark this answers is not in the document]
ltamaster: I will add that

rundeckapp/grails-app/services/rundeck/services/ScmService.groovy
gschueler: here's the problem.  set up import using "archive" UUID behavior.  When imported, the uuid is removed, replaced with a new uuid, and the source uuid is stored in the job metadata (now under the "scm-import" name).  Then setup SCM export, using the export UUID behavior "original", this is meant to use the original source id in the export pattern, such as `${job.sourceId}.${config.format}`. However, the scm-export metadata does not contain the source Id anymore.  It used to work when the only metadata was stored with the "scm-import" metadata name.

gschueler (changes requested):
* [ ] After external rename, before import, The scm import status still shows as "1 unimported files found, 1 tracked files were deleted"
* [x] The diff shows "import needed" and "file has been deleted from scm"
* [x] Performing the scm import works correctly. It no longer deletes/adds the job and maintains the execution history of the job.

After external rename status:
Screen Shot 2021-07-06 at 9 55 49 AM

diff:

Screen Shot 2021-07-06 at 9 56 18 AM

rundeckapp/grails-app/services/rundeck/services/ScmService.groovy
ltamaster: @gschueler I added more code to this case. The idea is to check (when the SCM export is enabled) if there is an SCM import metadata with a `sourceId`  attribute set

gschueler (approved):
- [x] project status shows job rename state. This seems to work now. 

I did see an instance of the same result where the tooltip for the project status showed "1 unimported file(s) found, 1 tracked file(s) were deleted", however when I refreshed the job list, it correctly showed "1 file was renamed" result, so I think it may have been resolving the status when I first loaded the page.

- [x] scm export correctly uses sourceId from original import now

PR #7031, opened 2021-05-04
Fix: externally deleted job not reflected in scm import status

gschueler:

Fixes https://github.com/rundeckpro/rundeckpro/issues/1730

* if a job is deleted in another cluster member, the scm import status is not reflected

detect jobs that were deleted and emit an event so that the scm plugin listeners are notified of deleted jobs

* does not fix issue with hibernate cache causing 500 error

**Additional context**
There may be two events emitted for deleted jobs, if both import and export plugins are configured for a project. In that case both import and export loader tasks will emit a DELETED job change event, but the handlers of those events are idempotent, and so it will not affect the behavior. a future refactor could consolidate the loader code for import/export plugins and emit a single event


ronaveva (approved):
It now shows the import action in other members.
LGTM

PR #7045, opened 2021-05-10
Export Variables available to Notifications

imad6639:
Enhancement - Allow export variables to be available for use in notifications for jobs and referenced jobs

When notifications are triggered, and map of exported variables are sent to the notification service

**Additional context**
fix rundeckpro/rundeckpro#1625
fix rundeckpro/rundeckpro#1626
fix rundeckpro/rundeckpro#1621
fix rundeckpro/rundeckpro#1616


rundeckapp/grails-app/services/rundeck/services/NotificationService.groovy
```diff
@@ -787,6 +789,9 @@ public class NotificationService implements ApplicationContextAware{
             execution:[id:exec.id,status:state,user:exec.user],
             notification:[trigger:trigger]
         ]
+        if(export)
+            props << export
```
ProTip, line 793: `export` contains a `job` key at this point which ends up replacing the hand-crafted `props` key from line 788. I'm not sure what the impact of that is but:

* It adds a lot more keys to `job`. Was the intention of line 788 to limit the keys available?
* In the case of `group` the null coalesce to empty string gets defeated.

ProTip (changes requested):

ProTip (approved):

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -3804,17 +3812,14 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
...
                         ]
                 )
             }
 
             result.sourceResult = wresult.result
 
-            Map<String, String> data = ((WorkflowExecutionResult) wresult.result)?.getSharedContext()?.getData(ContextView.global())?.get("export")
-            if (data) {
-                executionContext.getOutputContext().addOutput(ContextView.global(), "export", data)
-            }
```
gschueler, line 3817: i think this will break exported data vars for job references in the case where `jitem.ignoreNotifications` is true

PR #7047, opened 2021-05-11
Add loading skeletons

ProTip:
Adds loading skeletons to the following components:

* Project picker
* News

This is a sort of draft implementation. CSS backgrounds are used for the repeatability along the Y-axis without the need to do any height calculations.

If we can't make this work with the desired skeleton style, it may need another solution.

Preference would be to not bring in an external dep just for this, and the needs of each component are rather bespoke if closely matching the loaded content shapes is desired.

skeleton


mishingo (changes requested):
Adjust color of skeleton list and refine sizing
https://github.com/rundeck/rundeck/pull/7062

mishingo (approved):

Issue #1720, opened 2016-03-03
Scheduled jobs are executed twice.


chswtj:
[trigger.example.zip](https://github.com/rundeck/rundeck/files/156982/trigger.example.zip)

We have Rundeck 2.6.2-1 deployed as war file on Tomcat running behind Apache 2.2 on Red Hat Enterprise Linux Server release 6.5 (Santiago).
As DB we use Oracle DB 11gR2 on a second server.

On this setup we are currently executing approx. 4000 jobs per day on our production environment.

Since some days, we see, that a number of scheduled jobs are triggered twice within 1-60 seconds. i.e. a job is triggered and some seconds later again.
We did not see, that is was triggered 3 or more times. Only twice.

All scheduled jobs are marked as "Multiple Executions : NO"
Most of the scheduled jobs are scheduled to run once 1 day, the other ones are scheduled to run each 10 or 15 minutes.
Most of the jobs are running very short time, because they are trigger long running client jobs asynchronously.

In the attachment there is one of thees jobs, that is scheduled for every 10 minutes.
Furthermore there is a snippet of the rundeck log, where it is shown, that the job is triggered twice.

This is only 1 example for all the other jobs, described above.

ahonor:
@chswtj Did you configure the instances to be in cluster mode? I think that will be necessary since they are both pointing to the same DB.

In rundeck-config.properties:
`rundeck.clusterMode.enabled=true`

In framework.properties:
`rundeck.server.uuid=<value generated by uuidgen>`

chswtj:
Thanks for your fast reply.
I followed your advice and let the system run over the weekend.
No all jobs ran only once.
But I noticed, that on Saturday and Sunday night 2 jobs still ran twice.
I exported the jobs, removed them from rundeck and than reloaded them.
Now they seem to run only once as defined.

When I added the properties you mentioned above, I restarted rundeck.
But then now scheduled job was running. I had to reload all scheduled jobs again to get it run.
I hope that this is the normal behaviour.

Thanks again.

PR #7060, opened 2021-05-14
RUN-130: Allowed apostrophe character on first name and last name fields

alexander-variacode:
Issue:
https://github.com/rundeckpro/rundeckpro/issues/1720

Fix description:
Allowed apostrophe character on first name and last name fields.


rundeckapp/grails-app/domain/rundeck/User.groovy
```diff
@@ -41,8 +41,8 @@ class User {
     static hasMany = [reportfilters:ReportFilter,jobfilters:ScheduledExecutionFilter,nodefilters:NodeFilter]
     static constraints={
         login(matches: '^[a-zA-Z0-9\\.,@\\(\\)\\s_\\\\/-]+$')
```
qualman, line 43: Did we want to expand this regex to include `\\p{L}\\p{M}` as well, so as to include all UTF-8 alpha-numeric characters, and/or just remove the validation? That way we address both #1720 and #1806?

jtobard: In my opinion, if we have clients that are using characters outside that regex that come from external integrations, as is the case rundeckpro/rundeckpro#1806 we should leave this field with the new regex that is more universal expecting more users using different alphabets.

qualman: @fdevans , how do you feel about expanding this regex to something like `login(matches: '^[a-zA-Z0-9\\p{L}\\p{M}\\.,@\\(\\)\\s_\\\\/-]+$')` (suggested by @alexander-variacode) in order to address rundeckpro/rundeckpro#1806 and allow users with additional character sets to use their normal alphabet as a login name?

gschueler: let's just add what is needed to support the additional unicode characters for now

qualman:
No errors thrown in my testing as mentioned in the issue. Looks good!!

PR #7069, opened 2021-05-18
Fix project picker view all link

ProTip:
Add app link for `menuHome` and use this for the link. This should account for `rdbase`.

Using `rdbase` as the `href` as it redirects to the projects list. This didn't seem explicit enough..


imad6639:
tested this with a deployed war in tomcat and as a standalone as a war with a servlet context path and the `View All Projects` issue is resolved.  However, the `Create New Project` has the same issue with redirecting to ${port}:resources/createProject without the servlet context path

ProTip:
@imad6639 I don't see a `Create New Project` link. Is this somewhere else?

image

imad6639:
Seems to be only available when there are no projects

Screen Shot 2021-05-18 at 1 31 31 PM

ProTip:
@imad6639 The most recent commit should address that issue.

imad6639 (approved):

PR #7087, opened 2021-05-25
Refactor: auth success user profile updates decoupled from UserService

gschueler:
* Use grails event to trigger user profile updates instead of directly calling UserService
* Fix spring dependency chain issue

Fixes https://github.com/rundeckpro/rundeckpro/issues/1796

Instead of depending on UserClass directly, change `RundeckPreauthSuccessEventHandler` and `RundeckJaasAuthenticationSuccessEventListener` to send a grails event with user detail for updating user profile. Modify UserService to react to the event.

Attempt to alter spring dependency chain to be less strict at load time did not work.


sjrd218 (changes requested):
Looks like lines 691 and 692 of resources.groovy need to be removed and a ref to configurationService needs to be added.

gschueler:
@sjrd218 updated

sjrd218 (approved):
Worked for me.

PR #7089, opened 2021-05-26
Fix: bundled groovy plugins should be overwritten when extracted

gschueler:

Fix: bundled groovy plugins are not named with version number in the name, so should overwrite previous versions if they exist.

Set overwrite flag to true when extracting groovy plugins

**Additional context**

For https://github.com/rundeckpro/rundeckpro/issues/1798


sjrd218 (approved):

fdevans:
Will this work as a backport or does it need a separate PR?

PR #7106, opened 2021-06-09
Fix rundeckpro-1824 ent logstorage config [RUN-78]

gschueler:

Fixes https://github.com/rundeckpro/rundeckpro/issues/1824

* A groovy plugin containing no definition would be registered but later cause NPE loading the plugin

* prevent NPE when loading spring bean plugins
* avoid registering groovy script plugins until after they are constructed


sjrd218 (approved):
Works without error for me.

PR #7115, opened 2021-06-14
RUN-75: Adding maxsize and expiration time for fileUploadService config on RundeckConfigBase

carlosrfranco:
Fixes https://github.com/rundeckpro/rundeckpro/issues/1826

It was not possible to upload larger files since the config prop settings `rundeck.fileUploadService.tempfile.maxsize` was not working.

The property should be included on RundeckConfigBase class

**Additional context**
The grails config properties `grails.controllers.upload.maxFileSize` and `grails.controllers.upload.maxRequestSize` must also be defined since the default value on Grails application is approximately 25Mib or 26214400


gschueler (approved):
Please add documentation about the grails config to the docs https://docs.rundeck.com/docs/administration/configuration/config-file-reference.html#job-file-option-uploads

PR #7118, opened 2021-06-15
RUN-89 Add UI profile and search to plugin list endpoint

ProTip:
This enhancement adds UI profile(`iconUrl`, `providerMetadata`) to the public `plugin/list` API endpoint along with the query parameter `service` to filter the results by provider.

**Additional Context**
This is part of ongoing efforts to move functionality out of UI specific endpoints and into the public API.

### DIFF
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/PluginController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/PluginController.groovy
@@ -1,5 +1,6 @@
 package rundeck.controllers
 
+import com.dtolabs.rundeck.app.api.plugins.ApiPluginListProvider
 import com.dtolabs.rundeck.app.support.PluginResourceReq
 import com.dtolabs.rundeck.core.authorization.AuthContext
 import com.dtolabs.rundeck.core.plugins.configuration.PropertyScope
@@ -8,7 +9,6 @@ import org.rundeck.app.spi.AuthorizedServicesProvider
 import org.rundeck.core.auth.AuthConstants
 import com.dtolabs.rundeck.core.plugins.PluginValidator
 import com.dtolabs.rundeck.core.plugins.configuration.PluginAdapterUtility
-import com.dtolabs.rundeck.server.plugins.services.UIPluginProviderService
 import grails.converters.JSON
 import groovy.transform.CompileStatic
 import org.springframework.web.multipart.MultipartFile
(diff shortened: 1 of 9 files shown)


rundeckapp/src/test/groovy/rundeck/services/PluginApiServiceSpec.groovy
```diff
@@ -63,11 +66,14 @@ class PluginApiServiceSpec extends Specification implements ServiceUnitTest<Plug
...
         when:
         1 * service.rundeckPluginRegistry.getPluginMetadata(_,_) >> fakeMeta
+        1 * service.uiPluginService.getProfileFor(_, _) >> [icon: 'foo']
+        1 * service.grailsLinkGenerator.link(_) >> 'http://localhost:8080/icon'
         def response = service.listPlugins()
         def service = response[0]
         def entry = service.providers[0]
 
         then:
+        println(entry)
```
ProTip, line 76:
```suggestion
```

rundeckapp/grails-app/services/rundeck/services/PluginApiService.groovy
```diff
@@ -268,7 +272,23 @@ class PluginApiService {
                  rundeckCompatibilityVersion: rdVer,
                  targetHostCompatibility: tgtHost,
                  pluginDate   : toEpoch(dte),
-                 enabled      : true]
+                 enabled      : true] as LinkedHashMap<Object, Object>
+
+                if(service != "UI") {
```
gschueler, line 277: need to do an import but preferred to using constant string:

```suggestion
                if(service != ServiceNameConstants.UI) {
```

but, is there a reason to skip UI plugins here?

gschueler (approved):

rundeckapp/grails-app/controllers/rundeck/controllers/PluginController.groovy
```diff
@@ -138,10 +138,15 @@ class PluginController extends ControllerBase {
     }
 
     def listPlugins() {
+        String service = params.service
+
         def providers = []
         pluginApiService.listPlugins().each { svc ->
+            if (service && service != svc.service)
```
gschueler, line 145: perhaps this needs an API version bump and version check here

rundeckapp/grails-app/services/rundeck/services/PluginApiService.groovy
ProTip: That's a very good question. I brought this bit of logic over from the non-public endpoint that was being used. IIRC I got an error without the check, but in either case I don't enjoy this bit as it's not clear at all why "UI" is an exception. Would need to do further debugging to figure out why.

rundeckapp/grails-app/controllers/rundeck/controllers/PluginController.groovy
ProTip: This should function the same without the new, optional service parameter supplied. Are we still bumping API version for non-breaking changes?

gschueler: the api bump is so that folks using previous versions don't see the API docs and expect the results returned here. so in this case it's primarily to communicate a required version level for the new features, allowing them to be documented accurately

ProTip:
@gschueler I've:

* Added a typed response object using ApiResource
* Implemented a new functional test to ensure the fields are present/missing based on API version endpoint used(and to add a spot for future tests)
* Updated to use service constant

test/selenium/src/__tests__/api/plugins.test.ts
```diff
@@ -0,0 +1,45 @@
...
+        const resp = await cluster.client.apiRequest({
+            method: 'GET',
+            pathTemplate: 'api/39/plugin/list'
+        });
+
+        const found = (<Array<any>>resp.parsedBody).find(i => i.iconUrl || i.providerMetadata)
+        expect(found).toBeFalsy()
+    })
+
+    it('Returns v40 fieldd', async () => {
```
ProTip, line 28:
```suggestion
    it('Returns v40 field', async () => {
```

gschueler (approved):

PR #7120, opened 2021-06-16
RUN-124: Scope Ace setup to execution plugin tab

ProTip:
fixes rundeckpro/rundeckpro#1846
[RUN-124]

**Bug**: This Ace setup was introduced in #7082 to attach the editor to lifecycle plugin property fields. It caused duplicate editors in at least one case.

I scoped the jQuery selector to the `#tab_execution_plugins` element so it would not apply to other DOM locations. I'm actually unsure if this satisfies the original intent.

[RUN-123]: https://pagerduty.atlassian.net/browse/RUN-123

[RUN-124]: https://pagerduty.atlassian.net/browse/RUN-124


gschueler (approved):
limiting to the execution plugins is fine. We may want it enabled for some of the other tabs as well, but that was the purpose of the original change

PR #7121, opened 2021-06-16
RUN-64: Fix: optionValuePlugins list disappears when when a validation errors

imad6639:
bugfix, optionListValue was not sent back to opsEdit view on validation error
return the optionValuesPlugins to the view on validation error exactly as it's sent during the initial edit


gschueler (approved):
would be good to add a test on the controller, but approving it as is

fdevans:
Added #7135 to capture the test request from above.

PR #7122, opened 2021-06-16
RUN-232: Job import does not preserve option value ordering

### LINKED ISSUE #6364: job definition load does not preserve option value ordering
**Describe the bug**

When importing a job using the `rd jobs load` command, defined option values get sorted during the import process.

**My Rundeck detail**

* Rundeck version: 3.1.0
* Rundeck CLI version: 1.3.1
* install type: rpm
* OS Name/version: CentOS 7
* DB Type/version: mysql

**To Reproduce**

1. Create a new job in RunDeck with a single option containing allowed values "E,D,C,B,A", inspecting the job shows the ordering for the option as defined:
   initial job
2. Export the job using `rd jobs list --project Sandbox --idlist 83efa98c-faa8-4ffc-b169-ecd9cad4fd05 --format yaml --file test.yaml`.  The resulting yaml for the option is in the defined order:
   ```yaml
     options:
     - enforced: true
       name: Ordering
       required: true
       value: C
       values:
       - E
       - D
       - C
       - B
       - A
       valuesListDelimiter: ','
   ```
3. Import the job using `rd jobs load --project Sandbox --format yaml --duplicate update --file test.yaml`, inspecting the job following import shows the ordering for the option is now alphabetic:
   post import

**Expected behavior**

After importing the job I expect the Allowed Values to have the order originally defined and correctly captured in the yaml file. I.e. E,D,C,B,A

 - OS: Ubuntu 20.04
 - Browser: Firefox / Chrome
 - Version: N/A

**Additional context**

None

ajxb:
Fixes #6364 

A potential bugfix for #6364 - I thought it would be simpler and clearer to submit a PR with the proposed changes.

I've revised my solution as I realised that it wasn't simple to change `values` from Set to List without unintended consequences.

I've updated the code to assign `data.values` to `optionValues` at the same time as the assignment to `values`.

I've also updated public method `productValuesList()` to use `optionValues` to populate the `valuesList` if `optionValues` is set, otherwise it drops back to the default functionality of using `values` if it is set.

**Additional context**

I'm unsure if this change will have unintended consequences due to my lack of experience with Java and the size of the Rundeck codebase.  Please be gentle!

### DIFF
--- a/rundeckapp/grails-app/domain/rundeck/Option.groovy
+++ b/rundeckapp/grails-app/domain/rundeck/Option.groovy
@@ -270,6 +270,7 @@ public class Option implements Comparable{
             }else{
                 opt.valuesListDelimiter=DEFAULT_DELIMITER
             }
+            opt.optionValues=new ArrayList<String>(data.values);
             opt.valuesList = opt.produceValuesList()
             opt.values = null
         }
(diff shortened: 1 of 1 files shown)


fdevans:
Added this to our review queue.

imad6639 (approved):

PR #7123, opened 2021-06-16
Fix: option values enabled by default dev mode

gschueler:
Fix dev mode issue: option values plugin feature flag was not enabled in dev mode by default.

also clean up the feature flag configs in application.groovy


sjrd218:
Should we put all these in application.yml and stop using application.groovy entirely? It'd be nice to have 1 file as the source for initial configuration.

sjrd218 (approved):

Issue #7130, opened 2021-06-22
3.4.0 UI bug : When displaying overview page with a big amount of projects ( over 15 ) the last ones may not load "Action" button correctly 


mazzella-c:
**Describe the bug**

The "Projects" overview page may not display correctly the interface button "Action" and "Activity"for the last projects in the list.

**My Rundeck detail**

* Rundeck version: 3.4.0
* install type: WAR
* OS Name/version: RHEL 7.7
* DB Type/version: postgre

**To Reproduce**

1. Log into a Rundeck instance with over 15 projects
2. Click on "Home" to list all the Projects
3. Scroll down to display more projects, "Activity" is empty and "Action" button is unavailable

**Expected behavior**

After scrolling to see the remaining projects, the remaining projects should load the buttons initially outside of the viewport.

**Desktop (please complete the following information):**

 - OS: Windows
 - Browser : Firefox 88, Chrome 87

**Additional context**

See #4149

MegaDrive68k:
I see a similar behavior resizing the browser window:

In my case:

1. Resize the browser Window.
2. Enter to Projects page.
3. Scroll Down.
4. GUI elements still loading:

![Screenshot_2](https://user-images.githubusercontent.com/51376003/122939232-75587b00-d341-11eb-8563-5a543567a511.png)

PR #7133, opened 2021-06-23
RUN-141: Fix: UI Bug with Projects list that hinders display over 15 projects in a list.

mishingo:
Bugfix: In the default project list on the Home page, if there are more than 15 projects the remaining projects do not load
https://github.com/rundeck/rundeck/issues/7130

It appears that there is a conflict in the `project-list-item` selector since the data attribute is added by knockout after load. By referencing a child element of the selector that is available in the DOM on load, allows for the waypoint to trigger.


ProTip:
Good find!

ProTip (approved):

Issue #7125, opened 2021-06-17
RUN-137: 3.4.0 UI Bugs


hs-hub-world:
Hello,
  Thank you for the new UI release. FYI...

Using Chrome browser:

![image](https://user-images.githubusercontent.com/20629911/122414807-448cd600-cf3c-11eb-9024-1d856872151b.png)

![image](https://user-images.githubusercontent.com/20629911/122414833-48b8f380-cf3c-11eb-82cd-95f6621f57c0.png)

![image](https://user-images.githubusercontent.com/20629911/122414847-4c4c7a80-cf3c-11eb-9c68-8232130fddf6.png)

Version:
![image](https://user-images.githubusercontent.com/20629911/122415112-7bfb8280-cf3c-11eb-9915-1ec62a527585.png)

Thanks,
HS

gschueler:
fixed by https://github.com/rundeck/rundeck/pull/7140

PR #7140, opened 2021-06-25
RUN-137: Fix 3.4.0 UI regressions

ProTip:
Addresses #7125

### Search modal freeze
This is not a regression in `3.4.x` and needs some further discussion. I have split this issue off into its own: #7141

### Job number tooltip
Adjusted positioning:
image

### Commands node info modal
A change was introduced to the `.table` class for a specific screen that caused the table columns to collapse on all work break opportunities. I moved this change to the `.table-fixed` class that was applied to the table from the previous fix to scope it more narrowly. This table needs some work but the previous, more usable display is restored:

image


mishingo:
@Mergifyio rebase

mishingo (approved):

PR #7142, opened 2021-06-28
RUN-63 and RUN-155: Fixes Sort list of plugins by key order.

imad6639:
enh, sort list of plugins by key in order for easy readability.  Add a test for the controller method
add groovy sort closure to collection used to display plugin list

### DIFF
--- a/rundeckapp/grails-app/views/scheduledExecution/_optEdit.gsp
+++ b/rundeckapp/grails-app/views/scheduledExecution/_optEdit.gsp
@@ -388,7 +388,7 @@
                         </div>
                 <feature:enabled name="optionValuesPlugin">
                     <!--List OptionValuesPlugins here -->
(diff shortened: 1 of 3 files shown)


rundeckapp/grails-app/views/scheduledExecution/_optEdit.gsp
```diff
@@ -388,7 +388,7 @@
                         </div>
                 <feature:enabled name="optionValuesPlugin">
                     <!--List OptionValuesPlugins here -->
-                    <g:each in="${optionValuesPlugins}" var="optionValPlugin">
+                    <g:each in="${optionValuesPlugins?.sort{a,b->a.key<=>b.key}}" var="optionValPlugin">
```
ltamaster, line 391: just curious @imad6639 
maybe this sort can be done in the controller/service instead of the view (it could be better to test it)

ltamaster (approved):
tested

PR #7153, opened 2021-07-12
RUN-222: SCM Loader enhancement: avoid failed plugin will be init on background iteration

ltamaster:
enhancement:
avoid SCM Loader try to load a failed plugin integration in each iteration (for example because the SCM folder is wrong or for authentication issues)

- add a list that saves failed plugins 
- each time the SCM background process fails to init an SCM plugin, that failed plugin is storage in the list
- if the plugin changes and it doesn't fail, it will be removed from the list.

**Additional context**

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/scm/ScmLoaderService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/scm/ScmLoaderService.groovy
@@ -34,6 +34,8 @@ class ScmLoaderService implements EventBusAware {
     ConfigurationService configurationService
     public static final long DEFAULT_LOADER_DELAY = 0
     public static final long DEFAULT_LOADER_INTERVAL_SEC = 20
+    public static final long INIT_RETRY_TIMES = 5
+    public static final long INIT_RETRY_TIMES_DELAY = 1000
 
     /**
      * scheduledExecutor to load job SCM cache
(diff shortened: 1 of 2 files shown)


gschueler (changes requested):
This works as described: if the scm config does not work, startup no longer continually spits out errors, and it shows as Disabled in the GUI.

But there is a case where the repeated attempts was good: when the remote server is temporarily down.  When it retries repeatedly at least eventually it reconnects and is able to init the plugin.  

I think we should add a retry limit, where it repeats at least a few times (3-5 times) before disabling the plugin.  It should also perhaps implement a backoff to less frequently if it fails repeatedly.

gschueler:
I will also file a backlog item to add additional UI feature to show admin when the init fails.  Otherwise there is no way to know that the plugin was disabled because of an init failure.

rundeckapp/grails-app/services/rundeck/services/scm/ScmLoaderService.groovy
```diff
@@ -115,18 +117,33 @@ class ScmLoaderService implements EventBusAware {
...
+                            } catch (Throwable t) {
+                                retryCount++
+                                log.error("processMessages error: $project/$integration: ${t.message} retrying ${retryCount}/${retryTimes}")
+                                Thread.sleep(retryDelay)
+
+                                if(retryCount>retryTimes){
+                                    scmFailedProjectInit.put(projectIntegration, pluginConfigData)
+                                    process = true
+                                    removingLoaderProcess(project, integration)
+                                }
```
gschueler, line 143: would probably be better to work like this pseudocode:

```groovy
if(retryCount>retrytimes){
  //...
}else{
  log.error("...retrying...")
  Thread.sleep()
}
```

rundeckapp/grails-app/services/rundeck/services/scm/ScmLoaderService.groovy
```diff
@@ -115,18 +117,33 @@ class ScmLoaderService implements EventBusAware {
...
+                                process = true
 
-                        } catch (Throwable t) {
-                            log.error("processMessages error: $project/$integration: ${t.message}")
-                            scmFailedProjectInit.put(projectIntegration, pluginConfigData)
-                            removingLoaderProcess(project, integration)
+                            } catch (Throwable t) {
+                                retryCount++
+                                log.error("processMessages error: $project/$integration: ${t.message} retrying ${retryCount}/${retryTimes}")
+                                Thread.sleep(retryDelay)
```
gschueler, line 137: the retry delay is fixed here, but maybe it should back off like: `Thread.sleep(retryDelay*retryCount)`

PR #7158, opened 2021-07-15
RUN-121: Remove 1 second timeout for request

### LINKED ISSUE #7116: Loading skeletons have minimum display time
**Describe the bug**
The project selector and news widget appear to have about a **1s** minimum loading display time left-over in the UI stores from development.

These delays should be removed to improve the responsiveness of the components.

**My Rundeck detail**

* Rundeck version: 3.4.0

**To Reproduce**
1. Start an empty Rundeck instance and create a test project
2. Click the project selector to expand
3. Refresh the page and repeat `2`
4. Notice that even though the projects response comes back fast from the server, the loading animation is always shown for the same amount of time.

**Expected behavior**
Content should display immediately when available.

mishingo:
Fixes #7116 

Bugfix: Project Picker always took 1 sec to load 

Remove the timeout that cause the project picked to be forced to load after 1 sec

### DIFF
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/stores/Projects.ts
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/stores/Projects.ts
@@ -17,7 +17,6 @@ export class ProjectStore {
 
         this.projects = []
         const resp = (yield this.client.projectList()) as ProjectListResponse
-        yield new Promise(res => {setTimeout(res, 0)})
         resp.forEach(p => {
             this.projects.push(Project.FromApi(p))
         })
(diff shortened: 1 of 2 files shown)


rundeckapp/grails-spa/packages/ui-trellis/src/stores/Projects.ts
```diff
@@ -17,7 +17,7 @@ export class ProjectStore {
 
         this.projects = []
         const resp = (yield this.client.projectList()) as ProjectListResponse
-        yield new Promise(res => {setTimeout(res, 1000)})
+        yield new Promise(res => {setTimeout(res, 0)})
```
ProTip, line 20: It might be okay to remove this line if you'd like.

ProTip (commented):
IIRC there should be an identical sleep for the community news.

ProTip (approved):

PR #7166, opened 2021-07-17
RUN-255: Enable Cookie name and domain customization

laubstein:
It's an enhancement. With `rundeck.multiURL.enabled` (https://github.com/rundeck/rundeck/pull/6760) we can use multiple url's, but we still need to login on each of them.

The code basically externalize two cookie properties: `name` and `domainNamePattern` ([spring docs](https://docs.spring.io/spring-session/docs/current/reference/html5/guides/java-custom-cookie.html#custom-cookie-options))

No other solutions.

**Additional context**
Example:

```
rundeck.multiURL.enabled=true
rundeck.security.cookie.name=RDK_SESSION
rundeck.security.cookie.domainNamePattern=^.+?\\.(\\w+\\.[a-z]+)\$
```

rdk_cookies

### DIFF
--- a/rundeckapp/grails-app/conf/spring/resources.groovy
+++ b/rundeckapp/grails-app/conf/spring/resources.groovy
@@ -150,20 +150,21 @@ beans={
         }
     }
 
-    def defaultCookieSerializer = new DefaultCookieSerializer()
-
-    if (application.config.rundeck.security?.cookie?.domainNamePattern?.trim()) {
-        defaultCookieSerializer.setDomainNamePattern(application.config.rundeck.security.cookie.domainNamePattern.toString().trim());
+    rundeckCookieSerializer(DefaultCookieSerializer) {
+        if (application.config.rundeck.security?.cookie?.domainNamePattern?.toString()?.trim()) {
+            domainNamePattern = application.config.rundeck.security.cookie.domainNamePattern.toString().trim()
+        }
+        cookieName = application.config.rundeck.security?.cookie?.name?.toString()?.trim() ?: "JSESSIONID"
     }
 
-    def cookieName = application.config.rundeck.security?.cookie?.name?.trim() ?: "JSESSIONID"
-    defaultCookieSerializer.setCookieName(cookieName)
+    rundeckMapSessionRepository(MapSessionRepository, new ConcurrentHashMap())
 
-    def sessionFilter = new MapSessionRepository(new ConcurrentHashMap<>())
-    def sessionIdResolver = new CookieHttpSessionIdResolver()
-    sessionIdResolver.setCookieSerializer(defaultCookieSerializer)
-    springSessionRepositoryFilter(SessionRepositoryFilter, sessionFilter) {
-        httpSessionIdResolver =  sessionIdResolver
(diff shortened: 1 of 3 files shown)


rundeckapp/grails-app/conf/spring/resources.groovy
```diff
@@ -144,6 +149,23 @@ beans={
...
     }
+
+    def defaultCookieSerializer = new DefaultCookieSerializer()
+
+    if (application.config.rundeck.security?.cookie?.domainNamePattern?.trim()) {
+        defaultCookieSerializer.setDomainNamePattern(application.config.rundeck.security.cookie.domainNamePattern.toString().trim());
+    }
+
+    def cookieName = application.config.rundeck.security?.cookie?.name?.trim() ?: "JSESSIONID"
+    defaultCookieSerializer.setCookieName(cookieName)
```
gschueler, line 160: the preferred way to define beans is with the bean builder syntax, instead of instantiating beans in this location

for example:

```.groovy
rundeckCookieSerializer(DefaultCookieSerializer){
    if (application.config.rundeck.security?.cookie?.domainNamePattern?.trim()) {
        domainNamePattern=application.config.rundeck.security.cookie.domainNamePattern.toString().trim()
    }
    cookieName = application.config.rundeck.security?.cookie?.name?.trim() ?: "JSESSIONID"
}
```

then for the session filter it would be something like this:

```.groovy
rundeckMapSessionRepository(MapSessionRepository, new ConcurrentHashMap())
rundeckSessionIdResolver(CookieHttpSessionIdResolver){
    cookieSerializer = ref('rundeckCookieSerializer')
}
springSessionRepositoryFilter(SessionRepositoryFilter, ref('rundeckMapSessionRepository')) {
        httpSessionIdResolver =  ref('rundeckSessionIdResolver')
}
```

gschueler:
@laubstein It seems we are already doing some cookie config for spring boot and the embedded Jetty in our app config here: https://github.com/rundeck/rundeck/blob/main/rundeckapp/grails-app/conf/application.yml#L309-L319

I'm not sure what the effect of your pr will have on those settings.  Does the newly defined bean make those obsolete?

gschueler:
@laubstein I don't think we can accept this pr as it currently stands.  From what I understand of the spring-session-core beans that are being added, the `springSessionRepositoryFilter` bean doesn't support session expiration.  Is that correct?
Perhaps the pr can only configure the beans if a specific configuration flag is enabled.

laubstein:
> @laubstein I don't think we can accept this pr as it currently stands. From what I understand of the spring-session-core beans that are being added, the `springSessionRepositoryFilter` bean doesn't support session expiration. Is that correct?
> Perhaps the pr can only configure the beans if a specific configuration flag is enabled.

@gschueler you're right. As `jetty.initParams` are used, I'm changing the code to keep using this.

(sorry for the late answer, my last month was crazy)

fdevans:
Hi @laubstein.  How is this update going?  We'd like to confirm it's still coming or close the PR.  Can you let us know by next week?

gschueler:
closing this pr for lack of updates

PR #7173, opened 2021-07-20
RUN-194: Fix errors from 1000+ Executions on Oracle Database

carlosrfranco:
Fixes https://github.com/rundeckpro/rundeckpro/issues/1873

Fix ORA-01795 error when using oracle database if there are more than 1000 executions for a referenced job

The executions id list should be partitioned if there are more than 1000 ids to be included on the query. This issue happens only if using oracle database


gschueler (approved):
I did not test with oracle database, but the code is acceptable

PR #7179, opened 2021-07-21
RUN-190: Remove old release version arg from setversion.sh

qualman:
Checked in with @gschueler and it seems like we no longer need to use this old release number argument. I removed it, so as to simplify script usage.

### DIFF
--- a/setversion.sh
+++ b/setversion.sh
@@ -1,29 +1,19 @@
 #!/bin/bash
 
 CUR_VERSION="$(grep version.number= "$PWD/version.properties" | cut -d= -f 2)"
-CUR_RELEASE="$(grep version.release.number= "$PWD/version.properties" | cut -d= -f 2)"
 CUR_TAG="$(grep version.tag= "$PWD/version.properties" | cut -d= -f 2)"
 
 echo "current NUMBER: $CUR_VERSION"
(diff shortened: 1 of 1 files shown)


gschueler (approved):

qualman:
Not sure why the tests are failing here...:
```
OK: [rundeckpro-enterprise-3.4.2-SNAPSHOT.war] WEB-INF/rundeck/plugins/aws-s3-model-source-v1.0.6.jar.properties MUST exist. Result: (true)
FAIL: [rundeckpro-enterprise-3.4.2-SNAPSHOT.war] WEB-INF/rundeck/plugins/py-winrm-plugin-2.0.12.zip MUST exist. Result: (false)
FAIL: [rundeckpro-enterprise-3.4.2-SNAPSHOT.war] WEB-INF/rundeck/plugins/py-winrm-plugin-2.0.12.zip.properties MUST exist. Result: (false)
OK: [rundeckpro-enterprise-3.4.2-SNAPSHOT.war] WEB-INF/rundeck/plugins/openssh-node-execution-2.0.1.zip MUST exist. Result: (true)
```

Any ideas? I branched from an up-to-date copy of main (master for the plugins)

PR #7182, opened 2021-07-22
RUN-262: Possibility to use redis to store sessions

laubstein:
It's an enhancement. While working on #7166 I realized that distributed sessions can be a useful feature to the community.

This implementation introduces 8 new options:

- rundeck.session.storeType - at this moment, only `redis` will change session storage behavior.
- rundeck.session.redis.host - redis host. If not specified, a warning will be logged (`Redis session not enabled. Please set rundeck.session.redis.host.`) and store type goes to default (memory).
- rundeck.session.redis.port - redis port. Default: 6379
- rundeck.session.redis.password - redis password.
- rundeck.session.redis.database - redis database. Default: 0
- rundeck.session.redis.maxTotal - redis pool max total connections. Default: 100
- rundeck.session.redis.maxIdle - redis pool max idle connections. Default: 100
- rundeck.session.redis.minIdle - redis pool min idle connections. Default: 10

No other solutions.

**Additional context**

### DIFF
--- a/docker/official/remco/templates/rundeck-config.properties
+++ b/docker/official/remco/templates/rundeck-config.properties
@@ -12,10 +12,24 @@ server.address={{ getv("/rundeck/server/address", "0.0.0.0") }}
 server.servlet.context-path={{ getv("/rundeck/server/contextpath", "/") }}
 grails.serverURL={{ getv("/rundeck/grails/url", "http://127.0.0.1:4440") }}
 rundeck.multiURL.enabled={{ getv("/rundeck/multiurl/enabled", "false") }}
+
+server.servlet.session.timeout={{ getv("/rundeck/server/session/timeout", "3600") }}
+
+# session cookie configuration
 rundeck.security.cookie.name={{ getv("/rundeck/security/cookie/name", "") }}
 rundeck.security.cookie.domainNamePattern={{ getv("/rundeck/security/cookie/domainnamepattern", "") }}
 
-server.servlet.session.timeout={{ getv("/rundeck/server/session/timeout", "3600") }}
+# memory or redis
+rundeck.session.storeType={{ getv("/rundeck/session/storetype", "") }}
+
+{% if getv("/rundeck/session/storetype", "") == "redis" %}
+# redis session configuration if rundeck.session.storeType is "redis"
+rundeck.session.redis.host={{ getv("/rundeck/session/redis/host", "") }}
+rundeck.session.redis.port={{ getv("/rundeck/session/redis/port", "") }}
+rundeck.session.redis.maxTotal={{ getv("/rundeck/session/redis/maxtotal", "100") }}
+rundeck.session.redis.maxIdle={{ getv("/rundeck/session/redis/maxidle", "100") }}
+rundeck.session.redis.minIdle={{ getv("/rundeck/session/redis/minidle", "10") }}
+{% endif %}
 
 {% if exists("/rundeck/database/create") %}
 dataSource.dbCreate = {{ getv("/rundeck/database/create", "") }}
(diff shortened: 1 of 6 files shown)


laubstein:
While doing more tests, I've found situations where the default serializer throws exceptions because some classes does not implements Serializable.

Changed the PR to WIP.

ProTip:
I know that we "abuse" the session pretty heavily by storing per-user caches and other information in it between requests. The serialization issue has come up before if FS sessions are accidently configured(Tomcat), but the resource usage of session(size, access pattern) may come into play as well if out of process storage is used.

laubstein:
@ProTip based on your feedback I did some changes:

1. switched from Jdk serializer to Xml serializer (using [xstream](https://github.com/x-stream/xstream)). This fixed the problems with serialization.
2. switched redis client from [jedis](https://github.com/redis/jedis) to [lettuce](https://github.com/lettuce-io/lettuce-core) as lettuce allows ["read from secondaries write on primary"](https://lettuce.io/core/release/reference/#readfrom.redis-cluster) strategy.

The code is usable¹ and I think that is good enough to a first delivery. I used the following architecture to perform tests:

```ascii
          ┌────────────────────────────────────┐
          │                                    │
          │       ┌───────────┐ ┌──────────┐ ┌─┴─────────┐
          │  ┌────┤ rundeck1  ├─► ldap     ◄─┤ rundeck2  │
          │  │    │           │ │          │ │           │
          │  │    └─┬──────┬──┘ └──────────┘ ├───┬───────┘
          │  │      │      │                 │   │
          │  │      │      │          ┌──────┘   │
          │  │      │      │rw        │rw        │ro
          │  │      │      │          │          │
          │  │      │      │          │          │
          │  │      │      │          │          │
 ┌────────▼──▼────┐ │ ┌────▼──────────▼─┐   ┌────▼────────────┐
 │                │ │ │                 │   │                 │
 │ postgres       │ │ │ redis primary   ├───┤ redis secondary │
 │                │ │ │ read/write      │   │ read only       │
 └────────────────┘ │ └─────────────────┘   └▲────────────────┘
                    │                        │
                    └────────────────────────┘
                      ro

  rundeck1 urls: rundeck1.rundeck.local
                 dev.rundeck1.rundeck.local

  rundeck2 urls: rundeck2.rundeck.local
                 dev.rundeck2.rundeck.local
```
I was able to login on `rundeck1.rundeck.local:4440` and go to the other url's without relogin.

#### Some details about the choices I did

_to solve the serialization problem_

I've tried to change classes and add `implements Serializable` on "all of them". At some point I got that `com.codahale.metrics.Meter` can't be deserialized :disappointed:.

Then I've tried [Jackson](https://github.com/FasterXML/jackson) with some modules and other configurations... again, no success (`No suitable constructor found ...` :disappointed: ).

My last shot was try to use XML, and I got success on this¹.

_to deal with "resource usage of session"_

Changing redis client allows the user to configure redis with secondaries and loadbalance the read operations.

-----

¹ I know that the security framework from xstream is not enabled and the version is a bit old. If we upgrade the version, we start to get errors while unmarshalling. The good part is that xstream evicts all the points of problem and we can create a whitelist.

Example using xstream 1.4.11

```
message             : Security alert. Unmarshalling rejected.
class               : javax.crypto.spec.SecretKeySpec
required-type       : javax.crypto.spec.SecretKeySpec
converter-type      : com.thoughtworks.xstream.XStream$InternalBlackList
path                : /org.rundeck.web.infosec.HMacSynchronizerTokensHolder/manager/secretKey
```

gschueler:
I think we would have to re-evaluate our internal use of session-stored data to see whether this makes sense

PR #7185, opened 2021-07-25
RUN-238: Add pluginSecurity feature to disable plugin uploads

chrismcg14:
Added a feature flag: rundeck.feature.pluginSecurity.enabled

When set to true, plugin upload will not be allowed and the upload plugin menu item will not be displayed

[RUN-238]

[RUN-238]: https://pagerduty.atlassian.net/browse/RUN-238


ltamaster:
@chrismcg14 something important to add is a template entry in the docker official image to set the config variable in rundeck-config.properties

ltamaster:
> @chrismcg14 something important to add is a template entry in the docker official image to set the config variable in rundeck-config.properties
there is a feature template so we don't need it 
https://github.com/rundeck/rundeck/blob/main/docker/official/remco/templates/rundeck-config-features.properties

ltamaster (approved):
LGTM
tested with a war file and docker 
also tested trying to  upload a plugin using the rundeck cli

gschueler:
The `PluginController.installPlugin` action should also check the feature flag @chrismcg14 @ltamaster

PR #7192, opened 2021-07-28
RUN-165: Fix issue when scheduled jobs runs twice in a cluster

carlosrfranco:
fixes https://github.com/rundeckpro/rundeckpro/issues/1678

When a job is edited using the instance of the cluster that does not own the job and this instance becomes the owner after editing, the takeout message is sent to itself, then the owner never removes it from the schedule in quartz. 

Also, after a job is edited and saved, the remote execution node is assigned twice and then, two takeout messages are sent.

Another problem found was that, when unassigning the job from a schedule, the schedule was not being removed because scheduling at the project level does not change the `scheduled` attribute of ScheduledExecution to `true`. Therefore, if another instance of the cluster unassignee the job, the instance that owns it will not know that this job was previously scheduled and does not delete the scheduling in quartz.

The job will assign the remote node only once after being edited and then a new attribute has been added to the message to inform who was the last owner of the job. Thus, all instances that are not the destination of the takeout message will check if it should remove any quartz triggers and remove the message if it was the last owner.

The case when unassigning the job from a schedule was fixed checking if the local instance has any trigger scheduled for itself in addition to checking if the `scheduled` attribute in ScheduledExecution was `true` previously

**Additional context**
PR related: https://github.com/rundeckpro/rundeckpro/pull/1919

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
@@ -3354,15 +3354,9 @@ class ScheduledExecutionService implements ApplicationContextAware, Initializing
 
 
         def modify = true
+        boolean schedulingWasChanged = oldjob.schedulingWasChanged(scheduledExecution)
         if(frameworkService.isClusterModeEnabled()){
-
-            if (oldjob.originalCron != scheduledExecution.generateCrontabExression() ||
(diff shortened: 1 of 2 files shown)


gschueler:
@carlosrfranco i pushed a change, adding a test that fails, to show an existing issue with the `updateScheduleOwner` call https://github.com/rundeck/rundeck/pull/7192/commits/2337aebbb37d503f3c997bd75c9fbc00b88b2838

gschueler (approved):

PR #7199, opened 2021-08-02
RUN-138: Fix: Improve Limit log feature for better behavior with large logs

alexander-variacode:
Issue:
https://github.com/rundeckpro/rundeckpro/issues/1753

Fix description:
Added three new parameters at rundeck-config.properties applied over all projects.

## rundeck-config.properties

rundeck.execution.logs.output.limit = 5MB
rundeck.execution.logs.output.limitAction = truncate

- `rundeck.execution.logs.output.limit` - Show an error when that limit is exceeded. Could be entered either maximum total line-count (e.g. "100"), maximum per-node line-count ("100/node"), or maximum log file size ("100MB", "100KB", etc.), using "GB","MB","KB","B" as Giga- Mega- Kilo- and bytes.
- `rundeck.execution.logs.output.limitAction`: Is the action applied when the limit is achieved. It is triggered only if there is a value at framework.project.log.output.limit. The values accepted are ‘truncate' or 'halt’. The default value if it does not set is 'halt'.


alexander-variacode:
Tested by Jonathan Garces

rundeckapp/grails-app/services/rundeck/services/logging/ThresholdLogWriter.groovy
```diff
@@ -40,13 +47,20 @@ class ThresholdLogWriter extends FilterStreamingLogWriter {
...
         getWriter().addEvent(event)
 
+        if (!warning && threshold.isWarningSizeReached() && warningReached.compareAndSet(false, true)) {
+            LOG.warn(MSG_WARN_SIZE  , threshold.warningSize, threshold.getValue())
+        }
+
         if (!limit && threshold.isThresholdExceeded() && limitReached.compareAndSet(false, true)) {
-            getWriter().addEvent(LogUtil.logError("Log output limit exceeded: " + threshold.description))
+            String msgError = "Log output limit exceeded: ${threshold.description}"
+            LOG.error(msgError)
```
gschueler, line 62: the `LOG.warn` and `LOG.error` do not seem very useful, they will log to the system log, but do not contain any information about which execution has a log log.

rundeckapp/grails-app/services/rundeck/services/FrameworkService.groovy
```diff
@@ -130,6 +147,43 @@ class FrameworkService implements ApplicationContextAware, ClusterInfoService {
...
+
+    /**
+     * @return the log size warning from framework.properties file
+     */
+    String getProjectLogSizeWarning() {
+        return rundeckFramework
+                .getPropertyRetriever()
+                .getProperty(PROJECT_LOG_SIZE_WARNING)
+    }
+
```
gschueler, line 186: avoid using framework properties for global config.  it should use configurationService to retrieve config values

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -1042,8 +1042,18 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
 //
 
         //set up log output threshold
-        def thresholdMap = ScheduledExecution.parseLogOutputThreshold(scheduledExecution?.logOutputThreshold)
-        def threshold = LoggingThreshold.fromMap(thresholdMap,scheduledExecution?.logOutputThresholdAction)
+        boolean logGlobalConfig = frameworkService.getProjectLogGlobalConfig()
+
+        boolean limitFramework = false
```
gschueler, line 1047: the `limitFramework` variable will not work the way I think you are trying to use it.  Please write some tests.

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -1248,6 +1258,35 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
...
+        } else {
+            outputLimit = scheduledExecution?.logOutputThreshold
+            if (!outputLimit) {
+                outputLimit = frameworkService.getProjectLogOutputLimit()
+                limitFramework = true
+            }
+        }
+        log.info("Logging output limit: ${outputLimit}")
+        return outputLimit
+    }
```
gschueler, line 1275: This logic appears to be: if enabled, use only the global limit. if not enabled, and no job limit is enabled, use the global limit. This allows the job limit to exceed the global limit, or enforces a global limit.

I think it should be changed to:

1. if no global limit is set, apply any job limit that is set.
2. if a global limit is set, and a job limit is set, apply `min(global,job)` to the determine the limit. The global limit cannot be exceeded by the job limit if it is set.
3. if only a global limit is configured, use it.

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -1248,6 +1258,35 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
...
+        if (logGlobalConfig) {
+            limitAction = frameworkService.getProjectLogLimitAction()
+        } else {
+            if (limitFramework) {
+                limitAction = frameworkService.getProjectLogLimitAction() ?: scheduledExecution?.logOutputThresholdAction
+            }
+        }
+        log.info("Logging limit action: ${limitAction}")
+        return limitAction ?: LoggingThreshold.ACTION_HALT
+    }
```
gschueler, line 1288: The logic here should also be changed.

1. if a global limit is set, it should default to action Truncate
2. if a job limit is set, it should use the job action, but a global action of Halt should take precedence

i.e.
job limit & no global limit => job action
job limit & global limit & global action halt => halt
job limit & global limit & global action truncate => job action
job limit & global limit & no global action =>  job action

gschueler (changes requested):

rundeckapp/grails-app/services/rundeck/services/logging/ThresholdLogWriter.groovy
alexander-variacode: Done

rundeckapp/grails-app/services/rundeck/services/FrameworkService.groovy
alexander-variacode: Done

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
alexander-variacode: Done... added an extra validation:
no job limit & global limit => global limit action (default: truncate)

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
alexander-variacode: Done

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
alexander-variacode: Done... variable removed

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -1248,6 +1256,51 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
...
+    private String getOutputLimit(ScheduledExecution scheduledExecution) {
+        String outputLimit = ""
+        String globalLimit = configurationService.getString(LoggingThreshold.EXECUTION_LOGS_OUTPUT_LIMIT)
+        String jobLimit = scheduledExecution?.logOutputThreshold
+
+        if (!globalLimit) {
+            outputLimit = jobLimit
+        }
+        else if (globalLimit && jobLimit) {
+            outputLimit = String.valueOf(Math.min(Long.valueOf(globalLimit), Long.valueOf(jobLimit)))
```
gschueler, line 1268: this won't work if the limit is e.g. `1MB/node`.  the comparison is better done inside LoggingThreshold.   Load the job LoggingThreshold, and the global LoggingThreshold (if present), and compare the two.

rundeckapp/grails-app/services/rundeck/services/logging/ThresholdLogWriter.groovy
```diff
@@ -20,33 +20,55 @@ import com.dtolabs.rundeck.core.logging.FilterStreamingLogWriter
...
  * Used to log that a threshold was reached, and optionally truncate further output
  */
 class ThresholdLogWriter extends FilterStreamingLogWriter {
+    static final Logger LOG = LoggerFactory.getLogger(ThresholdLogWriter.class)
+
+    static final String MSG_WARN_SIZE = "The log file size achieved more than {} bytes - current size: {} bytes"
+
     LoggingThreshold threshold
     final boolean truncate
+    Execution execution
```
gschueler, line 39: it seems the only use for this `execution` is retrieving the job name.  It would be better to pass in a String to include in the warning message at this location https://github.com/rundeck/rundeck/pull/7199/files#diff-784bda3682bff4a5c6ff2ed5b7db3bd2676701746ed28e6a3fbfd82244a7b185R184

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -1248,6 +1256,51 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
         }
     }
 
+    private String getOutputLimit(ScheduledExecution scheduledExecution) {
```
gschueler, line 1259: please add tests for getOutputLimit

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -1248,6 +1256,51 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
...
+            outputLimit = String.valueOf(Math.min(Long.valueOf(globalLimit), Long.valueOf(jobLimit)))
+        }
+        else if (globalLimit && !jobLimit) {
+            outputLimit = globalLimit
+        }
+        log.info("Logging output limit: ${outputLimit}")
+        return outputLimit
+    }
+
+    private String getLimitAction(ScheduledExecution scheduledExecution) {
```
gschueler, line 1277: please add tests for getLimitAction

gschueler (changes requested):

rundeckapp/grails-app/services/rundeck/services/logging/ThresholdLogWriter.groovy
alexander-variacode: Done

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
alexander-variacode: Done

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
alexander-variacode: Done

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
alexander-variacode: Done

rundeckapp/grails-app/services/rundeck/services/logging/LoggingThreshold.groovy
```diff
@@ -72,6 +98,68 @@ class LoggingThreshold implements ThresholdValue<Long>, ValueWatcher<Long> {
...
+        }
+        else if (jobLimitMap && globalLimitMap && !globalLimitAction) {
+            limitAction = jobLimitAction
+        }
+        else if (!jobLimitMap && globalLimitMap) {
+            limitAction = globalLimitAction ?: ACTION_TRUNCATE
+        }
+
+        return limitAction
+    }
```
gschueler, line 161: It's not really possible to compare like this

```
String globalLimit = globalLimitMap.maxSizeBytes ?: globalLimitMap.maxLines
            String jobLimit    = jobLimitMap.maxSizeBytes ?: jobLimitMap.maxLines
```
because the max size in bytes vs maximum lines are different units.  The comparison should probably have this precedence order: size, lines, linesPerNode.  So if to find the minimum of two thresholds, it would work more like:

* if types are not the same, choose the one with higher precedence type
* if types are the same, choose lowest value

rundeckapp/grails-app/services/rundeck/services/logging/ThresholdLogWriter.groovy
```diff
@@ -20,33 +20,55 @@ import com.dtolabs.rundeck.core.logging.FilterStreamingLogWriter
...
     void addEvent(final LogEvent event) {
         def limit = limitReached.get()
+        boolean warning = warningReached.get()
         if (truncate && limit) {
             return
         }
         getWriter().addEvent(event)
 
+        if (!warning && threshold.isWarningSizeReached() && warningReached.compareAndSet(false, true)) {
+            LOG.warn(MSG_WARN_SIZE  , threshold.warningSize, threshold.getValue())
```
gschueler, line 64: this warning is still not useful

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -1041,11 +1041,24 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
...
+        String limitAction = LoggingThreshold.getLimitAction(
+                globalThresholdMap,
+                getGlobalLimitAction(),
+                jobThresholdMap,
+                getJobLimitAction(scheduledExecution))
+        LoggingThreshold threshold = LoggingThreshold.fromMap(thresholdMap, limitAction)
+
+        String warnSize = configurationService.getString(LoggingThreshold.EXECUTION_LOGS_SIZE_WARNING, LoggingThreshold.WARN_SIZE_DEFAULT)
+        Map<String, Long> warnSizeMap = ScheduledExecution.parseLogOutputThreshold(warnSize)
+        threshold?.warningSize = warnSizeMap[LoggingThreshold.MAX_SIZE_BYTES] ?: warnSizeMap[LoggingThreshold.MAX_LINES]
```
gschueler, line 1059: it seems like you are trying to define a secondary threshold for warning, using the same format as the normal threshold.  However, then you are setting the size value based on that format, when the threshold unit value may be different.  

Was the warning part of the customer request? It doesn't seem very useful, I think it should be removed

rundeckapp/grails-app/services/rundeck/services/logging/LoggingThreshold.groovy
gschueler: I'd recommend parsing the two LoggingThreshold (global and job values), then have a method in LoggingThreshold that can synthesize the correct threshold like:

```groovy
LoggingThreshold createMinimum(LoggingThreshold other){
    //1. if the same type, use the lowest maxValue
    //2. if not the same type, use the one with size in bytes first, then lines, then lines per node
    //3. compare the actions, choose halt over truncate
    //4. create a new LoggingThreshold with the new values
}
```

gschueler (changes requested):

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
jtobard: I have the answer to that, the warning came from the customer and CS as a way to monitor this situation using Splunk or any monitor tool. Not mandatory, just valuable for the customer.

jtobard:
PR moved to: #7248

PR #7205, opened 2021-08-04
Job tags not copied when duplicating job 

imad6639:
Retrieve JobComponent values when copying job and add it to the model. fix rundeckpro/rundeckpro#1764 fix run-260

bugfix.  Copy Job did not copy values for JobComponents.  
In the copy method, Job Component Values are retrieved from job being copied from and set in model returned to view
N/A
**Additional context**


chrismcg14 (approved):
Tested

PR #7210, opened 2021-08-10
Allow to customize rundeck.gui.startpage when using Docker image

jacekkow:
In Docker image `rundeck.gui.startpage` configuration option is hardcoded to `projectHome`. This PR allows to modify it using environment variables, identically to all the other options.

### DIFF
--- a/docker/official/remco/templates/rundeck-config.properties
+++ b/docker/official/remco/templates/rundeck-config.properties
@@ -61,7 +61,7 @@ rundeck.api.tokens.duration.max={{ getv("/rundeck/api/tokens/duration/max", "30d
 
 rundeck.log4j.config.file={{ rundeckHome }}/server/config/log4j.properties
 
-rundeck.gui.startpage=projectHome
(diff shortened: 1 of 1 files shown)


fdevans:
Thank you @jacekkow.  Can you share how you will use this and an example? _(just curiousity)_. I will test and approve within this cycle.

jacekkow:
We have recently migrated from Rundeck 2.x (own container image) to Rundeck 3.x (official image) - during the process we have noticed that the default project page has changed from jobs view ("jobs") to dashboard view ("projectHome").

We want to restore previous behavior, but it is impossible without modifying the image. After merging this PR changing behavior would require adding only one parameter (`-e RUNDECK_GUI_STARTPAGE...`):
```
docker run \
  --name rundeck \
  -p 4440:4440 \
  -v data:/home/rundeck/server/data \
  -e RUNDECK_GUI_STARTPAGE=jobs \
  rundeck/rundeck:3.4.2
```

PR #7215, opened 2021-08-12
RUN-310 Fixing modal to save node filter when editing/creating a job. Adding config property to define max number of matched nodes.

carlosrfranco:
fixes https://github.com/rundeckpro/rundeckpro/issues/1894

The modal to save a node filter when editing a job was not being displayed. Another issue was that the maximum matched nodes on filter results was 100 and it was not possible to change this value (Job editing page)

The modal was included on page to be shown and was included a new config property on rundeck-config.properties to define the max count of matched nodes (default is 100)


carlosrfranco:
Docs PR: https://github.com/rundeck/docs/pull/876

PR #7247, opened 2021-09-03
RUN-387: fix: new Ace editor supports min/maxLines for auto sizing of editor

### LINKED ISSUE #7149: Expand yaml editor box
**Describe the issue**
When using the YAML editor in Rundeck UI (for example with Ansible inline playbooks), there are only 7 lines of code visible.
It would be great if this can be extended (with a slider at the corner of the window, or by setting a configuration option).

In the past the field was way larger.
Since I see this in both Ansible and Kubernetes integrations I think it has to do with the UI framework.

**Version**
I believe this started somewhere in Rundeck 3.3; the following two versions have this limitation: for sure
* Rundeck 3.3.10
* Rundeck 3.4.0

rijnier:
This change fixes #7149
The new version of the Ace editor (introduced since 3.3) no longer supports the old method of resizing the editor window.
I've also added minLines so you don't have a single line editor when the textfield is empty.

### DIFF
--- a/rundeckapp/grails-app/assets/javascripts/application.js
+++ b/rundeckapp/grails-app/assets/javascripts/application.js
@@ -383,52 +383,13 @@ function _setupAceTextareaEditor(textarea, callback, autoCompleter) {
 
   //create editor
   var editor = ace.edit(generateId(_shadow));
-  editor.$blockScrolling = Infinity;
-
-  var checkResize;
-  if (data.aceResizeMax) {
-    var heightPx = parseInt(height.replace(/px$/, ''));
-    var lineheight = editor.renderer.lineHeight;
-    var checkSize = Math.floor(heightPx / lineheight);
-    var checkSizeMax = parseInt(data.aceResizeMax);
-    if (checkSize < checkSizeMax) {
-
-      var checkSizeInc = 5;
-      var pxInc = lineheight;
-      checkResize = function (editor) {
-        "use strict";
-        var lineCount = editor.session.getLength();
-        if (lineCount > checkSize && checkSize < checkSizeMax) {
-          var diff = Math.min(checkSizeMax - checkSize, lineCount - checkSize);
-          var increment = Math.max(checkSizeInc, diff);
-          checkSize += increment;
-          heightPx += increment * pxInc;
-          _shadow.css({
-            height: heightPx + 'px'
-          });
-          editor.resize();
-        }
-      };
-
-      if (data.aceResizeAuto) {
-        checkResize(editor);
-      }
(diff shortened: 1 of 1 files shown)


rundeckapp/grails-app/assets/javascripts/application.js
```diff
@@ -383,52 +383,13 @@ function _setupAceTextareaEditor(textarea, callback, autoCompleter) {
...
-    }
-  }
+  editor.setOptions({
+    maxLines: Infinity,
+    minLines: 10
+  });
 
   setAceSyntaxMode(data.aceSessionMode, editor);
   editor.setTheme("ace/theme/chrome");
-  editor.getSession().on('change', function (e) {
```
ProTip, line 421: Is this entire event handler no longer necessary?

rijnier: It seems to only be used to trigger the resize action.
I've tested it without (only using Chrome so far) and I don't see any difference in behaviour so far.

gschueler: I believe the `callback` at least still needs to be triggered by change events, it is used on other pages which setup ace textareas.  I'm not sure about the "synthetic change" part of this, I think @ProTip you were the one who added that code here https://github.com/rundeck/rundeck/commit/95d84adcbfd493538d11108bbf96d0b976718f7a

gschueler: I'd suggest keeping the old behavior, and simply remove the resize part if it is no longer necessary

gschueler (changes requested):
I'd request keeping the old Ace "change" event behavior, and simply remove the resize part if it is no longer necessary

rundeckapp/grails-app/assets/javascripts/application.js
ProTip: @gschueler Sorry for getting back to this so late; the synthetic event was added so plugins could listen to changes more easily. Removing just resize if it is no longer needed is probably the way.

qualman:
I just tested this as a new set of eyes, and it looks good. It looks like we just need you to address the change @gschueler mentioned, @rijnier .

gschueler:
@qualman i made the change

gschueler:
@Mergifyio rebase

qualman:
I resolved the merge conflict for this PR on a new branch, and that is https://github.com/rundeck/rundeck/pull/7375. I am going to close this one, as it is now a slightly stale duplicate, and we'll go ahead and merge in the other.

PR #7248, opened 2021-09-03
RUN-138: Fix: Improve Limit log feature for better behavior with large logs

alexander-variacode:
Fix description:
Added three new parameters at rundeck-config.properties applied over all projects.

## rundeck-config.properties

rundeck.execution.logs.output.limit = 5MB
rundeck.execution.logs.output.limitAction = truncate

- `rundeck.execution.logs.output.limit` - Show an error when that limit is exceeded. Could be entered either maximum total line-count (e.g. "100"), maximum per-node line-count ("100/node"), or maximum log file size ("100MB", "100KB", etc.), using "GB","MB","KB","B" as Giga- Mega- Kilo- and bytes.
- `rundeck.execution.logs.output.limitAction`: Is the action applied when the limit is achieved. It is triggered only if there is a value at framework.project.log.output.limit. The values accepted are ‘truncate' or 'halt’. The default value if it does not set is 'halt'.


alexander-variacode:
@Mergifyio rebase

rundeckapp/grails-app/services/rundeck/services/logging/LoggingThreshold.groovy
```diff
@@ -80,6 +89,66 @@ class LoggingThreshold implements ThresholdValue<Long>, ValueWatcher<Long> {
...
+        t.action = ACTION_HALT
+        if (!job?.action && global?.action) {
+            t.action = global.action
+        }
+        else if (job?.action && !global?.action) {
+            t.action = job.action
+        }
+        else if (!job?.action && global?.action) {
+            t.action = global.action
+        }
```
gschueler, line 143:
```suggestion
```

gschueler: duplicate if branch

gschueler (approved):

PR #7255, opened 2021-09-08
Patch styles

ProTip:
* Fix follow button not showing text after click in execution viewer
* Fix black text in execution show action menu dropdown


gschueler:
that was a load-bearing style

gschueler (approved):

PR #7259, opened 2021-09-10
fix to include payload format and http method on notification importing

carlosrfranco:
fixes https://github.com/rundeckpro/rundeckpro/issues/1830

The payload format and http method was not being included when importing a job

Correction in property name and adjustments in imported data for notification creation


ronaveva (approved):
Tried and it works
image

LGTM

PR #7264, opened 2021-09-14
RUN-345: Add functionality to specify a blocklist of plugins

chrismcg14:
Creates a config entry: rundeck.plugins.providerBlockListFile
If a file if provided to the config entry, then it will parse the file for any blacklisted plugins and not load those plugins.
The yaml will have two different sets of entries fileNameEntries and providerNameEntries. fileNameEntries will be looking for the name of the plugin jar, and will not install it if present. providerNameEntries looks for the service type and the provider names for that service type. This allows to block individual plugins, should numerous be bundled and only certain ones need to be removed.
Format for the yaml:

fileNameEntries:
  - rundeck-ansible
  - rundeckpro-cyberark
  - openssh-node-execution
  - rundeck-copyfile-plugin
  - rundeck-localexec
  - rundeck-script-plugin

providerNameEntries:
  - NodeExecutor: 
    - local
  - FileCopier:
    - ssh-copier
    - jsch-scp
    - script-copy
  - WorkflowNodeStep:
    - nixy-waitfor-local-ping
    - nixy-waitfor-local-port-open
    - filetransfer
  - ResourceModelSource:
    - file
    - directory
    - script
    - local
  - WorkflowStep:
    - github-script-step


gschueler (changes requested):
RundeckPluginBlocklist seems to load and parse the blocklist every time it is queried, it should probably load it once when necessary and cache the data so it is not repeatedly parsing the list.

gschueler (approved):

chrismcg14:
@Mergifyio rebase

PR #7267, opened 2021-09-16
enhancement: Enable system config menuitem beans to use groupings

gschueler:
Adds a way for menuitem bean definitions to define groupings in the System Config menu.

### DIFF
--- a/rundeckapp/grails-app/views/common/_navData.gsp
+++ b/rundeckapp/grails-app/views/common/_navData.gsp
@@ -119,11 +119,11 @@
 
                     <g:if test="${params.project ?: request.project}">
                         <g:ifMenuItems type="PROJECT" project="${projectName}">
-                            <g:forMenuItems type="PROJECT" var="item" project="${projectName}">
+                            <g:forMenuItems type="PROJECT" var="item" project="${projectName}" groupvar="group">
                 {
                     type: 'link',
                     id: 'nav-${item.title.toLowerCase().replace(' ', '-')}-link',
-                    group: 'main',
+                    group: '${enc(attr:group?.id?:'main')}',
                     priority: '${enc(attr: item.priority)}',
                     class: '${enc(attr: item.iconCSS ?: 'fas fa-plug')}',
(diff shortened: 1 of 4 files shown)


rundeckapp/grails-app/views/common/_navData.gsp
```diff
@@ -119,11 +119,11 @@
 
                     <g:if test="${params.project ?: request.project}">
                         <g:ifMenuItems type="PROJECT" project="${projectName}">
-                            <g:forMenuItems type="PROJECT" var="item" project="${projectName}">
+                            <g:forMenuItems type="PROJECT" var="item" project="${projectName}" groupvar="group">
                 {
                     type: 'link',
                     id: 'nav-${item.title.toLowerCase().replace(' ', '-')}-link',
-                    group: 'main',
+                    group: '${enc(attr:group?.id?:'main')}',
```
gschueler, line 126: @ProTip I looked at the NavBar.vue which seems to hardcode only 'main' and 'bottom' groups.  So I wasn't sure whether I should add this change here.  The GroupedMenuItem allows some grouping behavior in the System Config menu (also includes a label for the group), however the dropdown menu has different usage than the project side navbar, not sure if dynamic group name should be enabled here

gschueler, line 126:
```suggestion
                    group: 'main',
```

rundeckapp/grails-app/views/common/_navData.gsp
```diff
@@ -119,11 +119,11 @@
 
                     <g:if test="${params.project ?: request.project}">
                         <g:ifMenuItems type="PROJECT" project="${projectName}">
-                            <g:forMenuItems type="PROJECT" var="item" project="${projectName}">
+                            <g:forMenuItems type="PROJECT" var="item" project="${projectName}" groupvar="group">
```
gschueler, line 122:
```suggestion
                            <g:forMenuItems type="PROJECT" var="item" project="${projectName}">
```

PR #7350, opened 2021-10-28
RUN-465 Fixes duplicated job scheduling after renaming job/group

carlosrfranco:
fixes https://github.com/rundeckpro/rundeckpro/issues/2093

Rundeck is not able to handle scheduling after renaming job making the job run multiple times for the same schedule

Was included the original job/group name on the job take over message in order to handle scheduling after renaming a job

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/jobs/JobReference.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/jobs/JobReference.java
@@ -50,4 +50,14 @@ public interface JobReference {
      * @return server UUID
      */
     String getServerUUID();
+
+    /**
+     * @return original job name in case of renaming
+     */
+    default String getOriginalJobName(){ return null; }
+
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/dtolabs/rundeck/core/jobs/JobReference.java
```diff
@@ -50,4 +50,14 @@
...
+
+    /**
+     * @return original job name in case of renaming
+     */
+    default String getOriginalJobName(){ return null; }
+
+    /**
+     * @return original group name in case of renaming
+     */
+    default String getOriginalGroupName(){ return null; }
```
gschueler, line 62: I think you should rename these to "originalQuartzJobName" and "originalQuartzGroupName" because it can be confusing on what "job" it is referring to.

gschueler (approved):

PR #7357, opened 2021-11-02
RUN-479: Including runAtTime parameter on Run Job Later feature

carlosrfranco:
fixes https://github.com/rundeckpro/rundeckpro/issues/2105

When scheduling a job via `Run Job Later` feature, the job runs immediately (does not wait for the scheduled time)

The parameter `RunAtTime` should be included on the request


ronaveva (approved):
LGTM

PR #7358, opened 2021-11-02
fix height for input

mishingo:
Bugfix: height of select inputs

add class override


MegaDrive68k (approved):
Fixed!

Screenshot_1

PR #7378, opened 2021-11-12
add classes to fix layout

mishingo:
Enhancement: Adjust page layouts to prevent headers that block visibility of the main content

Move from Grid to absolute positioned items, and enable scrolling to the full main content


qualman:
Looking good!

Screen Shot 2021-11-22 at 14 05 39

qualman (approved):

PR #7383, opened 2021-11-16
RUN-493 Add warning message if changing job scheduling on a cluster

carlosrfranco:
fixes https://github.com/rundeckpro/rundeckpro/issues/1895

When changing the scheduling and saving the job very close the next execution of the old scheduling (like 2 or 3 seconds), the job  will execute considering the old scheduling until the take over message is handled.

Since the scheduling changes takes effect after the job take over message is processed (if cluster mode is enabled), a message is shown on the Gui to warn user that the scheduling will take effect after few seconds.


alexander-variacode (approved):
It looks fine and the message appears when the schedule time is changed on job edition form

PR #7386, opened 2021-11-18
Fix issue serving user assets

sjrd218:
When attempting to access users assets in a sub directory eg /user-assets/sub1/asset1.png

Grails does not populate the controllerName correctly so the interceptor helper is
blocking the attempt with a 403. This changes works around the problem by
adding a prefix check for /user-assets if they have been enabled by the user.


chrismcg14 (approved):
LGTM

Issue #2112, opened 2016-10-14
Rundeck Cluster Setup


prashanthwge:
# Bug report

(Note: Please fill in the blank `___` about your rundeck installation below. You can remove this line.)

**My Rundeck detail**
LB is setup 
Clustermode=enabled
UUID : <separate for each server>
- Rundeck version: 2.6.7-1
- install type: war
- OS Name/version: linux 2.6

**Expected Behavior**
When we schedule job it is associated to a specific host under the cluster as these nodes are under LB

**Actual Behavior**
If the servers are up and running all the jobs works fine.  But when one of the node goes down the jobs associated to that node does not execute till the server is back up.  How do we setup a Active/Active HA setup.

**How to reproduce Behavior**
2 servers bring up one instance schedule the job and then stop the server on which the job was scheduled.
# Enhancement request

(_For enhancements_: Please search the existing Issues and look at the [Trello board](https://trello.com/b/sn3g9nOr/rundeck-development) for your idea before posting.)

...

dandunckelman:
As described in ["known shortcomings and caveats"](http://rundeck.org/docs/administration/scaling-rundeck.html#known-shortcomings-and-caveats), this is not something built-in to Rundeck.

You have to do the failover yourself (more of that is documented in the aforementioned documentation).

PR #7389, opened 2021-11-22
RUN-513: option default value bad parsing

L2JE:
editOption gsp template was doing an unnecesary html-encoding to the option default value so it was removed.

Fixes:
[Job Option default value bad parsing #2104](https://github.com/rundeckpro/rundeckpro/issues/2104)
[rundeckpro/rundeckpro #2112](https://github.com/rundeckpro/rundeckpro/issues/2112)

Test:
A test will be added to the qa framework to check that the option data is rendered corretly.


alexander-variacode (approved):
It looks ok for me.

PR #7402, opened 2021-12-01
RUN-519 Allows file option uploaded for execution to be reused when retry execution feature is enabled.

carlosrfranco:
fixes https://github.com/rundeckpro/rundeckpro/issues/2110

If a job has a file type option and retry execution is enabled, the execution get stuck in running state and not attempt to retry since the file is marked as used and then it is deleted on the first attempt

If the execution should retry if it fails, the file should be kept to be used on subsequent attempts


ronaveva (approved):
LGTM

PR #7420, opened 2021-12-09
RUN-541: Fix: Job options misbehaviors when reordering/duplicating

L2JE:
Now we do a shift right before inserting the duplicated option.

Fix:
https://github.com/rundeckpro/rundeckpro/issues/2037


alexander-variacode:
@Mergifyio rebase

PR #7434, opened 2021-12-13
RUN-533: Fixed ldap user information on profile

alexander-variacode:
Issue:
https://github.com/rundeckpro/rundeckpro/issues/2155

Fix description:
Fixed problem when Rundeck tried to save ldap information at the DB.


ronaveva (approved):
LGTM

PR #7436, opened 2021-12-14
RUN-548: Export to another instance fix

L2JE:
Bootstrap extracts the modal from its container and puts it in another. Thus, when the modal shows up it has no parent form to submit.

Now, there's a script that copies the form data into the modal before bootstrap shows it.
 
Fix: [Export to another instance issue](https://github.com/rundeckpro/rundeckpro/issues/2187)


ronaveva (approved):
LGTM

PR #7450, opened 2021-12-22
DRAFT: RUN-551: Convert Workflow Options to Vue

qualman:
An enhancement to convert all of the Workflow Options section of Jobs to a Vue front-end from GSP.

A mostly one-to-one mapping that brings over all UI, and converts basic TagLibrary helpers and interactivity into Vue methods and functionality.

The biggest alternative on the table was bringing more layers over into Vue. For now, this conversion still leverages some JS libraries that were relied upon to bring reactivity into the GSP pages. Down the line, we can and should likely bring those closer to this Vue world. For now, we will knock on them via `window`.

**Additional context**
This relies upon data that is put into/taken out of the global `window._rundeck` data payload. One of the next goals of this conversion should likely be to leverage API calls directly to the controller. This conversion softly attempts that, but still relies on the payload.


qualman:
@imad6639 reintroduced this work in a new PR, so this PR is no longer needed. Closing

Issue #5780, opened 2020-02-11
Enable Job GUI pagination causes API results to be truncated


gschueler:
**Describe the bug**

Setting `rundeck.gui.paginatejobs.enabled=true` ([docs](https://docs.rundeck.com/docs/administration/configuration/gui-customization.html#gui-customization) ) makes the job list in a project paginated.  However the API results are then also paginated, where the job list API doesn't support pagination controls

**My Rundeck detail**

* Rundeck version: 3.2.x

**To Reproduce**
1. set configuration `rundeck.gui.paginatejobs.enabled=true`
2. load more than the pagination max jobs into a project
3. use API or `rd jobs list -p myproject` and see only the pagination max number of results.

**Expected behavior**

job list API should return all results.

**Additional context**

Job list API should be updated to support pagination controls, however the GUI configuration should not affect the current behavior of the API.

Related: #5621

PR #7461, opened 2022-01-03
RUN-565: API job list GUI pagination

ronaveva:
fixes https://github.com/rundeck/rundeck/issues/5780

Added a new flag in JobQueryInput to know whether the query comes from an API call or GUI request. If it comes from the API it will ignore the pagination configuration for the GUI, but it can still handle pagination using request parameters (max and offset), it just prevents the use of default pagination parameters.

Examples:

With rundeck GUI pagination config:
rundeck.gui.paginatejobs.enabled=true
rundeck.gui.paginatejobs.max.per.page=3

(All the 7 jobs are returned)
Screen Shot 2022-01-03 at 11 02 27

Using "max" parameter as 2 it only returns 2 jobs
Screen Shot 2022-01-03 at 11 03 18


gschueler:
I think the pagination should become an input option to the listWorkflows method, so that it can be requested if required, so then where it is called can determine what pagination to use.

ronaveva:
> I think the pagination should become an input option to the listWorkflows method, so that it can be requested if required, so then where it is called can determine what pagination to use.

The BaseQueryInput already have the pagination parameters so i don't think we need a new input parameter in listWorkflows, what we could do is to set all the required values before calling the method and then just use these values

L2JE:
ScheduledExecutionService was decoupled from its consumers:
* The pagination setup was moved from ScheduledExecutionService to MenuController (excepting the offset=0 for compatibility).
* Added a new test for pagination on MenuController.jobs() endpoint.
* Added the alternative to disable pagination for api (it will ignore pagination parameters if the `grailsApplication.config.rundeck.api.paginatejobs.enabled` property is set to false explicitly)
* Cleaned just a little some old code.

PR #7462, opened 2022-01-03
Added quotes to the nodeFilter before getting it

L2JE:
This issue appeared after migrating to vue. When getting the node filter from the component, it didn't add quotes to the filter value if if has spaces in it.
Fix: https://github.com/rundeckpro/rundeckpro/issues/2172


ronaveva (approved):
LGTM

PR #7465, opened 2022-01-05
Fixes Project Nav Remains when View All Projects link is clicked #RUN-405 

ehe-pd:
Fixes RUN-405

Root cause:
- The `View All Projects` is rendered as an <a> element with the attribute `href` set to the URL of the current project. When clicked the default behavior of <a> tag got triggered instead of executing the registered javascript handler. This causes the page to always be rendered with an active project so the sidebar is always displayed.
- There is another issue with the ProjectSelect Vue template definition. The allProjectsLink `eventHandler` used a non-existent handler name `handleSelectAll`. Since this event never got triggered due to the above issue, so it was undetected until this fix was applied.

Solution:
- Prevent the default behavior of the <a> tag by using Vue modifier e.g. change `@click` to `@click.prevent`.
- Change the select all projects event handler to the right name `allClicked`

Milestone: 3.4.10


qualman:
Looks good to me!

PR #7476, opened 2022-01-12
Add command line -m option to run db migrations then stop the app

sjrd218:
Run the app with `-m` and it will apply any unapplied database migrations, then shut down.


rundeckapp/grails-app/init/rundeckapp/cli/CommandLineSetup.groovy
```diff
@@ -117,6 +117,10 @@ class CommandLineSetup {
                 .withDescription("Down migrate or rollback to previous db versions.")
                 .create('r');
 
+        Option migrate =   OptionBuilder.withLongOpt("migrate")
```
gschueler, line 120: maybe `migrate-only` ?

gschueler (approved):

PR #7479, opened 2022-01-13
RUN-565: Fix/issue 5780 api gui pagination

L2JE:
fixes https://github.com/rundeck/rundeck/issues/5780

ScheduledExecutionService was decoupled from its consumers:
* The pagination setup was moved from ScheduledExecutionService to MenuController _(excepting the offset=0 for compatibility)_.
* Added a new test for pagination on MenuController.jobs() endpoint.
* Added the alternative to disable pagination for api _(it will ignore pagination parameters if the `rundeck.api.paginatejobs.enabled` property is set to false explicitly)_
* Cleaned just a little some old code.

Examples:

With rundeck GUI pagination config:
rundeck.gui.paginatejobs.enabled=true
rundeck.gui.paginatejobs.max.per.page=3

(All the 7 jobs are returned)
Screen Shot 2022-01-03 at 11 02 27

Pagination parameters work perfectly fine (decoupled from others configuration), unless `rundeck.api.paginatejobs.enabled` exists and is false
image

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/config/RundeckConfigBase.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/config/RundeckConfigBase.java
@@ -66,7 +66,7 @@ public class RundeckConfigBase {
     @Data
     public static class RundeckApiConfig {
         ApiTokensConfig tokens;
-        PaginateJobs paginateJobs;
+        PaginateJobs paginatejobs;
 
         @Data
         public static class PaginateJobs {
--- a/core/src/main/java/com/dtolabs/rundeck/core/config/RundeckConfigBase.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/config/RundeckConfigBase.java
@@ -554,7 +554,7 @@ public class RundeckConfigBase {
                 Per per;
                 @Data
                 public static class Per {
-                    String page;
+                    Integer page;
                 }
             }
         }
--- a/core/src/main/java/com/dtolabs/rundeck/core/config/RundeckConfigBase.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/config/RundeckConfigBase.java
@@ -66,6 +66,12 @@ public class RundeckConfigBase {
     @Data
     public static class RundeckApiConfig {
         ApiTokensConfig tokens;
+        PaginateJobs paginateJobs;
+
+        @Data
+        public static class PaginateJobs {
+            Boolean enabled;
+        }
 
         @Data
         public static class ApiTokensConfig {
@@ -542,7 +548,15 @@ public class RundeckConfigBase {
         @Data
         public static class PaginateJobs {
             Boolean enabled;
-            String maxPerPage;
+            Max max;
+            @Data
+            public static class Max {
+                Per per;
+                @Data
+                public static class Per {
+                    String page;
+                }
+            }
         }
 
         @Data
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
@@ -2921,6 +2921,7 @@ class MenuController extends ControllerBase implements ApplicationContextAware{
 
         }
 
+        query.apiCall = true
         query.projFilter = params.project
         //test valid project
 
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
@@ -303,9 +303,10 @@ class MenuController extends ControllerBase implements ApplicationContextAware{
             }
         }
 
-        if(guiPaginationEnabled)
+        if(configurationService.getBoolean('gui.paginatejobs.enabled',false)) {
             query.paginatedRequired = true
-        else{
+            query.max = configurationService.getInteger('gui.paginatejobs.max.per.page', 10)
+        }else{
             query.max = null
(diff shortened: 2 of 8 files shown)


rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
```diff
@@ -577,11 +593,32 @@ class MenuController extends ControllerBase implements ApplicationContextAware{
...
+
+    private boolean getGuiPaginationEnabled() {
+        grailsApplication.config.rundeck.gui.paginatejobs.enabled in ["true",true]
+    }
+
+    private boolean getApiPaginationEnabled() {
+        if(grailsApplication.config.rundeck.api.paginatejobs.enabled != null)
+            return grailsApplication.config.rundeck.api.paginatejobs.enabled in ["false",false]
+        
+        return true
```
gschueler, line 612: should use the `configurationService` instead of grailsApplication.config

rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
```diff
@@ -577,11 +593,32 @@ class MenuController extends ControllerBase implements ApplicationContextAware{
...
+    }
+
     private def listWorkflows(ScheduledExecutionQuery query,AuthContext authContext,String user) {
         long start=System.currentTimeMillis()
         if(null!=query){
             query.configureFilter()
         }
+
+        query.max = getConfiguredMaxPerPage(10)
```
gschueler, line 621: I think this should be removed.  Wouldn't this override the max provided in the query object, even for api requests?

rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
```diff
@@ -2912,6 +2949,12 @@ class MenuController extends ControllerBase implements ApplicationContextAware{
      * API: /api/2/project/NAME/jobs, version 2
      */
     def apiJobsListv2 (ScheduledExecutionQuery query) {
+
+        if (!apiPaginationEnabled){
+            query.max = null
+            query.offset = null
+        }
+
```
gschueler, line 2957: Since job pagination is a new "feature" it needs to only apply by default in a new API version.  You will need to increment the API version, and if the version in the request >= that version, enable it by default, otherwise it should be disabled.

Also, the job list API should then support appropriate pagination data, such as including the `offset` and `max` in the response data (both XML and JSON)

rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
```diff
@@ -422,7 +406,7 @@ class ScheduledExecutionService implements ApplicationContextAware, Initializing
             if(query && query.sortBy && xfilters[query.sortBy]){
                 order(xfilters[query.sortBy],query.sortOrder=='ascending'?'asc':'desc')
             }else{
-                if(paginationEnabled) {
+                if(query.paginatedRequired) {
```
gschueler, line 409: I think the groupPath ordering doesn't have to be only if pagination is required.  Suggest we just include the `order("groupPath", "asc")` for all queries

gschueler (changes requested):

rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
L2JE: Actually this is just for keeping previous behavior, you could ask the api for number of jobs...

This works perfectly in 3.4.9:
`curl "http://localhost:8080/api/40/project/P/jobs?max=2&authtoken=dSQvmNJG7UPNfd1LggQ5hxSXnPoH4PSp" | xml_pp -l`

The only problem was that when setting gui pagination properties it affected api responses...

gschueler: ok 👍

rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
L2JE: Doesn't this change the previous expected behavior? May be for the next version...

rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
L2JE: I had to add the classes Max and Per to access paginatejobs.max.per.page property from ConfigurationService without affecting compatibility...

core/src/main/java/com/dtolabs/rundeck/core/config/RundeckConfigBase.java
```diff
@@ -542,7 +548,15 @@ public Enabled(final Boolean enabled) {
...
         public static class PaginateJobs {
             Boolean enabled;
-            String maxPerPage;
+            Max max;
+            @Data
+            public static class Max {
+                Per per;
+                @Data
+                public static class Per {
+                    String page;
```
gschueler, line 557: should be an `Integer`?

gschueler (approved):

core/src/main/java/com/dtolabs/rundeck/core/config/RundeckConfigBase.java
L2JE: Yep, It's obvious it was a copy paste haha

L2JE:
We have to add the following property to the docs:
`rundeck.api.paginatejobs.enabled` _(=true if not defined)_

PR #7493, opened 2022-01-21
RUN-611: Ldap name and last name cannot sync

alexander-variacode:
Issue:
Rundeck cannot sync the name and last name from user.

Fix description:
Changed property name way to obtain the correct value from rundeck-config.properties


ronaveva (approved):
LGTM

PR #7495, opened 2022-01-24
RCLOUD-181: Add the ability to authenticate webhooks using the Authorization header

sjrd218:
POSTs to webhooks can now be authenticated, by a value specified in the Webhook data or a value from the key store.

### DIFF
--- a/grails-webhooks/grails-app/controllers/webhooks/WebhookController.groovy
+++ b/grails-webhooks/grails-app/controllers/webhooks/WebhookController.groovy
@@ -3,16 +3,15 @@ package webhooks
 import com.dtolabs.rundeck.core.authorization.AuthContext
 import com.dtolabs.rundeck.core.authorization.AuthContextEvaluator
 import com.dtolabs.rundeck.core.authorization.AuthContextProvider
-import com.dtolabs.rundeck.core.authorization.AuthorizationUtil
 import com.dtolabs.rundeck.core.authorization.UserAndRolesAuthContext
-import com.dtolabs.rundeck.core.storage.ResourceMeta
 import com.dtolabs.rundeck.core.webhook.WebhookEventException
 import com.dtolabs.rundeck.plugins.webhook.WebhookDataImpl
-import com.fasterxml.jackson.databind.ObjectMapper
 import grails.converters.JSON
 import groovy.transform.PackageScope
 import org.rundeck.core.auth.AuthConstants
+import webhooks.authenticator.AuthorizationHeaderAuthenticator
 
+import javax.servlet.http.HttpServletRequest
 import javax.servlet.http.HttpServletResponse
 
 class WebhookController {
@@ -23,7 +22,6 @@ class WebhookController {
     AuthContextEvaluator rundeckAuthContextEvaluator
     AuthContextProvider rundeckAuthContextProvider
     def apiService
-    def storageService
 
     def admin() {}
 
@@ -148,7 +146,7 @@ class WebhookController {
             return
         }
 
-        if(!authorizeWebhookSecret(authContext, hook, request.getHeader(AUTH_HEADER))) {
+        if(hook.authConfigJson && !authorizeWebhook(hook, request)) {
             sendJsonError("Failed webhook authorization")
             return
         }
@@ -184,22 +182,8 @@ class WebhookController {
     }
 
     @PackageScope
-    boolean authorizeWebhookSecret(AuthContext authContext, Webhook hook, String headerAuthValue) {
-        if(!hook.secret || hook.secret.isEmpty()) return true
-        if(hook.secret.startsWith("keys/")) {
-            try {
-                def keyStorageService = storageService.storageTreeWithContext(authContext)
-                ResourceMeta contents = keyStorageService.getResource(hook.secret).getContents();
-                ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
-                contents.writeContent(byteArrayOutputStream);
+    boolean authorizeWebhook(Webhook hook, HttpServletRequest request) {
+        return new AuthorizationHeaderAuthenticator().authenticate(hook, request)
 
-                return headerAuthValue == new String(byteArrayOutputStream.toByteArray());
-            } catch (IOException e) {
-                log.warn("Failed to get webhook storage key: ${hook.secret}", e)
-                return false
-            }
-        } else {
-            return hook.secret == headerAuthValue
-        }
     }
 }
--- a/grails-webhooks/grails-app/controllers/webhooks/WebhookController.groovy
+++ b/grails-webhooks/grails-app/controllers/webhooks/WebhookController.groovy
@@ -5,6 +5,7 @@ import com.dtolabs.rundeck.core.authorization.AuthContextEvaluator
 import com.dtolabs.rundeck.core.authorization.AuthContextProvider
 import com.dtolabs.rundeck.core.authorization.AuthorizationUtil
 import com.dtolabs.rundeck.core.authorization.UserAndRolesAuthContext
+import com.dtolabs.rundeck.core.storage.ResourceMeta
 import com.dtolabs.rundeck.core.webhook.WebhookEventException
 import com.dtolabs.rundeck.plugins.webhook.WebhookDataImpl
 import com.fasterxml.jackson.databind.ObjectMapper
@@ -15,13 +16,14 @@ import org.rundeck.core.auth.AuthConstants
 import javax.servlet.http.HttpServletResponse
 
 class WebhookController {
+    static final String AUTH_HEADER = "Authorization"
     static allowedMethods = [post:'POST']
 
     def webhookService
-    def frameworkService
     AuthContextEvaluator rundeckAuthContextEvaluator
     AuthContextProvider rundeckAuthContextProvider
     def apiService
+    def storageService
 
     def admin() {}
 
@@ -146,6 +148,11 @@ class WebhookController {
             return
         }
 
+        if(!authorizeWebhookSecret(authContext, hook, request.getHeader(AUTH_HEADER))) {
+            sendJsonError("Failed webhook authorization")
+            return
+        }
+
         WebhookDataImpl whkdata = new WebhookDataImpl()
         whkdata.webhookUUID = hook.uuid
         whkdata.webhook = hook.name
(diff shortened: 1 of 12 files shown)


grails-webhooks/grails-app/controllers/webhooks/WebhookController.groovy
```diff
@@ -175,4 +182,24 @@ class WebhookController {
...
+        if(!hook.secret || hook.secret.isEmpty()) return true
+        if(hook.secret.startsWith("keys/")) {
+            try {
+                def keyStorageService = storageService.storageTreeWithContext(authContext)
+                ResourceMeta contents = keyStorageService.getResource(hook.secret).getContents();
+                ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
+                contents.writeContent(byteArrayOutputStream);
+
+                return headerAuthValue == new String(byteArrayOutputStream.toByteArray());
+            } catch (IOException e) {
```
gschueler, line 197: I think this also needs to catch `StorageException` to handle auth failure and not found

grails-webhooks/grails-app/services/webhooks/WebhookService.groovy
```diff
@@ -159,6 +159,7 @@ class WebhookService {
         hook.uuid = hookData.uuid ?: hook.uuid
         hook.name = hookData.name ?: hook.name
         hook.project = hookData.project ?: hook.project
+        if(hookData.secret != null) hook.secret = hookData.secret
```
gschueler, line 162: should the secret be stored as a hashed value if it is not a key storage path?

sjrd218: If we hash the secret we can't display the secret on edit, so the user can't go back and check the secret value. Definitely more secure, but not sure if the UI would make sense.

fdevans: I vote hash it and they can change it.

sjrd218: We'll probably need to give the user a warning that when they put a value in the Secret/Authorization String field, if it doesn't refer to a key store value, it will be hashed and the original value will no longer be visible. Thoughts?

fdevans:
Let's change the field description to:
`[Optional] A Webhook Authentication string can be provided to increase security of this webhook.  Put a value in the field or select a KeyStorage entry.  All posts will need to include the same string in the Authorization header.`

Also what are thoughts on calling the field "Authorization String" instead of Secret to avoid confusion on what type of secret it is and reinforce that it needs to be whatever full string they are sending.  (It tripped me up in PostMan when I thought I could just type in a password value for one of bearer or api token entries as the secret, but the string needed `token1` or `bearer` as well on the Rundeck side.)

fdevans (changes requested):
see previous comment

grails-webhooks/grails-app/services/webhooks/WebhookService.groovy
fdevans: So if it's a key-storage entry it would still be possible to see the string.

If it's not a key-storage entry it would be obscured with ***** right?

How will it know it's a key-storage string?  Should we could have a separate box for selecting from KS and then the straight string box would be **** even while they are typing.  That would imply they won't be able to see it in the future (like a password) and I can cover off the field being obscured in docs.

gschueler: I think we should not overload the input field with both key storage path and "secret" value.  maybe it's simpler to do one or the other.  Currently we don't have a simple UI for that kind of behavior.  I notice @sjrd218 that access to key storage depends on the auth context defined for the webhook request - will that auth context have correct read access to key storage?

fdevans: Then let's just use Key Storage.

rundeckapp/grails-spa/packages/ui/src/pages/webhooks/views/WebhooksView.vue
```diff
@@ -340,6 +388,15 @@ export default observer(Vue.extend({
...
     },
+    confirmRemoveAuth() {
+      var self = this
+      self.$confirm({
+        title:"Confirm",
+        content:"Are you sure you want to remove this webhook authorization string?"
+      }).then(() => {
+        this.handleSave()
+        }).catch(() => {})
+    },
```
gschueler, line 399: this method doesn't seem to be called anywhere

rundeckapp/grails-spa/packages/ui/src/pages/webhooks/views/WebhooksView.vue
```diff
@@ -57,6 +57,19 @@
                     <CopyBox style="max-width: 800px" v-if="!curHook.new" :content="postUrl()"/>
                     <span class="form-control fc-span-adj font-italic" style="height: auto;" v-if="curHook.new">{{$t('message.webhookPostUrlPlaceholder')}}</span>
                   </div>
+                  <div class="card-content">
+                    <label>{{ $t('message.webhookAuthLabel') }}</label>
+                    <div class="help-block"> {{$t('message.webhookGenerateSecretCheckboxHelp')}}</div>
+                    <div class="checkbox"><input type="checkbox" v-model="curHook.useAuth" @click="confirmAuthToggle" class="form-control"><label>{{ $t('message.webhookGenerateSecurityLabel') }}</label></div>
```
gschueler, line 63: Label needs a `for` attribute, does not work to click the label

gschueler: fyi: i see the same problem with the "Enabled" checkbox, so actually a general problem with checkboxes in this UI

rundeckapp/grails-spa/packages/ui/src/pages/webhooks/views/WebhooksView.vue
```diff
@@ -193,20 +208,24 @@ var i18nInstance = new VueI18n({
 export default observer(Vue.extend({
   name: "WebhooksView",
   components: {
+    HelpIcon,
```
gschueler, line 211: this component was added, but is not used?

rundeckapp/grails-spa/packages/ui/src/pages/webhooks/i18n.js
```diff
@@ -20,6 +20,11 @@ const translationStrings = {
       webhookUserHelp:'The authorization username assumed when running this webhook. All ACL policies matching this username will apply.',
       webhookRolesLabel:'Roles',
       webhookRolesHelp:'The authorization roles assumed when running this webhook (comma separated). All ACL policies matching these roles will apply.',
+      webhookAuthLabel: 'HTTP Authorization String',
+      webhookGenerateSecurityLabel: "Use Authorization String",
+      webhookGenerateSecretCheckboxHelp: "[Optional] A Webhook Authentication string can be generated to increase security of this webhook. All posts will need to include the generated string in the Authorization header.",
```
gschueler, line 25: this messages says "Authentication" but the other messages say "Authorization".  I think they should all say "Authentication"

rundeckapp/grails-spa/packages/ui/src/pages/webhooks/i18n.js
```diff
@@ -20,6 +20,11 @@ const translationStrings = {
       webhookUserHelp:'The authorization username assumed when running this webhook. All ACL policies matching this username will apply.',
       webhookRolesLabel:'Roles',
       webhookRolesHelp:'The authorization roles assumed when running this webhook (comma separated). All ACL policies matching these roles will apply.',
+      webhookAuthLabel: 'HTTP Authorization String',
```
gschueler, line 23: suggest labelling it "HTTP Authentication Header" instead of "String"
```suggestion
      webhookAuthLabel: 'HTTP Authentication Header',
```

rundeckapp/grails-spa/packages/ui/src/pages/webhooks/i18n.js
gschueler: i take that back, even though it is an authentication strring, the actual HTTP header says "Authorization" so let's use "Authorization Header" consistently

rundeckapp/grails-spa/packages/ui/src/pages/webhooks/i18n.js
```diff
@@ -20,6 +20,11 @@ const translationStrings = {
...
+      webhookGenerateSecurityLabel: "Use Authorization String",
```
gschueler, line 24:
```suggestion
      webhookGenerateSecurityLabel: "Use Authorization Header",
```

gschueler (commented):
-[x] Auth string generation and interaction works as expected
- some minor text and UI corrections should be made

Another observation:

* After saving/creating a webhook where the auth string is to be generated an revealed, for some reason there is a delay after the "Saved!" message appears, and seeing the auth string in the UI.  Looking at the network calls, the POST to save succeeds quickly, then there is a ~4s long delay while the WebhookEvent plugin list API is called, finally the auth string is shown in the UI.  This delay may be my local issue, but it seems odd to have the plugin listing API delay the UI updates

rundeckapp/grails-spa/packages/ui/src/pages/webhooks/views/WebhooksView.vue
gschueler:
```suggestion
                    <div class="checkbox"><input type="checkbox" v-model="curHook.useAuth" @click="confirmAuthToggle" class="form-control" id="wh-authtoggle"><label for="wh-authtoggle">{{ $t('message.webhookGenerateSecurityLabel') }}</label></div>
```

rundeckapp/grails-spa/packages/ui-trellis/src/components/help/HelpIcon.vue
```diff
@@ -0,0 +1,34 @@
+<template>
+    <div class="help-icon">
+        <i :id="`help-${this._uid}`" class="fas fa-sm fa-question-circle text-muted" style="cursor:pointer;margin-left:5px"/>
+        <popover :title="title" :target="`#help-${this._uid}`" custom-class="popover-wide" viewport="#main-panel">
```
gschueler, line 4:
```suggestion
        <popover :title="title" :target="`#help-${this._uid}`" custom-class="popover-wide" viewport="#section-main">
```

see #7510

rundeckapp/grails-spa/packages/ui/src/pages/webhooks/i18n.js
fdevans: To help distinguish.... The "feature" will be labeled, documented _Webhook Authentication_.   For the field below was my thinking on calling it Authorization String.

I had asked previously to call it string (instead of secret) to distinguish it from other types of Authorization Headers common in tools like postman etc. where that term shows up.  Trying to make it clear that just the string we generate is the header value and does not include "token" or "secret" as part of the value string.

fdevans: I think for the main ON/Off button it makes sense to call it "Use Authorization Header" as that is what's needed, but maybe the value is still called "Authorization String"?

gschueler: that's fine

fdevans (approved):

gschueler (commented):
there is still a timing issue, wrt the plugins list/webhook list refresh. If i save a webhook, it shows the auth string immediately now, but I can click to a different webhook, it and keeps showing the same auth string.  It shows that auth string until the plugin list/webhook list refresh completes (e.g. ~4s or so for me).  I will push a change to improve this.

gschueler (approved):

PR #7508, opened 2022-01-28
Fix 1850 logoutput unresponsive

perezo-pd:
Reduce amount of lines displayed during execution in Log Output view after maxLogSize has been reached to prevent destabilizing the app

Fix: https://github.com/rundeckpro/rundeckpro/issues/1850

### DIFF
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/components/execution-log/logViewer.vue
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/components/execution-log/logViewer.vue
@@ -154,7 +154,7 @@ export default class LogViewer extends Vue {
     @Prop({default: 'dark'})
     theme?: string
 
-    @Prop({default: 614400})
+    @Prop({default: 3145728})
     maxLogSize!: number
 
     @Prop()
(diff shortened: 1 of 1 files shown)


rundeckapp/grails-spa/packages/ui-trellis/src/components/execution-log/logViewer.vue
```diff
@@ -417,15 +417,11 @@ export default class LogViewer extends Vue {
...
-      this.execCompleted = res.execCompleted
-      this.completed = res.completed
-      this.nextProgress = res.completed ?
-        100:
-        Math.round((parseInt(res.offset) / res.totalSize) * 100)
-
-      if (res.entries.length == 0)
+    private handleExecutionLogResp() {
+      const defaultMaxLogSize = 3145728
+      if (this.maxLogSize == defaultMaxLogSize) {
```
carlosrfranco, line 422: Seems like the default value is just used to compare with the value from controller to handle or not the execution log. It would be preferable to check if the property has been configured or not on rundeck-config.properties file

carlosrfranco (changes requested):

rundeckapp/grails-spa/packages/ui-trellis/src/components/execution-log/logViewer.vue
perezo-pd: @carlosrfranco code has been updated to use a separate property (trimOutput)

carlosrfranco (approved):
LGTM

alexander-variacode:
@Mergifyio rebase

PR #7530, opened 2022-02-11
RUN-438: Fix SSO configs. Upgrade grails spring security plugin and libs.

sjrd218:
Add configs for SSO properties to RundeckConfigBase.
Updated grails spring security plugin to the grails 5 version.
Updated spring security to the latest version.
Fixed lombock dep in core.
Updated jackson and snake yaml to version shipped with the spring boot version we are using.


rundeckapp/grails-app/views/user/login.gsp
```diff
@@ -134,7 +134,7 @@
                       </div>
                     </g:if>
                     <!--SSO Login Feature-->
-                    <g:set var="loginButtonEnabled" value="${cfg.getString(config: "sso.loginButton.image.enabled") in [true,'true']}"/>
+                    <g:set var="loginButtonEnabled" value="${cfg.getString(config: "sso.loginButton.enabled") in [true,'true']}"/>
```
gschueler, line 137: preferred to use getBoolean
```suggestion
                    <g:set var="loginButtonEnabled" value="${cfg.getBoolean(config: "sso.loginButton.enabled", default: false)}"/>
```

PR #7535, opened 2022-02-14
Tweak task definition to work if environment is undefined

gschueler:
Fixing an issue that sometimes happens with IntelliJ in debug mode - it fails to evaluate this gradle file.
* fix issue with `+=` 
* fix evaluating `project.environment` if environment property is not defined


fdevans:
Closing per @gschueler as the change was implemented elsewhere.

PR #7553, opened 2022-02-23
RUN-695: fix issue with API access before server fully starts

gschueler:
Fix an issue (seen in CI) where API requests before server fully starts prevent all API requests in the session, when using username/password API login.

* Handle exception caused by missing API response marshaller configuration before server has completed startup by responding with 503 (Service Unavailable)
* Fix issue where "error/fiveHundred" response caused by exception would block API access for the session when using username/password login
* Additional change was necessary, due to order of Interceptors: ApiVersionInterceptor should not return false immediately, but adds a request attribute, so that the SetUserInterceptor executes cleanly and sets up appropriate session data about authenticated user. The AuthorizationInterceptor then handles the 503 response based on request attribute


chrismcg14 (approved):
LGTM

PR #7554, opened 2022-02-24
RUN-603: Enhance UserController to support more flexible error message

ehe-pd:
**Is this a bug fix or an enhancement? Please describe.**
Enhancement for RUN-603

The original design of the Error action only displays a fixed message for any user login failure. By adding logic to retrieve error messages from query parameters we can make the error page display custom error messages. It provides a better user experience e.g. for RUN-603 we can display the specific rate-limiting error to end-users. 

**Additional context**
Currently, this enhancement is only used by the enterprise version for the feature of login rate-limiting.

### DIFF
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/UserController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/UserController.groovy
@@ -59,8 +59,18 @@ class UserController extends ControllerBase{
         redirect(action:"login")
     }
 
+    /**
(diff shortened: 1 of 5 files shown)


gschueler:
I altered the behavior so that a flash context "loginErrorCode" can be preset instead of passing a URL parameter

ehe-pd:
> 

I have tested. The improvement works well.

gschueler (approved):

PR #7567, opened 2022-03-04
RUN-707: Use-Client-Side-FileReader-API-Load-Content-From-File

ehe-pd:
- Fixes RUN-707. It's an security enhancement by obsolete file upload feature for import storage key. 

- The uploading file feature is replaced by read the file content in the client-side through FileReader API so there is no file uploaded to the server. 

- We considered obsolete the file upload feature completely. But it will change the user experience significantly. To avoid confusing user we choose the way with minimum impact to the user experience.

**Additional context**
There is no noticeable UI change.

### DIFF
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/StorageController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/StorageController.groovy
@@ -260,6 +260,8 @@ class StorageController extends ControllerBase{
      *
      */
     def keyStorageUpload(StorageParams storageParams){
+        if(!requestHasValidToken()) return
+
         if (storageParams.hasErrors()) {
             flash.errors=storageParams.errors
             return redirect(controller: 'menu',action: 'storage',params: [project:params.project])
@@ -269,11 +271,6 @@ class StorageController extends ControllerBase{
         }
         AuthContext authContext = getAuthContextForPath(session.subject, storageParams.resourcePath)
         def resourcePath = storageParams.resourcePath
-
-        withForm { }.invalidToken{
-            flash.errorCode= 'request.error.invalidtoken.message'
-            return redirect(controller: 'menu', action: 'storage', params: [project: params.project])
-        }
 
         def contentType= null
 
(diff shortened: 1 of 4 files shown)


gschueler (commented):
We will need to change the `KeyStorageSelector.vue` UI as well

gschueler (commented):
* When using the File Upload button, the file name cannot be changed, the original file name is always used
* Selecting a non-text file (e.g. a PNG file) will fill the textarea with binary data. (The saved file seems to be corrupted and does not match the original file when downloaded.)

ehe-pd:
> * When using the File Upload button, the file name cannot be changed, the original file name is always used
It follows the original design when using file upload, the filename is not editable. I will change the field to make it editable.

> * Selecting a non-text file (e.g. a PNG file) will fill the textarea with binary data. (The saved file seems to be corrupted and does not match the original file when downloaded.)
The binary file is broken because the content is always read as ASCII text.  If a user wants to use an arbitrary file as a key, this should be an expected result. This is a user experience topic, in my opinion, we can provide a text hint to inform users to provide a correct text file instead of accommodating anything they choose to upload.

ehe-pd:
@gschueler 
> We will need to change the `KeyStorageSelector.vue` UI as well

I have tested the feature in Node Wizard and checked the source code of the component, this component has already used the same client-side way to upload keys instead of uploading a file. 

```
handleFileUpload(e: any){
                var files = e.target.files || e.dataTransfer.files;
                if (!files.length)
                    return;

                const file = files[0];
                this.upload.file = file.name;

                let reader = new FileReader();
                reader.onload = (event:any) => {
                    let text = event.target.result;
                    this.upload.fileContent = text;
                    if(this.upload.errorMsg!=null){
                        this.upload.errorMsg = null;
                    }
                };
                reader.onerror = (event:any) => {
                    this.upload.errorMsg = "file cannot be read";
                    this.upload.file = null;
                };
                reader.readAsText(file);

            },
```

rundeckapp/grails-app/controllers/rundeck/controllers/StorageController.groovy
```diff
@@ -268,19 +269,13 @@ class StorageController extends ControllerBase{
...
-        }.invalidToken{
+
+        withForm { }.invalidToken{
             flash.errorCode= 'request.error.invalidtoken.message'
             return redirect(controller: 'menu', action: 'storage', params: [project: params.project])
         }
-        if(!valid){
-            return
-        }
+
```
gschueler, line 277: the `valid` variable is still necessary. without that check, execution of `keyStorageUpload` method will not return.  the `return` inside the closure does not return from the outer method

gschueler (approved):

PR #7569, opened 2022-03-04
RUN-706: Add feature flag to disable "Job File Option" uploads

ltamaster:
**Is this a bugfix or an enhancement? Please describe.**
enhancement:
Adding feature flag for file upload plugin. It will be enabled by default.
In case the flag is false, input option with the type `file` cannot be created.
eg:
````
rundeck.feature.fileUploadPlugin.enabled=false
`````

* new feature flag for file upload plugin
* return an error in job import if the feature flag is disabled and the job has an input file type
* not display an input options file in an existing job
* in option form, disable the input file type 

**Additional context**


chrismcg14 (approved):
LGTM, tested and works as expected

PR #7570, opened 2022-03-04
RUN-688 special char handling

perezo-pd:
Handling special char when loading remote URL option. Not using UTF-8 encoding to avoid decoding issues when dealing with passwords
Fix: https://github.com/rundeckpro/rundeckpro/issues/2188


perezo-pd:
Replaced by https://github.com/rundeck/rundeck/pull/7609

PR #7573, opened 2022-03-08
Remove tests from release builds

chrismcg14:
Don't require tests to publish


gschueler:
seems circle config is not defined correctly based on the error in circle

gschueler (approved):

PR #7577, opened 2022-03-10
RUN-667: Upgrading h2 library version to 2.x

carlosrfranco:
Enhancement to upgrade the h2 database version to 2.x

1 - KEYWORDS

Was applied a workaround the issue appending ';NON_KEYWORDS=MONTH,HOUR,MINUTE,YEAR,SECONDS' on dataSource.url

2 - Varchar type

The varchar maximum size was decreased to 1048576.

3 - AUTO_INCREMENT

The column attribute 'autoIncrement' defined as 'true' will create the column as 'ID AUTO_INCREMENT NOT NULL'.
Then, when saving an item, the sql statement generated by gorm includes 'null' on the ID column and it will fail.
The ID column should be created like: 'BIGINT GENERATED BY DEFAULT AS IDENTITY DEFAULT ON NULL' to avoid failing when saving from gorm.

4 - Tests

Unit tests that use database to save data was changed so that the migrations runs before each of these tests


ehe-pd (commented):
I have tested the changes and run it locally. There are a few findings that make me hesitate to approve the PR.
1. We have to change the setting at 5 different places, but to run the application successfully I still have to make the same change to `rundeck-runtime/server/config/rundeck-config.properties`. The change on the H2 database URL violates the Don't Repeat Yourself principle. I think a single point configuration should be a preferred solution and I am still studying the possible ways to implement it. 
2. The 2.0 driver is not compatible with 1.x database if there is one already existing. The server will fail to start due to the compatible error and I have to delete the database. We may need to add some reminders in README to make developers aware of this issue.

ehe-pd (approved):
It works. The only thing left to complete is extra notes in the documentation about how to correctly setup H2 database URL with the newly introduced NON_KEYWORDS parameter.

stepfair:
Is there a reason that the version wasn't bumped to 5.0 when this is not a passive change via semantic versioning?

fdevans:
Major version updates with Grails/Gradle and other foundational upgrades of the product as well as API deprecations, and larger feature announcements.

H2 is not recommended for production use and the Critical vulnerability drove the urgency to release the updated library.

smiba:
FYI the latest version broke our rundeck after a simple APT upgrade without warning because of the new h2 format. In my honest opinion this should've been handled better, you usually don't expect your entire tooling to be broken after.

`[2022-05-12T15:56:53,487] ERROR pool.ConnectionPool [main] - Unable to create initial connections of pool.
org.h2.jdbc.JdbcSQLNonTransientConnectionException: Unsupported connection setting "MVCC" [90113-210]
`

PR #7587, opened 2022-03-15
RUN-741: fix flaky test in ExecutionController2Spec

gschueler:
* Fix a flaky test with a race condition
* refactor the test to only test the controller behavior
* add a new test to test service behavior


carlosrfranco (approved):
LGTM

PR #7608, opened 2022-03-24
RUN-786: Fix: CPU usage showing negative value

L2JE:
Fix: https://github.com/rundeckpro/rundeckpro/issues/2273

Used to use systemLoadAverage but this always return a negative value in windows.
Now changed this to work with getSystemCpuLoad() which works perfectly on windows and linux. 

image


ltamaster (approved):
tested

PR #7630, opened 2022-04-05
RUN-688: Handling special characters in user:password

perezo-pd:
Handling special chars when loading remote URL option.
Using clean URL when assembling the URI object and use parsed credentials (URL object) when assembling the basic auth header
Fix: https://github.com/rundeckpro/rundeckpro/issues/2188


alexander-variacode (approved):
LGTM

PR #7647, opened 2022-04-11
RUN-791 Descriptive console and API output

DarwisNarvaezDev:
## Bugfix to complement console or API output
According to the issue: [https://github.com/rundeckpro/rundeckpro/issues/2272], if there are problems importing projects with invalid jobs, the console did not give more details of what happened, with the proposed change, there could be more information about the job that fails, its UUID and the related workflow with the bug, As we can see in the following:

Test case: Trying to import a project that has invalid project and job references.

API output **before** the fix:

Screen Shot 2022-04-11 at 16 02 34

Console output **before** the fix:

Screen Shot 2022-04-11 at 16 03 56

API output **after** the fix:

Screen Shot 2022-04-11 at 16 12 07

Console output **after** the fix:

Screen Shot 2022-04-11 at 16 13 35

**Solution**
During the loading process of each job, we extracted from the possible errors more details about the failure, also added the full job log to the error's list to alert the user in the GUI what happened.

## Bugfix to complement GUI output
According to the issue: [https://github.com/rundeckpro/rundeckpro/issues/2270], if there are problems importing projects with invalid jobs, the GUI doesn't show any clues about the causes:

GUI output **before** the fix:

Screen Shot 2022-04-11 at 17 13 08

GUI output **after** the fix:

Screen Shot 2022-04-11 at 14 25 01

**Solution**
We managed to extract from the error list, the one that corresponds to the job that could not be imported and show it in the interface.


jtobard (approved):
Tested and it work 👍

PR #7648, opened 2022-04-13
RUN-828: Can't open referenced job in GUI, when has thousands of log executions

L2JE:
Fix: https://github.com/rundeckpro/rundeckpro/issues/2404

The problem was that we were looping over a list of all the authorized reports 3 times. First to generate a list of authorized ReferencedExecutions and then twice on the `ReportService` method.

Findings: 
1. It wasn't checking authorization to referencedExecution reports correctly (shown below in a [review](https://github.com/rundeck/rundeck/pull/7648/files))
2. The report href were generated wrong: https://github.com/rundeck/rundeck/pull/7649

The solution resulted on generating the referencedExecutions ids in a subquery for the ReportService.getReports query, so it runs all in the db at once. The authorization check is done through the list of projects.

Tested with 107542 refex reports.

Advice:
If rundeck had to keep performance having even more reports, we should add a limit for the queries on the db through properties so that if we have more than a certain amount `<N>` of reports the gui shows `+<N>reports` in the activity count and the service just queries the table in `<N>` sized batches instead of the full table.


L2JE:
The fix works on H2 at least but when testing with postgres, the db outputs the following error:
```
db_pg14     | 2022-05-05 17:22:48.791 -03 [260] ERROR:  operator does not exist: character varying = bigint at character 1005
db_pg14     | 2022-05-05 17:22:48.791 -03 [260] HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.
```

 
This is the subquery that has the issue:
```

this_.jc_exec_id in (
                select this_.execution_Id as y0_ 
                from referenced_execution this_ 
                where this_.scheduled_Execution_Id=$1
            )
```

`base_report.jc_exec_id is varchar`
`referenced_execution.execution_Id is int(8)`


Can’t find a way to generate a cast function in the query from grails.

PR #7649, opened 2022-04-13
Fix: 404 page when clicking on referenced execution from a different parent

L2JE:
Fix: https://github.com/rundeck/rundeck/issues/7670
Case Tested in: https://github.com/rundeckpro/rundeck-tester/pull/149

Solution:
Generate report items using the report's ctxProject value instead of the received params.project.


jtobard (changes requested):
Missing test, you can create one at rundeck-tester to increase coverage of the feature

L2JE:
@Mergifyio rebase

PR #7657, opened 2022-04-14
Creating mechanism to upgrade from old version h2 database (1.x) to latest version (2.1.210)

carlosrfranco:
This PR implements a mechanism to upgrade h2 database (created using h2 v1.x lib) automatically on prebootsrap

- it check the upgrade mechanism is enabled
- export database to a script
- create a backup for the database files
- delete old db files
- create new database using h2 v2.x
- importa all data

Before start the application:
- download h2v1 jar
- set env var `RUNDECK_H2_UPGRADE_ENABLE` as true
- set RUNDECK_H2_UPGRADE_LIB_PATH with the path where the h2v1 jar was copied (`rdeck.base/server/lib/h2-1.4.200.jar` as default)
- Start the application

### DIFF
--- a/rundeckapp/grails-app/init/rundeckapp/init/prebootstrap/UpgradeH2DatabasePreBootstrap.groovy
+++ b/rundeckapp/grails-app/init/rundeckapp/init/prebootstrap/UpgradeH2DatabasePreBootstrap.groovy
@@ -1,0 +1,154 @@
+package rundeckapp.init.prebootstrap
+
+import com.dtolabs.rundeck.core.properties.CoreConfigurationPropertiesLoader
+import com.zaxxer.hikari.HikariConfig
+import com.zaxxer.hikari.HikariDataSource
+import groovy.sql.Sql
+import org.rundeck.app.bootstrap.PreBootstrap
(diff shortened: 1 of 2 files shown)


rundeckapp/grails-app/init/rundeckapp/init/prebootstrap/UpgradeH2DatabasePreBootstrap.groovy
```diff
@@ -0,0 +1,173 @@
...
+        }
+    }
+
+    private boolean exportScriptOldDatabase(){
+        logger.info("Exporting database to a script file")
+
+        boolean result = false
+        try {
+
+            def cmd = ['/bin/sh', '-c', "java -cp " + libPath + " org.h2.tools.Script -url 'jdbc:h2:file:" + dbPath + ";DB_CLOSE_ON_EXIT=FALSE' -user 'sa' -password ''"]
```
ehe-pd, line 139: Username and password should not be hardcoded. Like dbUrl, the value of these two variables can be extracted from the configurations.

fdevans:
I don't think we want to support this long term.  Let's focus on providing via documentation.

PR #7663, opened 2022-04-18
Run 835

DarwisNarvaezDev:
# Bugfix - Expanding tokens in file workflow step #
As described in: https://github.com/rundeckpro/rundeckpro/issues/2148, option tokens were not expanded when used within a URL in the step file. This PR solves that problem.

**About the solution**
We created a method that returns a boolean, if a certain regex identifies that there are tokens inside a string, we use this method to evaluate the path of the file that enters as a parameter for the execution of the step. So, if it finds tokens, we use the "replaceDataReferencesInString" method to expand the tokens inside the path.

**Exhibits**
After the fix:
Screen Shot 2022-04-18 at 14 24 36


ronaveva (approved):
LGTM

PR #7671, opened 2022-04-25
RUN-845: Export to instance not working when preserving UUID

DarwisNarvaezDev:
# Bugfix to get the value on the "export to instance" form
A described here: https://github.com/rundeckpro/rundeckpro/issues/2478 there was an error while we try to export a project to other instance of rundeck when the checkbox "preserve uuid" was selected. Technically, the problem was the checkbox used in the form, which was not getting the given value.

**Solution**
What we made was change the current html tag to a built-in GSP component named "g:checkbox", and bind the component to the actual param that will travel to the controller.

**Exhibits**
After the fix, we are able to see the export process result.
Screen Shot 2022-04-25 at 20 17 11

**QA Framework tests**
https://github.com/rundeckpro/rundeck-tester/pull/146


jtobard (approved):
Approved, test here: https://github.com/rundeckpro/rundeck-tester/pull/146

PR #7674, opened 2022-04-26
RUN-842: Remove options when importing Job Definitions

Jesus-Osuna-M:
Bugfix
Fix: https://github.com/rundeckpro/rundeckpro/issues/2469

**Description**
When a job was created with some defined options and is exported in a YAML or XML job, removing the option field or enforcing to an empty array (options:[]) doesn't delete the old options.

**Solution** 
Clear the job options and use the new options given on the file.

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
@@ -3041,6 +3041,10 @@ class ScheduledExecutionService implements ApplicationContextAware, Initializing
         }
     }
 
+    /**
+     *Remove and delete all existing options from a scheduledExecution, setting options param to null
+     * @param scheduledExecution
+     */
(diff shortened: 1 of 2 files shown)


rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
```diff
@@ -3227,6 +3227,7 @@ class ScheduledExecutionService implements ApplicationContextAware, Initializing
                 !it.key.startsWith( 'nodeExclude')
             }
         }else{
+            deleteExistingOptions(scheduledExecution)
```
alexander-variacode, line 3230: Try to add a javadoc to deleteExistingOptions() method explaining its purpose (I know the method name says everything but extra information is not bad and also now a javadoc is generated in every release)

alexander-variacode (commented):
I left a comment for you Jesus :)

alexander-variacode (approved):
LGTM

PR #7677, opened 2022-04-28
RUN-740: Expose spring actuator health check endpoints

ehe-pd:
Enhancement

Configure Spring Boot Actuator Health Indicator out-of-box feature to expose liveness and readiness endpoints. Manually configure DataSourceHealthIndictor to fit into the Rundeck application.

A fully in-house developed solution could be an option but it cost too much effort unnecessarily when we already have a standard Spring Boot out-of-box framework.

**Additional context**
How to test it in a local environment:
1. The default configuration has already been set in `application.yml` file under the rundeckapp/grails-app/conf/ folder.
Screen Shot 2022-04-28 at 10 30 39 AM

2. By default, the Health indicator endpoints are disabled. There are a few ways to turn it on:
    - Use OS Environment variable:
      ```env "management.endpoint.health.enabled=true" gradle bootRun```
    - Pass it as a JVM parameter
      ```gradle bootRun --system-prop="management.endpoint.health.enabled=true"```
    - If you are using a docker image, you can provide this env parameter when starting the container.
   
After the application started. You should be able to access the endpoints:
- {RUNDECK_BASE_URL}/actuator/health/readiness
- {RUNDECK_BASE_URL}/actuator/health/liveness

If everything is good. The endpoints should return JSON content with the `status` UP and an HTTP Status Code `200`.
If something goes wrong:
- You either can not access the endpoint successfully 
- Or you will receive the return JSON content with `status` DOWN or OUT_SERVICE. And the HTTP status code is `503` service unavailable.

For detail please refer to the Spring Boot Actuator documentation: https://docs.spring.io/spring-boot/docs/current/reference/html/actuator.html?query=health%27%20target=_blank%3E%3Cb%3Ehealth%3C/b%3E%3C/a%3E-groups

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/config/RundeckConfigBase.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/config/RundeckConfigBase.java
@@ -58,6 +58,7 @@ public class RundeckConfigBase {
     RundeckNotificationConfig notification;
     RundeckApiConfig api;
     ScmLoader scmLoader;
+    RundeckHealthIndicatorConfig health;
 
     @Data
     public static class UserSessionProjectsCache {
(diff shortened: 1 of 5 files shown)


carlosrfranco (approved):
It worked for me

PR #7699, opened 2022-05-10
RUN-849 Allow editing the endpoint in Pagerduty Plugins

DarwisNarvaezDev:
# Enhancement to allow Rundeck users to customize the target PagerDuty service URL #
Rundeck Pro PR: https://github.com/rundeckpro/rundeckpro/pull/2523
As seen in issue: https://github.com/rundeckpro/rundeckpro/issues/2464 it was not possible to change the target URL of the PagerDuty service. .

### Solution ###
We made changes to make possible the customization of the URL through the "framework.properties" file. In this file will be the property: "framework.pagerduty.service.url" and its value will enter as a parameter in the request of PagerDuty plugins.

*In this repo:* we added the property section in "framework.properties" file template to allow the user to input the wanted value and added the default value of the property in the first run.

### Exhibits ###
After fix, basic workflow step:
Screen Shot 2022-05-10 at 12 35 44

Test execution 1, with default value:
Screen Shot 2022-05-10 at 12 34 32

Changing the value in the props file:
Screen Shot 2022-05-10 at 12 09 44

Test execution 2, same Workflow Step parameters, different output:
Screen Shot 2022-05-10 at 12 41 38

Bonus execution:
Screen Shot 2022-05-10 at 12 47 23

### DIFF
--- a/core/src/main/java/com/dtolabs/launcher/Setup.java
+++ b/core/src/main/java/com/dtolabs/launcher/Setup.java
@@ -50,6 +50,7 @@ public class Setup implements CLIToolLogger {
     public static final boolean FORCE_FLAG = true;
     private Parameters parameters = new Parameters();
     public static final String TEMPLATE_RESOURCES_PATH = "com/dtolabs/launcher/setup/templates";
+    public static final String PAGERDUTY_DEFAULT_SERVICE_URL = "https://events.pagerduty.com";
 
     /**
      * default constructor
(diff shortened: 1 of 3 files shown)


jtobard:
Moving this to 4.4.0 since we need to work on how to get the properties correctly on notifications

fdevans:
We can't pull this from 4.3.  See internal chat

chrismcg14:
@DarwisNarvaezDev This PR is unneeded. Instead of injecting the value into the framework properties, check if it is set in the job step in your pro PR, and if it isn't then check the framework, like here: https://github.com/rundeckpro/cloud-aws-plugins/blob/7e1dd22af9e63dfdd9a2547cd67ee1d71bcadd17/src/main/groovy/com/rundeck/plugins/aws/util/AwsPluginUtil.groovy#L142

PR #7701, opened 2022-05-10
RUN-877: remove asset-pipeline config from plugin

### LINKED ISSUE #7700: RUN-877: bootRun/dev mode UI is broken
**Describe the bug**

when running rundeck oss in dev mode/bootRun, the UI is broken. The app fails to load all assets with 404 responses

cause: The “grails-execution-mode-timer” plugin has incorrect config for asset-pipeline plugin, causing it to create a manifest file resource.  this causes the asset-pipeline grails plugin to use only the assets listed in the manifest file, instead of dynamically in non-war mode.

fix: remove asset-pipeline incorrect usage from grails-execution-mode-timer plugin. Since this plugin doesn't use assets, just remove all asset-pipeline config from the gradle build file.

**My Rundeck detail**

* Rundeck version: main branch dev mode

**To Reproduce**
1. execute gradle :rundeckapp:bootRun task
2. Load app in browser
3. see all css/js cannot be loaded

**Expected behavior**
UI works properly in devmode/bootRun

gschueler:
Fix #7700
* remove asset-pipeline config from execution-mode-timer plugin

### DIFF
--- a/grails-execution-mode-timer/build.gradle
+++ b/grails-execution-mode-timer/build.gradle
@@ -6,14 +6,12 @@ buildscript {
     }
     dependencies {
         classpath "org.grails:grails-gradle-plugin:$grailsGradlePluginVersion"
-        classpath "com.bertramlabs.plugins:asset-pipeline-gradle:$assetPluginVersion"
     }
 }
 
 apply plugin:"eclipse"
 apply plugin:"idea"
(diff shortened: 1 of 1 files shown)


ehe-pd (approved):
Tested. Problem fixed.

PR #7703, opened 2022-05-11
RUN-887: Fix local file copier config error

gschueler:
Fix: if `local` file copier is configured for a project, GUI fails with an error message
* fix npe when a provider definition is not found
* fix `local` file copier cannot be selected as default project config for file copier


ehe-pd (commented):
block list test failed

ehe-pd (approved):

PR #7704, opened 2022-05-11
RUN-873:Fix: bug ansible plugins dont show properly

Jesus-Osuna-M:
Fix: https://github.com/rundeckpro/rundeckpro/issues/2430

Problem
A CSS property was being changed to block when it should remain on none, that class was  "autocomplete-suggestions"

Solution
Add a the CSS property display none to the class autocomplete-suggestions


L2JE (approved):
LGTM

PR #7726, opened 2022-05-23
RUN-878: Grant appadmin user access to executionMode page and license Info page

carlosrfranco:
Appadmin users should see license info page and be able to change executionMode


ehe-pd (approved):
Tested.

PR #7728, opened 2022-05-24
Validation for permissions is added in the import of projects. The va…

avelasquezr:
Fix https://github.com/rundeckpro/rundeckpro/issues/2497

Project Import fails when you add the project name as reference

Validation for permissions is added in the import of projects.
The validation added is to allow projects to be imported without validating reference to other projects.


L2JE (approved):
LGTM

PR #7729, opened 2022-05-24
RUN-925: Fix stored acls not require app admin auth

chrismcg14:


gschueler (approved):
added some changes:
* MenuController.acls access level should allow ops_admin
* fix Access Control menu item display to match the same access levels

PR #7732, opened 2022-05-25
RUN-913: When recreating storage tree, include converter plugin props

chrismcg14:
Fix RUN-913


ehe-pd (approved):
Tested with @L2JE. It works.

PR #7735, opened 2022-05-26
RUN-917: update ExecReport table for better queries

### LINKED ISSUE #7725: RUN-917: base_report table should use bigint(20) for foreign keys
We have a very large (~500,000 executions / day) Rundeck installation with an unfortunately-massive accumulation of execution logs due to negligence a few years ago when we set up the software. The execution history cleaner has stopped working at all, even with a small maximum execution count, I believe because the initial SELECT that retrieves database rows locks up the Java process.

One of the culprits here may be the `jc_exec_id` field in the `base_report` table. This field references `execution.id`, but it uses `varchar(255)` instead of `bigint(20)`. Performing a MySQL Join between those two tables results in **every** `base_report` row being scanned. e.g.:

```
MySQL [core_rundeck]> EXPLAIN SELECT
    ->     e.id AS e_id,
    ->     br.id AS br_id
    -> FROM
    ->     execution e
    -> JOIN log_file_storage_request lfsr ON lfsr.execution_id = e.id
    -> JOIN base_report br ON br.jc_exec_id = e.id
    -> WHERE e.id = 122594920\G

*************************** 1. row ***************************
           id: 1
  select_type: SIMPLE
        table: e
         type: const
possible_keys: PRIMARY
          key: PRIMARY
      key_len: 8
          ref: const
         rows: 1
        Extra: Using index
*************************** 2. row ***************************
           id: 1
  select_type: SIMPLE
        table: lfsr
         type: const
possible_keys: UK_trqsa9so0qcv6okcd6fan88yf,UC_LOG_FILE_STORAGE_REQUESTEXECUTION_ID_COL
          key: UK_trqsa9so0qcv6okcd6fan88yf
      key_len: 8
          ref: const
         rows: 1
        Extra: Using index
*************************** 3. row ***************************
           id: 1
  select_type: SIMPLE
        table: br
         type: index
possible_keys: EXEC_REPORT_IDX_2
          key: EXEC_REPORT_IDX_2
      key_len: 768
          ref: NULL
         rows: 179198857
        Extra: Using where; Using index
```

I included the `log_storage_file_request` table to show the contrast (1 row retrieved instead of 179198857). I can confirm that altering the table to use `bigint(20)` makes the index work properly.

I feel that the `jc_exec_id` field (and other FK fields in this table) should be converted to an Integer type in order for the index to work properly, although I'm aware that doing so could make the upgrade process quite lengthy for large databases.

gschueler:
Fix #7725 - queries to clean up executions can be slow due to incorrect datatype in base_report table

* change the execution ID field from a varchar to appropriate datatype
* rename field `jcExecId` to `executionId`
* add db migration scripts
* update tests

**Additional context**
This change may cause a long re-index when the db migration script is applied depending on the DB contents.


rundeckapp/grails-app/services/rundeck/services/ReportService.groovy
```diff
@@ -302,11 +302,11 @@ class ReportService  {
                             or {
                                 List execIdFilterPartioned = Lists.partition(query.execIdFilter, 1000)
                                 execIdFilterPartioned.each { List partition ->
-                                    'in'('jcExecId', partition)
+                                    'in'('executionId', partition.collect{Long.parseLong(it)})
```
ronaveva, line 305: To further improve the performance we could avoid the use of "in" sentence by using DetachedCriteria, and instead of receive the execution id list we could just receive the authorized projects list (that should be fairly smaller) to restrict the join and then attach the subquery with exist statement

Something like this:

``` 
subquery = DetachedCriteria.forClass(ReferencedExecution.class, "re").with{
                setProjection Projections.property('re.execution.id')
                add Restrictions.conjunction().add(Restrictions.eq('re.executionId', 'this_.jcExecId'))
                add Restrictions.conjunction().add(Restrictions.eq('re.scheduledExecution.id', se.id))
                if(query.execProjects){
                    add Restrictions.conjunction().add(Restrictions.in('this.ctxProject', query.execProjects))
                }
            }
```

rundeckapp/grails-app/migrations/core/ExecReportJcExecIdToExecutionId.groovy
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    changeSet(author: "gschueler (generated)", id: "1653344096108-51", dbms: 'mysql,mariadb') {
+        sql("""update base_report set execution_id = (IF(jc_exec_id REGEXP '^[0-9]+\$', CAST(jc_exec_id as unsigned),
+ null))""")
+    }
+    changeSet(author: "gschueler (generated)", id: "1653344096108-52", dbms: 'h2,postgresql') {
+        sql("""update base_report set execution_id = CAST(JC_EXEC_ID as bigint)""")
+    }
+    changeSet(author: "gschueler (generated)", id: "1653344096108-53", dbms: 'mssql') {
```
ronaveva, line 24: It looks like this change could use the same as h2 and postgresql

fdevans:
Was this tested for upgrade scenarios?

ronaveva:
> Was this tested for upgrade scenarios?

i've tested the upgrade with mysql and h2. It looks ok

PR #7744, opened 2022-06-03
RUN-944: Add bootstrap storage tree rewrite feature

gschueler:
this feature will automatically re-store any key storage resource that is not encrypted,
when the jasypt-encryption converter plugin is in use. 

If the jasypt-encryption plugin is not configured, this feature will not do anything.
If it is in use, and the enabled flag is set, each resource that is not already encrypted will be restored to encrypt it.

configuration:

enabled (default false):

    rundeck.feature.storageRewrite.enabled=true

base path (default "keys"):

    rundeck.storage.rewrite.basePath=keys

**Additional context**

### DIFF
--- a/rundeckapp/src/main/groovy/com/dtolabs/rundeck/app/tree/TreeUpdater.groovy
+++ b/rundeckapp/src/main/groovy/com/dtolabs/rundeck/app/tree/TreeUpdater.groovy
@@ -1,0 +1,60 @@
+package com.dtolabs.rundeck.app.tree
+
+import com.dtolabs.rundeck.core.storage.ResourceMeta
+import com.dtolabs.rundeck.core.storage.StorageTree
+import groovy.transform.CompileStatic
+import groovy.util.logging.Slf4j
+import org.rundeck.storage.api.Resource
+/**
+ * Iterates over given storage tree, and updates resources according to the config
+ */
+@CompileStatic
+@Slf4j
+class TreeUpdater implements Updater {
+
+    void updateTree(StorageTree storageTree, String basePath, UpdaterConfig updaterConfig) {
+        if (!updaterConfig.shouldPerform()) {
+            return
+        }
+        //detect if configuration contains jasypt-encryption
+        log.info("Storage Tree Updater: begin: ${updaterConfig.name}")
+        int count = 0
+        List<String> paths = [basePath]
+        while (paths.size() > 0) {
+            String dirPath = paths.remove(0)
+
+            if (!storageTree.hasDirectory(dirPath)) {
+                log.warn("Skipping a path ${dirPath} - it was not a directory")
+                continue
+            }
+
+            Set<Resource<ResourceMeta>> dirList = storageTree.listDirectory(dirPath)
+            paths.addAll(
+                dirList.findAll {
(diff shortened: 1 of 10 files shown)


rundeckapp/src/main/groovy/com/dtolabs/rundeck/app/tree/TreeUpdater.groovy
```diff
@@ -0,0 +1,74 @@
...
+import org.rundeck.storage.api.Resource
+/**
+ * Iterates over given storage tree, and updates resources according to the config
+ */
+@CompileStatic
+@Slf4j
+class TreeUpdater implements Updater {
+    StorageTree storageTree
+
+    Boolean enabled
```
ehe-pd, line 16: I feel thse two configuration parameters better belong to the `UpdaterConfig` subclass: `JasyptEncryptionEnforcerUpdaterConfig`. 

```
public JasyptEncryptionEnforcerUpdaterConfig implements UpdaterConfig {
    Boolean enabled
    String basePath

    @Override
    boolean shouldPerform() {
       if (!enabled) {
            return false
        }

        def configuration = treeCreator.storageConfigMap
        return configuration.entrySet().find {
            it.key =~ /^converter\.\d+\.type/ && it.value == 'jasypt-encryption'
        }
    }
}
```

This way, we can separate the concern on config from the TreeUpdater logic which can be stateless

ehe-pd (commented):
The logic works well. 

Here is some idea maybe can simplify the code structure a bit:

1. Make the TreeUpdater a stateless class
```
class TreeUpdater implements Updater {
   void performTreeUpdate(StorageTree storageTree, UpdaterConfig updaterConfig) {
       ... ...
    }
}
```

2. Put all configuration related login into `JasyptEncryptionEnforcerUpdaterConfig`. So in the future, if we have different configuration implementations we can just replace the bean. 

3. GormTreeUpdater is the event listener connecting all moving pieces together and trigger the update procedure:
```
class GormTreeUpdater {
    Updater updater
    StorageTree storageTree
    UpdaterConfig updaterConfig

    @Subscriber("rundeck.bootstrap")
    void bootstrapEvent() {
        updater.performTreeUpdate(storageTree, updaterConfig)
    }
}
```

3. In the resource.groovy configuration: 
```
rundeckJasyptConverterUpdaterConfig(JasyptEncryptionEnforcerUpdaterConfig){
    treeCreator = ref('rundeckStorageTreeCreator')
    enabled = grailsApplication.config.getProperty('rundeck.feature.storageRewrite.enabled', Boolean.class, false)
     basePath = grailsApplication.config.getProperty('rundeck.storage.rewrite.basePath', String.class, 'keys')
 }

    rundeckTreeStorageTreeUpdater(TreeUpdater) 

    rundeckGormTreeUpdater(GormTreeUpdater){
        storageTree = ref('rundeckStorageTree')
        updaterConfig = ref('rundeckJasyptConverterUpdaterConfig')
        updater = ref('rundeckTreeStorageTreeUpdater')
    }
```

gschueler:
Updated addressing comments

ehe-pd (approved):
Tested. It works.

PR #7750, opened 2022-06-07
RUN-909: Fix: Rundeck gives error on job email notification using coma and spaces between emails

L2JE:
Fix: https://github.com/rundeckpro/rundeckpro/issues/2545

Ignored whitespace after comma if present.


alexander-variacode:
@mergifyio rebase

ronaveva (changes requested):
i think a better way to approach this would be to simply remove every space in the mail list and also store it like that. Pehaps also to restrict the UI to avoid the use of blank spaces

ronaveva (changes requested):
@L2JE this needs to be applied at backend level also. For example, if a user wants to import a project from 3.4.X it should be able to.

ronaveva (approved):
LGTM

L2JE:
Must add to release notes:
Serialization changes from 3.4.x
If a customer upgrades from 3.4.X, they will have to commit for changes on email notification's recipients.

PR #7763, opened 2022-06-08
RUN-950: Changing default SCM SSH authentication to SSHJ

ltamaster:
enhancement:
Replacing default SCM SSH authentication (using JSCH) by SSHJ

**Additional context**
GitHub has tightened up their SSH key requirements, and wants SSH-RSA 2 keys. The new default for ssh-keygen via OpenSSH is not supported by JSCH, and is likely why we saw an influx of issues involving the error "invalid privatekey"


plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/ssh/SshjSessionFactory.groovy
```diff
@@ -0,0 +1,92 @@
...
+        if (config == null)
+            config = OpenSshConfig.get(fs)
+
+        final OpenSshConfig.Host hc = config.lookup(host)
+        if (port <= 0)
+            port = hc.getPort()
+        if (user == null)
+            user = hc.getUser()
+
+        sshClient = createConnection(user, host,port)
```
ehe-pd, line 55: Is there any specific reason to make `sshClient` a property of the factory instance? The property value is overwritten every time `getSession` gets called, which means the old one got discarded.

A session factory is better to be an immutable singleton object without saving a SSHClient as its internal state.

According to the document about SSHClient and RemoteSession, they seem to be an abstraction of the same thing - an established SSH tunnel connected to a host. I think it's natural that an SSHClient instance should be part of the SshjSession instance.  

```suggestion
        return new SshjSession(createConnection(user, host, port)
```

plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/ssh/SshjSessionFactory.groovy
```diff
@@ -0,0 +1,92 @@
...
+        return new SshjSession(sshClient, uri)
```
ehe-pd, line 56:
```suggestion
        
```

plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/ssh/SshjSessionFactory.groovy
```diff
@@ -0,0 +1,92 @@
...
+            ssh.loadKnownHosts()
+        }else{
+            ssh.addHostKeyVerifier(new PromiscuousVerifier())
+        }
+
+        if (port != null) {
+            ssh.connect(host, port)
+        } else {
+            ssh.connect(host);
+        }
```
ehe-pd, line 82: SSHClient is an auto-closable resource. It's better to be enclosed in a try-with-resources block.
```
try(ssh.connect(host, port)) {
}
```

carlosrfranco:
Tested using Github and it worked

plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/ssh/SshjSession.groovy
```diff
@@ -0,0 +1,157 @@
...
+        this.uri = uri
+
+        this.sshClient = createConnection()
+
+    }
+
+    @Override
+    Process exec(String commandName, int timeout) throws IOException {
+        this.session = sshClient.startSession()
+        return new SshjProcess(commandName, timeout)
```
ehe-pd, line 39: To utilize the autoclosable feature of Session and Command:
```
try(Session session = sshClient.startSession()
    Session.Command command = session.exec(commandName)
) {
   return new SshProcess(command, timeout)
}
```

plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/ssh/SshjSession.groovy
```diff
@@ -0,0 +1,157 @@
...
+import org.eclipse.jgit.errors.TransportException
+import org.eclipse.jgit.transport.OpenSshConfig
+import org.eclipse.jgit.transport.RemoteSession
+import org.eclipse.jgit.transport.URIish
+import org.eclipse.jgit.util.io.IsolatedOutputStream
+
+@CompileStatic
+class SshjSession implements RemoteSession {
+
+    Session session
```
ehe-pd, line 19: This property can be removed.

plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/ssh/SshjSession.groovy
```diff
@@ -0,0 +1,157 @@
...
+
+    private class SshjProcess extends Process {
+
+        int timeout
+        Session.Command cmd
+        InputStream inputStream
+        OutputStream outputStream
+        InputStream errorStream
+
+        SshjProcess(String commandName, int tms) throws TransportException, IOException {
```
ehe-pd, line 99:
```
SshjProcess(Session.Command command, int tms) throws TransportException, IOException {
```

plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/ssh/SshjSession.groovy
```diff
@@ -0,0 +1,157 @@
...
+            try {
+                timeout = tms
+                cmd = session.exec(commandName)
```
ehe-pd, line 102: cmd = command

plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/ssh/SshjSession.groovy
```diff
@@ -0,0 +1,157 @@
...
+                throw new IllegalStateException()
+            }
+            return cmd.getExitStatus()
+        }
+
+        @Override
+        void destroy() {
+            cmd.close()
+            closeOutputStream()
+            session.close()
```
ehe-pd, line 134: Can be changed to:
```suggestion: 
         if(cmd != null) cmd.close() 
         closeOutputStream()
```

ehe-pd (changes requested):
The refactor looks good. 
There are a few improvements I think are worth applying.

plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/ssh/SshjSession.groovy
ltamaster: I tried this but I got an error: Stream is closed, so maybe we can leave it as it is now

ehe-pd (approved):

PR #7770, opened 2022-06-14
instance name color fixed at login and logout

Jesus-Osuna-M:
Partially fix: https://github.com/rundeckpro/rundeckpro/issues/1868

This PR only fix property rundeck.gui.instanceNameLabelColor.  
 
Problem:
When using rundeck.gui.instanceNameLabelColor didn't change the color of the instance name on the login page nor on the logout page. A CSS class was missing on the HTML.

Solution:
Change the CSS class to match other views where this class worked as expected.  

**Before fix:**
imagen

**After fix:**
imagen


DarwisNarvaezDev:
LGTM

PR #7784, opened 2022-06-24
Checking if jquery element has length greater than zero to avoid access an attribute on a undefined object

carlosrfranco:
fixes https://github.com/rundeckpro/rundeckpro/issues/2609

An javascript error was preventing acl ui manager to be opened

This PR check if jquery element has length greater than zero to avoid access an attribute on a undefined object


carlosrfranco:
@Mergifyio rebase

gschueler (approved):

PR #7785, opened 2022-06-28
RUN-1006: Fix Notification Property Value type when SCM Export YAML format.

Jesus-Osuna-M:
Fix: https://github.com/rundeckpro/rundeckpro/issues/2212

Problem
Exporting a job on YAML format using SCM a notification property was showing a string value when a boolean was expected. 

Solution
Change the value set from 'true' to true.


jtobard (changes requested):
Need unit or functional test

Jesus-Osuna-M:
@Mergifyio rebase

PR #7786, opened 2022-06-28
Fix to enable job editing when SCM fails

avelasquezr:
An additional property was added to recognize the enabling of SCM and activate the loader in case it is removed/has no permissions.

Bugfix: https://github.com/rundeckpro/rundeckpro/issues/2567


avelasquezr:
RUN-947

jtobard (changes requested):
This need some test, if it is not possible to do it on core should be done on rundeck-tester

jtobard (approved):

PR #7791, opened 2022-06-30
RUN-1045: fixed broken error message that appears when you have not stored a password

mcat115:
I was using the debugger to trace an error I've been stuck on all day. Once I found it, I realized it was a bug. I found the message in the java file TypedStorageTreeImpl.java, but the string is hard coded with what looks to be interpolation, but since it's a java file and you cannot do that in Java, I'm literally getting an error with the interpolation syntax instead of any data. Because of this, it would print ${path} instead of printing the actual path that variables holds. So I fixed it to work with Java syntax instead of the broken Groovy syntax it was using. I also removed a period from the end of another plugin description to fit with the style I have been implementing in the moving cloud repos ticket

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/storage/TypedStorageTreeImpl.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/storage/TypedStorageTreeImpl.java
@@ -46,8 +46,9 @@ public class TypedStorageTreeImpl extends StorageTreeImpl implements TypedStorag
         if (resource.getContents().getContentType().equals(contentType)) {
             return resource;
         }
+        String errorMessage = "Path " + path + " does not store a password, content-type: " + contentType;
         throw new WrongContentType(
-                "Path ${path} does not store a password, content-type: ${resource.contents.contentType}",
+                errorMessage,
                 StorageException.Event.READ,
                 path
         );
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/dtolabs/rundeck/core/storage/TypedStorageTreeImpl.java
```diff
@@ -46,8 +46,9 @@ public Resource<ResourceMeta> getResourceWithType(Path path, String contentType)
         if (resource.getContents().getContentType().equals(contentType)) {
             return resource;
         }
+        String errorMessage = "Path " + path + " does not store a password, content-type: " + contentType;
```
ehe-pd, line 49: It's a good fix. 
The original error message is quite confusing; based on my understanding the problem is `Path <path> does not store a resource for content-type: <contentType>`.

ehe-pd (changes requested):
Maybe we can make the error message more accurate and readable?

eric-leihe (approved):

ehe-pd (approved):

fdevans:
Is this customer visible?  If so, can we get a description or title that highlights what was fixed from their perspective?  If not, we can add the `release-notes/exclude` label to it.

mcat115:
If the customer does not provide a password as needed for certain credentials, the error message will now say which password key path path is incorrectly configured instead of an incorrectly interpolated "path" variable. It is now functioning as was originally intended

PR #7802, opened 2022-07-11
RUN-1013: Fix global log limit issue

alexander-variacode:
Issue:
https://pagerduty.atlassian.net/browse/RUN-872

Fix description:
Fixed global log limit adding a Failed status on a job.


ronaveva (changes requested):
I think a couple of things could be improved

- Method name does not match what it actually does, it could be "setMaxValueInfo" or so, since it doesn't just add the value, but also description and type
- New method needs javadoc
- It lacks tests

ronaveva (approved):
LGTM

alexander-variacode:
@mergifyio rebase

PR #7805, opened 2022-07-14
RUN-1041: Fix Date Format differences in Context Variable output

avelasquezr:
# RUN-1041 - Bugfix for date different formats

As seen here: https://github.com/rundeckpro/rundeckpro/issues/2617 there was a problem while displaying context variables with date data.

**Solution**
There is a difference between the date format of the classes: Date and Timestamp, which variable "dateStarted" is in Timestamp and "dateEnded" is in Date. It was changing to the same format using the SimpleDateFormat class, but the most comfortable and light solution is to use the class:

**Exhibits**
_Pre-fix:_
image
_Post-fix_
image

PD: RUN-1041


jtobard (changes requested):
Missing test

jtobard (approved):

PR #7812, opened 2022-07-19
RUN-1057: Fix: Unable to save passwords kV version=1 hashicorp vault

L2JE:
Upgrade vault plugin version

Depends on: https://github.com/rundeck-plugins/vault-storage/pull/43
Fix: https://github.com/rundeck-plugins/vault-storage/issues/26


L2JE:
@Mergifyio rebase

ltamaster:
@L2JE I think we don't need this. The `prefix` setting is a property of the vault plugin, so we don't need this code here. Also, I tested the vault plugin PR (https://github.com/rundeck-plugins/vault-storage/pull/43)  without setting the `prefix` attribute and it worked OK (using the main brach of Rundeck)

PR #7820, opened 2022-07-27
BUGFIX RUN-1084: unable to update webhook

DarwisNarvaezDev:
# Reactive Bugfix to Webhooks
As seen here: https://github.com/rundeckpro/rundeckpro/issues/2680, the GUI was not editing any webhooks.

## The problem
The method 'saveWebhook' of the 'WebhookService', is the responsible for a) saving a new webhook, b) Importing a new webhook and c) editing an existing webhook, so, when this PR: https://github.com/rundeck/rundeck/pull/7737, prevented the import process for webhooks with the same name, also caused the error when we call 'saveWebhook' with the same name for editing.

## The solution
When the 'importWebhook' is called, it puts an extra attribute inside the 'hook' object to be passed to 'saveWebhook' method, so, what we did was evaluate if that property exists in the hook in order to permit **'saving' for editing and not for importing**.

### DIFF
--- a/grails-webhooks/grails-app/services/webhooks/WebhookService.groovy
+++ b/grails-webhooks/grails-app/services/webhooks/WebhookService.groovy
@@ -160,7 +160,8 @@ class WebhookService {
         }
         hook.uuid = hookData.uuid ?: hook.uuid
         int countByNameInProject = Webhook.countByNameAndProject(hookData.name,hookData.project)
-        if(countByNameInProject > 0) {
+        def importFlag = (hook.hasProperty('importData') && hook.importData) ? hook.importData : false
+        if(countByNameInProject > 0 && importFlag ) {
             return [err: " A Webhook by that name already exists in this project"]
         }
         hook.name = hookData.name ?: hook.name
--- a/grails-webhooks/grails-app/services/webhooks/WebhookService.groovy
+++ b/grails-webhooks/grails-app/services/webhooks/WebhookService.groovy
@@ -159,13 +159,9 @@ class WebhookService {
             hook.uuid = UUID.randomUUID().toString()
         }
         hook.uuid = hookData.uuid ?: hook.uuid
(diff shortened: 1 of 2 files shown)


grails-webhooks/grails-app/services/webhooks/WebhookService.groovy
```diff
@@ -159,9 +159,17 @@ class WebhookService {
...
+        // If there's an import, we validate that the name of the WH is not duplicated in the project.
+        int countByNameInProject = Webhook.countByNameAndProject(hookData.name, hookData.project)
+        if (countByNameInProject > 0) {
+            // If we find during the import a WH with the same name, we check for duplicated UUID.
+            int countByUuidInProject = Webhook.countByUuidAndProject(hookData.uuid, hookData.project)
+            // If we find the UUID, it means that is a different webhook with duplicated name, so we prevent the import
+            if( countByUuidInProject == 0 ) return [err: "A Webhook by that name already exists in project"]
+        }else{
+            // If we don't find the WH by name in the project, we check if the WH exists in other projects to prevent errors
+            int countByUuidInDb = Webhook.countByUuid(hookData.uuid)
```
jtobard, line 171: This count prevent renaming the job, need to exclude current project

jtobard (changes requested):
int countByUuidInDb = Webhook.countByUuid(hookData.uuid) this line

grails-webhooks/grails-app/services/webhooks/WebhookService.groovy
jtobard: Looking at the previous code, the current behavior is to allow having the same uuid on multiple projects, the uuid only is verified if exists on the current project so this need to be removed

ronaveva (approved):
LGTM

jtobard (approved):

PR #7821, opened 2022-07-28
RSE-9-job-remote-option-url-retry-connection-parameter-is-not-honored

Jesus-Osuna-M:
Fix: https://github.com/rundeckpro/rundeckpro/issues/2644

Problem: 

When a job option was set from a remote URL the retry param was not taken in count, any retry was made.
There are two ways of setting those configurations, directly on the URL or the config.properties, the ones from the config.properties weren't being used. 

Solution:

On the RundeckBaseConfig add the missing params to be accessible. 
Implement on the Rundeck side a retry method if something fails when trying to obtain the values from an URL. 

A similar solution is already implemented on the code on rundeckapp/grails-app/services/rundeck/services/NotificationService.groovy on the postDataUrl method

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
@@ -3921,9 +3921,7 @@ class ScheduledExecutionService implements ApplicationContextAware, Initializing
             try {
                 //validate is no the first attempt
                 if ( retryCount > count ){
-                    if(count>0){
-                        Thread.sleep(contimeout*1000)
-                    }
+                    Thread.sleep(contimeout*1000)
                 }
                 def framework = frameworkService.getRundeckFramework()
                 def projectConfig = framework.frameworkProjectMgr.loadProjectConfig(scheduledExecution.project)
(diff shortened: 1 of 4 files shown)


Jesus-Osuna-M:
@Mergifyio rebase

rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
```diff
@@ -3911,32 +3913,49 @@ class ScheduledExecutionService implements ApplicationContextAware, Initializing
...
-            log.error("getRemoteJSON error: URL ${cleanUrl} : ${remoteResult.error}");
-        }
-        logRemoteOptionStats(remoteStats, [jobName: scheduledExecution.generateFullName(), id: scheduledExecution.extid, jobProject: scheduledExecution.project, optionName: mapConfig.option, user: username])
+        int count = retryCount
+
+        //cycle to retry getting the JSON from a remoteURL
+        do{
+            try {
+                //validate is no the first attempt
+                if ( retryCount > count ){
```
ronaveva, line 3923: @Jesus-Osuna-M there is no need for 2 separated "if" conditions

ronaveva (changes requested):
It lacks a test when getting an exception from getRemoteJSON method

PR #7793, opened 2022-06-30
RUN-999: Add audit tracking for jobs and acls.

ryunge:
This PR adds tracking events to the [AuditEventListener](https://docs.rundeck.com/docs/developer/audit-events-listeners.html#audit-events-listener-plugins) plugin interface to track the following additional events:

*  Job events: Create, Update, Delete, Run
*  ACL events: System & Project Acl update & delete

Also adds a default plugin for logging the events to the log4j service logger, and adding also a logs/rundeck.audit.events.log file.

PR #7826, opened 2022-08-01
RUN-1082: Add audit tracking for jobs and acls

ahormazabal:
Imported from: https://github.com/rundeck/rundeck/pull/7793

Original author: @ryunge

--

This PR adds tracking events to the [AuditEventListener](https://docs.rundeck.com/docs/developer/audit-events-listeners.html#audit-events-listener-plugins) plugin interface to track the following additional events:

    Job events: Create, Update, Delete, Run
    ACL events: System & Project Acl update & delete

Also adds a default plugin for logging the events to the log4j service logger, and adding also a logs/rundeck.audit.events.log file.


rundeckapp/templates/config/log4j2.properties.template
```diff
@@ -35,6 +35,18 @@ appender.audit.policies.type = Policies
...
+appender.audit_events.name = audit_events
+appender.audit_events.fileName = ${baseDir}/rundeck.audit.events.log
+appender.audit_events.append = true
+appender.audit_events.bufferedIO = true
+appender.audit_events.filePattern = ${baseDir}/rundeck.audit.events.log.%d{yyyy-MM-dd}.gz
+appender.audit_events.layout.type = PatternLayout
+appender.audit_events.layout.pattern = ${prefix} - %m%n
+appender.audit_events.policies.type = Policies
+appender.audit_events.policies.time.type = TimeBasedTriggeringPolicy
+appender.audit_events.policies.time.interval = 1
```
gschueler, line 48: i think all of the log4j config updates also need to be added to the [packaging repo](https://github.com/rundeck/packaging/blob/main/lib/common/etc/rundeck/log4j2.properties)

plugins/audit-logging-plugin/src/main/groovy/org/rundeck/plugin/audit/AuditLoggerPlugin.groovy
```diff
@@ -0,0 +1,35 @@
...
+  @Override
+  void init() {
+    LOGGER.debug("Audit events logging plugin initialized.")
+  }
+
+  /**
+   * Log received events
+   */
+  void onEvent(AuditEvent event) {
+    LOGGER.info("Audit Event: " + event)
```
gschueler, line 31: perhaps a future enhancement could use the MDC context variables to allow the log4j logging template to better specify the format of the data

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -2536,6 +2540,15 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
             throw new ExecutionServiceException('Job "' + se.jobName + '" {{Job ' + se.extid + '}}: Limit of running executions has been reached.', 'conflict')
         }
 
+        // Publish audit event for new job run.
+        if(auditEventsService) {
```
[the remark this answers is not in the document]
ahormazabal: Makes sense, but from other point of view, wouldn't it be better to log the attempt of running a job even if it failed launching?

plugins/audit-logging-plugin/src/main/groovy/org/rundeck/plugin/audit/AuditLoggerPlugin.groovy
ahormazabal: I understand the intent was to provide a basic default behavior. Indeed there's a lot of room for improvement here.

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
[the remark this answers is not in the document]
ahormazabal: As discussed in slack, the call was moved after the execution creation and the execution id was added to the event.

gschueler (approved):

rundeckapp/templates/config/log4j2.properties.template
gschueler: created PR https://github.com/rundeck/packaging/pull/36

PR #7829, opened 2022-08-03
RUN-1105: Upgrade grails to 5.1.9

ahormazabal:
Upgrade grails to 5.1.9 to address CVE-2022-35912

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -18,9 +18,11 @@
 #####################################################################
 group=org.rundeck
 currentVersion = 4.5.0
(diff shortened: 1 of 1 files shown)


gschueler (approved):

mukeshag80:
we are using 4.12 and 4.15 Rundeck community edition , how do we get rid of Grail 4.0.1 and 4.0.0 from the package level as scanner is identifying critical findings at package level or post installation it doesnt get installed ?

PR #7825, opened 2022-07-29
RUN-1092: Update okhttp to 4.9.2

ahormazabal:
Update okhttp to address PRISMA-2022-0239 https://github.com/square/okhttp/issues/6738

PR #7835, opened 2022-08-09
RUN-1092: Force okhttp version to 4.9.2 to address security issues

ahormazabal:


ahormazabal:
Cancelled because https://github.com/rundeck/rundeck/pull/7825

PR #7842, opened 2022-08-11
RUN-224: Fix: Enabling/Disabling both Scheduling and Executions at the same time using timer mode

L2JE:
Fix: https://github.com/rundeckpro/rundeckpro/issues/1311


ronaveva (approved):
LGTM

PR #7856, opened 2022-08-16
RSE-9-job-remote-option-url-retry-connection-parameter-not-honored

Jesus-Osuna-M:
Fix: https://github.com/rundeckpro/rundeckpro/issues/2644

Problem:

When a job option was set from a remote URL the retry param was not taken in count, any retry was made.
There are two ways of setting those configurations, directly on the URL or the config.properties, the ones from the config.properties weren't being used.

Solution:

On the RundeckBaseConfig add the missing params to be accessible.
Implement on the Rundeck side a retry method if something fails when trying to obtain the values from an URL.

A similar solution is already implemented on the code on rundeckapp/grails-app/services/rundeck/services/NotificationService.groovy on the postDataUrl method


ronaveva (approved):
LGTM

PR #7873, opened 2022-08-23
fix: RSE-34 validation added to prevent invalid captured pattern

Oalbarran94:
fix: https://github.com/rundeckpro/rundeckpro/issues/2661
The validation was added when you are trying to save a new capture pattern:

Case 1, pattern with no captured group: 

image

Case 2, valid pattern group with no name:

image

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/plugins/configuration/PropertyValidator.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/plugins/configuration/PropertyValidator.java
@@ -32,7 +32,14 @@ import java.util.*;
 public interface PropertyValidator {
     public boolean isValid(String value) throws ValidationException;
 
-    default boolean isValid(String value,Map<String,Object> props) throws ValidationException {
+    /**
+     * Custom validator to check for any property from a defined plugin.
+     * @param value Regex to validate
+     * @param props Plugin properties
+     * @return Boolean indicating validity of presented values
+     * @throws ValidationException
+     */
+    default boolean isValid(String value, Map<String,Object> props) throws ValidationException {
         return isValid(value);
     }
 }
--- a/core/src/main/java/com/dtolabs/rundeck/core/plugins/configuration/PropertyValidator.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/plugins/configuration/PropertyValidator.java
@@ -31,4 +31,8 @@ import java.util.*;
  */
 public interface PropertyValidator {
     public boolean isValid(String value) throws ValidationException;
+
+    default boolean isValid(String value,Map<String,Object> props) throws ValidationException {
+        return isValid(value);
+    }
 }
--- a/rundeckapp/src/main/groovy/com/dtolabs/rundeck/server/plugins/logging/SimpleDataFilterPlugin.groovy
+++ b/rundeckapp/src/main/groovy/com/dtolabs/rundeck/server/plugins/logging/SimpleDataFilterPlugin.groovy
@@ -106,6 +106,25 @@ See the [Java Pattern](https://docs.oracle.com/javase/8/docs/api/java/util/regex
     String invalidKeyPattern
 
     static class RegexValidator implements PropertyValidator {
+
+        @Override
+        boolean isValid(String value, Map<String,Object> props) throws ValidationException {
+            try {
+                def compile = Pattern.compile(value)
+                Matcher m = compile.matcher("");
+
+                if(m.groupCount() == 0){
+                    throw new PatternSyntaxException("Patter must have at least one group", value, -1)
+                }
+                if(m.groupCount() == 1 && !props.containsKey("name")){
+                    throw new PatternSyntaxException("It must be defined the field name", value, -1)
+                }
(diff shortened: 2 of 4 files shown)


rundeckapp/src/main/groovy/com/dtolabs/rundeck/server/plugins/logging/SimpleDataFilterPlugin.groovy
```diff
@@ -106,6 +105,32 @@ See the [Java Pattern](https://docs.oracle.com/javase/8/docs/api/java/util/regex
...
+        boolean isValid(String value, Map<String,Object> props) throws ValidationException {
+            try {
+
+                def invalidKeyPattern = "";
+
+                if(props.containsKey("invalidKeyPattern")){
+                    invalidKeyPattern = props.get("invalidKeyPattern")
+                }
+
+                def compile = Pattern.compile(value)
```
gschueler, line 119: just wrap the Pattern.compile in the try/catch for PatternSyntaxException

rundeckapp/src/main/groovy/com/dtolabs/rundeck/server/plugins/logging/SimpleDataFilterPlugin.groovy
```diff
@@ -106,6 +105,32 @@ See the [Java Pattern](https://docs.oracle.com/javase/8/docs/api/java/util/regex
...
+                Matcher m = compile.matcher("");
+
+                if(value != invalidKeyPattern && m.groupCount() == 0){
+                    throw new PatternSyntaxException("Patter must have at least one group", value, -1)
```
gschueler, line 123:
```suggestion
                    throw new ValidationException("Pattern must have at least one group", value, -1)
```

rundeckapp/src/main/groovy/com/dtolabs/rundeck/server/plugins/logging/SimpleDataFilterPlugin.groovy
```diff
@@ -106,6 +105,32 @@ See the [Java Pattern](https://docs.oracle.com/javase/8/docs/api/java/util/regex
...
+                }
+                if(value != invalidKeyPattern && m.groupCount() == 1 && !props.containsKey("name")){
+                    throw new PatternSyntaxException("It must be defined the field name", value, -1)
```
gschueler, line 126:
```suggestion
                    throw new ValidationException("The Name field must be defined when only one capture group is specified", value, -1)
```

core/src/main/java/com/dtolabs/rundeck/core/plugins/configuration/PropertyValidator.java
```diff
@@ -31,4 +31,8 @@
  */
 public interface PropertyValidator {
     public boolean isValid(String value) throws ValidationException;
+
+    default boolean isValid(String value,Map<String,Object> props) throws ValidationException {
```
gschueler, line 35: should add some javadoc

rundeckapp/src/main/groovy/com/dtolabs/rundeck/server/plugins/logging/SimpleDataFilterPlugin.groovy
```diff
@@ -106,6 +105,32 @@ See the [Java Pattern](https://docs.oracle.com/javase/8/docs/api/java/util/regex
...
+                def invalidKeyPattern = "";
+
+                if(props.containsKey("invalidKeyPattern")){
+                    invalidKeyPattern = props.get("invalidKeyPattern")
+                }
+
+                def compile = Pattern.compile(value)
+                Matcher m = compile.matcher("");
+
+                if(value != invalidKeyPattern && m.groupCount() == 0){
```
gschueler, line 122: this is potentially flaky, it would be better to have two separate validators for `invalidKeyPattern`, and the `regex` properties.  The invalidKeyPattern just needs a simple Regex validator, while the other has to check the number of groups in the regex. could subclass the simple Regex validator for the second case.

gschueler (changes requested):

rundeckapp/src/main/groovy/com/dtolabs/rundeck/server/plugins/logging/SimpleDataFilterPlugin.groovy
```diff
@@ -105,40 +105,42 @@ See the [Java Pattern](https://docs.oracle.com/javase/8/docs/api/java/util/regex
...
-                    throw new PatternSyntaxException("It must be defined the field name", value, -1)
-                }
                 return true
             } catch (PatternSyntaxException e) {
                 throw new ValidationException(e.message, e)
             }
         }
+    }
+
+    static class RegexPropertyValidator implements PropertyValidator {
```
gschueler, line 119: `RegexPropertyValidator` doesn't seem to be an accurate name, it should probably be something like `NamePropertyValidator` because it only applies to the "name" property

gschueler (approved):
minor class name change requested, but otherwise it looks good

Issue #7223, opened 2021-08-18
Kill job doesn't work


demogorgonz:
**Describe the bug**
Kill job doesn't kill actual job, process keeps running in background.

**My Rundeck detail**

* Rundeck version: Community 3.4.0
* install type: [chocolatey](https://community.chocolatey.org/packages/rundeck)
* OS Name/version: Win Server 2019

**To Reproduce**
1. Create a step `script` which run script `.\test.ps1 ` with loop task like `while(1) { sleep 1; Get-Random | Out-File 1.log }`
2. Set invocation string `pwsh -NonInteractive -NoProfile -ExecutionPolicy bypass -File`
3. Set file extension `.ps1`
4. Run the job, kill the job.
5. See the script running in process list, or check file content as it keeps changing (script continue to execute after "kill job")

![Capture](https://user-images.githubusercontent.com/5041776/129865288-3bc7e7bc-63c5-4aad-9e3a-3b75d88c314b.PNG)

**Expected behavior**
To kill the job ?

**Desktop (please complete the following information):**
 - OS: [e.g. Win 10]
 - Browser chrome
 - Version 92.0.4515.159

majkinetor:
This happens to me to. Previously this only happened with exe's that are spawned from within shell but now shell itself, that is spawned by Rundeck is not killed which is a major problem.

Khelben-Arunsun:
Yeah there should be a Rundeck option to Hook the kill button, catch the event, and Kill the remote process. 
The Knockout JS should deliver the button click event to end a job to a rundeck option, and then wait for the child process has ended successful callback , then end the current execution. 

`bool Rundeck-Option-kill-job == False;
While (job is running){ 
if (kill job on-click = true){

  try {
              HttpResponse<String> response = client.send(request, HttpResponse.BodyHandlers.ofString());
              if (response.statusCode()==200) {
                  Rundeck-Option-kill-job == true;
              } else
                  System.out.println(response.statusCode());
                  System.out.println(response.body());
                  }
  
          } catch (IOException | InterruptedException e) {
              System.out.println("caught exception... Job Did not kill properly");
              e.printStackTrace();
          }

    }
}`

demogorgonz:
Any update on this bug?

sanchobouillant:
Hi, 

Same issue with PHP scripts :)

demogorgonz:
Bumping this issue, the bug was submitted on 18th of August and is still not addressed...

Issue #1038, opened 2014-12-01
rundeck "Kill Job" does not terminate tree of child processes


lruslan:
It seem rundeck "Kill Job" logic does not ensure that child processes killed and exited, leaving child process running in background in detached state.

Environment:
Rundeck version: latest compiled from main 2.4.0-0-SNAPSHOT also official release 2.2.3-1
OS:Ubuntu 14.04.1 LTS , Red Hat Enterprise Linux Server release 6.5

Use case:
Job called locally or remotely ( using script node executor because sometimes we need pseudo terminal to be created ssh -q -tt ${node.username}@${node.hostname} ${exec.command})
Yaml job definition:

```
- project: test
  sequence:
    keepgoing: false
    strategy: node-first
    commands:
    - exec: /tmp/1.sh
  name: job1
```

Script 1.sh

```
#!/bin/bash
python /tmp/1.py
```

Script /tmp/1.py

```
import time
while 1 > 0:
    time.sleep(5)
    print "Hi"
```

Actual behavior:
After job is started press "Kill job" .
Job terminates with log entry:

```
Result: 2147483647
Failed: NonZeroResultCode: Result code was 2147483647
Execution failed: 28: [Workflow step failures: {1=Dispatch failed on 1 nodes: [localhost: NonZeroResultCode: Result code was 2147483647]}, Node failures: {localhost=[NonZeroResultCode: Result code was 2147483647]}]
```

ps axu, and pstree display process still alive and running:

```
#ps ax
rundeck   5386  0.0  0.0  17976  1432 ?        S    12:33   0:00 /bin/bash /tmp/1.sh
rundeck   5387  0.0  0.1  29840  4996 ?        S    12:33   0:00 python /tmp/1.py

~# pstree -a
init
  ├─1.sh /tmp/1.sh
  │   └─python /tmp/1.py
```

Similar problem with script node executor : 
Executor script:

```
ssh -q -tt ${node.username}@${node.hostname} ${exec.command}  
```

After kill pressed rundeck reports job as aborted
Log shows message:

```
java.lang.InterruptedException
Failed: Unknown: java.lang.InterruptedException
Execution failed: 7: [Workflow step failures: {1=Dispatch failed on 1 nodes: [docker: Unknown: java.lang.InterruptedException]}, Node failures: {docker=[Unknown: java.lang.InterruptedException]}]
```

pstree still reports running process 

```
[...]

PR #7881, opened 2022-08-25
make script-plugin terminate processes on interrupt

siilike:
Fixes or provides a workaround for #1038, #7223, and many others with a similar issue that have been closed as stale or duplicate.

Previously when killing a job (manually or automatically) the currently executing process actually continued executing. Now upon an exception or interrupt the process is killed.

This pull request fixes it only in the script plugin, but the same issue exists in the localexec plugin and possibly elsewhere.


siilike:
This is not stale, but rather very embarassing that nothing has been done to implement a 6-line fix for something that was reported in 2014!

siilike:
Not stale.

PR #7885, opened 2022-08-26
RSE-78: Unable to edit job when SCM is still activated

avelasquezr:
Fixes: https://github.com/rundeckpro/rundeckpro/issues/2567
Related to this PR too: https://github.com/rundeck/rundeck/pull/7786

# Bugfix for displaying menu option for jobs

This solution performs the improvement to cover the specific scenarios, where to remove the key storage, and then the latter is added again, the job options menu will not be opened to edit it.

**Steps for this specific scenario:**
1) Create a key storage with the password from the git. (path: keys/admin/git.pass)
2) Create a job
3) Configure SCM Export with the referenced key. (key path: keys/${user.login}/git.pass)
4) Check that the options of the job are called.
5) Delete the key storage.
6) Verify that the job options are called.
7) Add the same key storage again
8) Verify that the job options are called (here the error occurred)


jtobard:
This was approved for some reason GH didn't reflect the status but its ok.
Selenium test hest: https://github.com/rundeckpro/rundeck-tester/pull/178

PR #7886, opened 2022-08-26
RUN-1163:  Fix Autocomplete job context variables in job notifications

ltamaster:

Fix autocomplete job context variables in job notifications

- rollback https://github.com/rundeck/rundeck/pull/7704
- set ace autocomplete per element
- in vue plugin component, use the UIV autocomplete instead of using jquery (for compatibility)

**Additional context**

### DIFF
--- a/rundeckapp/grails-app/views/framework/_pluginConfigPropertyFormField.gsp
+++ b/rundeckapp/grails-app/views/framework/_pluginConfigPropertyFormField.gsp
@@ -268,6 +268,7 @@
         <g:set var="syntaxSelectable" value="${prop.renderingOptions?.(StringRenderingConstants.CODE_SYNTAX_SELECTABLE)}"/>
         <g:set var="syntaxModes" value="${prop.renderingOptions?.(StringRenderingConstants.CODE_SYNTAX_MODES_ALLOWED)}"/>
 
+        <g:set var="extraInputCss" value="${extraInputCss?.replaceAll("context_var_autocomplete","")}"/>
 
         <g:set var="scriptTokenAutocomplete" value="${prop.renderingOptions?.(StringRenderingConstants.SCRIPT_TOKEN_AUTOCOMPLETE)}"/>
         <g:set var="scriptTokenAutocompleteCss" value="${scriptTokenAutocomplete&&StringRenderingConstants.ScriptTokenAutocomplete.ENABLED.equalsOrString(scriptTokenAutocomplete)?'_wfscriptitem':''}"/>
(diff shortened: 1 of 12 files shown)


rundeckapp/grails-app/views/framework/_pluginConfigPropertyFormField.gsp
```diff
@@ -268,6 +268,7 @@
         <g:set var="syntaxSelectable" value="${prop.renderingOptions?.(StringRenderingConstants.CODE_SYNTAX_SELECTABLE)}"/>
         <g:set var="syntaxModes" value="${prop.renderingOptions?.(StringRenderingConstants.CODE_SYNTAX_MODES_ALLOWED)}"/>
 
+        <g:set var="extraInputCss" value="${extraInputCss?.replaceAll("context_var_autocomplete","")}"/>
```
gschueler, line 271: i think the only usage for `extraInputCss` is to set it to `context_var_autocomplete` currently, so could probably just remove it from the handling for CODE display type, i.e. remove it from line 284

ltamaster: I did that first but then I changed it hehhe

mishingo (approved):

test/selenium/src/pages/jobCreate.page.ts
```diff
@@ -126,8 +126,13 @@ export const Elems = {
     otherTab  : By.css('#job_edit_tabs > li > a[href=\'#tab_other\']'),
     multiExecFalse: By.css('input#multipleFalse'),
     multiExecTrue: By.css('input#multipleTrue'),
-    retryInput: By.css('input#schedJobRetry')
-    // Job Queue tab
+    retryInput: By.css('input#schedJobRetry'),
+
+    contextAutocomplete  : By.css('div[class="autocomplete-suggestions"]'),
+    notificationContextAutocomplete  : By.xpath('//*[@id="notification-edit-config"]/div/div/div[1]/div/div[1]/div/div[1]/section/ul/li/a'),
```
gschueler, line 132: this kind of xpath selector is going to be very brittle to changes in the html.

gschueler (commented):
Functionality mostly works, there are a few issues:

* in Notifications: typing `${` at the first part of the input field works.  However if you type anything else first, it seems the autocomplete doesn't work elsewhere in the text field
* (minor) Autocomplete window appears to float down below the input field sometimes

Notification: works at start of input field:

Screen Shot 2022-09-13 at 9 35 40 AM

Notification: doesn't work after typing some text:
Screen Shot 2022-09-13 at 9 35 45 AM

ltamaster:
@gschueler it seems the autocomplete component works like that (match with the beginning of the string). I have been checking other VUE autocomplete components and most of them work like that. I will continue checking options.

ehe-pd (approved):
Tested. Autocomplete works.

PR #7891, opened 2022-09-01
RUN-933: Add note about new example plugins repo

qualman:
This is an enhancement to notify users about the new location for [Rundeck plugin examples](https://github.com/rundeck/rundeck-plugin-examples) once it is released.

README note in the root folder of example content.

N/a

**Additional context**
**Shouldn't be merged until https://github.com/rundeck/rundeck-plugin-examples/pull/1 is merged, and we're ready to open that example plugin repo to the public (currently internal only).**


jsboak (approved):
looks good 👍

PR #7894, opened 2022-09-05
RUN-1206: Update snakeyaml to 1.31 to address CVE-2022-25857

ahormazabal:
Update snakeyaml to 1.31 because CVE-2022-25857


ltamaster (approved):
LGTM

PR #7929, opened 2022-09-26
RSE-50: fix: allowing user to see logs after the job is deleted

Oalbarran94:
**##### Problem**

When you delete a job from the job list page, and you try to see the log output of that job executions, an empty log is shown (it only happens when deleting the job from that page)

**##### Solution**

It was added a way to know if the job was deleted, if it was, Rundeck is going to use the path stored for that job in the execution table.

fix [RSE-50](https://pagerduty.atlassian.net/browse/RSE-50)
fix https://github.com/rundeckpro/rundeckpro/issues/2354

With this fix the user can watch the logs for a specific job even when this was deleted. The previous behavior did not allow to watch those logs even though they were stored in the file system. 

Execution panel with job deleted: 
deletedJob

Activity screen for deleted job: 
activityScreen

Logs from deleted job:
logDeleted

This fix was also tested with a custom file system as minio 

Screen Shot 2022-09-26 at 16 47 20

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/LogFileStorageService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/LogFileStorageService.groovy
@@ -1184,7 +1184,7 @@ class LogFileStorageService
             boolean checkPartial = false
     )
     {
-        File file = getFileForExecutionFiletype(execution, filetype, false, false)
+        File file = getFileForExecutionFiletype(execution, filetype, execution.scheduledExecution == null, false)
         def key = logFileRetrievalKey(execution,filetype)
         def keyPartial = logFileRetrievalKey(execution, filetype, true)
 
@@ -1522,7 +1522,7 @@ class LogFileStorageService
         log.debug("requestLogFileLoad(${e.id},${filetype}): file state: ${state}: ${result}")
         switch (state) {
             case ExecutionFileState.AVAILABLE:
-                file = getFileForExecutionFiletype(e, filetype, false, false)
+                file = getFileForExecutionFiletype(e, filetype, e.scheduledExecution == null, false)
                 break
             case ExecutionFileState.AVAILABLE_PARTIAL:
                 file = getFileForExecutionFiletype(e, filetype, false, true)
--- a/rundeckapp/grails-app/services/rundeck/services/LogFileStorageService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/LogFileStorageService.groovy
@@ -1525,7 +1525,7 @@ class LogFileStorageService
                 file = getFileForExecutionFiletype(e, filetype, e.scheduledExecution == null, false)
                 break
             case ExecutionFileState.AVAILABLE_PARTIAL:
-                file = getFileForExecutionFiletype(e, filetype, false, true)
+                file = getFileForExecutionFiletype(e, filetype, e.scheduledExecution == null, true)
                 retryBackoff = Math.max(getBackoffForPartialFile(e, filetype), retryBackoff)
                 if (performLoad && result.remotePartialState == LogFileState.AVAILABLE_PARTIAL) {
                     //intiate another partial retrieval if delay interval has passed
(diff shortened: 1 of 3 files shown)


Oalbarran94:
@Mergifyio rebase

Oalbarran94:
retest this please

rundeckapp/grails-app/services/rundeck/services/LogFileStorageService.groovy
```diff
@@ -1184,7 +1184,7 @@ class LogFileStorageService
             boolean checkPartial = false
     )
     {
-        File file = getFileForExecutionFiletype(execution, filetype, false, false)
+        File file = getFileForExecutionFiletype(execution, filetype, execution.scheduledExecution == null, false)
```
L2JE, line 1187: Use a variable for all these `execution.scheduledExecution == null` in different calls, (use a good name that shows why you need it)

rundeckapp/src/test/groovy/rundeck/services/LogFileStorageServiceSpec.groovy
```diff
@@ -1016,11 +1031,10 @@ class LogFileStorageServiceSpec extends RundeckHibernateSpec implements ServiceU
 
             def exec = new Execution(
                     dateStarted: new Date(),
-                    dateCompleted: null,
+                    dateCompleted: new Date(),
                     user: 'user2',
                     project: 'test',
-                    serverNodeUUID: 'D0CA0A6D-3F85-4F53-A714-313EB57A4D1F',
-                    outputfilepath: '/tmp/file'
+                    serverNodeUUID: 'D0CA0A6D-3F85-4F53-A714-313EB57A4D1F'
```
L2JE, line 1037: Why did you remove `exec.outputFilePath` in this and the following tests _(L1100, L1164)_?

L2JE (changes requested):
If possible, add the issue scenario _"Execution log not found after job deletion"_ as a test case. (If not add a comment explaining why)

rundeckapp/grails-app/services/rundeck/services/LogFileStorageService.groovy
Oalbarran94: We are not supposed to get "Execution log not found after job deletion" since the goal of the card was to show the logs even after a job was deleted. This is possible because Rundeck is saving all executions pointing to where the logs are located.

rundeckapp/src/test/groovy/rundeck/services/LogFileStorageServiceSpec.groovy
Oalbarran94: As we changed the way of validating (execution.scheduleExecution == null) if the execution was going to get the file by getting it from "outputfilepath" or by forming the filepath (as showed below) , that part of the test needed to be remove in order to follow its normal steps. Since within those test we do not use a valid execution.scheduleExecution, the tests itself were getting another path which they were not created for.

image

rundeckapp/grails-app/services/rundeck/services/LogFileStorageService.groovy
L2JE: What I mean by covering that use case is to make a test that verifies if the logs are still accessible after deletion trying to use the interfaces that the application is currently using to show them on the screen.

alexander-variacode: Osman try to save this operation `execution.scheduledExecution == null` in a variable because It just like that it didn’t explain why is there.

ronaveva: this scenario would still be possible if someone just deletes the log file manually, but i think that is covered by the tests 

https://github.com/rundeck/rundeck/blob/fff68b5a30fd3d06b611e011e8e19ece38b14c1a/rundeckapp/src/test/groovy/rundeck/services/LogFileStorageServiceSpec.groovy#L802

L2JE: Please confirm that with the current tests, in case the logs of deleted jobs don't appear again as before, the tests will complain about it.

L2JE (approved):
LGTM

alexander-variacode (approved):
LGTM

ronaveva (approved):
LGTM

PR #7962, opened 2022-10-17
RSE-184: Fix: Log error when SCM tries to reconnect

Jesus-Osuna-M:
Bugfix on Log 

Problem: When the SCM fails rundeck by default will retry to reconnect 5 times, On the screenshot below we can see that the retry is being made 6 times and the log shows a weird message saying retry 6/5. 

Solution: Change the evaluation condition to end the cycle when the las retry is done. 

Before fix:
imagen

After fix:
imagen


ronaveva (changes requested):
Please add tests for the method

Jesus-Osuna-M:
> Please add tests for the method
It wasn't possible to add a unit test for the method **startScmLoader** since is being executed inside a closure, inside that context the mocked values don't receive the expected values.  This method could be tested on the QA framework

ronaveva (approved):
LGTM

PR #7967, opened 2022-10-18
RSE-118: Fix: select-all-function

avelasquezr:
# BugFix: RSE-118

**Describe the bug**

When the “Select-all Values by default” is enabled within a Job option: when you change an upstream option, the cascading options don't render as selected, it toggles between select-all and non select all.

Select-all Values by defaul is related to:
1. Click to Edit this Job
2. Go to Workflow -> when adding an option you have this option when delimiting a list by commas or another character:

image

An additional validation is added when the values selected by default come and these are not considered in a subsequent validation.

**Additional context**

Before change:
Before_Change

After change:
After_Change


DarwisNarvaezDev (approved):
LGTM

PR #7969, opened 2022-10-19
RUN-1295: Allow form data in webhooks

sjrd218:
This PR fixes the bug where Grails eats the inputStream if x-www-form-urlencoded is the content type. The form data is passed to the webhook service in a new formData field in the WebhookData interface.

### DIFF
--- a/grails-webhooks/src/test/groovy/webhooks/WebhookControllerSpec.groovy
+++ b/grails-webhooks/src/test/groovy/webhooks/WebhookControllerSpec.groovy
@@ -61,6 +61,40 @@ class WebhookControllerSpec extends Specification implements ControllerUnitTest<
         "1234#test" | '{"msg":"ok"}' | new DefaultJsonWebhookResponder([msg:"ok"])
     }
 
+    def "test post content types"() {
+        given:
+        controller.rundeckAuthContextProvider = Mock(AuthContextProvider)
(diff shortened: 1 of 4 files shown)


grails-webhooks/src/test/groovy/webhooks/WebhookControllerSpec.groovy
```diff
@@ -61,6 +61,40 @@ class WebhookControllerSpec extends Specification implements ControllerUnitTest<
...
+        request.contentType = contentType
+        controller.post()
+
+        then:
+        1 * controller.webhookService.getWebhookByToken(_) >> { new Webhook(name:"test",authToken: "1234")}
+        1 * controller.rundeckAuthContextProvider.getAuthContextForSubjectAndProject(_, _) >> { new SubjectAuthContext(null, null) }
+        1 * controller.rundeckAuthContextEvaluator.authorizeProjectResourceAny(_,_,_,_) >> { return true }
+        1 * controller.webhookService.processWebhook(_,_,_,_,_) >> { args ->
+            if(isFormData) {
+                args[2].formData["key1"] == "val1"
```
gschueler, line 87: this won't fail as a test unless you use `assert`, otherwise since it is within the closure it isn't recognized as a comparison check

sjrd218: Thanks, fixed!

PR #7970, opened 2022-10-20
Add aws-mysql-jdbc to gradle build

jolexa:
This is an enhancement to better support AWS Aurora mysql database. The mariadb driver works already, but not [officially supported in newer versions](https://aws.amazon.com/blogs/database/using-the-mariadb-jdbc-driver-with-amazon-aurora-with-mysql-compatibility/) and advanced Aurora features like failover do not work for Rundeck with mariadb driver.

https://github.com/awslabs/aws-mysql-jdbc

AWS MYSQL JDBC driver should be in the java classpath.

This has been tested by manually downloading the jar and placing it in the classpath. That is an acceptable alternative but requires operator involvement.

**Other Notes**
aws-mysql-jdbc is licensed under GPLv2 with a FOSS exception


fdevans:
Are we legally allowed to distribute this?

PR #7971, opened 2022-10-20
RSE-159 Fix: Activity Tab no progress percentage displayed in a Progress-bar

Jesus-Osuna-M:
**Problem:**
On the activity tab, the percentage on the progress bar moved to the right when the percentage was more than 200% making it impossible to see the percentage.

imagen

**Solution:**
Add to the CSS class **.progress-bar** a max-width property to limit the growth of the progress-bar to 100% making the percentage shown on the center whatever the percentage is.

**Before Fix:**
imagen

The progress bar grew with the percentage.
imagen

**After Fix:**
The percentage is always shown in the center.
imagen

The progress bar only grows until 100%.
imagen


DarwisNarvaezDev (approved):
LGTM

PR #7979, opened 2022-10-25
Convert AuthToken criteria queries in UserController to use data API

sjrd218:
Convert AuthToken criteria queries in UserController to use data API


imad6639 (approved):

rundeckapp/src/main/groovy/org/rundeck/app/data/RdPageable.groovy
```diff
@@ -0,0 +1,16 @@
+package org.rundeck.app.data
+
+import org.rundeck.app.data.model.v1.page.Pageable
+
+class RdPageable implements Pageable{
```
gschueler, line 5: by default we should use `@CompileStatic` on all new groovy classes

rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormTokenDataProvider.groovy
```diff
@@ -107,6 +109,16 @@ class GormTokenDataProvider implements TokenDataProvider {
         tokens
     }
 
+    @Override
+    List<AuthenticationToken> findAllByUser(String userId) {
+        List<AuthenticationToken> tokens = []
+        List<AuthToken> authTokens = AuthToken.findAllByUser(User.get(userId.toLong()))
```
gschueler, line 115: should validate User.get return value

rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormTokenDataProvider.groovy
```diff
@@ -161,17 +173,137 @@ class GormTokenDataProvider implements TokenDataProvider {
...
+        }
+    }
+
+    @Override
+    @GrailsCompileStatic(TypeCheckingMode.SKIP)
+    Integer countTokensByType(AuthenticationToken.AuthTokenType type) {
+        return AuthToken.createCriteria().count {
+            or {
+                eq("type", type)
+                isNull("type")   //type could be null if token was created before we recorded token types
```
gschueler, line 252: I think we should remove support for "legacy" user tokens at some point.

rundeckapp/src/main/groovy/org/rundeck/app/data/RdPageable.groovy
sjrd218: done

rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormTokenDataProvider.groovy
sjrd218: Agreed, unless someone is upgrading from a very legacy version we shouldn't need to check this. Not sure what the priority would be on removing the extra check.

gschueler: i can backlog it

rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormTokenDataProvider.groovy
```diff
@@ -161,17 +175,137 @@ class GormTokenDataProvider implements TokenDataProvider {
...
     }
 
     @Override
-    AuthenticationToken tokenLookupWithType(final String token, AuthenticationToken.AuthTokenType type){
-        def authToken = AuthToken.tokenLookup(token, type);
-        return authToken ?: null
+    @GrailsCompileStatic(TypeCheckingMode.SKIP)
+    Integer countTokensByUser(String userId) {
+        return AuthToken.createCriteria().count {
+            eq("user", User.get(userId.toLong()))
```
gschueler, line 228:
```suggestion
        User user = User.get(userId.toLong())
        if(!user) throw new DataAccessException("Couldn't find user: ${userId}")
        return AuthToken.createCriteria().count {
            eq("user", user)
```

gschueler (approved):

PR #7983, opened 2022-10-27
RUN-1314: Update ubuntu packages on docker build

ahormazabal:


ehe-pd (approved):
Tested. Packages are upgraded to their latest version.

PR #7988, opened 2022-11-02
AuthToken fixes/updates

imad6639:
Bugfixes for matching auth token functionality.  Remove substring on encoded hash, add clear token for created token responses.  Add gormTokenProvider needed for plugins

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/UserService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/UserService.groovy
@@ -62,7 +62,6 @@ class UserService {
     }
 
     String getOwnerName(Long userId) {
-        User.createCriteria().get {}
         User.createCriteria().get {
             eq("id", userId)
             projections {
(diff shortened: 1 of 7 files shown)


rundeckapp/grails-app/services/rundeck/services/UserService.groovy
```diff
@@ -61,6 +61,16 @@ class UserService {
         return user
     }
 
+    String getOwnerName(Long userId) {
+        User.createCriteria().get {}
```
sjrd218, line 65: extra?

sjrd218 (changes requested):
I think an extra change that needs to be made is on the _token.gsp 

lines 40-45 should now be: 
```
<td width="10%" title="Creator: ${token.creator}">
            ${token.ownerName}
            <g:if test="${!token.creator.equalsIgnoreCase(token.ownerName)}">
                (${token.creator})
            </g:if>
        </td>
```

sjrd218 (approved):
lgtm

PR #7997, opened 2022-11-08
RSE-186: syncOauthUser properties don't have hot reload

L2JE:
Ticket: https://pagerduty.atlassian.net/browse/RSE-186

#### Problem

The following settings work through Configuration Management but they require a restart:

`rundeck.security.syncOauthUser=true`

#### Solution

Added the property as part of the solution mentioned in enterprise PR


alexander-variacode (approved):
LGTM

PR #8000, opened 2022-11-09
Rcloud 662

sjrd218:
This PR removes the use of the DataManager in the Rundeck services that are managed by Spring. Spring wiring makes the data providers much easier to use and makes advanced Spring bean management much more convenient.

I've added a DataProvider marker interface to the SPI that should be used by all data providers in the future so that we can easily identify them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,4 @@ rundeckapp/out
 rundeckapp/rundeck-runtime
 /rundeckapp/grails-app/assets/node_modules
 /rundeckapp/grails-app/assets/**/node_modules
+/rundeckapp/grails-app/conf/application-development.yml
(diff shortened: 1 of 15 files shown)


.gitignore
```diff
@@ -58,3 +58,4 @@ rundeckapp/out
 rundeckapp/rundeck-runtime
 /rundeckapp/grails-app/assets/node_modules
 /rundeckapp/grails-app/assets/**/node_modules
+/rundeckapp/grails-app/conf/application-development.yml
```
gschueler, line 61: intended change/

gschueler (approved):

.gitignore
sjrd218: Yes, sometimes I use an application-development.yml to override defaults. Since these settings should be developer specific the file shouldn't be checked into source control.

Issue #6711, opened 2021-01-07
Notifications payload format are not imported properly by the API and the UI


aylham:
Issue type: Bug report

**My Rundeck detail**

* Rundeck version: 3.3.7.20201208
* install type: rpm
* OS Name/version: centos7
* DB Type/version: postgres10

**Expected Behavior**

When importing a job definition in YAML the field `notification/format` isn't  properly read and result in a `null` in the `payload format` field after the import.

```
  notification:
    onfailure:
      format: json
      urls: http://rundeck-server:4440/api/36/webhook/TOKEN#webhook
```

**How to Reproduce**

Import the following job :

```
- defaultTab: nodes
  description: Job
  executionEnabled: true
  id: 1fadb795-b378-4e3d-8b46-4b344eadc111
  loglevel: INFO
  name: MyJob
  nodeFilterEditable: false
  notification:
    onfailure:
      format: json
      urls: http://rundeck-server:4440/api/36/webhook/TOKEN#webhook
    onsuccess:
      format: json
      urls: http://rundeck-server:4440/api/36/webhook/TOKEN#webhook
  notifyAvgDurationThreshold: null
  options:
  - description: Numbers of days alarms are conserved
    name: alarms
    regex: \d+
    value: '30'
  - description: ' Numbers of days pb are conserved'
    name: pb
    regex: \d+
    value: '30'
  plugins:
    ExecutionLifecycle: null
  scheduleEnabled: true
  sequence:
    commands:
    - args: --alarms ${option.alarms} --pbehaviors ${option.pbehaviors} -f
      scriptfile: /var/lib/rundeck/libext/myscript
    keepgoing: false
    strategy: node-first
  uuid: 1fadb795-b378-4e3d-8b46-4b344eadc111
```

The field `payload format` will not be set as expected by the definition.

samyscoub:
:+1:

MegaDrive68k:
Confirmed and happening on 3.3.8 too.

And saving the job again,  produces a "null" payload format on the notification.

Thanks for your feedback @samyscoub!

aylham:
Hello,
Does someone has any news on this problem?

spaced0gg:
Seeing this is 3.3.8 as well.

**My Rundeck detail**

    Rundeck version: 3.3.8
    install type: rpm
    OS Name/version: RHEL6
    DB Type/version: mysql

[...]

PR #8005, opened 2022-11-10
RSE-133: Fix: Notifications payload format are not imported properly by the API and the UI

avelasquezr:
## BugFix: Notifications payload format are not imported properly by the API and the UI

Fix to modify the LinkedHashMap mapping and insert a single map to the array. Related to https://github.com/rundeck/rundeck/issues/6711

#### Considering the following YAML:
``` yaml
# YAML
...
notification:
  onfailure:
    format: json
    method: post
    urls: http://rundeck-server:4440/api/36/webhook/TOKEN#webhook
  onsuccess:
    format: json
    method: post
    urls: http://rundeck-server:4440/api/36/webhook/TOKEN#webhook
```
### The Problem:
When the yaml is identified as coming in Map format (without "-", this symbol reference a list type), each map is inserted into the list:
code reference:
```
iteration {
   map.put(key,value)
   array.add(map)
}
```
### The Fix:
The mapping of the LinkedHashMap is modified to insert a single map to the array.
code reference:
```
iteration {
   map.put(key,value)
}
array.add(map)
```
Now list types and map types are accepted
``` yaml
# YAML
...
notification:
  onfailure:
    - format: json
       method: post
       urls: http://rundeck-server:4440/api/36/webhook/TOKEN#webhook
    - format: json
       method: post
       urls: http://rundeck-server:4440/api/36/webhook/TOKEN#webhook
  onsuccess:
    format: json
    method: post
    urls: http://rundeck-server:4440/api/36/webhook/TOKEN#webhook
```

### DIFF
--- a/rundeckapp/src/test/groovy/org/rundeck/app/components/JobYAMLFormatSpec.groovy
+++ b/rundeckapp/src/test/groovy/org/rundeck/app/components/JobYAMLFormatSpec.groovy
@@ -218,7 +218,7 @@ class JobYAMLFormatSpec extends Specification {
     }
 
     @Unroll
-    def "should return a notification map with email and webhook notifs"() {
+    def "should return a notification list of map with email and webhook notifs"() {
         given:
         def input = "" +
                 "- defaultTab: nodes\n" +
@@ -234,15 +234,41 @@ class JobYAMLFormatSpec extends Specification {
                 "        attachLogInFile: true\n" +
                 "        recipients: leojesus.juarez@gmail.com\n" +
                 "        subject: RD-SUCCESS\n" +
-                "    - email:\n" +
-                "        attachLog: true\n" +
-                "        attachLogInline: true\n" +
-                "        recipients: 2@gmail.com\n" +
-                "        subject: RD-SUCCESS\n" +
                 "    - format: xml\n" +
                 "      httpMethod: get\n" +
                 "      urls: http://localhost:4440/project\n" +
-                "    - format: json\n" +
+                "  notifyAvgDurationThreshold: null\n" +
+                "  plugins:\n" +
+                "    ExecutionLifecycle: null\n" +
+                "  scheduleEnabled: true\n" +
+                "  schedules: []\n" +
+                "  sequence:\n" +
+                "    commands:\n" +
+                "    - exec: asd\n" +
+                "    keepgoing: false\n" +
+                "    strategy: node-first"
+        def sut = new JobYAMLFormat()
+        when:
+        def result = sut.decode(new StringReader(input))
+        then:
+        result[0].notification['onsuccess'].size() == 2
+        result[0].notification['onsuccess'].findAll{ it['email'] != null }.size() == 1
+        result[0].notification['onsuccess'].findAll{ it['urls'] != null }.size() == 1
+    }
+
+    @Unroll
+    def "should return a notification list with one webhook notification"() {
+        given:
+        def input = "" +
+                "- defaultTab: nodes\n" +
+                "  description: ''\n" +
+                "  executionEnabled: true\n" +
+                "  loglevel: INFO\n" +
+                "  name: a\n" +
+                "  nodeFilterEditable: false\n" +
+                "  notification:\n" +
+                "    onsuccess:\n" +
+                "      format: json\n" +
                 "      httpMethod: post\n" +
                 "      urls: http://localhost:4440/project\n" +
                 "  notifyAvgDurationThreshold: null\n" +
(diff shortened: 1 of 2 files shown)


rundeckapp/src/test/groovy/org/rundeck/app/components/JobYAMLFormatSpec.groovy
```diff
@@ -264,6 +264,93 @@ class JobYAMLFormatSpec extends Specification {
         result[0].notification['onsuccess'].findAll{ it['urls'] != null }.size() == 2
     }
 
+    @Unroll
+    def "should return a notification map webhook url notifs"() {
```
L2JE, line 268: The decode method always returns list of notifications for each trigger. Example of better name: `"should return a notif list with one webhook notif"`

L2JE (changes requested):
It seems to not be working on the case of having two different notifications in the same trigger without the list style:
```
  notification:
    onstart:
      format: json
      httpMethod: post
      urls: http://localhost:4440/project
      email:
        attachLog: 'true'
        attachLogInFile: true
        recipients: tom@example.com
        subject: JOB-STARTED
```

Link to official docs example:
https://docs.rundeck.com/docs/manual/document-format-reference/job-yaml-v12.html#another-valid-format

rundeckapp/src/test/groovy/org/rundeck/app/components/JobYAMLFormatSpec.groovy
avelasquezr: Test - renamed

rundeckapp/src/test/groovy/org/rundeck/app/components/JobYAMLFormatSpec.groovy
```diff
@@ -264,6 +264,166 @@ class JobYAMLFormatSpec extends Specification {
...
+                "          attachLog: true\n" +
+                "          attachLogInline: true\n" +
+                "          recipients: 2@gmail.com\n" +
+                "          subject: RD-SUCCESS\n" +
+                "      - format: xml\n" +
+                "        httpMethod: get\n" +
+                "        urls: http://localhost:4440/project\n" +
+                "      - format: json\n" +
+                "        httpMethod: post\n" +
+                "        urls: http://localhost:4440/project\n" +
```
L2JE, line 327: one or two elements should be enough to represent the list for this test case

rundeckapp/src/test/groovy/org/rundeck/app/components/JobYAMLFormatSpec.groovy
```diff
@@ -264,6 +264,166 @@ class JobYAMLFormatSpec extends Specification {
...
+        def sut = new JobYAMLFormat()
+        when:
+        def result = sut.decode(new StringReader(input))
+        then:
+        result[0].notification['onsuccess'].size() == 1
+        result[0].notification['onsuccess'].findAll{ it['urls'] != null }.size() == 1
+    }
+
+    @Unroll
+    def "should return a notification list of map and single map with email and webhook notifs"() {
```
L2JE, line 301: This name sounds confusing.

This test seems to be trying to check that either a list of notifications or a notification map will result on a list of notifications for each trigger always. So I think a more suitable name (thinking on the testcase and not on the exact code) for this test would be something like:
`should return list of notifications for each trigger when some triggers have notifs in yaml list and some not in a list`

rundeckapp/src/test/groovy/org/rundeck/app/components/JobYAMLFormatSpec.groovy
```diff
@@ -264,6 +264,166 @@ class JobYAMLFormatSpec extends Specification {
...
+        result[0].notification['onsuccess'].size() == 4
+        result[0].notification['onsuccess'].findAll{ it['email'] != null }.size() == 2
+        result[0].notification['onsuccess'].findAll{ it['urls'] != null }.size() == 2
+
+        result[0].notification['onfailure'].size() == 1
+        result[0].notification['onfailure'].findAll{ it['urls'] != null }.size() == 1
+    }
+
+    @Unroll
+    def "should return a notification list of map and single map with email, plugin and webhook notify"() {
```
L2JE, line 355: Same as previous one, confusing name, thinking from the use case point of view, a better name for this case could be:
`should return list of notifications for each trigger when all triggers have notifs not in yaml lists` or 
`should return list of notifications for each trigger when all triggers have notifs in yaml maps`

L2JE (changes requested):

L2JE (approved):
LGTM

PR #8007, opened 2022-11-11
Fixes support widget button

carlosrfranco:


mishingo (approved):

rundeckapp/grails-app/views/layouts/base.gsp
```diff
@@ -61,9 +61,9 @@
     <asset:javascript src="bootstrap-all.js"/>
     <g:set var="includePrototypeJs" value="${true}" scope="page"/>
 
-    <g:ifPageProperty name="meta.skipPrototypeJs">
-        <g:set var="includePrototypeJs" value="${false}" scope="page"/>
-    </g:ifPageProperty>
+%{--    <g:ifPageProperty name="meta.skipPrototypeJs">--}%
+    <g:set var="includePrototypeJs" value="${false}" scope="page"/>
+%{--    </g:ifPageProperty>--}%
```
gschueler, line 66: it seems you are disabling the flag that enables some pages to avoid including prototype.  is there a reason for this? we want to eventually get to a point where we remove prototype completely, so disabling this seems like a step backwards @mishingo @carlosrfranco

carlosrfranco: @gschueler actually this change disables prototype for all pages

gschueler: I see, if that's the goal, you should just remove that code, and the code below which includes prototype

PR #8010, opened 2022-11-16
RUN-1200: Enabled feature  query API endpoint

ehe-pd:
This is an enhancement to the existing enabled feature query API which is in the incubator.

Remove the incubation part from the feature name query.

The `feature` API returns a JSON object:

1. If the query is against a specific feature `/feature/{featureName}`, below will return:
```
{
    "name": "feature name"
    "enabled": true/false
} 
```
2. If the query has no featureName parameter provided `/feature`, below list of JSON object will return:
```
[{
    "enabled":false,
    "name":"pluginSecurity"
 },{
   "enabled":true,
   "name":"projectKeyStorage"
},...,
{
    "enabled":true,
    "name":"projectManagerServiceBootstrapWarmupCache"
}]
```


rundeckapp/grails-app/controllers/rundeck/controllers/ApiController.groovy
```diff
@@ -206,59 +209,62 @@ class ApiController extends ControllerBase{
...
+     *      the request will be `/feature/runner`. The result of this query is an JSON object { "enabled": true/false }
+     *
+     * If `featureName` parameter is absent, the query will return all system features' status as an JSON object where the key is the feature's name and value is { "enabled": true/false }
      */
-    def featureToggle={
+    @Get(
+        uri= "/feature/{featureName}",
+        produces = MediaType.APPLICATION_JSON
+    )
+    def featureQuery(@PathVariable(name = "featureName", required = false) String featureName) {
```
gschueler, line 253: Should remove `required=false`, because it produces invalid OpenAPI spec document.

```suggestion
    def featureQuery(@PathVariable(name = "featureName") String featureName) {
```

rundeckapp/grails-app/controllers/rundeck/controllers/ApiController.groovy
```diff
@@ -206,59 +209,62 @@ class ApiController extends ControllerBase{
...
+                )
+            )
+        ]
+    )
+    @ApiResponse(
+        responseCode = "200",
+        description = "Map of all system features' status if no specific one specified",
+        content = @Content(
+            mediaType = "application/json",
+            schema = @Schema(implementation = Map)
```
gschueler, line 236: There should be two `@ApiResponse`s, and each should point to a specific datatype for the `@Schema(implementation=DataType)`, instead of just Map.
One for the List result, and one for a FeatureEnabledResult type.

The Controller should also respond with the actual data type object.

gschueler (changes requested):
A few changes should be made to the OpenAPI spec annotations:

rundeckapp/grails-app/controllers/rundeck/controllers/ApiController.groovy
```diff
@@ -206,62 +211,115 @@ class ApiController extends ControllerBase{
...
+                    allowEmptyValue = true,
+                    required = false,
+                    schema = @Schema(
+                            type='string'
+                    )
+            )
+        ],
+        responses = [
+            @ApiResponse(
+                responseCode = "200",
```
ehe-pd, line 249: OpenAPI annotation cannot have two ApiResponse with the same `responseCode`, the last one will be used to generate the document. So I split the API into two: single feature query and list all features.

gschueler: that's fine to do it that way.

btw, I think you can do it like this within one response code `@Schema(oneOf=[Type1,Type2])`

rundeckapp/grails-app/controllers/rundeck/controllers/ApiController.groovy
```diff
@@ -211,46 +213,91 @@ class ApiController extends ControllerBase{
...
+     */
+    @Get(
+        uri= "/feature/{featureName}",
+        produces = MediaType.APPLICATION_JSON
+    )
     @Operation(
         method = "GET",
         summary = "Get Rundeck System Feature Status",
-        description = "Return feature's on/off status",
+        description = "The result of this query is a JSON object { name: \"featureName\", \"enabled\": true/false }",
```
gschueler, line 234: you don't need to outlay the json format here, it will be represented in the response schema generation

rundeckapp/grails-app/controllers/rundeck/controllers/ApiController.groovy
```diff
@@ -211,46 +213,91 @@ class ApiController extends ControllerBase{
...
-        uri= "/feature/{featureName}",
-        produces = MediaType.APPLICATION_JSON
+            uri= "/feature",
+            produces = MediaType.APPLICATION_JSON
     )
-    def featureQuery(@PathVariable(name = "featureName") String featureName) {
+    @Operation(
+            method = "GET",
+            summary = "List all System Feature on/off Status",
+            description = "The query will return all system features' status as a list of JSON objects [{ name: \"featureName\", \"enabled\": true/false }, ...]",
```
gschueler, line 286: also, can remove the json schema here

gschueler, line 286:
```suggestion
            description = "The query will return all system features' status",
```

rundeckapp/grails-app/controllers/rundeck/controllers/ApiController.groovy
gschueler, line 234:
```suggestion
        description = "Return whether a feature is enabled or disabled.",
```

rundeckapp/grails-app/controllers/rundeck/controllers/ApiController.groovy
```diff
@@ -206,62 +211,115 @@ class ApiController extends ControllerBase{
...
+    )
+    @CompileStatic
+    def featureQuery(@PathVariable(name = "featureName") String featureName) {
+        if (!apiService.requireApi(request, response)) {
+            return
+        }
+
+        if (!apiService.requireVersion(request, response, ApiVersions.V42)) {
+            return
+        }
```
gschueler, line 266: can consolidate to one call
```suggestion
        if (!apiService.requireApi(request, response, ApiVersions.V42)) {
            return
        }
```

gschueler (approved):

PR #8022, opened 2022-11-23
RUN-1352 fix invalid property name

carlosrfranco:
An invalid property name was included in the code causing an error when changing and saving project settings

Change to `project.name` to access project name value

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/ProjectManagerService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ProjectManagerService.groovy
@@ -639,7 +639,7 @@ class ProjectManagerService implements ProjectManager, ApplicationContextAware, 
     {
         if(properties['project.description'] != null ) {
             def description = properties['project.description']
-            RdProject dbproj = projectDataProvider.findByName(projectName)
+            RdProject dbproj = projectDataProvider.findByName(project.name)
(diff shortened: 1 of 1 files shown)


ltamaster:
we need to add a test

rundeckapp/src/test/groovy/rundeck/services/ProjectManagerServiceSpec.groovy
```diff
@@ -721,6 +725,38 @@ class ProjectManagerServiceSpec extends Specification implements ServiceUnitTest
...
+            }
+        }
+        service.rundeckNodeService=Mock(NodeService)
+
+        when:
+        service.mergeProjectProperties(rundeckProject, properties, null)
+
+        then:
+        rundeckProject.projectConfig.name == "project"
+        rundeckProject.nodesFactory
```
ltamaster, line 757: do we expect that `rundeckProject.nodesFactory` will be set? maybe we can that more explicit (like `rundeckProject.nodesFactory!=null`)

carlosrfranco: Actually it expects to be equals to `rundeckNodeService`. Just updated it

ltamaster (approved):
LGTM

PR #8025, opened 2022-11-28
Fix dynamic properties when using plugin groups

chrismcg14:
When loading the dynamic properties, the plugin service will now populate the projectAndFramework values from the values set in the plugin group.

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/PluginService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/PluginService.groovy
@@ -25,6 +25,7 @@ import com.dtolabs.rundeck.core.plugins.configuration.AcceptsServices
 import com.dtolabs.rundeck.core.plugins.configuration.Description
 import com.dtolabs.rundeck.core.plugins.configuration.DynamicProperties
 import com.dtolabs.rundeck.core.plugins.configuration.PluginAdapterUtility
+import com.dtolabs.rundeck.core.plugins.configuration.Property
 import com.dtolabs.rundeck.core.plugins.configuration.PropertyResolver
 import com.dtolabs.rundeck.core.plugins.configuration.PropertyResolverFactory
 import com.dtolabs.rundeck.core.plugins.PluggableProviderService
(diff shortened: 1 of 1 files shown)


rundeckapp/grails-app/services/rundeck/services/PluginService.groovy
```diff
@@ -425,6 +376,26 @@ class PluginService implements ResourceFormats, PluginConfigureService {
...
+            PluggableProviderService<T> service,
+            PropertyResolverFactory.Factory resolverFactory,
+            PropertyScope defaultScope,
+            Services servicesProvider
+    )
+    {
+
+        def result = rundeckPluginRegistry?.configurePluginByName(name, service, resolverFactory, defaultScope)
+
+        if (result.instance != null) {
```
gschueler, line 390: result might be null
```suggestion
        if (result?.instance != null) {
```

rundeckapp/grails-app/services/rundeck/services/PluginService.groovy
```diff
@@ -187,74 +193,19 @@ class PluginService implements ResourceFormats, PluginConfigureService {
...
-        if(!pluginDescriptor){
-            return null
-        }
+        PropertyRetriever retriever = PropertyResolverFactory.instanceRetriever([:])
+        PropertyRetriever projectRetriever = PropertyResolverFactory.instanceRetriever(rundeckFramework.getFrameworkProjectMgr().getFrameworkProject(project).getProperties())
+        PropertyResolverFactory.Factory resolverFactory = PropertyResolverFactory.pluginPrefixedScoped(retriever,projectRetriever,rundeckFramework.getPropertyRetriever())
+        def plugin = configurePluginWithoutValidation(type, createPluggableService(getPluginTypeByService(serviceName)), resolverFactory, PropertyScope.Instance, null)
 
-        def instance = pluginDescriptor.instance
+        def instance=plugin.instance
```
gschueler, line 202: `plugin` might be null here, if the configure fails

gschueler (approved):

PR #8031, opened 2022-12-05
RSE-259 Fix:Ruleset Plugin fails to evaluate conditions if negative number are used

Jesus-Osuna-M:
Problem: Rundeck didn't recognize negative numbers on the ruleset and converted any negative number to  0 causing failings when evaluating conditions like **if:option.1>=0** .

Solution: Fix the method in charge of parse values from the ruleset to float numbers to accept negative numbers.


jtobard (approved):
I see the selenium test on rundeck-tester, we're good with that

PR #8033, opened 2022-12-06
RUN-1298: generate copyright year at build time

gschueler:
Defines copyright string with current year at build time


chrismcg14 (approved):
LGTM

PR #8035, opened 2022-12-07
RCLOUD-666: Convert access of Storage entity to be pass through a data provider

imad6639:
Convert access of Storage entity to be pass through a data provider
Enhancement to use data spi pattern
Created an interface and provider for Storage access


gschueler:
@mergifyio rebase

rundeck-data-model/src/main/java/org/rundeck/app/data/model/v1/storage/RundeckStorage.java
```diff
@@ -0,0 +1,71 @@
...
+import org.rundeck.app.data.utils.Utils;
+
+import java.io.Serializable;
+import java.util.*;
+
+public interface RundeckStorage {
+    Serializable getId();
+    String getNamespace();
+    String getDir();
+    String getName();
```
gschueler, line 30: i think the interface should return a single Path object, instead of dir and name, that is the more direct translation of the datatype

rundeck-data-model/src/main/java/org/rundeck/app/data/model/v1/storage/RundeckStorage.java
```diff
@@ -0,0 +1,71 @@
...
+    String getJsonData();
+    String getPathSha();
```
gschueler, line 32: using a SHA of the path as a unique key in the db, is an implementation detail of the Gorm storage, it doesn't really make sense to expose it in the SPI. Also, using json to store the Map of metadata is also an implementation detail of the gorm entity, I think that also should not be translated to the interface, instead a Map should be used, and the SHA and JSON stuff would just belong in the gorm provider

gschueler (approved):

PR #8036, opened 2022-12-07
RCLOUD-634 Rundeck Job Data API

sjrd218:
Introduce Rundeck Job Data API

This PR attempts to leave the current ScheduleExecution domain/controller/service layer as untouched as possible while introducing the Job API. Ideally this will reduce potential disruptions to the current feature set to a minimum while setting up the codebase for future migration to the JobData interface.

A new Grails plugin has been added to facilitate better sharing of data API implementations. This allows the implementation of JobData to be shared by elements in both the core and enterprise product.

I've run into some places where I had to introduce bridge code because the core project does not have a dependency on the data model. We should potentially revisit the decision to keep the libraries separate because it precludes using the data API in structures defined in the core library.

The new job data API deviates from the "canonical job map". This is to introduce additional uniformity and consistency to some structures(job steps, job components) and to increase the organization of other parts of job data (node, logging, scheduling). I believe this restructuring allows users and developers to more easily reason about the pieces of job data without having to go look at the code to understand it.

New API endpoints
- Job crud -
GET, POST, DELETE
http://localhost:4440/api/40/incubator/job

- Job Query -
GET http://localhost:4440/api/40/incubator/jobs?max=-1&offset=0&projFilter=fresh


rundeckapp/grails-app/domain/rundeck/CommandExec.groovy
```diff
@@ -203,10 +227,13 @@ public class CommandExec extends WorkflowStep  {
...
             ce.pluginConfig = data.plugins
         }
-        return ce
     }
 
     public boolean isNodeStep(){
         return true;
     }
+
+    public Boolean getNodeStep(){
```
jbrookspd, line 236: Is this used anywhere? Looks like it always returns true?

rundeckapp/grails-app/services/rundeck/services/JobEventsService.groovy
```diff
@@ -69,4 +74,42 @@ class JobEventsService {
             listener.jobChangeEvent(e, serializer)
         }
     }
+
+    @Subscriber('log.job.change.event')
+    def logJobChange(LogJobChangeEvent e) {
+        println "got a log job change event"
```
jbrookspd, line 80: Is this println neeeded?

rundeckapp/grails-app/services/rundeck/services/RdJobService.groovy
```diff
@@ -0,0 +1,54 @@
...
+    JobData getJobByIdOrUuid(Serializable id) {
+        JobData found = null
+        if (id instanceof Long) {
+            return getJobById(id)
+        } else if (id instanceof String) {
+            //attempt to parse as long id
+            try {
+                def idlong = Long.parseLong(id)
+                found = getJobById(idlong)
+            } catch (NumberFormatException e) {
```
jbrookspd, line 23: Should we at least log.info these in case an exception is thrown?

core/src/main/java/com/dtolabs/rundeck/core/plugins/PluginDescriptionService.java
```diff
@@ -0,0 +1,13 @@
+package com.dtolabs.rundeck.core.plugins;
+
+import com.dtolabs.rundeck.core.plugins.configuration.Description;
+import com.dtolabs.rundeck.core.plugins.configuration.PropertyScope;
+
+import java.util.Map;
```
imad6639, line 6: remove

rundeck-data-model/src/main/java/org/rundeck/app/data/model/v1/job/JobDataSummary.java
```diff
@@ -0,0 +1,15 @@
+package org.rundeck.app.data.model.v1.job;
+
+import java.io.Serializable;
```
imad6639, line 3: remove

grails-rundeck-data-shared/grails-app/views/error.gsp
```diff
@@ -0,0 +1,39 @@
+<!doctype html>
+<html>
```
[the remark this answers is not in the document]
sjrd218: looks like I can safely remove them.

rundeckapp/grails-app/domain/rundeck/CommandExec.groovy
sjrd218: This method conforms to groovy's expected format for getters and setters so that an instance of this class can use the form `command.nodeStep`. Unfortunately the form `isBoolean()` for boolean getters isn't sufficient.

rundeckapp/grails-app/services/rundeck/services/RdJobService.groovy
sjrd218: In this case it's fine to ignore this error. In a future PR I have marked this method of returning the job as deprecated because we should only be using the uuid in the future and not the internal.

jbrookspd (approved):

PR #8047, opened 2022-12-15
RSE-244 Fix: edit/view jobs when SCM Import Plugin is forbidden for user

DarwisNarvaezDev:
# Fix: Edit/View Jobs when SCM Import Plugin is Forbidden for User
Previously if the user don't have permissions to the key configured in the SCM Import Plugin, an internal server error could be seen and the basic job actions couldn't be performed.

## The Problem
1. Since th SCM service use the SESSION user to perform SCM operations, an error occurs when the system search for the keys in the plugin's config

## The Solution
1. In the controller of the menu (who renders job page and its functions) ask if the current user has permission to the key, if not disable the plugin and populate a message to the frontend.
2. Log at ERROR level the incidence.
3. Flash the unauthorized user with a message, letting the user now what happened with the SCM.

## Exhibits
a) If an unauthorized user with the SCM import enabled sees the jobs page, an error message will be displayed
Screenshot from 2022-12-15 15-50-18

b) But if we go ahead and try to edit the job or see its details, the system will allow us since the SCM is disabled
Screenshot from 2022-12-15 15-50-42

c) A line of log could be seen in the general log
Screenshot from 2022-12-15 15-51-16

### DIFF
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
@@ -81,6 +81,7 @@ class MenuController extends ControllerBase implements ApplicationContextAware{
     JobListLinkHandlerRegistry jobListLinkHandlerRegistry
     AuthContextEvaluatorCacheManager authContextEvaluatorCacheManager
     FeatureService featureService
+    StorageService storageService
 
     def configurationService
     ScmService scmService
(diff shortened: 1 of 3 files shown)


ronaveva (approved):
LGTM

gaddman:
The problem with this implementation is that if one user of a project has no key it will disable SCM for everyone in the project?

PR #8054, opened 2022-12-19
RUN-1385: Fix null definitions in plugin groups

chrismcg14:
Confirm config has been provided before enabling plugin group, otherwise do not enable


ltamaster:
@chrismcg14 Testing this fix, I am wondering if it is something that we can validate in the GUI.
For example, now the GUI allows you to save the plugin group with empty values and save the project changes. 
When you go back to the plugin tab, the "Plugin Group" entry disappears which seems odd.

Maybe the best approach could avoid the user saving the Plugin Group session if the user doesn't type any value. For example not enabling the "save" button if all plugin group attributes are empty.

ltamaster (approved):
TMLG

PR #8055, opened 2022-12-20
Run 1386-fix plugin groups deleted when not modified

chrismcg14:
load plugins config everytime page is loaded for edge cases where there are no events. i.e save, delete a plugin group


ltamaster (approved):
TMLG

PR #8062, opened 2022-12-21
RUN-1397: modular annotation authZ 

gschueler:
define ServiceLoader interface for new method interceptor authorization checks

Allows extending authZ checks that can be applied to controller actions with new annotations


ehe-pd (approved):

gschueler:
@mergifyio rebase

PR #8074, opened 2023-01-06
RUN-1385: Plugin group null definitions

chrismcg14:
Plugin groups were able to be saved even if no value was provided. This PR fixes that by confirming the config is set before saving. This PR also updated the emissions to the parent component to only use one method upon loading or when the user saves or deletes.

### DIFF
--- a/rundeckapp/grails-spa/packages/ui/src/pages/project-config/PluginSetConfig.vue
+++ b/rundeckapp/grails-spa/packages/ui/src/pages/project-config/PluginSetConfig.vue
@@ -65,8 +65,6 @@ export default Vue.extend({
   },
   methods:{
     inputReceived(pluginConfigs: ProjectPluginConfigEntry[]) {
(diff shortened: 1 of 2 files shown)


rundeckapp/grails-spa/packages/ui/src/pages/project-config/PluginSetConfig.vue
```diff
@@ -68,38 +63,12 @@ export default Vue.extend({
...
-      });
-      return data
-
-    }
-  },
   methods:{
-    pluginsConfigWasSaved(pluginConfigs: ProjectPluginConfigEntry[]) {
-      this.$emit("saved");
+    inputReceived(pluginConfigs: ProjectPluginConfigEntry[]) {
+      console.log("input received")
```
ltamaster, line 68: this need to be cleaned

ltamaster (approved):
LGTM

PR #8091, opened 2023-01-16
RSE-280 Fix: Cannot Edit Job Schedules

DarwisNarvaezDev:
# RSE-280 Fix: Cannot Edit Job Schedules
when a description is set in a job, the changes made to job schedules are not applied or persisted in DB.

## The Problem
By an unknown reason to date, some of the 'ScheduledExcecution" model's attributes, weren't marked as dirty if the description is set. In other words, if the description is set, the scheduled doesn't turn into an 'updated' value and not persisted in grails "save()" method.

## Solutions Iterated
- Annotate the 'ScheduleExcecution' model with @DirtyCheck, to track the changes. (not working)
- Validating the pre-saved object, to see if the property is dirty; triggered by other validation. (resource-expensive and not working)
- Passing another param to the save method (failOnError: true) to make a rollback of the save action (since the app works as expected before, this didn't work)

## The Solution
Mark the imported (new) job to be updated in db as dirty if the scheduling was changed.

### DIFF
--- a/rundeckapp/grails-app/domain/rundeck/ScheduledExecution.groovy
+++ b/rundeckapp/grails-app/domain/rundeck/ScheduledExecution.groovy
@@ -831,12 +831,12 @@ class ScheduledExecution extends ExecutionContext implements EmbeddedJsonData {
         }else{
             months = params.selectedMonths?.trim() ? valueStringToIndexString(params.selectedMonths, monthsofyearlist) : parseCheckboxFieldFromParams("month", params,true, monthsofyearlist)
         }
-        this.month = months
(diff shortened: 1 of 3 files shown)


ronaveva (changes requested):
instead of marking the object dirty, you could try to use the setters in the entity for value assignation
ScheduledExecution.populateTimeDateFields
```

setMonth(months)

setDayOfWeek(daysOfWeek ?: '?')

setDayOfMonth(daysOfMonth)

setSeconds(params.seconds?params.seconds:"0")

setYear(params.year?params.year:"*")

setCrontabString(null)
```

ronaveva (approved):
LGTM

PR #8100, opened 2023-01-18
RSE-253 Fix: Cannot Add Node Wizard Sources

DarwisNarvaezDev:
# RSE-253 Fix: Cannot Add Node Wizard Sources
Node wizard resource model throws 500 error when size limit is reached. [Ticket](https://pagerduty.atlassian.net/browse/RSE-253)

## The Problem
Since the DB column that hold the bytes for the Node Wizard sources has the 'blob' data type, only certain quantity of bytes can be persisted, so when we exceed this quantity, an error could be seen in Rundeck.

## The Solution
Prevent the error, redirect and notify the user with the error


rundeckapp/grails-app/controllers/rundeck/controllers/FrameworkController.groovy
```diff
@@ -3500,5 +3509,14 @@ class FrameworkController extends ControllerBase implements ApplicationContextAw
         }
         render(status: HttpServletResponse.SC_NO_CONTENT)
     }
+
+    private def nodeSourceFileExceededMaxSize(ByteArrayInputStream bais){
```
ronaveva, line 3513: since this is a new method, please use @CompileStatic annotation

ronaveva (changes requested):

rundeckapp/grails-app/controllers/rundeck/controllers/FrameworkController.groovy
DarwisNarvaezDev: Done, due @CompileStatic, had to change the bais.count to bais.available() to avoid conflicts

ronaveva (approved):
LGTM

rundeckapp/grails-app/controllers/rundeck/controllers/FrameworkController.groovy
```diff
@@ -3500,5 +3510,15 @@ class FrameworkController extends ControllerBase implements ApplicationContextAw
         }
         render(status: HttpServletResponse.SC_NO_CONTENT)
     }
+
+    @CompileStatic
+    private def nodeSourceFileExceededMaxSize(ByteArrayInputStream bais){
+        def nodesMaxSize = 65472
```
ahormazabal, line 3516: * This would limit the file size to 64kb. How was this limit decided? 
IMO it should be not equal but a little less than the value at https://github.com/rundeck/rundeck/blob/b9db4cf0ff290a9bc88141412627d89382868db3/rundeckapp/grails-app/domain/rundeck/Storage.groovy#L44

* Also this value cannot be hardcoded here. It should be at least moved to a constant so its easier to find and mantain.

rundeckapp/grails-app/controllers/rundeck/controllers/FrameworkController.groovy
```diff
@@ -2059,6 +2060,15 @@ class FrameworkController extends ControllerBase implements ApplicationContextAw
 
 
         def bais = new ByteArrayInputStream(params.fileText.toString().getBytes("UTF-8"))
+        def fileValidationResult = nodeSourceFileExceededMaxSize(bais)
+        if( fileValidationResult?.result ){
```
ahormazabal, line 2064: This logic should be reversed. Its too conterintuitive that the validation gives "false" if the validation passes. Would be confusing for a future mantainer.

rundeckapp/grails-app/controllers/rundeck/controllers/FrameworkController.groovy
```diff
@@ -3500,5 +3510,15 @@ class FrameworkController extends ControllerBase implements ApplicationContextAw
...
         render(status: HttpServletResponse.SC_NO_CONTENT)
     }
+
+    @CompileStatic
+    private def nodeSourceFileExceededMaxSize(ByteArrayInputStream bais){
+        def nodesMaxSize = 65472
+        if( bais.available() > nodesMaxSize ){
+            return [result: "Nodes exceed the maximum capacity of subscribed nodes. Max: ${nodesMaxSize}, Input: ${bais.available()}"]
+        }else{
+            return null
```
ahormazabal, line 3520: As already stated, is very counterintuitive that this method returns null if the validation is ok, and an object if the validation fails. In this case would be better to just always return an object with a `passed` flag or something alike.
Example:
`if(exeeds) return [result: false, message: "this file is wrong"]
else return [result: true, message: null]`

ahormazabal (changes requested):

rundeckapp/grails-app/controllers/rundeck/controllers/FrameworkController.groovy
DarwisNarvaezDev: Hey Alberto, thanks for helping out to improving this code; Regarding this comment, although we state in the ORM mappings that the max size of that attribute will be '_n_' quantity, when we connect rundeck with an external DB (like mysql) GORM creates for that attribute, the column DATA with BLOB datatype, which can hold up to 65,535 bytes of data. That is because we declared that limit.

rundeckapp/grails-app/controllers/rundeck/controllers/FrameworkController.groovy
DarwisNarvaezDev: WIP

rundeckapp/grails-app/controllers/rundeck/controllers/FrameworkController.groovy
DarwisNarvaezDev: WIP

ronaveva (approved):
LGTM

ahormazabal (approved):

PR #8112, opened 2023-01-24
RSE-290: Fix: cannot update jobs after upgrade 3.3 to 4.8 job (error 500)

L2JE:
Ticket: https://pagerduty.atlassian.net/browse/RSE-290

### Problem
Roles in Active Directory can have spaces as part of their name, this results on the following error after upgrading from 3.3.x to 4.x (`3.3.10 -> 4.4.8` tested) and editing-saving a job.

image

These are the steps to reproduce the same issue:
1. In a 3.3.X instance (tested on 3.3.10) create a project and a simple job. (WAR/RPM install used)

2. Perform the [upgrade](https://docs.rundeck.com/docs/upgrading/upgrading.html) to 4.X.X version (4.8.0 used)

2. Within the DB schema, change the user_role_list table row of that job to this value (old version because coming from format): `1234ExampleOfRole1,4321ExampleOfRole2`

3. Try to update the job on the UI (just click save without any actual changes) > updates as expected and the list is updated to the new JSON format

4. Within the DB schema, change the user_role_list table row of that job to this value: `1234 ExampleOfRole1,4321 ExampleOfRole2`

5. Try to update the job on the UI (just click save without any actual changes) > see error, job is not updated neither the user_role_list in DB.

### Solution
The library that was used for parsing the userRoleList was replaced with another which is throwing a different type of exception (`com.fasterxml.jackson.databind.exc.MismatchedInputException`) for this case. The solution was to consider a supertype of this exception (`com.fasterxml.jackson.core.JsonProcessingException`) in the `getUserRoles` method implementation.

### DIFF
--- a/rundeckapp/grails-app/domain/rundeck/ScheduledExecution.groovy
+++ b/rundeckapp/grails-app/domain/rundeck/ScheduledExecution.groovy
@@ -24,6 +24,7 @@ import com.dtolabs.rundeck.core.dispatcher.DataContextUtils
 import com.dtolabs.rundeck.core.jobs.JobOption
 import com.dtolabs.rundeck.core.jobs.JobReference
 import com.fasterxml.jackson.core.JsonParseException
+import com.fasterxml.jackson.core.JsonProcessingException
 import org.rundeck.util.Sizes
 import rundeck.services.JobReferenceImpl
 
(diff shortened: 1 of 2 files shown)


ronaveva (changes requested):
@L2JE please add a test for the exception scenario

ronaveva (approved):
LGTM

PR #8116, opened 2023-01-26
RUN-1447: Job with Job references does not use correct Runners in referenced jobs

ltamaster:
Core changes for issue with Job references and the runner( does not use correct Runners in referenced jobs )

**Additional context**


core/src/main/java/com/dtolabs/rundeck/core/jobs/JobReferenceItem.java
```diff
@@ -0,0 +1,158 @@
...
+        this.failOnDisable = failOnDisable;
+        this.importOptions = importOptions;
+        this.uuid = uuid;
+        this.useName = useName;
+        this.ignoreNotifications = ignoreNotifications;
+        this.childNodes = childNodes;
+    }
+
+    @Override
+    public String getLabel() {
```
gschueler, line 70: Let's use Lombok instead

ahormazabal (approved):

PR #8118, opened 2023-01-30
RSE-266 Fix: Cannot remove notifications from a job by loading an XML file

Jesus-Osuna-M:
Problem: When trying to update a job by uploading a job definition, it will not remove existing notifications if the new job definition does not have any notifications. This problem was generated by the way the notifications were trying to be deleted, not correctly removing it from the object. 

Solution: Create a method to remove notifications based on the method deleteExistingNotification, to correctly remove notifications from the updated job if the input job doesn't have any notifications.


ronaveva (approved):
LGTM

PR #8124, opened 2023-02-02
RUN-1445: Upgrade the version of sshj-plugin to 0.1.4 which fixed the…

ehe-pd:
This is a bugfix. 

A bug caused by new version of sshj lib was fixed in `sshj-plugion:v0.1.4`, we need to update the `build.yaml` file to reflect the version change.


fdevans:
Can we get more details what the issue is that this fixes?

gschueler (approved):

gschueler:
from the ticket:
running commands via sshj plugin, subsequent commands show "no ouput" even when successful.
caused by a change in the sshj library

PR #8140, opened 2023-02-09
RUN-1471 change update from storage data service to gorm data object save to f…

imad6639:
Bugfix to switch saving existing domain object from data services to gorm object.

### DIFF
--- a/rundeckapp/src/main/groovy/org/rundeck/app/data/providers/storage/GormStorageDataProvider.groovy
+++ b/rundeckapp/src/main/groovy/org/rundeck/app/data/providers/storage/GormStorageDataProvider.groovy
@@ -72,7 +72,7 @@ class GormStorageDataProvider implements StorageDataProvider {
         storage.lastUpdated = new Date()
 
          try {
-             storageDataService.save(storage)
+             storage.save(flush: true)
         } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


rundeckapp/src/main/groovy/org/rundeck/app/data/providers/storage/GormStorageDataProvider.groovy
```diff
@@ -72,7 +72,7 @@ class GormStorageDataProvider implements StorageDataProvider {
         storage.lastUpdated = new Date()
 
          try {
-             storageDataService.save(storage)
+             storage.save(flush: true)
```
ahormazabal, line 75: I'd put this inside StorageDataService as a custom method, like it was done here: https://github.com/rundeckpro/rundeckpro/pull/2848/commits/e45b4fae6bc3dd7603e1cd61f8e0c850e18093c3

ahormazabal (changes requested):

rundeckapp/src/main/groovy/org/rundeck/app/data/providers/storage/GormStorageDataProvider.groovy
ahormazabal: This change will be deferred for a later version

ahormazabal (approved):

PR #8149, opened 2023-02-16
Adding PluginMeta spi

ocerda:
Move PluginMeta interaction of the database behind data SPI

A new RdPluginMeta interface was created and all of the references were moved to new entity

**Additional context**
Jira ticket https://pagerduty.atlassian.net/browse/RCLOUD-794

### DIFF
--- a/rundeck-data-model/src/main/java/org/rundeck/app/data/model/v1/pluginMeta/RdPluginMeta.java
+++ b/rundeck-data-model/src/main/java/org/rundeck/app/data/model/v1/pluginMeta/RdPluginMeta.java
@@ -1,0 +1,12 @@
+package org.rundeck.app.data.model.v1.pluginMeta;
+
+import java.util.Date;
+
+public interface RdPluginMeta {
+    Long getId();
+    String getKey();
+    String getProject();
+    String getJsonData();
+    Date getDateCreated();
+    Date getLastUpdated();
+}
--- a/rundeck-data-model/src/main/java/org/rundeck/app/data/model/v1/pluginMeta/RdPluginMeta.java
+++ b/rundeck-data-model/src/main/java/org/rundeck/app/data/model/v1/pluginMeta/RdPluginMeta.java
@@ -1,12 +1,13 @@
 package org.rundeck.app.data.model.v1.pluginMeta;
 
 import java.util.Date;
+import java.util.Map;
 
 public interface RdPluginMeta {
     Long getId();
     String getKey();
     String getProject();
-    String getJsonData();
     Date getDateCreated();
     Date getLastUpdated();
+    Map<String,Object> getPluginData();
 }
--- a/rundeckapp/grails-app/services/rundeck/services/JobMetadataService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/JobMetadataService.groovy
@@ -88,7 +88,7 @@ class JobMetadataService {
      * @param project project
      */
     def removeProjectPluginMeta(final String project, final String type) {
-        pluginMetaDataProvider.deleteByProjectAndDataKey(project, type)
+        pluginMetaDataProvider.deleteAllByProjectAndKeyLike(project, "%/${type}")
     }
 
     /**
--- a/rundeckapp/grails-app/services/rundeck/services/JobMetadataService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/JobMetadataService.groovy
@@ -97,7 +97,7 @@ class JobMetadataService {
      * @param id jobid
      */
     def removeJobPluginMetaAll(final String project, final String id) {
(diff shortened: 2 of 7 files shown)


rundeck-data-model/src/main/java/org/rundeck/app/data/model/v1/pluginMeta/RdPluginMeta.java
```diff
@@ -0,0 +1,12 @@
+package org.rundeck.app.data.model.v1.pluginMeta;
+
+import java.util.Date;
+
+public interface RdPluginMeta {
+    Long getId();
+    String getKey();
+    String getProject();
+    String getJsonData();
```
sjrd218, line 9: Let's define this as `Map<String,Object> getPluginData()` which is found on the PluginMeta domain class. That will allow the data to be transferred as json rather then json embedded inside json. 

It's unfortunate that we have to use a map here to hold the data, but since we truly don't know what it is and it could be anything we have to break our informal 'don't use maps in the api' rule.

sjrd218 (changes requested):

rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormPluginMetaDataProvider.groovy
```diff
@@ -0,0 +1,57 @@
...
+    @Override
+    Integer deleteAllByProject(String project) {
+        return PluginMeta.executeUpdate('delete PluginMeta where project=:project', [project: project], [flush: true])
+
+    }
+
+    @Override
+    Integer deleteByProjectAndDataKey(String project, String type) {
+        return PluginMeta.executeUpdate('delete PluginMeta where project=:project and data_key like :data_key' , [project: project, data_key: "%/${type}"], [flush: true])
+    }
```
gschueler, line 44: `deleteAllByProjectAndKeyLike` and `deleteByProjectAndDataKey` do almost the same thing, it doesn't seem like we need two different methods here. The callers of `deleteByProjectAndDataKey` can be modified to conform to the input for `deleteAllByProjectAndKeyLike` and just call that method

gschueler (changes requested):

rundeckapp/grails-app/services/rundeck/services/JobMetadataService.groovy
```diff
@@ -88,7 +88,7 @@ class JobMetadataService {
      * @param project project
      */
     def removeProjectPluginMeta(final String project, final String type) {
-        pluginMetaDataProvider.deleteByProjectAndDataKey(project, type)
+        pluginMetaDataProvider.deleteAllByProjectAndKeyLike(project, type)
```
gschueler, line 91: original `deleteByProjectAndDataKey` converted parameter `type` into a LIKE query using `%/${type}`, so that should be what is passed in here.

rundeckapp/grails-app/services/rundeck/services/JobMetadataService.groovy
```diff
@@ -97,7 +97,7 @@ class JobMetadataService {
      * @param id jobid
      */
     def removeJobPluginMetaAll(final String project, final String id) {
-        pluginMetaDataProvider.deleteAllByProjectAndKeyLike(project, id + '/%')
+        pluginMetaDataProvider.deleteAllByProjectAndKeyLike(project, id)
```
gschueler, line 100: I think this should be reverted, the parameter should be the LIKE query, so `id + '/%'` is correct

rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormPluginMetaDataProvider.groovy
```diff
@@ -26,7 +26,7 @@ class GormPluginMetaDataProvider implements PluginMetaDataProvider{
 
     @Override
     void deleteAllByProjectAndKeyLike(String project, String key) {
-        def found = PluginMeta.findAllByProjectAndKeyLike(project, key)
+        def found = PluginMeta.findAllByProjectAndKeyLike(project, key + '/%')
```
gschueler, line 29: this should accept the parameter "key" as the LIKE query , so should not append anything. Also I suggest renaming the parameter to `keyLike` to be clear

gschueler (changes requested):

ltamaster (approved):
tested

rundeckapp/grails-app/services/rundeck/services/JobMetadataService.groovy
ltamaster: @ocerda this need to be added

gschueler (approved):

sjrd218 (approved):

PR #8152, opened 2023-02-19
RUN-1472: No Message On GUI For Scheduled Job On Unhealthy Runner

ehe-pd:
Bugfix

If a scheduled job tries to run on an unhealthy Runner, the execution will be recorded in the database with a proper error message. The previous implementation just throw an exception which caused the execution record discarded so the user cannot know what happened. 
But for Adhoc execution, the failed execution record won't be saved but an error message will be thrown and user can see the error immediately on the page.


rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -2517,6 +2537,14 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
         }
 
         execution.scheduledExecution=se
+
+        // If there is a preExecutionCheckError, put that error into the execution's extraMetadataMap.
+        if(!beforeExecutionResult.isSuccessful() && beforeExecutionResult.isTriggeredByJobEvent(JobPreExecutionEvent)) {
+            Map<String, String> extraMeta = new HashMap<>()
+            extraMeta.put(JobPreExecutionEvent.getName(), beforeExecutionResult.getErrorMessage())
+            execution.setExtraMetadataMap(extraMeta)
```
gschueler, line 2545: this would overwrite any other extra metadata, I think it should merge with any existing Map data instead

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -1412,6 +1421,8 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
     public static String EXECUTION_MISSED = "missed"
     public static String EXECUTION_QUEUED = "queued"
 
+    public static String BEFORE_EXECUTION_CHECK_FAILURE = "before-execution-check-failure"
```
gschueler, line 1424: is this unused now?

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
ehe-pd: Good point. I will make it a merge operation. Thanks.

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
ehe-pd: Yes. Should be deleted.

gschueler (approved):

PR #8157, opened 2023-02-23
RUN-1479: Establish new section in execution head page for feature use.

carlosrfranco:
A new socket is needed to include runner info related to #8108 

Includes new section in execution head page for the runner info socket

### DIFF
--- a/rundeckapp/grails-app/views/execution/show.gsp
+++ b/rundeckapp/grails-app/views/execution/show.gsp
@@ -471,6 +471,14 @@ search
 
                         </section>
 
+                        <section>
+                            <div class="vue-ui-socket">
+                                <div>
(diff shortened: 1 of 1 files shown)


rundeckapp/grails-app/views/execution/show.gsp
```diff
@@ -471,6 +471,14 @@ search
 
                         </section>
 
+                        <section>
+                            <div class="vue-ui-socket">
+                                <div>
+                                    <ui-socket section="job-runner-execution" location="top" :event-bus="EventBus" />
```
gschueler, line 477: the section should be something like `execution-show` location `header`
```suggestion
                                    <ui-socket section="execution-show" location="header" :event-bus="EventBus" />
```

gschueler (changes requested):

gschueler (approved):

PR #8167, opened 2023-02-28
RUN-1492: Fix: Webhooks not showing proper information After Create

mishingo:
Bugfix: Upon creation of a webhook, the page wouldn't show the correct information, and appear to have had an issue.

Currently the UUID that is used to select the webhook is incorrect, and has been modified to use the correct UUID from the API.


gschueler (approved):
It seems the webhooks API has some issues, since it should be returning the created Webhook's UUID in the response, but it is not. I will create a separate issue to fix the API, but this change does solve the immediate problem.  There is a chance this code will not be correct, if the new webhook is not the final item in the list.

PR #8169, opened 2023-03-01
RSE-335: Fix: Enable schedules from project settings don't work as expected in clustermode

L2JE:
Ticket: https://pagerduty.atlassian.net/browse/RSE-335

#### Problem
Cluster member wasn't able to reschedule jobs that are not owned by it because it was reading the `project.disable.schedule` property from cache (which was invalid after the schedule state was changed from other member and before the schedule change message was processed)

#### Solution
Allow `FrameworkService` to return project configuration after reloading it.


alexander-variacode (approved):
LGTM

PR #8170, opened 2023-03-01
RSE-254: Enh: add multiple node wizard sources

L2JE:
Ticket: https://pagerduty.atlassian.net/browse/RSE-254

#### Problem
Adding a second Node Wizard resource points towards the first one

##### Side issue fixed
1. Delete a `node wizard` resource that has a node definition in it and save project
2. Add a new `node wizard` resource and save project
3. Go to edit node page and see the new node wizard has the old wizard content (It should've been deleted)

#### Solution
* Add `AutogenInstanceId` property for a property that need to be generated on front-end when plugin that uses it is created.
* Send `project.plugins.removed.<PROVIDER_NAME>` event to allow node resources subscribe to it and delete stored data if necessary.


alexander-variacode (approved):
LGTM

PR #8171, opened 2023-03-01
issue rse-288. SCM plugin does not release .pack files from ScmExport…

Oalbarran94:
Issue [RSE-288](https://pagerduty.atlassian.net/browse/RSE-288)

### DIFF
--- a/plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/GitExportPlugin.groovy
+++ b/plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/GitExportPlugin.groovy
@@ -143,6 +143,7 @@ class GitExportPlugin extends BaseGitPlugin implements ScmExportPlugin {
 
     @Override
     void cleanup() {
+        repo?.close()
         git?.close()
         git?.getRepository()?.close()
     }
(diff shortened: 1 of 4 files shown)


plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/GitExportPlugin.groovy
```diff
@@ -144,18 +146,28 @@ class GitExportPlugin extends BaseGitPlugin implements ScmExportPlugin {
...
+        git?.close()
         File base = new File(config.dir)
-        try {
-            FileUtils.delete(base, FileUtils.RECURSIVE)
-        } catch(IOException e){
-            logger.error("Failed to delete repo folder")
+
+        if(base.exists()){
+            try {
+                if (System.getProperty("os.name").toLowerCase().contains("windows")) {
```
ronaveva, line 160: A test is needed to check for this line to be true at some point and validate that "windowCacheConfig.install()" is called 1 time and also, that is called 0 time if the condition is not met

ronaveva (changes requested):
it needs some unit tests, see comment

ronaveva (approved):
LGTM

PR #8178, opened 2023-03-03
RUN-1542: Fix: scm import/export options for job JSON

gschueler:
* match behavior of YAML formatter
* add tests
* add indented output by default in JSON document fix RUN-1540


ltamaster (approved):
TMLG

PR #8193, opened 2023-03-10
RUN-199: Improve Authentication Options for Remote URL Option Input

ltamaster:
Enhancement: authentication method options as part of the Remote URL Option Input.

**Basic Authentication**
Use a username and password (from key storage) to authenticate

Screenshot 2023-03-22 at 20 24 09

**API Key Authentication**

Use a key name and token to authenticate. The key can be passed from a header or a query parameter

Screenshot 2023-03-22 at 20 25 40

**Bearer Authentication**

Use an authorization header like `Authorization: Bearer the-token-value`

Screenshot 2023-03-22 at 20 29 11

**Additional context**


gschueler (changes requested):

rundeckapp/src/test/groovy/rundeck/controllers/ScheduledExecutionControllerSpec.groovy
```diff
@@ -3909,4 +3913,113 @@ class ScheduledExecutionControllerSpec extends RundeckHibernateSpec implements C
...
+        configRemoteUrl.token = "123"
+        configRemoteUrl.keyName = "apiKey"
+        configRemoteUrl.apiTokenReporter = ApiTokenReporter.HEADER
+
+        when:
+        controller.getRemoteJSON(url, configRemoteUrl, 100, 100, 5,false)
+
+        then:
+        1 * controller.logger.debug("getRemoteJSON using authentication")
+        1 * controller.logger.debug("API key authentication with Header apiKey")
```
gschueler, line 3965: Should avoid relying on side-effect of using the logger.debug in the test.

Perhaps introduce an interface that can be mocked.  E.g. `interface HttpClientCreator{ HttpClient<HttpResponse> createClient() }` then you can replace [this line](https://github.com/rundeck/rundeck/pull/8193/files#diff-0588b6c107c25b5c990676b0df69cfec52009719523a901e54dd5eb9feb91622R850) with a call to the HttpClientCreator. 

Then you can test the `setUri`, `setHeader` etc all have the right values

gschueler (changes requested):

rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
```diff
@@ -3985,7 +3990,11 @@ class ScheduledExecutionService implements ApplicationContextAware, Initializing
                 def projectConfig = framework.frameworkProjectMgr.loadProjectConfig(scheduledExecution.project)
                 boolean disableRemoteOptionJsonCheck = projectConfig.hasProperty(REMOTE_OPTION_DISABLE_JSON_CHECK)
 
-                remoteResult = ScheduledExecutionController.getRemoteJSON(srcUrl, timeout, contimeout, retryCount, disableRemoteOptionJsonCheck)
+                remoteResult = ScheduledExecutionController.getRemoteJSON(new HttpClientCreator(){
+                    HttpClient<HttpResponse> createClient(){
+                        new ApacheHttpClient()
+                    }
```
gschueler, line 3996: I think you could just pass a closure, `{->new ApacheHttpClient()}` or even just `ApacheHttpClient.&newInstance` instead of an anonymous class

rundeckapp/src/test/groovy/rundeck/controllers/ScheduledExecutionControllerSpec.groovy
```diff
@@ -3909,4 +3925,188 @@ class ScheduledExecutionControllerSpec extends RundeckHibernateSpec implements C
...
+                }
+                getEntity()>>Mock(HttpEntity){
+                    getContent()>>new ByteArrayInputStream('{}'.getBytes());
+                }
+                getFirstHeader("Content-Type")>>Mock(Header){
+                    getValue()>>"application/json"
+                }
+            }
+            processor.accept(rsp)
+        })
```
gschueler, line 3967: so Spock will evaluate this by expecting the closure inside the method call argument to work like a constraint and evaluates the [statements like assertions](https://spockframework.org/spock/docs/2.1/interaction_based_testing.html#_code_constraint) (in this case it still succeeds), but for side-effects like calling processor.accept here, you probably want to do it like this here and in the other tests: 

```suggestion
        1 * client.execute(_) >> {
            RequestProcessor processor = it[0]
            HttpResponse rsp = Mock(HttpResponse){
                getStatusLine()>>Mock(StatusLine){
                    getStatusCode()>>200
                }
                getEntity()>>Mock(HttpEntity){
                    getContent()>>new ByteArrayInputStream('{}'.getBytes());
                }
                getFirstHeader("Content-Type")>>Mock(Header){
                    getValue()>>"application/json"
                }
            }
            processor.accept(rsp)
        }
```

gschueler (approved):

ltamaster:
@sjrd218 I added small changes for the new `rundeck-data-shared` in order to avoid parse issues with `configMap` field.
I think the long-term solution is to include `JobOptionConfigData configData` as a field instead `configMap` but I have some dependency issues in  `rundeck-data-model`

grails-rundeck-data-shared/src/main/groovy/rundeck/data/job/RdOption.groovy
```diff
@@ -85,4 +86,12 @@ class RdOption implements JobOption, OptionData, Comparable<OptionData>, Validat
     }
 
     static ObjectMapper mapper = new ObjectMapper()
+
+    @Override
+    Map<String, String> getConfigMap() {
+        if(!configData){
+            return null
+        }
+        configData.getJobOptionConfigEntries().stream().findFirst().orElse(null).toMap()
```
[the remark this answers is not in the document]
ltamaster: @sjrd218 sent a commit to fix that

PR #8208, opened 2023-03-22
Fix plugin registry loading behavior

sjrd218:
Make rundeck plugin registry check for null objects rather than rely on groovy truthy checks which can cause unexpected behavior

### DIFF
--- a/rundeckapp/src/test/groovy/com/dtolabs/rundeck/server/plugins/RundeckPluginRegistrySpec.groovy
+++ b/rundeckapp/src/test/groovy/com/dtolabs/rundeck/server/plugins/RundeckPluginRegistrySpec.groovy
@@ -6,11 +6,13 @@ import com.dtolabs.rundeck.core.common.IRundeckProject
 import com.dtolabs.rundeck.core.common.ProjectManager
 import com.dtolabs.rundeck.core.common.PropertyRetriever
 import com.dtolabs.rundeck.core.execution.service.NodeExecutor
+import com.dtolabs.rundeck.core.logging.LogEvent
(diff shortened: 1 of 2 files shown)


rundeckapp/src/test/groovy/com/dtolabs/rundeck/server/plugins/RundeckPluginRegistrySpec.groovy
```diff
@@ -919,6 +922,103 @@ class RundeckPluginRegistrySpec extends Specification implements GrailsUnitTest
...
+        registry.pluginDirectory = pdir
+        registry.metaClass.findBean = { String beanName -> return new TestStreamingLogReaderPlugin() }
+        registry.pluginRegistryMap = ["StreamingLogReader:mystreamingreader": "mystreamingreader"]
+        def actual = registry.loadBeanDescriptor("mystreamingreader","StreamingLogReader")
+        then:
+        actual
+
+    }
+
+    static class TestStreamingLogReaderPlugin implements StreamingLogReaderPlugin, Describable {
```
JoseOrtiz, line 939: Do we need this to be `Describable`?

sjrd218: yes, otherwise it will return null which is not what we want

JoseOrtiz: Ohh, right, for the loadBeanDescription

imad6639 (approved):

PR #8217, opened 2023-03-28
RUN-1615: Invalidate user sessions upon password change/reset 

chrismcg14:
Inject session beans even if maxsessions isn't enforced excluding concurrentSessionControlAuthenticationStrategy

Needed for https://github.com/rundeckpro/rundeckpro/pull/2946

### DIFF
--- a/rundeckapp/grails-app/conf/spring/resources.groovy
+++ b/rundeckapp/grails-app/conf/spring/resources.groovy
@@ -774,46 +774,32 @@ beans={
         SpringSecurityUtils.registerLogoutHandler("cookieClearingLogoutHandler")
 
     }
-
+    sessionRegistry(SessionRegistryImpl)
+    concurrentSessionFilter(ConcurrentSessionFilter, sessionRegistry)
+    registerSessionAuthenticationStrategy(RegisterSessionAuthenticationStrategy, ref('sessionRegistry')) {}
+    sessionFixationProtectionStrategy(SessionFixationProtectionStrategy) {
+        migrateSessionAttributes = grailsApplication.config.getProperty("grails.plugin.springsecurity.sessionFixationPrevention.migrate", String.class,'')
+        // true
+        alwaysCreateSession = grailsApplication.config.getProperty("grails.plugin.springsecurity.sessionFixationPrevention.alwaysCreateSession", String.class, '')
+        // false
+    }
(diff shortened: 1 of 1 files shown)


rundeckapp/grails-app/conf/spring/resources.groovy
```diff
@@ -797,6 +797,21 @@ beans={
                 [concurrentSessionControlAuthenticationStrategy, sessionFixationProtectionStrategy, registerSessionAuthenticationStrategy]
         )
     }
+    else {
```
ahormazabal, line 800: I think it would be better to just put all this out of the if() instead of adding duplicated code, and just leave the concurrentSessionControlAuthenticationStrategy inside the if condition. As this is security stuff it should be kept simple and to the point. Otherwise a future mantainer could miss to fix one of the code parts and fail to fix a potential security issue.

ahormazabal: Could be something like (id did not test this though)
```

    sessionRegistry(SessionRegistryImpl)
    concurrentSessionFilter(ConcurrentSessionFilter, sessionRegistry)
    registerSessionAuthenticationStrategy(RegisterSessionAuthenticationStrategy, ref('sessionRegistry')) {}

    sessionFixationProtectionStrategy(SessionFixationProtectionStrategy) {
        migrateSessionAttributes = grailsApplication.config.getProperty("grails.plugin.springsecurity.sessionFixationPrevention.migrate", String.class, '')
        // true
        alwaysCreateSession = grailsApplication.config.getProperty("grails.plugin.springsecurity.sessionFixationPrevention.alwaysCreateSession", String.class, '')
        // false
    }
    
    var authenticationStrategies = [sessionFixationProtectionStrategy, registerSessionAuthenticationStrategy]
    if (grailsApplication.config.getProperty("rundeck.security.enforceMaxSessions", Boolean.class, false)) {
        concurrentSessionControlAuthenticationStrategy(
                ConcurrentSessionControlAuthenticationStrategy,
                ref('sessionRegistry')
        ) {
            exceptionIfMaximumExceeded = false
            maximumSessions = grailsApplication.config.getProperty("rundeck.security.maxSessions", Integer.class, 1)
        }
        authenticationStrategies.add(0, concurrentSessionControlAuthenticationStrategy)
    }
    
    sessionAuthenticationStrategy(
            CompositeSessionAuthenticationStrategy,
            authenticationStrategies
    )

```

ahormazabal (approved):

PR #8220, opened 2023-03-30
RSE-321: Fix: ERROR while running a dir command in a path with Japanese files (Upgrade plugin version)

L2JE:
Ticket: https://pagerduty.atlassian.net/browse/RSE-321

Fixed in: https://github.com/rundeck-plugins/py-winrm-plugin/pull/94

Upgrading py-winrm plugin version, fix is present in v2.1.2


ronaveva (approved):
LGTM

PR #8221, opened 2023-03-31
RUN-1616: Set Cookie SemeSite Flag To Strict

ehe-pd:
It's a security enhancement. 

Add a spring configuration `server.servlet.session.cookie.same-site` to application.yml file. Spring boot will automatically add `SameSite=Strict` header into the HTTP cookie.

**Additional context**
https://github.com/rundeck/rundeck/pull/new/RUN-1616-Cookie-Missing-SameSite-Security-Flag
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#strict

### DIFF
--- a/rundeckapp/grails-app/conf/application.yml
+++ b/rundeckapp/grails-app/conf/application.yml
@@ -335,6 +335,7 @@ server:
             persistent: false
             cookie:
                 http-only: true
+		same-site: "Strict"
             tracking-modes: 'cookie'
 rundeck:
     web:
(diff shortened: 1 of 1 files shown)


chrismcg14 (approved):

RolandVExp:
Hello, this feature good from a security point of view, but it has a drawback that should be fixed.
If you link a rundeck job from another domain, you go to the login page which is now normal, but once logged in you end up on the home. 
The target page is lost.

PR #8227, opened 2023-04-04
RSE-269 Fix: Remove Webhook Debug Data

DarwisNarvaezDev:
# RSE-269 Fix: Remove Webhook Debug Data
Some of the clients of Advanced Webhooks feature were reporting performance issues caused by excessive amount of db records for webhook data as stored events.

# The Problem
For each call to the advanced webhook, two records are created in db, this data caused performance issues.

# The Fix
Querying the database to delete events related with webhooks when a particular webhook is removed. 

## Implications
- Users who would like to remove event records belonging to webhooks that have been removed from a project can run this query:
```sql
DELETE FROM STORED_EVENT WHERE 
TOPIC LIKE '%debug%'
OR
TOPIC LIKE '%recent%'
AND
SUBSYSTEM LIKE 'webhooks'
AND
PROJECT_NAME LIKE '<PROJECT NAME>'
```

### DIFF
--- a/grails-webhooks/grails-app/services/webhooks/WebhookService.groovy
+++ b/grails-webhooks/grails-app/services/webhooks/WebhookService.groovy
@@ -103,7 +103,7 @@ class WebhookService {
         Long queryResultForDebug = deleteEvents(TOPIC_DEBUG_EVENTS, webhook)
         Long queryResultForRecentEvents = deleteEvents(TOPIC_RECENT_EVENTS, webhook)
         def totalAmountOfRowsAffected = queryResultForDebug + queryResultForRecentEvents
-        log.info("Rows affected in the query: ${totalAmountOfRowsAffected}")
+        log.info("${totalAmountOfRowsAffected} events deleted related to the webhook: ${webhook.name}")
     }
 
     /**


fdevans:
Can we revisit this with Product involved?  I am tempted to include a Debug On/Off switch with a default of "off", but would like to discuss further.

grails-webhooks/grails-app/services/webhooks/WebhookService.groovy
```diff
@@ -87,6 +96,33 @@ class WebhookService {
         return plugin.onEvent(context,data) ?: new DefaultWebhookResponder()
     }
 
+    Long handleWebhookEventsData(EventQueryType queryType, RdWebhook webhook){
```
ronaveva, line 99: I don't see the point on having "EventQueryType" if it's just delete. Also the method name should be something related to what is doing.
Also i think there's no need to return the number of the records deleted, instead it should be logged here and just be void.
Please add javadoc to the method

ronaveva (changes requested):
See comments

grails-webhooks/grails-app/services/webhooks/WebhookService.groovy
DarwisNarvaezDev: done

ronaveva (approved):
LGTM

PR #8256, opened 2023-04-14
fix: project context not shown in storage view

gschueler:

fix: project context not shown in storage view (new Vue implementation)

* don't render key-storage-view until after mount of key-storage-page, so that project param is used (other solution would be to watch the project prop inside key-storage-view)
* use correct key storage path in API requests, based on rootPath


gschueler:
changes in other pr

PR #8265, opened 2023-04-18
RUN-1627: Deferred project delete

ahormazabal:
- Adds ability to defer the project deletion process to happen in background.
- Adds framework support for disabling projects.

To enable this feature, set the `rundeck.projectService.deferredProjectDelete=true` property in your `rundeck-config.properties` file.

### DIFF
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
@@ -2868,6 +2868,15 @@ Since: V18''',
 
         def ScheduledExecution scheduledExecution = scheduledExecutionService.getByIDorUUID(params.id)
 
+        if (scheduledExecution?.project && frameworkService.isFrameworkProjectDisabled(scheduledExecution.project)) {
+            return apiService.renderErrorFormat(response, [
+                    status: HttpServletResponse.SC_NOT_FOUND,
+                    code: 'api.error.project.disabled',
+                    args: [scheduledExecution.project],
+                    format: response.format
+            ])
+        }
+
         if (!apiService.requireExists(response, scheduledExecution, ['Job ID', params.id])) {
             return
         }
@@ -2899,14 +2908,6 @@ Since: V18''',
                             args  : [response.format]
                     ]
             )
-        }
-        if (frameworkService.isFrameworkProjectDisabled(scheduledExecution.project)) {
-            return apiService.renderErrorFormat(response, [
-                    status: HttpServletResponse.SC_NOT_FOUND,
-                    code: 'api.error.project.disabled',
-                    args: [scheduledExecution.project],
-                    format: response.format
-            ])
         }
         def extra = [:]
         def clusterModeEnabled = frameworkService.isClusterModeEnabled()
@@ -3005,6 +3006,15 @@ Format is a string like `2d1h4n5s` using the following characters for time units
 
         def ScheduledExecution scheduledExecution = scheduledExecutionService.getByIDorUUID(params.id)
 
+        if (scheduledExecution?.project && frameworkService.isFrameworkProjectDisabled(scheduledExecution.project)) {
+            return apiService.renderErrorFormat(response, [
+                    status: HttpServletResponse.SC_NOT_FOUND,
+                    code: 'api.error.project.disabled',
+                    args: [scheduledExecution.project],
+                    format: response.format
+            ])
+        }
+
         if (!apiService.requireExists(response, scheduledExecution, ['Job ID', params.id])) {
             return
         }
@@ -3037,14 +3047,7 @@ Format is a string like `2d1h4n5s` using the following characters for time units
                     ]
             )
         }
-        if (frameworkService.isFrameworkProjectDisabled(scheduledExecution.project)) {
-            return apiService.renderErrorFormat(response, [
-                    status: HttpServletResponse.SC_NOT_FOUND,
-                    code: 'api.error.project.disabled',
-                    args: [scheduledExecution.project],
-                    format: response.format
-            ])
-        }
+
         def extra = [:]
 
         //future scheduled executions forecast
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/ProjectController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/ProjectController.groovy
@@ -902,7 +902,7 @@ Authorization required: `delete` access for `project` resource type or `admin` o
                     @Parameter(
                             name = 'deferred',
                             in = ParameterIn.QUERY,
-                            description = 'Deferred Delete',
+                            description = 'Deferred Delete. Since: v45',
                             allowEmptyValue = false,
                             required = false,
                             schema = @Schema(implementation = Boolean.class)
@@ -934,14 +934,11 @@ Authorization required: `delete` access for `project` resource type or `admin` o
         def authorizing = authorizingProject
         def project1 = authorizing.resource
 
-        Boolean deferred = (request.api_version < ApiVersions.V45) ? false :
-                configurationService.getBoolean("projectService.deferredProjectDelete", true)
+        Boolean deferred = false
 
-        if(params.containsKey("deferred")) {
-            if (!apiService.requireApi(request, response, ApiVersions.V45)) {
-                return
-            }
-            deferred = params.getBoolean("deferred")
+        if (request.api_version >= ApiVersions.V45) {
+            deferred = params.getBoolean("deferred",
+                    configurationService.getBoolean("projectService.deferredProjectDelete", true))
         }
 
         ProjectService.DeleteResponse result = projectService.deleteProject(
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/ScheduledExecutionController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/ScheduledExecutionController.groovy
@@ -673,6 +673,15 @@ if the step is a node step. Implicitly `"true"` if not present and not a job ste
 
         def ScheduledExecution scheduledExecution = scheduledExecutionService.getByIDorUUID( params.id )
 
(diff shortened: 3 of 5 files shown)


rundeckapp/grails-app/controllers/rundeck/controllers/FrameworkController.groovy
```diff
@@ -928,8 +930,12 @@ class FrameworkController extends ControllerBase implements ApplicationContextAw
...
             projectDescriptionError = message(code: "project.description.can.only.contain.these.characters")
             errors << projectDescriptionError
-        } else if (framework.getFrameworkProjectMgr().existsFrameworkProject(project)) {
-            projectNameError = "Project already exists: ${project}"
+        } else if (projectManager.isFrameworkProjectDisabled(project)) {
+            projectNameError = message(code: "project.disabled", args: [project])
+            log.error(projectNameError)
+            errors << projectNameError
+        } else if (projectManager.existsFrameworkProject(project)) {
+            projectNameError = message(code: "project.exists", args: [project])
```
gschueler, line 938: wow, I think this validation checks (name, exist, disabled) should go much earlier in this method, so we can fail fast without doing all of the other stuff.  Not directly related to your change, but seems like basic validation that should come early

rundeckapp/grails-app/services/rundeck/services/ProjectService.groovy
```diff
@@ -1714,13 +1710,52 @@ class ProjectService implements InitializingBean, ExecutionFileProducer, EventPu
      */
     @GrailsCompileStatic
     DeleteResponse deleteProject(IRundeckProject project, IFramework framework, AuthContext authContext, String username) {
+        log.info("Requested deletion of project ${project.name}")
```
gschueler, line 1713: should this method verify the project is not already disabled? (idempotency)

rundeckapp/grails-app/migrations/core/Project.groovy
```diff
@@ -30,4 +30,14 @@ databaseChangeLog = {
...
-}
\ No newline at end of file
+    changeSet(author: "Alberto Hormazabal", id: "4.13.0-add-project-state") {
+        preConditions(onFail: "MARK_RAN") {
+            not {
+                columnExists(tableName: "project", columnName: 'state')
+            }
+        }
+        addColumn(tableName: "project") {
+            column(name: 'state', type: '${text.type}')
```
gschueler, line 40: should also add an index on this column

rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormProjectDataProvider.groovy
```diff
@@ -33,7 +33,10 @@ class GormProjectDataProvider implements RundeckProjectDataProvider {
     }
 
     Long create(RdProject data) throws DataAccessException {
-        Project project = new Project(description: data.getDescription(), name: data.getName())
+        Project project = new Project(
+                name: data.getName(),
+                description: data.getDescription(), 
+                state: data.getState())
```
gschueler, line 39: will we create a project in disabled state? should we just use enabled state as default here?

rundeckapp/grails-app/services/rundeck/services/ProjectService.groovy
```diff
@@ -1721,13 +1717,61 @@ class ProjectService implements InitializingBean, ExecutionFileProducer, EventPu
...
+            }
+        } catch (UnsupportedOperationException e) {
+            log.warn("Could not disable project. Aborting project delete operation: ${e.getMessage()}", e)
+            throw new UnsupportedOperationException("Could not delete project. Project disabling not supported by framework: ${e.getMessage()}", e)
+        }
+
+        log.info("Deferring deletion of project ${project.name} to background task.")
+
+        def promise = Promises.task {
+            return Project.withNewTransaction {
```
sjrd218, line 1744: Move `Project.withNewTransaction` to Gorm data provider implementation

rundeckapp/grails-app/services/rundeck/services/ProjectService.groovy
```diff
@@ -1721,13 +1717,61 @@ class ProjectService implements InitializingBean, ExecutionFileProducer, EventPu
...
+    protected DeleteResponse deleteProjectInternal(IRundeckProject project, IFramework framework, AuthContext authContext, String username) {
+        log.info("Starting deletion of project ${project.name} by username $username")
         notify('projectWillBeDeleted', project.name)
+        def result = new DeleteResponse(success: false)
 
         //disable scm
         scmService.removeAllPluginConfiguration(project.name)
 
-        ScheduledExecution.withTransaction { TransactionStatus status ->
+        Project.withTransaction { TransactionStatus status ->
```
sjrd218, line 1774: Would rather leave this as a reference to ScheduledExecution so that it can be refactored out of a domain class specific implementation later.

rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormProjectDataProvider.groovy
```diff
@@ -80,13 +81,14 @@ class GormProjectDataProvider implements RundeckProjectDataProvider {
...
     @Override
-    @CompileStatic(TypeCheckingMode.SKIP)
-    Collection<String> getFrameworkProjectNames() {
-        def c = Project.createCriteria().list {
-            projections {
-                property "name"
-            }
-        }
+    RdProject findByNameAndState(String name, RdProject.State state) {
+        RdProject project = projectDataService.getByName(name)
```
sjrd218, line 85: Using data services instead of criteria queries causes weird data problems. There seems to be a bug in the Gorm internals that isn't fixed, so we should stay away from data services until it's resolved.

rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormProjectDataProvider.groovy
```diff
@@ -100,16 +102,24 @@ class GormProjectDataProvider implements RundeckProjectDataProvider {
...
-    @CompileDynamic
-    String getProjectDescription(String name){
-        def c = Project.createCriteria()
-        c.get {
-            eq('name', name)
-            projections {
-                property "description"
-            }
+    Collection<String> getFrameworkProjectNames() {
+        return projectDataService.findProjectName()
```
sjrd218, line 106: Use criteria query instead.

rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormProjectDataProvider.groovy
```diff
@@ -100,16 +102,24 @@ class GormProjectDataProvider implements RundeckProjectDataProvider {
...
+    }
+
+    @Override
+    Collection<String> getFrameworkProjectNamesByState(RdProject.State state) {
+        def namelist = []
+        // If searching for enabled, we must include legacy projects with null in their field.
+        if(state == RdProject.State.ENABLED) {
+            namelist.addAll(projectDataService.findProjectNameByState(null))
```
sjrd218, line 114: Don't use a data service.

rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormProjectDataProvider.groovy
```diff
@@ -100,16 +102,24 @@ class GormProjectDataProvider implements RundeckProjectDataProvider {
...
         }
+        namelist.addAll(projectDataService.findProjectNameByState(state))
+        return namelist
+    }
 
+    @Override
+    String getProjectDescription(String name){
+        return projectDataService.findProjectDescription(name)
```
sjrd218, line 122: Don't use a data service.

sjrd218 (changes requested):

rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormProjectDataProvider.groovy
ahormazabal: This point is just data passing, so the default is indeed set by the caller who provides the RdProject object.

rundeckapp/grails-app/services/rundeck/services/ProjectService.groovy
```diff
@@ -1721,13 +1717,61 @@ class ProjectService implements InitializingBean, ExecutionFileProducer, EventPu
...
+            success: false,
+            error: "Cannot delete an already disabled project.")
+        }
+
+        if(!configurationService.getBoolean("projectService.deferredProjectDelete", false)) {
+            return deleteProjectInternal(project, framework, authContext, username)
+        }
+
+        try {
+            Project.withNewTransaction {
```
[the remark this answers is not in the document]
ahormazabal: Regretfully i need these two transactions in this point so the data is consistent between the two threads handling the process deferral. I can remove the direct reference to the gorm class but cannot remove the transaction creation.

rundeckapp/grails-app/services/rundeck/services/ProjectService.groovy
ahormazabal: This process must happen in a separate transaction so it works correctly across multiple threads, and spans many data domains so i cannot move it inside the data provider. I can remove the domain class reference though

rundeckapp/grails-app/services/rundeck/services/ProjectService.groovy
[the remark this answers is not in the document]
ahormazabal: Got it working with @Transactional instead

rundeckapp/grails-app/services/rundeck/services/ProjectService.groovy
ahormazabal: Reverted

gschueler (changes requested):
Some places the check if project is disabled should happen earlier in the method.  One API endpoint is missing a response in the error case

gschueler (approved):

PR #8284, opened 2023-04-26
RUN-1701: key storage vue conversion

chrismcg14:
The legacy code is still available behind a feature flag. By setting:
rundeck.feature.vueKeyStorage.enabled=false 
the UI will be rendered using the legacy code. that feature is enabled by default, but can be turned off


ehe-pd (approved):
Tested.

PR #8285, opened 2023-04-27
RSE-271 Fix: Successful remote job import from API but getting error in SCM API response

avelasquezr:
# RSE-271 Fix: Successful remote job import from API but getting error in SCM API response
When you import remote jobs with SCM configured using the API, you get an error in the response but the job WAS imported.

# The problem
When you consume the `performAction` method to import the remote job, you first check if you're going to delete any jobs, and if you don't delete any jobs, the response is set to `success = false`.
# The Fix:
The field `success = true` is added after successfully importing the remote job.

## Exhibits
**Before fix:**
image
**After fix:**
image

### DIFF
--- a/plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/imp/actions/ImportJobs.groovy
+++ b/plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/imp/actions/ImportJobs.groovy
@@ -134,6 +134,7 @@ class ImportJobs extends BaseAction implements GitImportAction {
                     jobsChanged.add(importResult.getJob())
                 }
                 plugin.importTracker.trackJobAtPath(importResult.job,walk.getPathString())
+                success = true
                 sb << ("Succeeded importing ${walk.getPathString()}: ${importResult}")
             }
         }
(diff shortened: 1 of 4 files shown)


plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/imp/actions/ImportJobs.groovy
```diff
@@ -134,6 +134,7 @@ class ImportJobs extends BaseAction implements GitImportAction {
                     jobsChanged.add(importResult.getJob())
                 }
                 plugin.importTracker.trackJobAtPath(importResult.job,walk.getPathString())
+                success = true
```
jtobard, line 137: This is weird, in the line 73 `success` is initialized as true, why this needs to be set as true again if nothing fails in between?

avelasquezr: In the line 84, success is set to false.

alexander-variacode (commented):
Add a test about this scenario

alexander-variacode (approved):
LGTM

PR #8291, opened 2023-04-29
RUN-1625: Select Runner For Adhoc Command

ehe-pd:
Enhancement. Let user can select Runner when execution ad-hoc command

### DIFF
--- a/rundeckapp/grails-app/views/framework/adhoc.gsp
+++ b/rundeckapp/grails-app/views/framework/adhoc.gsp
@@ -147,7 +147,6 @@ search
               </div>
             </div>
           </div>
-
           <g:ifExecutionMode active="true" project="${params.project ?: request.project}">
             <div class="card">
               <div class="card-content">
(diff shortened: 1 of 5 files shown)


rundeckapp/grails-app/views/framework/adhoc.gsp
```diff
@@ -202,19 +201,21 @@ search
...
-                <g:jsonToken id="adhoc_req_tokens" url="${request.forwardURI}"/>
-                <g:form action="adhoc" params="[project: params.project]">
                   <div data-ko-bind="nodeFilter">
                     <g:render template="nodeFiltersHidden" model="${[params: params, query: query]}"/>
                   </div>
+              <div class="card">
+                <div class="card-content vue-ui-socket">
+                  <ui-socket section="job-runner-adhoc-selector" location="top" :event-bus="EventBus" />
+                </div>
+              </div>
```
gschueler, line 218: this should be a simple ui-socket without any predefined structure around it.  I would also put it in a spot more like right below the existing "card" section for node filter results, e.g line 202 https://github.com/rundeck/rundeck/pull/8291/files#diff-0e9491e8f22c3143e9208dc704801932a77fb2d89a5a736f07e446564c7a69e9R202
so for example

```
<ui-socket section="adhoc-commands" location="node-results"></ui-socket>
```

ehe-pd: I put the component here to make it inside the form and `#runbox` so I don't need to add any logic to extract the value from the runner selector for submission.

gschueler (approved):

PR #8292, opened 2023-05-01
RUN-1623 Add doNodedispatch property to execution reference

carlosrfranco:
Include doNodedispatch property to the executionReference to get that info from executionContext


carlosrfranco:
This PR is no longer needed

PR #8295, opened 2023-05-02
RSE-410 Fix: SCM enabled if user dont have the key to SCM import.

DarwisNarvaezDev:
# RSE-410 Fix: SCM enabled if user dont have the key to SCM import
As a side effect of [this](https://github.com/rundeck/rundeck/pull/8047), the SCM gets disabled when a user don't have the rights to see the key configured in the SCM config.

# The Problem
The "disable" method gets called when the user don't have the rights to the key.

# The Fix
Prevent the method call and populate the warning object as a result, not updating the GUI with the SCM results (not calling the SCM loader)

## Implications
:heavy_check_mark: Only users with rights to the key can perform SCM functions.
:heavy_check_mark: Users without key rights can edit/view/run jobs. (original issue)

## Exhibits
Admin:
Screenshot from 2023-05-02 13-04-11

Non-Admin:
Screenshot from 2023-05-02 13-03-21


ronaveva:
it looks like it's only checking for sshKey access, but what happens if the scm was configured with password instead of key?

`storageService.hasPath(authContext, scmService.loadScmConfig(params.project, 'import')?.config?.sshPrivateKeyPath)`

I dont see this validation for scm export, why is that?

PR #8300, opened 2023-05-04
RSE-410 & RSE-418 SCM General Disconnection Issues ☹

DarwisNarvaezDev:
# RSE-410 & RSE-418 SCM General Disconnection Issues ☹
This PR Summarizes the fix of the SCM's disconnection issues: (a) SCM Disabled if user don't have rights to use the SSH configured key and (b) SCM Disabled if the user use variables in ssh or password path.

## Eurekas! 🤘
:heavy_check_mark:  Users without permissions will see a notification use the GUI as usual.
:heavy_check_mark: The SCM wont be disconnected/disabled if the user don't have permissions to the key or password.
:heavy_check_mark: Variables in key path will be expanded in key or password path.


ronaveva:
it looks like it's only checking for sshKey access, but what happens if the scm was configured with password instead of key?

`storageService.hasPath(authContext, scmService.loadScmConfig(params.project, 'import')?.config?.sshPrivateKeyPath)`

I dont see this validation for scm export, why is that?

ronaveva:
On a second thought, i think these changes should not go into the controller (not all the changes at least) but into the scm implementation.
A better approach would be to have a new method on ScmExportPlugin and ScmImportPlugin interfaces to let the implementation decide whether the user has access to perform scm actions or not

ronaveva (approved):
LGTM

PR #8308, opened 2023-05-05
RUN-1719: convert project node source edit page to Vue

ltamaster:
Convert Project Node Source Edit page to Vue

* Enabled fallback to the original page by using the feature flag `rundeck.feature.legacyUi.enabled=true` or URL parameter `?legacyUi=true`
* Fix a bug in NodesYaml generator, which would cause 500 error if the nodeset was empty
* disable "web workers" in Ace. if syntax mode is "json", causes an error in browser console.
* Add "soft wrap" control option to the Ace component, to retain same behavior from non-Vue use of Ace.


ltamaster:
tested, it worked OK

gschueler (approved):

PR #8312, opened 2023-05-09
RUN-1727: Correct loading indicator text

ehe-pd:
Bugfix. Use String literal `Loading...` to replace the malfunction i18n translation. This is a temporary fix, we should implement a i18n solution to fully translate the component.


nkrithivasan (approved):
lgtm

PR #8320, opened 2023-05-11
RUN-1537: Runner support for Key Storage plugins

ltamaster:
move StoragePluginProviderService to core

**Additional context**


core/src/main/java/com/dtolabs/rundeck/core/plugins/BaseScriptPlugin.java
```diff
@@ -263,8 +261,60 @@ public SecretBundle prepareSecretBundleWorkflowNodeStep(ExecutionContext context
...
 
     private SecretBundle generateBundle(ExecutionContext context, Map<String, Object> configuration){
+
+        List<String> listStoragePaths = generateListStoragePath(context, configuration);
+        DefaultSecretBundle bundle = new DefaultSecretBundle();
+
+        for (String propValue : listStoragePaths) {
+            Resource<ResourceMeta> r = context.getStorageTree().getResource(propValue);
+            if(r != null) {
+                ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
```
ehe-pd, line 302: Maybe we should use `try with` block to auto close the stream.

ehe-pd (approved):

PR #8322, opened 2023-05-12
RSE-326: Fix: Error responses including stacktraces

L2JE:
Ticket: https://pagerduty.atlassian.net/browse/RSE-326

#### Problem
When there's an internal error, the responses include information about the internal system (stacktraces).
image
or
image

#### Solution
Use an Interceptor that matches responses with internal server errors and render a default error page for any of them:
image

##### Previous behavior can be re-enabled for debug:
```
# In rundeck-config.properties to disable the interceptor
rundeck.feature.debug.showTracesOnResponse=false
```

### DIFF
--- a/rundeckapp/src/test/groovy/rundeck/interceptors/InternalErrorsInterceptorSpec.groovy
+++ b/rundeckapp/src/test/groovy/rundeck/interceptors/InternalErrorsInterceptorSpec.groovy
@@ -2,8 +2,7 @@ package rundeck.interceptors
 
 import grails.testing.web.interceptor.InterceptorUnitTest
 import org.apache.http.HttpStatus
-import org.grails.taglib.GrailsTagException
-import org.springframework.context.ApplicationContext
+import rundeck.services.ConfigurationService
 import spock.lang.Specification
 
 class InternalErrorsInterceptorSpec extends Specification implements InterceptorUnitTest<InternalErrorsInterceptor> {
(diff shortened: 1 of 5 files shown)


ronaveva:
1. Is it possible to use the ConfigurationService? (instead of grailsApplication.config)
2. Is it possible to log the exception? sometimes this information is actually useful to find the root cause of an issue.

L2JE:
> 1. Is it possible to use the ConfigurationService? (instead of grailsApplication.config)
> 
> 2. Is it possible to log the exception? sometimes this information is actually useful to find the root cause of an issue.

1. I thought it could impact throughput but now I will run a stress test with and without using the ConfigurationService to see if it really does before making a change.

2. Even the unhandled exceptions are logged as usual before the `Interceptor.after` method is reached. The logging will depend on the rest of the infrastructure.

L2JE:
There's no notable difference between using a controller in the interceptor to access properties and not using it:

#### 10 concurrent users accessing `/user/login` 10000 times with the default behavior (`rundeck.feature.debug.showTracesOnResponse=false`)
image

**RC1**: where properties are accessed using `grailsApplication.config.getProperty('rundeck.feature.debug.showTracesOnResponse', 'false')`

**RC2**: where properties are accessed using `this.configurationService.getBoolean('feature', 'debug', 'showTracesOnResponse', false)`

ronaveva (approved):
LGTM, just make sure to add the new property to the doc.

rundeckapp/src/test/groovy/rundeck/interceptors/InternalErrorsInterceptorSpec.groovy
```diff
@@ -0,0 +1,51 @@
...
+        defineBeans {
+            configurationService(ConfigurationService) {
+                grailsApplication = grailsApplication
+            }
+        }
+
+        withRequest(controller:"demo", action: 'index')
+
+        then:
+        !interceptor.doesMatch()
```
ronaveva, line 48: shouldn't the response have some sort of exception?

ronaveva (changes requested):
please check the 2 minor comments

ronaveva (approved):
LGTM

PR #8346, opened 2023-05-26
RUN-1766:Fix: Key Storage Making Call to List Keys Twice

chrismcg14:
the check parent dir method is the same as load keys. this removed that method and the single call to it that was resulting in duplicate calls to the backend to list the keys


ehe-pd (approved):
LGTM

PR #8347, opened 2023-05-26
RSE-430 Fix: HTTP Strict Transport Security (HSTS) Header Not Used

Jesus-Osuna-M:
Problem:
HSTS headers were not being applied to the assets path because the doFilterInternal method from the RundeckSecurityHeadersFilter class was unable to apply the headers to those paths since they were outside of its scope.

Solution:
Delegate the HSTS to the embedded Jetty server by adding a SecureRequestCustomizer

### DIFF
--- a/rundeckapp/grails-app/init/rundeckapp/init/servlet/JettyServletHstsCustomizer.groovy
+++ b/rundeckapp/grails-app/init/rundeckapp/init/servlet/JettyServletHstsCustomizer.groovy
@@ -12,7 +12,6 @@ public class JettyServletHstsCustomizer implements JettyServerCustomizer {
     long stsMaxAgeSeconds
     Boolean stsIncludeSubdomains
 
-
     /**
      * Customizes the given server by adding a secure request customizer for SSL configurations
      *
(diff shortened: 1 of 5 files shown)


rundeckapp/grails-app/init/rundeckapp/init/servlet/JettyServletHstsCustomizer.groovy
```diff
@@ -0,0 +1,45 @@
...
+        }
+    }
+
+    /**
+     * Checks if SSL is enabled in the HTTP configuration.
+     * @param httpConfig the HTTP configuration to check
+     * @return true if SSL is enabled, false otherwise
+     */
+    public static boolean checkSSL(HttpConfiguration httpConfig) {
+        int securePort = httpConfig.getSecurePort()
```
ronaveva, line 35: this could be changed to a single line

```suggestion
        return (httpConfig.getSecurePort() > 0)
```

ronaveva (approved):
Looks ok, just a single minor suggestion

rundeckapp/grails-app/init/rundeckapp/init/servlet/JettyServletHstsCustomizer.groovy
Jesus-Osuna-M: ty

DarwisNarvaezDev:
@Jesus-Osuna-M Please, add the documentation.

PR #8352, opened 2023-05-29
RSE-461 Fix: Job Step plugin that uses ${config.*} variables within are no longer resolving

Oalbarran94:
# Job Step plugin that uses ${config.*} variables within are no longer resolving

Since version 4.9.0 plugins that used ${config.*} variables, were no longer resolving configured values, replacing aforementioned values with blank strings. 

[RSE-461](https://pagerduty.atlassian.net/browse/RSE-461)

# Fix

There were a field to indicate whether the application needed to resolve the config path, being always true. Meaning we were gonna replace ${config.*} values  for blank.

We added the validation to replace that field for whatever we defined in the plugin configuration.

image

[RSE-461]: https://pagerduty.atlassian.net/browse/RSE-461?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/plugins/AbstractDescribableScriptPlugin.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/plugins/AbstractDescribableScriptPlugin.java
@@ -79,6 +79,8 @@ public abstract class AbstractDescribableScriptPlugin implements Describable {
     public static final String CONFIG_RENDERING_OPTIONS = "renderingOptions";
     public static final String SETTING_MERGE_ENVIRONMENT = "mergeEnvironment";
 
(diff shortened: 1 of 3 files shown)


ronaveva (changes requested):
What you think about making true as default value for "blankIfUnexpanded" variable?

It should also include tests in "AbstractDescribableScriptPluginSpec" class
It should also include tests for "NodeStepPluginAdapter" to avoid future issues like this one

Oalbarran94:
Done with requested tests... About making the variable true, I did not adde it since the property builder already have this defined .
image

ronaveva (approved):
LGTM

PR #8356, opened 2023-05-30
RUN-1261: Dynamic default values for properties plugins

carlosrfranco:
Plugin enhancement to define dynamic default values for plugin properties

It was added a `Map<String, Object> dynamicDefaults(
        Map<String, Object> projectAndFrameworkValues,
        Services services
    )` method to the DynamicProperties interface.

This method should return a map with string default values for all the dynamic properties 

**Additional context**
Also updates project config save/update process to enforce more validation of plugin property values


rundeckapp/grails-app/controllers/rundeck/controllers/ProjectController.groovy
```diff
@@ -2160,6 +2163,31 @@ key2=value'''
...
+                        errors << (
+                                report.errors ?
+                                        "${provider} configuration was invalid: " + report.errors :
+                                        "${provider} configuration was invalid"
+                        )
+                        return apiService.renderErrorFormat(response,[
+                                status: HttpServletResponse.SC_BAD_REQUEST,
+                                message:errors,
+                                format:respFormat
+                        ])
```
gschueler, line 2185: the response should not be rendered here, it is within ".each" loops and possibly would render multiple times, in addition the `return` statement will not exit the loops because they are not for loops but closures.  This should be deferred to outside of both loops.

rundeckapp/grails-app/controllers/rundeck/controllers/ProjectController.groovy
```diff
@@ -2425,7 +2453,33 @@ Authorization required: `configure` access for `project` resource type or `admin
...
```
gschueler, line 2477: the same problem here with returning within the closures, and rendering the error response

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/pluginConfig.vue
```diff
@@ -135,353 +135,377 @@ import PluginInfo from './PluginInfo.vue'
...
-      inputShowIcon: this.showIcon !== null ? this.showIcon : true,
-      inputShowDescription: this.showDescription !== null ? this.showDescription : true,
-      inputValues: {} as any,
-      inputSaved: {} as any,
-      inputSavedProps: (typeof this.savedProps !== 'undefined') ? this.savedProps : ['type'],
-      rkey: 'r_' + Math.floor(Math.random() * Math.floor(1024)).toString(16) + '_',
-      groupExpand:{} as {[name:string]:boolean},
-      inputLoaded: false,
-      inputContextAutocomplete: this.contextAutocomplete !== null ? this.contextAutocomplete : false
+@Component({
```
gschueler, line 155: this needs to be reverted back to the "options API" style Vue component for now, we need this style to make it easier for the vue3 migration

gschueler (changes requested):

chrismcg14 (approved):
LGTM, tested

PR #8357, opened 2023-05-30
RSE-421: fix: Added a customizer to avoid trace method on root path

alexander-variacode:
Issue:
it was a problem when it was sent a trace http method to the root path because the app response with a cookie JSESSIONID and this addressed to a security concern.

Fix:
I was added a customizer to jetty to avoid trace http method requests and the response it's a 405 - Method not allowed and without any cookie associated to.

### DIFF
--- a/rundeckapp/grails-app/init/rundeckapp/init/servlet/BanConfigCustomizer.groovy
+++ b/rundeckapp/grails-app/init/rundeckapp/init/servlet/BanConfigCustomizer.groovy
@@ -1,0 +1,54 @@
+package rundeckapp.init.servlet
+
+import org.eclipse.jetty.http.HttpMethod
+import org.eclipse.jetty.http.HttpStatus
+import org.eclipse.jetty.server.Connector
+import org.eclipse.jetty.server.HttpConfiguration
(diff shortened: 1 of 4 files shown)


rundeckapp/grails-app/init/rundeckapp/init/servlet/BanConfigCustomizer.groovy
```diff
@@ -0,0 +1,54 @@
...
+
+    BanHttpCustomizer(List<HttpMethod> methods) {
+        this.banMethods = methods
+    }
+
+    @Override
+    void customize(Connector connector, HttpConfiguration channelConfig, Request request) {
+        HttpMethod currentMethod = HttpMethod.fromString(request.method)
+        String currentPath = request.pathInfo
+        if (banMethods.contains(currentMethod) && currentPath == banPath) {
```
ahormazabal, line 49: Maybe we shouldn't check for path, but for TRACE method only. And in that case also i'd rename this class to `BanHttpMethodCustomizer` so its even more self describing.

ahormazabal (changes requested):

ahormazabal (approved):

PR #8366, opened 2023-06-02
RSE-182:Fix: SCM Options in Job Menu Not Loading

DarwisNarvaezDev:
# RSE-182 Fix: SCM Options in Job Menu Not Loading
When user don't have permission to scm key or password, the job action dropdown list is not rendering properly.

# The Problem
A request is made to the ScheduledExecutionController to retrieve from SCM integrations info about the jobs covered by them, so if the user don't have the right credentials, the request can't be performed and the controller was not smart enough to handle this.

# The Fix
Checking if the user has the credentials to perform the request to the integrations before we send them.

# Exhibits
Unauthorized - Not rendering options
Screenshot from 2023-06-02 11-18-37

Unauthorized - Rendering options
Screenshot from 2023-06-02 11-18-49


ronaveva (approved):
LGTM

PR #8370, opened 2023-06-02
RUN-1782: Support local assets in UI plugins by using asset: prefix

gschueler:

* scripts/styles with `asset:` prefix will be treated as local assets and use asset URL directly
* allows asset plugin to manage caching


kjoyce77:
Do the enterprise plugins immediately start using this or is that a future version?  Looking forward to this upgrade the lack of caching is a big issue in our environment.

PR #8393, opened 2023-06-14
RSE-156: Fix/add changeset preconditions

ronaveva:
**Problem:**
Currently there are a lot of changeset without preconditions, relaying only on "DATABASECHANGELOG" table records.

**Solution:**

1. Upgrade db migration version
2. Upgrade liquibase version
3. Add preconditions to changesets
4. Since mysql requires constraint schema name for unique constraint validation, it is necessary to provide a new property that can be accessible from the migration scripts
5. "uniqueConstraintExists" does not work with mssql, so a separated sqlcheck is required


rundeckapp/grails-app/migrations/changelog.groovy
```diff
@@ -82,6 +82,13 @@ databaseChangeLog = {
         property name: "varchar3072.type", global: "true", value: "VARCHAR(3072)", dbms: "mysql,postgresql,mssql,h2,mariadb"
         property name: "varchar3072.type", global: "true", value: "VARCHAR2(3072 CHAR)", dbms: "oracle"
 
+        if('${catalogName}' == '${catalogName}'){
```
ahormazabal, line 85: As discussed, i think it isn't worth to support a custom catalog name. I'd just configure the default Catalog name and ditch that overly confusing If clause.

ahormazabal (changes requested):

ahormazabal (approved):

PR #8395, opened 2023-06-15
RSE-406:: LDAP Fix huge number of groups

avelasquezr:
## Describe Bug
LDAP: error code 4 - Sizelimit Exceeded when you activated the nestedGroups option and you had a large number of groups.

## Before fix:
image

## After fix:
image

#### All 1014 members can be obtained..


alexander-variacode (commented):
Add the unit test of your change

avelasquezr:
> Add the unit test of your change

There already existed 3 unit tests that covered the case. The necessary mocks are added to keep them working.

alexander-variacode (commented):
Add a test on JettyCachingLdapLoginModuleTest instead of JettyCachingLdapLoginModuleTest2. The idea is to have the new features/fixes testing on Spock.

alexander-variacode (approved):
LGTM

PR #8396, opened 2023-06-16
RSE-473: Doc: Run API on local environment

L2JE:
Ticket: https://pagerduty.atlassian.net/browse/RSE-473

Added docs on how to run tests in `test` directory the same as they are run in circleci.
Also added more console output to the scripts that run tests to better understand what's happening behind.


alexander-variacode (approved):
LGTM

PR #8404, opened 2023-06-22
RSE-569 Fix: Error importing job definition from 4.5 to 4.13

Jesus-Osuna-M:
Problem:
Jobs with a "step-first" workflow strategy couldn't be imported. This workflow strategy should be taken for Rundeck as a sequential since they are marked as a synonym.
 
Solution:
Evaluate if the job to be imported has a "step-first"  workflow strategy and replace it with a sequential workflow strategy.

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
@@ -23,6 +23,8 @@ import com.dtolabs.rundeck.core.audit.ResourceTypes
 import com.dtolabs.rundeck.core.authorization.AuthContext
 import com.dtolabs.rundeck.core.authorization.UserAndRoles
 import com.dtolabs.rundeck.core.authorization.UserAndRolesAuthContext
+import com.dtolabs.rundeck.core.execution.workflow.SequentialWorkflowStrategy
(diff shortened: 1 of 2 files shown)


rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
```diff
@@ -3508,7 +3510,7 @@ class ScheduledExecutionService implements ApplicationContextAware, Initializing
         if (!workflow.commands || workflow.commands.size() < 1) {
             return null
         }
-        def name = workflow.strategy
+        def name = (workflow.strategy.equalsIgnoreCase(WorkflowExecutionItem.STEP_FIRST)) ? SequentialWorkflowStrategy.PROVIDER_NAME : workflow.strategy
```
ronaveva, line 3513:
```suggestion
        def name = (WorkflowExecutionItem.STEP_FIRST.equalsIgnoreCase(workflow.strategy)) ? SequentialWorkflowStrategy.PROVIDER_NAME : workflow.strategy
```

ronaveva (changes requested):
just a minor change needed, overall looks ok

rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
Jesus-Osuna-M: 👍

PR #8414, opened 2023-06-28
Fix typescript build error

ehe-pd:
Bugfix. Typescript build failed due to class definition issue.

Correct the typescript class.


JoseOrtiz (approved):
I think this will be very helpful

PR #8423, opened 2023-07-11
 RSE-569 Fix: Error importing job definition from 4.5

Jesus-Osuna-M:
Problem:
Jobs with a "step-first" workflow strategy couldn't be imported. This workflow strategy should be taken for Rundeck as a sequential since they are marked as a synonym.

Solution:
Evaluate if the job to be imported has a "step-first" workflow strategy and replace it with a sequential workflow strategy.


ronaveva (approved):
LGTM

PR #8425, opened 2023-07-11
RSE-705: Fixing repeating warn log messages

Oalbarran94:
# Problem

A client made an upgrade to 4.14.1 and started to see a lot of warning messages related to SCM plugin.

The problem with this warning is that appears every 20 seconds.

[RSE-705](https://pagerduty.atlassian.net/browse/RSE-705)

# Solution

Rundeck was trying to close a jgit resource which was already closed. So that generated the repeated message. We fix the problem now the message is not showing anymore. 

image

[RSE-705]: https://pagerduty.atlassian.net/browse/RSE-705?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ronaveva (changes requested):
Please check if is it possible to add test to validate the calls to the "close" method

ronaveva (approved):
LGTM

PR #8433, opened 2023-07-14
Rcloud-665: execution to spi

sjrd218:
Add SPI for Execution data.

The data SPI adds a uuid to all executions, and associates jobs to execution by a jobUuid column. This allows non-gorm implementations of execution data to associate executions and jobs.

This PR should be fairly safe as it only adds new classes and interfaces and does not significantly change any code that uses the Execution domain class directly.

This PR also adds new incubator endpoints for accessing the new execution data structure.

### DIFF
--- a/grails-rundeck-data-shared/src/main/groovy/rundeck/data/util/ExecutionDataUtil.groovy
+++ b/grails-rundeck-data-shared/src/main/groovy/rundeck/data/util/ExecutionDataUtil.groovy
@@ -1,0 +1,34 @@
+package rundeck.data.util
+
+import org.rundeck.app.data.model.v1.execution.ExecutionData
+import rundeck.data.constants.ExecutionConstants
+
+class ExecutionDataUtil {
+
+    public static String getExecutionState(ExecutionData e) {
+        return e.cancelled ? ExecutionConstants.EXECUTION_ABORTED :
+                (null == e.dateCompleted && e.status == ExecutionConstants.EXECUTION_QUEUED) ? ExecutionConstants.EXECUTION_QUEUED :
+                        null != e.dateStarted && e.dateStarted.getTime() > System.currentTimeMillis() ? ExecutionConstants.EXECUTION_SCHEDULED :
+                                (null == e.dateCompleted && e.status!=ExecutionConstants.AVERAGE_DURATION_EXCEEDED) ? ExecutionConstants.EXECUTION_RUNNING :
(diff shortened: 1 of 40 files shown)


grails-rundeck-data-shared/src/main/groovy/rundeck/data/util/ExecutionDataUtil.groovy
```diff
@@ -0,0 +1,34 @@
...
+                                        (e.status == ExecutionConstants.AVERAGE_DURATION_EXCEEDED) ? ExecutionConstants.AVERAGE_DURATION_EXCEEDED:
+                                                (e.status in ['true', 'succeeded']) ? ExecutionConstants.EXECUTION_SUCCEEDED :
+                                                        e.cancelled ? ExecutionConstants.EXECUTION_ABORTED :
+                                                                e.willRetry ? ExecutionConstants.EXECUTION_FAILED_WITH_RETRY :
+                                                                        e.timedOut ? ExecutionConstants.EXECUTION_TIMEDOUT :
+                                                                                (e.status == 'missed') ? ExecutionConstants.EXECUTION_MISSED :
+                                                                                        (e.status in ['false', 'failed']) ? ExecutionConstants.EXECUTION_FAILED :
+                                                                                                isCustomStatusString(e.status) ? ExecutionConstants.EXECUTION_STATE_OTHER : e.status.toLowerCase()
```
JoseOrtiz, line 20: Could this be written in a swith-case logic? Too many nested ternary operator doesn't look good

rundeckapp/grails-app/migrations/core/Execution.groovy
```diff
@@ -159,4 +162,61 @@ databaseChangeLog = {
...
+            column(name: "job_uuid")
+        }
+    }
+
+    changeSet(author: "rundeckdev", id: "4.x-populate-job-uuid") {
+        preConditions(onFail: "MARK_RAN") {
+            tableExists(tableName: "execution")
+            tableExists(tableName: "scheduled_execution")
+        }
+        sql("update execution set job_uuid = (select scheduled_execution.uuid from scheduled_execution where scheduled_execution.id = execution.scheduled_execution_id) where job_uuid is null")
```
ltamaster, line 200: Maybe we need to document this in the release notes because this could take a couple of minutes with a large database when the APP will be updated

sjrd218: I have removed the data population migration statements. That data is not required at this point, and if we require it in the future we can decide the best way to populate those columns at that point.

ltamaster (approved):

PR #8436, opened 2023-07-18
RSE-406::Flag to enable/disable search for more than 1000 users

avelasquezr:
Depends and continue: https://github.com/rundeck/rundeck/commit/be518d9c4863dfcff5922e1d460e358e9466cb65

This flag is created with the intention of controlling if we want to obtain all the user's groups and those that come by default. Due to the large number of groups a user can have (example: 500k groups), it may reduce performance when logging in.

## Describe New Flag Functionality
LDAP: error code 4 - Sizelimit Exceeded when you activated the nestedGroups option and you had a large number of groups.
**New**: `allGroups` flag to enable/disable if getting all groups or default. Flag is **false** by default.

## With:
Flag: `allGroups` in `false`:

image

## And:
Flag: `allGroups` in `true`:

image

#### All 1014 members can be obtained..

### DIFF
--- a/rundeckapp/src/main/groovy/com/dtolabs/rundeck/jetty/jaas/JettyCachingLdapLoginModule.java
+++ b/rundeckapp/src/main/groovy/com/dtolabs/rundeck/jetty/jaas/JettyCachingLdapLoginModule.java
@@ -262,6 +262,8 @@ public class JettyCachingLdapLoginModule extends AbstractLoginModule {
 
     protected boolean _nestedGroups;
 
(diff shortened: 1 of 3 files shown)


ronaveva (approved):
LGTM

just make sure to add the docs for the new property

fdevans:
Please update the title of this PR to be clearer what this is implementing.

avelasquezr:
Documentation: https://github.com/rundeck/docs/pull/1292

PR #8439, opened 2023-07-19
RSE-466 Fix: Throwing a precise message when the data column limit is reached

DarwisNarvaezDev:
# RSE-466 Fix: Throwing a precise message when the data column limit is reached
Using MySql as a database backend and attempting to save large resource models cause a indescribable error in the import archive process.

# The problem
An exception that doesn't say so much.

# The solution
Checking the root cause and transforming the cause message to an user-friendly string.


gschueler (changes requested):
The DbStorageService is used for more than just the Node source storage, so changing the message here is not correct.

rundeckapp/grails-app/controllers/rundeck/controllers/ProjectController.groovy
```diff
@@ -468,6 +473,24 @@ class ProjectController extends ControllerBase{
...
+     * Give a hint based on the error message in the exception thrown during the import
+     *
+     * @param Throwable - The exception thrown
+     * @return String - null if is not supported yet, a message if it is.
+     *
+     * */
+    @CompileStatic
+    private String hintErrorCause(Throwable t){
+        final SIZE_CONSTRAINT_VIOLATION_STRING = 'Data too long for column \'data\''
+        final SIZE_CONSTRAINT_VIOLATION_MESSAGE = "The content of the node resource definitions could be too big to store it on the database, try to split it into multiple files before importing it."
```
gschueler, line 486: I think it is too specific.  The same exception could be caused if the resource is too large, e.g. project.properties, not necessarily the node sources.  Perhaps saying something like "Some content was too large, this may be caused by a node source definition, ... etc"

DarwisNarvaezDev: Agree :+1:

gschueler (approved):

Issue #8335, opened 2023-05-19
Execution display unresponsive with large number of nodes


alex-umani:
**Describe the bug**
When attempting to view the execution display for a job with a large number of nodes (whether trying to access live display of a running job or reviewing the log for a completed job), the display page goes unresponsive. Waiting eventually displays the results, but this can take between 5 and 10 minutes to complete. The display time for execution logs seems to increase as the number of nodes increases. 

With jobs targeting 100-300 nodes, the display is noticeably slow but manageable. As the number of nodes increases, the display time increases. Our largest jobs target 700-800 nodes and take 5-10 minutes to display (sometimes the page completely fails to load even with this amount of time). 

This appears to be independent of log size and entirely dependent on number of nodes. We have a job that targets a single node and produces an 8.5G log. This display page loads immediately (with the proper "whale" notice). A job that targets 750 nodes but only produces a 4.5G log takes the 5-10 minutes to display properly with the page going unresponsive while loading.

No errors appear in the service log or any of the rundeck logs. There is no appreciable impact on CPU or RAM on either the application host or the database host while these pages are loading. It appears to be almost entirely a client-side javascript loop.

The behaviour has been observed consistently on our server using Chrome, Firefox, and Edge browsers.

Full details and discussion, including confirmed reproduction of the issue, is here: https://groups.google.com/g/rundeck-discuss/c/7jX68I0iUxs

**My Rundeck detail**

* Rundeck version: 4.13.0
* install type: rpm
* OS Name/version: Red Hat 7
* DB Type/version: mysql

**To Reproduce**
1. Run a job or command execution against 700+ nodes
2. Go to the job Activity list
3. Click on the execution to load its display
4. The page goes unresponsive for a long time then eventually loads

**Expected behavior**
The page should load without significant delays, certainly within the timeout notice period of standard browsers.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome, Firefox, and Edge all confirmed

**Additional context**
The issue was first noted after we upgraded from 3.3.16 to 4.10.1 (following all interim upgrade steps as recommended). Recently upgraded to 4.13.0 and the issue persists.

I've run a Firefox profiler check on it and the entire delay seems to be a nested loop issue in /assets/static/components/uisockets-f616e545f28116cec9fa19bdff6c2cc3.js although I'm unclear from looking at the code what exactly that script is doing.

MegaDrive68k:
Confirmed.

[...]

PR #8444, opened 2023-07-20
RSE-693: Fix stuck log output

ronaveva:
**Issue:** When running a job with too many nodes (like 700) to dispatch it, rundeck would get stuck when trying to show the log output page.

**Solution:** Remove non working functionality that tried to attach the runner id associated to the node (also removes the "runned id" column)


carlosrfranco (approved):
LGTM

simon-c-msc:
#8335

PR #8448, opened 2023-07-24
RSE-700::Fix empty values key value data for nixy steps local command

avelasquezr:
## Describe Bug:
Jobs using nixy that were worked on 4.8 stopped working properly on 4.14.x and 4.15.x

**My Rundeck detail**

Rundeck version: [ 4.8.0 , 4.14 ]

install type: [ deb, war ]

OS Name/version: [ ubuntu 20, ubuntu 22, macOs, windows]

DB Type/version: [ Mariadb, H2 ]

**To Reproduce**  

1- Install rundeck 4.8 and upload the yaml that I leave here. 

2- Run that job and you will see that the output is ok

3- Install rundeck 4.14 and do the same

4- You will see that the jobs failed

image

**Expected behavior**  

We expected that the Nixy plugin works fine on both versions.

## Solution
If the value obtained from the metadata, field: `blankIfUnexpandable`, is different from null, the value obtained is used only. Set the value to false or keep the default value in true.

image


Oalbarran94 (changes requested):

avelasquezr:
A new flag is added to control if we want to show an unexpected value in empty

PR #8462, opened 2023-08-02
RSE-699 Fix: Metrics Call Fails in PSQL

DarwisNarvaezDev:
# RSE-699 Fix: Metrics Call Fails in PSQL
The API calls to metrics endpoint returned a 500 when user requested it.

## The Problem
When rundeck server query the database for metrics, the results came in 'INTERVAL' datatype and the mapping was set to a "TIMESTAMP" object.

## The Solution
Creating different criterias that will be iterated until the result is extracted.

## Tested in
:heavy_check_mark: H2
:heavy_check_mark: Postgres
:heavy_check_mark: Mysql
:heavy_check_mark: Microsoft SQL Server
:heavy_check_mark: Oracle

## To QA :white_check_mark: 
1. Start the server and set it to use PostgreSql as a database backend
2. Import a project that has more than 1k of execution records, _can be used [this](https://github.com/rundeckpro/rundeck-tester/tree/master/qa-framework-test-runner/environments/loadEnvironment/projectImport/EDLProject) QAF project_
3. Create an user token allowed to use the API
4. Make this call: `curl -X GET -H "X-RunDeck-Auth-Token: <your token>" -H "Accept: application/xml" http://localhost:4440/api/40/project/<your project>/executions/metrics`
5. (Error in past versions) An error is returned
6. (Solution) A payload with metrics.

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
@@ -117,8 +117,6 @@ import java.text.DateFormat
 import java.text.MessageFormat
 import java.text.ParseException
 import java.text.SimpleDateFormat
-import java.time.Duration
-import java.time.Instant
 import java.util.concurrent.CompletableFuture
 import java.util.concurrent.TimeUnit
 import java.util.function.Consumer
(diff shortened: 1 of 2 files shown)


jtobard:
Moved to 4.17.0 since we're waiting for support feedback

alexander-variacode (approved):
LGTM

PR #8469, opened 2023-08-03
RSE-700::A new flag is added to control if we want to show an unexpected value

avelasquezr:
## Describe Bug:
Jobs using nixy that were worked on 4.8 stopped working properly on 4.14.x and 4.15.x

**My Rundeck detail**

Rundeck version: [ 4.8.0 , 4.14 ]

install type: [ deb, war ]

OS Name/version: [ ubuntu 20, ubuntu 22, macOs, windows]

DB Type/version: [ Mariadb, H2 ]

**To Reproduce**  

1- Install rundeck 4.8 and upload the yaml that I leave here. 

2- Run that job and you will see that the output is ok

3- Install rundeck 4.14 and do the same

4- You will see that the jobs failed

image

**Expected behavior**  

We expected that the Nixy plugin works fine on both versions.

## Solution
Field: `blankIfUnexpandable` is parameterized in `rundeck.plugins.nixy-local-script.blank` in `framework.properties` . Set the value to **true** or keep the default value in **false**.

framework.properties
> rundeck.plugins.nixy-local-script.blank=true

image


alexander-variacode (commented):
This fix does not resolve the main problem related to "BlankIfUnexpandable" variable or property. This will be rejected to find a better solution.

PR #8476, opened 2023-08-04
Change default node executor and file copier to use SSHJ

chrismcg14:


ltamaster (approved):
we need to document the new properties to set the default node executor and file copier.

PR #8482, opened 2023-08-08
RSE-722 Fix: Expand Timeout Variable on Scheduled Job

DarwisNarvaezDev:
# RSE-722 Fix: Expand Timeout Variable on Scheduled Job
The 'timeout' variable wont get expanded when the job is triggered by a schedule.

## The Problem
The timeout was taken literally from the DB

## The Solution
Expanding the variable programmatically as the 'user ran' process workflow.

# To QA :heavy_check_mark: 
1. Create a job with an option named "timeout" and default value of 5 and
2. A step which lasts 15 seconds then
3. Go to "other" in job definition and input '${option.timeout}' in 'Timeout' field also
4. Schedule it to run each minute and save the job

:checkered_flag: Expected behavior: the job fails in 5 seconds after the schedule triggers it.


rundeckapp/grails-app/jobs/rundeck/quartzjobs/ExecutionJob.groovy
```diff
@@ -338,13 +336,23 @@ class ExecutionJob implements InterruptableJob {
...
         }
         if (!initMap.authContext) {
             throw new RuntimeException("authContext could not be determined")
         }
         return initMap
     }
 
+    @CompileStatic
+    private long expandTimeout(RunContext initMap){
+        def jobArguments=FrameworkService.parseOptsFromString(initMap.execution?.argString)
```
alexander-variacode, line 351: This is a property comes from framework.properties?

alexander-variacode (commented):
Add more test options like -1, null, 0

alexander-variacode (approved):
LGTM

PR #8485, opened 2023-08-09
RSE-724: Keep "allGroups" in true by default.

avelasquezr:
Keep the value true to fetch all groups in LDAP. If there is an error, just change it to false.


ronaveva (approved):
LGTM

PR #8486, opened 2023-08-10
RUN-1853: Fix: Button border width causes misaligned input group on hi-dpi screen

gschueler:
1.5px border width does not match with 1px border width on inputs and input-group-addon elements, causing a 1px alignment issue

fix a misalignment in input group with both input-group-addon and input-group-btn sections.

This border width is used in only two places, the `.input-group-btn .btn` and a pagination element.


chrismcg14 (approved):
LGTM now using docker image

PR #8487, opened 2023-08-10
RSE-737 Upgrade jdbc mssql driver

Jesus-Osuna-M:
Problem: The Microsoft JDBC Driver for SQL Server in use was deprecated.

Solution: Change it to a newer version with support.


ronaveva (approved):
LGTM

PR #8489, opened 2023-08-11
RUN-1950: Grails 6 Upgrade

ahormazabal:
Upgrade to Grails 6 plus associated library version upgrades
Includes upgrade to snakeyaml 2


ahormazabal:
Moved to: https://github.com/rundeck/rundeck/pull/8648

PR #8491, opened 2023-08-14
Fix: Not showing parent job output on node exec view when having jobref error handler 

L2JE:

#### Problem

_**STR**_
1. Create a "child job" that simply runs an echo command
2. Create a "parent" Job with the following:
        Step 1 with a Script Step that prints and exit with status different than 0 (so it always is marked as failed)
        an Error handler for the Step 1 that uses a job reference of the "child job"
3. Run the parent job
4. Go to the nodes view and expand step 1 to see log output

_**Actual Behavior**_
* Only the output of the child job error handler is shown
image

_**Expected behavior**_
* The output of both the "parent job" and the "child job" error handler should be present in the step node view

#### Solution
The `ExecutionOutput` class has an `entriesbyNodeCtx` map that is only used by the nodes view to render log output, this map now groups all the log entries using the "root" of the stepCtx (if an entry stepCtx is `2/1` then this entry is added to the map with the key `2` )

image


L2JE:
This pr was replaced with this other: https://github.com/rundeck/rundeck/pull/8494

PR #8494, opened 2023-08-15
RSE-438: Fix: node step with workflow error handler missing top level step state

gschueler:

#### Problem

_**STR**_
1. Create a "child job" that simply runs an echo command
2. Create a "parent" Job with the following:
        Step 1 with a Script Step that prints and exit with status different than 0 (so it always is marked as failed)
        an Error handler for the Step 1 that uses a job reference of the "child job"
3. Run the parent job
4. Go to the nodes view and expand step 1 to see log output

_**Actual Behavior**_
* Only the output of the child job error handler is shown
image

_**Expected behavior**_
* The output of both the "parent job" and the "child job" error handler should be present in the step node view

#### Solution
Make the nodes view show all the steps that run on an expanded node at the same level. 
Example 1:
_"Given a job with 4 steps being step 3 a job reference step and step 4 a job reference step pointing to a job with 2 steps"_
image

Example 2:
* Main Job Definition
image
* Child Job Definition
image
* child-job-w-handler Definition
image

Execution:
* Step 3 and 4 fail and trigger the error handlers
image

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/workflow/StateMapping.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/workflow/StateMapping.groovy
@@ -252,7 +252,8 @@ class StateMapping {
                     hasSubworkflow: state.hasSubWorkflow(),
                     workflow: mapOf(state.subWorkflowState, ident, nodestates, allNodes)
             ]
-        } else if (state.nodeStateMap) {
+        }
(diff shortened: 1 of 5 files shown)


gschueler:
👍 lgtm

ahormazabal (approved):

PR #8504, opened 2023-08-22
RSE-286: Fix: Project Components not exported when using export to another instance

L2JE:
Ticket: https://pagerduty.atlassian.net/browse/RSE-286
QA_FW pipeline: https://app.circleci.com/pipelines/github/rundeckpro/rundeck-tester/6888/workflows/40d44d92-cdbf-42f6-b6f2-97ca4dcc8321

#### Problem
These components are not exported when using "export to another instance" feature
-  Webhook
- Tour
- Calendar
-  Schedules
- Node Wizard

This happens beacause the api endpoint to import project and the client that starts the export actions where working with a fixed set of components (It was not updated when we started adding more enterprise components)

#### Solution
Use api client from https://github.com/rundeck/rundeck-cli which is currently maintained and up to date.


rundeckapp/grails-app/controllers/rundeck/controllers/ProjectController.groovy
```diff
@@ -3062,6 +3066,7 @@ If false, do not import the project configuration (default).''', schema = @Schem
 If false, do not import the ACL Policies (default).''', schema = @Schema(implementation = Boolean.class)),
                     @Parameter(name = 'importScm ', required = false, in = ParameterIn.QUERY, description = '''If true, import SCM configuration from the archive. 
 If false, do not import the SCM configuration (default).''', schema = @Schema(implementation = Boolean.class)),
+                    @Parameter(name = 'importComponents ', required = false, in = ParameterIn.QUERY, description = '''It imports components''', schema = @Schema(implementation = String.class)),
```
ronaveva, line 3069: it looks like this new parameter is intended to do the same as "importComponents.NAME"

ronaveva (changes requested):
Please add javadoc and tests to the new methods and check comment for ProjectController.groovyL3069

rundeckapp/grails-app/services/rundeck/services/ProjectService.groovy
```diff
@@ -653,20 +658,115 @@ class ProjectService implements InitializingBean, ExecutionFileProducer, EventPu
...
+
+        Response<ProjectImportStatus> status = rundeckClient.getService().importProjectArchive(exportArchiveParams.getTargetproject(),
+                exportArchiveParams.preserveuuid?'preserve':'remove',
+                exportArchiveParams.exportExecutions,
+                exportArchiveParams.exportConfigs,
+                exportArchiveParams.exportAcls,
+                exportArchiveParams.exportScm,
+                importWebhookOpt,
+                importWebhookOpt && !Boolean.getBoolean(exportArchiveParams.exportOpts[WebhooksProjectComponent.COMPONENT_NAME]['inludeAuthTokens']),
+                Boolean.getBoolean(exportArchiveParams.exportComponents['node-wizard'] as String),
```
gschueler, line 709: `node-wizard` is not the same as the parameter "importNodesSources" for this method.  the importNodesSources param, should use the `exportArchiveParams.importNodesSources`

L2JE: Okay, now, since the node sources are always included in the archive so far (and also there's no option for this in the export to instance gui) I set the parameter to true whenever the the exportConfig parameter is true (because node source data is present in the project config)

PR #8474, opened 2023-08-03
RUN-1851 refactor PluginInfo component

smartinellibenedetti:
**Is this a bugfix, or an enhancement? Please describe.**
Enhancement for further ui conversion - no changes in functionality

**Describe the solution you've implemented**
Refactor PluginInfo component to extract the description logic to its own component, to increase reusability

PR #8510, opened 2023-08-25
RUN-1851 split PluginInfo into 2 components (vue 3 conversion)

smartinellibenedetti:
Enhancement - this is the vue 3 version of the PR #8474  - need to split this component into 2 so that can reuse it on the license conversion work to vue

Refactor PluginInfo component to extract the description logic to its own component, to increase reusability

### DIFF
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/PluginDetails.vue
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/PluginDetails.vue
@@ -1,0 +1,84 @@
+<template>
+  <div>
+    <span
+        :class="descriptionCss"
+        v-if="showDescription"
+        style="margin-left: 5px"
+    >
+      {{ shortDescription }}
+    </span>
+    <details
+        class="more-info details-reset"
+        :class="extendedCss"
+        v-if="showDescription && showExtended && extraDescription"
+    >
+      <summary>
+        More...
+        <span class="more-indicator-verbiage more-info-icon">
+          <i class="glyphicon glyphicon-chevron-right" />
+        </span>
+        <span class="less-indicator-verbiage more-info-icon">
+          <i class="glyphicon glyphicon-chevron-down" />
+        </span>
+      </summary>
+
+      {{ extraDescription }}
+    </details>
+  </div>
+</template>
+
+<script lang="ts">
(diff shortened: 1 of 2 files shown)


rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/PluginDetails.vue
```diff
@@ -0,0 +1,84 @@
+<template>
+  <div>
```
JoseOrtiz, line 2: I think you could get rid of this `div` as Vue 3 accept multi root components

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/PluginDetails.vue
```diff
@@ -0,0 +1,84 @@
...
+    >
+      {{ shortDescription }}
+    </span>
+    <details
+        class="more-info details-reset"
+        :class="extendedCss"
+        v-if="showDescription && showExtended && extraDescription"
+    >
+      <summary>
+        More...
```
JoseOrtiz, line 16: Should this text be a translatable one?

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/PluginDetails.vue
```diff
@@ -0,0 +1,84 @@
...
+        return desc.substring(0, desc.indexOf("\n"));
+      }
+      return desc;
+    },
+    extraDescription(): string | null {
+      const desc = this.description;
+      if (desc && desc.indexOf("\n") > 0) {
+        return desc.substring(desc.indexOf("\n") + 1);
+      }
+      return null;
```
JoseOrtiz, line 75: I think is best to return an empty string, right?

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/PluginDetails.vue
smartinellibenedetti: I was trying to avoid doing structural changes to reduce the risk of breaking something  (as this string wasn't translated before I split the components), but I agree

JoseOrtiz (approved):

PR #8526, opened 2023-09-01
RUN-1923: fix: short description not shown for node/workflow steps in workflow step create picker

gschueler:
If a step plugin has extended description content, the first line is not shown in the list of plugins when adding a new step to a workflow.

- [x] specify firstLineOnly to show the first line

before:
Screenshot 2023-09-01 at 11 20 40 AM

after:

Screenshot 2023-09-01 at 11 19 40 AM

RUN-1923


smartinellibenedetti (approved):

smartinellibenedetti:
image

Confirmed :+1

PR #8532, opened 2023-09-06
RSE-534 Fix: Cannot Import Jobs with Blank Notification's Config

DarwisNarvaezDev:
# RSE-534 Fix: Cannot Import Jobs with Blank Notification's Config

## The Problem
If a import was attempted with a empty notification plugin, configured in a job an error could be seen: 
image

## The Solution
Removing the validation from the xml codec, which was the functionality that was preventing the import.

### DIFF
--- a/rundeckapp/src/test/groovy/rundeck/JobsXMLCodecTests.groovy
+++ b/rundeckapp/src/test/groovy/rundeck/JobsXMLCodecTests.groovy
@@ -6942,12 +6942,8 @@ void testDecodeBasic__no_group(){
 </joblist>
 """
 
-        try {
-            def jobs = JobsXMLCodec.decode(xml0)
-            assertNotNull(jobs)
-        } catch (Exception e) {
-            // Never going to get here but...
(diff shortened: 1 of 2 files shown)


rundeckapp/src/test/groovy/rundeck/JobsXMLCodecTests.groovy
```diff
@@ -6915,7 +6915,39 @@ void testDecodeBasic__no_group(){
...
+"""
 
-
+        try {
+            def jobs = JobsXMLCodec.decode(xml0)
+            assertNotNull(jobs)
+        } catch (Exception e) {
+            // Never going to get here but...
+            assertNotEquals("onstart plugin had blank or missing 'configuration' element", e.message)
+        }
```
gschueler, line 6951: you can remove the try/catch, so if there is an exception thrown, the test will fail

DarwisNarvaezDev: will do

gschueler (approved):

PR #8545, opened 2023-09-14
RUN-1947 key storage view does not show error when saving fails

carlosrfranco:
When an error is returned from promise trying to save a new key, it is not shown on the page

Change the code to catch result and error message from promise


smartinellibenedetti (approved):
Fix confirmed 👍 

image

PR #8561, opened 2023-09-20
RUN-1975: fix: execution mode is not retained after config refresh

gschueler:
Fix: execution mode state is not retained after config refresh

Don't rely on config value, which may have been reloaded from disk.  The execution mode should persist in memory instead.


rundeckapp/src/test/groovy/rundeck/services/ConfigurationServiceSpec.groovy
```diff
@@ -43,6 +43,24 @@ class ConfigurationServiceSpec extends Specification implements ServiceUnitTest<
         !service.executionModeActive
     }
 
+    void "executionMode remains passive after config reload"() {
```
gschueler, line 46:
```suggestion
    void "executionMode remains the same after config reload"() {
```

ehe-pd (approved):

PR #8575, opened 2023-09-27
RUN-1732: fix variable expansion for DYNAMIC_FORM property values

gschueler:
Fix variable expansion in Workflow Step, Node Step and Notification plugins that use the DYNAMIC_FORM rendering option

- [x] parse JSON formatted property values that use DYNAMIC_FORM option before expansion, and re-serialize back into json values afterwards

**Additional context**
Variable expansion within a DYNAMIC_FORM plugin property value cannot be done directly since the expanded variable can break the JSON formatting


chrismcg14 (approved):
LGTM

PR #8583, opened 2023-09-29
RSE-812: Fix: not importing logs from s3logstore

L2JE:

#### Problem
Not exporting logs if they are not present in the file system even if they are in the configured remote log storage

_**STR**_
1. Have an instance with a configured remote log storage
2. Create a Project with a job and run it a couple of times
3. remove some log files from the server file system (`<RDECK_HOME>/var/logs/rundeck/<PROJECT_NAME>/job/<JOB_UUID>/logs`)
4. Export the project and import it from another instance that has the same log storage configured.

_**Actual Behavior**_
* Error message shows up: some executions (the ones removed in step **3**) couldn't be loaded
image

_**Expected behavior**_
* No error should be displayed and execution logs should be loaded afterwards from remote log storage when accessed

#### Solution
* When exporting an execution if this is not present in the FS, the server will look for a configured log storage plugin for the execution's project and generate the path for the specific execution in the configured log storage using the format: `ext:<exec-id>:<path-to-resource-in-remote-storage>`

* When importing executions that have a remote output filepath, a log warning is displayed from the ProjectService: _"Log file for imported execution "<EXEC_ID>" is not present in archive. This logs will be loaded when accessed if its path "<OUTPUT_FILEPATH>" is present in configured log storage"_

* Once the execution is accessed, the log storage plugin is initialized with the execution context that now has 3 more variables: `outputfilepath`, `isRemoteFilePath` and `execIdForLogStore`. Plugins can now use the `outputfilepath` variable to get the execution log in case the `isRemoteFilePath` is `true` (meaning the file path can not be generated using the configured path). Besides, `execIdForLogStore` stores the original execution id that can be used to validate file metadata.

* If the plugin doesn't support this behavior the server will behave as usual, not retrieving the log file (and responding with "Workflow state and log output is not available" message).

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/LogFileStorageService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/LogFileStorageService.groovy
@@ -29,10 +29,8 @@ import com.dtolabs.rundeck.core.execution.logstorage.ExecutionFileLoader
 import com.dtolabs.rundeck.core.execution.logstorage.ExecutionFileLoaderService
 import com.dtolabs.rundeck.core.execution.logstorage.ExecutionFileState
 import com.dtolabs.rundeck.core.logging.*
-import com.dtolabs.rundeck.core.plugins.configuration.PluginAdapterUtility
 import com.dtolabs.rundeck.core.plugins.configuration.PropertyResolverFactory
 import com.dtolabs.rundeck.core.plugins.configuration.PropertyScope
-import com.dtolabs.rundeck.core.plugins.configuration.Description
 import com.dtolabs.rundeck.plugins.logging.ExecutionFileStoragePlugin
 import com.dtolabs.rundeck.server.plugins.services.ExecutionFileStoragePluginProviderService
 import grails.events.EventPublisher
@@ -51,7 +49,6 @@ import org.springframework.context.event.ContextClosedEvent
 import org.springframework.core.task.AsyncListenableTaskExecutor
 import org.springframework.core.task.TaskExecutor
 import org.springframework.scheduling.TaskScheduler
-import org.springframework.transaction.annotation.Propagation
 import rundeck.Execution
 import rundeck.services.events.ExecutionCompleteEvent
 import rundeck.services.execution.ValueHolder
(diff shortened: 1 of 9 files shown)


rundeckapp/grails-app/services/rundeck/services/LogFileStorageService.groovy
```diff
@@ -1369,6 +1373,60 @@ class LogFileStorageService
...
+     * @return the value of the framework property "framework.plugin.ExecutionFileStorage.<file-storage-plugin>.path"
+     */
+    String getStorePathForProject(String projectName){
+        PropertyResolverFactory.Factory resolverFactory = frameworkService.getFrameworkPropertyResolverFactory(projectName)
+        String pluginName = getConfiguredPluginName()
+        String storePath = null
+
+        if(pluginName != null) {
+            Description description = pluginService.getPluginDescriptor(pluginName, executionFileStoragePluginProviderService).getDescription()
+            storePath = PluginAdapterUtility.mapDescribedProperties(resolverFactory.create(executionFileStoragePluginProviderService.name, description), description, PropertyScope.Instance).get('path')
```
ronaveva, line 1387: it is not right to rely on a plugin property (right now might be path but it could change and could have any name for any implementation), instead it should be part of the exposed interfaces

ronaveva (changes requested):

rundeckapp/grails-app/services/rundeck/services/ProjectService.groovy
```diff
@@ -912,8 +923,12 @@ class ProjectService implements InitializingBean, ExecutionFileProducer, EventPu
                         dir('executions/') {
                             //export executions
                             //export execution logs
+                            String remotePathTemplate = null
+                            if(execs.size() > 0)
+                                remotePathTemplate = logFileStorageService.getStorePathTemplateForExecution(execs.get(0))
+
                             execs.each { Execution exec ->
-                                exportExecution zip, exec, "execution-${exec.id}.xml"
+                                exportExecution zip, exec, "execution-${exec.id}.xml", remotePathTemplate
```
L2JE, line 931: It obtains the configured remote storage path just once with the first execution for efficiency since the configured path for log storage exists at framework level being the same for all executions and getting this property value requires instancing a LogFileStoragePlugin object.

On a project with **13656** executions (all logs in remote storage), exporting all the executions only and measuring the time since pressing the _Export Archive_ button, until the archive is available to be saved.

* AVG time for getting the expanded path for each execution: **1' 16,28"**

* AVG time for getting the remotePathTemplate just once and expand it for each execution (chosen implementation): **0' 36.53"**

ronaveva (approved):
LGTM, just make sure to test this thoroughly with different log storage providers

PR #8589, opened 2023-10-03
RUN-2005: fix log viewer progress bar

smartinellibenedetti:
With vue 3 migration, Logviewer isn't rendering the progress bar with full width.

added 'align-self: stretch' to force the flex div to occupy full width.

**Additional context**

before:
image

after:
image


ltamaster:
@smartinellibenedetti I tested the fix and this fixes the progress bar with full width, but I cannot see the Runner label where the  execution is running

smartinellibenedetti:
> @smartinellibenedetti I tested the fix and this fixes the progress bar with full width, but I cannot see the Runner label where the execution is running

Oh I misunderstood what was missing (before I wasn't even seeing the localhost), on it. Just a question though, how do I add a runner label?

smartinellibenedetti:
as we spoke on standup, will fix the runner label in a separate PR.

ltamaster (approved):

PR #8592, opened 2023-10-04
RSE-761: testdeck replacement

ronaveva:
First run to replace testdeck suite with test containers and spock based tests

* Use testcontainers to run rundeck server image
* basic API client setup via OkHttp
* Migrate some API tests from old bash scripts to Spock
* Migrate selenium tests to spock

**Additional context**
Tests can be run using testcontainers to automatically start rundeck server in a docker container, or can be executed against any URL using environment parameters, see README file

**TODO**
- [x] run API tests against docker image or running instance 
- [x] migrate selenium tests
- [x] change circleci configuration new test module instead of testdeck
- [x] improve test logging


.circleci/config.yml
```diff
@@ -450,6 +450,50 @@ jobs:
...
+      - run:
+          command: |
+            # copy war file to 
+            TOMCAT_DIR='functional-test/src/test/resources/docker/file/tomcat'
+            WAR=$(find artifacts -name '*.war')
+            cp $WAR $TOMCAT_DIR/data/rundeck-launcher.war
+      - run:
+          command: ./gradlew :functional-test:tomcatTest --no-daemon --stacktrace --info --parallel
+          environment:
+            TEST_RUNDECK_IMAGE: TOMCAT_TAG:<< parameters.tomcat-vers >>
```
gschueler, line 494: I think the tomcat tag is not set correctly
```suggestion
            TOMCAT_TAG:<< parameters.tomcat-vers >>
```

functional-test/src/test/groovy/org/rundeck/util/container/RdDockerContainer.groovy
```diff
@@ -0,0 +1,55 @@
...
+                .forStatusCodeMatching(it -> it >= 200 && it < 500 && it != 404)
+                .withStartupTimeout(Duration.ofMinutes(5)))
+        if(PORT_BINDINGS != null && !PORT_BINDINGS.isEmpty()){
+            setPortBindings(ImmutableList.of(PORT_BINDINGS))
+        }
+        if(CUSTOM_ENVS != null && !CUSTOM_ENVS.isEmpty()){
+            CUSTOM_ENVS.split(',').each(){
+                withEnv(it.split(":")[0], it.split(":")[1])
+            }
+        }
```
gschueler, line 40: may want to use a log consumer
```
        withLogConsumer(new Slf4jLogConsumer(log))
```

.circleci/config.yml
```diff
@@ -491,7 +491,7 @@ jobs:
       - run:
           command: ./gradlew :functional-test:tomcatTest --no-daemon --stacktrace --info --parallel
           environment:
-            TEST_RUNDECK_IMAGE: TOMCAT_TAG:<< parameters.tomcat-vers >>
+            TEST_RUNDECK_IMAGE: << parameters.tomcat-vers >>
```
gschueler, line 494: should use an env var that makes sense, it is not a rundeck image, it is a tomcat tag. I think it makes more sense as TOMCAT_TAG

ronaveva: i think, instead of an specific name we should only use a generic one, this way we don't have multiple variables with the same purpose

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/BasicLoginSpec.groovy
```diff
@@ -19,7 +19,6 @@ class BasicLoginSpec extends SeleniumBase {
 
         then:
             currentUrl.contains("/menu/home")
-            pageSource =~ ~/\d+ Projects?/
     }
```
gschueler, line 22: why was this removed?

gschueler:
there is a fix for CI packaging tests in main, will need to update

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/BasicLoginSpec.groovy
ronaveva: it's better to validate against elements since pageSource gets evaluated before all of the javascript/jquery/ajax, and it wasn't working

gschueler: ah, ok.  I was trying to replicate the original test here https://github.com/rundeck/rundeck/blob/main/test/selenium/src/__tests__/selenium/login.test.ts#L31 but I was actually not doing it correctly anyway.  I will update it

gschueler (approved):
will need to wait until #8610 is merged

gschueler (changes requested):
Circle config needs to be updated to match new design

gschueler (approved):

PR #8595, opened 2023-10-05
RUN-2008: runner name

smartinellibenedetti:
Fix for runner name (need a second PR on pro side);

Originally had flattened the settings to solve a console error, but this caused another issue that was preventing the runner name from showing. I've undone the change and saw that the console error isn't showing up anymore. 

Will make a second PR on pro side to ensure that the runner name will show up

### DIFF
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/execution-log/logViewer.vue
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/execution-log/logViewer.vue
@@ -49,9 +49,11 @@
           <input type="checkbox" v-model="settings.stats" id="logview_stats">
           <label for="logview_stats">Display Stats</label>
         </div>
-        <ui-socket section="execution-log-viewer" location="settings"
-                   :event-bus="eventBus"
-        />
+        <div class="vue-ui-socket">
+          <ui-socket section="execution-log-viewer" location="settings"
+                     :event-bus="eventBus"
+          />
+        </div>
       </form>
     </RdDrawer>
     <div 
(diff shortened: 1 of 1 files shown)


rundeckapp/grails-spa/packages/ui-trellis/src/library/components/execution-log/logViewer.vue
```diff
@@ -49,9 +49,11 @@
...
         </div>
-        <ui-socket section="execution-log-viewer" location="settings"
-                   :event-bus="eventBus"
-        />
+        <div class="vue-ui-socket">
+          <ui-socket section="execution-log-viewer" location="settings"
+                     :event-bus="eventBus"
+          />
+        </div>
       </form>
```
gschueler, line 57: why this change, did it affect things? i don't think we should wrap ui-socket like that when it is already inside a Vue component

smartinellibenedetti: Just updated it to match the format recommended for the vue 3 migration (ui-socket inside of another element with vue-ui-socket class) but it doesn't make any difference how things are rendered in practice

smartinellibenedetti: I can undo it if you'd like

gschueler: yes, that format is only needed when embedding ui-socket into the page outside of a Vue component

gschueler (approved):

PR #8608, opened 2023-10-10
RCLOUD-1070: Update option processing

sjrd218:
Extract option parsing to a shared gradle project.
Convert some places that process options to use the JobData and OptionData models.


grails-rundeck-data-shared/src/main/groovy/rundeck/data/util/OptionsParserUtil.groovy
```diff
@@ -0,0 +1,41 @@
...
+        def String key=null
+        for(int i=0;i<tokens.length;i++){
+            if (key) {
+                optsmap[key] = tokens[i]
+                key = null
+            }else if (tokens[i].startsWith("-") && tokens[i].length()>1){
+                key=tokens[i].substring(1)
+            }
+        }
+        if(key){
```
jbrookspd, line 36: Is this needed?

jbrookspd (approved):

PR #8624, opened 2023-10-18
RUN-2019: project config plugin validation issue from API request

carlosrfranco:
When an individual project config is updated the config property is validating isolated and it is not considering the others current settings. It cause validation error for required properties

Now all the current settings is loaded before validation


gschueler (changes requested):
CI fails, tests need to be updated

gschueler (changes requested):
issue: the test also needs to test the case when validation fails, that is what originally caused the problem

gschueler (approved):

Issue #7875, opened 2022-08-23
Regression - job definition load does not preserve option value ordering


ajxb:
**Describe the bug**

Rundeck 4.4.0 has reintroduced a bug that had been fixed in Rundeck 3.4.2.  Defined option values automatically get sorted for a job when it is imported via the CLI tool or the Rundeck GUI.

**My Rundeck detail**

* Rundeck version: 4.4.0
* install type: rpm
* OS Name/version: CentOS7
* DB Type/version: mysql

**To Reproduce**

1. Create a new job in RunDeck with a single option containing allowed values "E,D,C,B,A", inspecting the job shows the ordering for the option as defined:
   ![image](https://user-images.githubusercontent.com/1756536/186180722-17c023be-9f63-41f3-aa66-33d6f7feeac6.png)
2. Export the job using the *Download job definition in YAML* option from the *Action* menu in the Rundeck GUI.  The resulting yaml for the option is in the defined order:
   ```yaml
     options:
     - enforced: true
       name: Ordering
       required: true
       value: C
       values:
       - E
       - D
       - C
       - B
       - A
       valuesListDelimiter: ','
   ```
3. Import the job using the *Upload Definition* option from the *Job Actions* menu in the Rundeck GUI, inspecting the job following import shows the ordering for the option is now alphabetic:
   ![image](https://user-images.githubusercontent.com/1756536/186182403-b636314a-8c40-4e86-bacc-add4d748882d.png)

**Expected behavior**

After importing the job I expect the Allowed Values to have the order originally defined and correctly captured in the yaml file. I.e. E,D,C,B,A

**Screenshots**

See above.

**Desktop (please complete the following information):**
 - OS: Kubuntu 22.04
 - Browser: Brave
 - Version: v1.42.97 (Aug 17th, 2022) based on Chromium to 104.0.5112.102

**Additional context**

This problem had been resolved via PR https://github.com/rundeck/rundeck/pull/7122 (although I realise now that this caused a separate issue https://github.com/rundeck/rundeck/issues/7773), it looks like changes to the code in this area have reverted the fix - see this commit https://github.com/rundeck/rundeck/commit/c2c2b3ff0e2e29c91813e3e1d28b7286b57d092b#diff-db1fbbfd262db697fd32faf5f406bc72e65c9ce9a89d9baa4503d698af92ebcbR264-R271.

I believe that the problem occurs as a result of using a TreeSet which automatically sorts values.

MegaDrive68k:
[...]

Issue #7773, opened 2022-06-16
issue with uploading YAML Job


KieranHolton:
**Describe the bug**
I have recently created a new Rundeck host with version 4.2 since using the new host it seems I'm having issues with uploading YAML templates for jobs.

The particular issues appears to be with options with multiple values. 
example the following job fails to upload;

```
- defaultTab: nodes
  description: testing multiple values
  executionEnabled: true
  id: 9525e0b4-065b-4d26-9399-a59c9d21a476
  loglevel: INFO
  name: MultiValue Test
  nodeFilterEditable: false
  options:
  - label: Multi 1
    name: multi1
    value: '1'
    values: '1,2,3,4,5,6,7,8'
    valuesListDelimiter: ','
  plugins:
    ExecutionLifecycle: null
  scheduleEnabled: true
  sequence:
    commands:
    - description: echo "blah"
      exec: blah
    keepgoing: false
    strategy: node-first
  uuid: 9525e0b4-065b-4d26-9399-a59c9d21a476

````

this returns error: 

`org.rundeck.app.components.jobs.JobDefinitionException: Unable to create Job: Could not find matching constructor for: java.util.ArrayList(String)`

In my previous instance of rundeck (VERSION : 3.4.0-20210614) this works fine.

I appreciate that if i change the values to be 1 per line this works fine, however we template this automatically as part of a deployment which makes it easier to to have them as a comma separated string

**My Rundeck detail**

* Rundeck version: 4.2.0-20220509
* install type: rpm
* OS Name/version: centos08
* DB Type/version: mysql

**To Reproduce**
1.go to project
2. select Job actions/Upload Definition
3. select YAML File
4.upload

**Expected behavior**
Job should upload with no errors

**Screenshots**
![image](https://user-images.githubusercontent.com/51415463/174052902-34cd6455-5328-4aa2-af05-70b5fd5d592d.png)

**Desktop (please complete the following information):**
 - OS: Ubuntu 20.04.4 LTS
 - Browser Chrome
 - Version 101.0.4951.64

PR #8659, opened 2023-11-02
Fix import of option values to retain order specified in source file

ajxb:

A potential bugfix for #7875.

Use an `ArrayList<String>` in place of a `TreeSet` when casting `data.values` into a suitable Collection for `opt.optionValues`.  A TreeSet automatically sorts the values passed, this is undesired behaviour that causes option values to get automatically sorted on import regardless of the `sortValues` setting.

I've also simplified the logic by combining the cast and assignment of opt.optionValues - this removes the need for the second cast operation found on line [#L293](https://github.com/ajxb/rundeck/blob/main/rundeckapp/grails-app/domain/rundeck/Option.groovy#L293).

N/A

**Additional context**

This is a reimplementation of PR #7122  that was accepted into the codebase and subsequently reverted.

I realise that my original implementation broke the import when the values parameter was a comma separated list (rather than an array) as was reported in #7773 - the reimplementation represented by this PR ensures that both use cases work as expected.

I've tested the import in the following situations:
1. values is a comma separated string
2. values is an array
3. sortValues is set to false
4. sortValues is set to true


fdevans:
@ajxb - This is still in "Draft" mode.  Are you still working on this?  When it's ready for review please update the branch from 'main', then mark it as ready for review.  Thank you.

PR #8662, opened 2023-11-02
RSE-863: Fix: API listing running executions fail using wildcards

alexander-variacode:
❌ **Problem**
A client performed an upgrade to 4.15 and now the API to listing the executions using the wildcard fails. Only works if you use the name of the project instead the wildcard.

✅ **Fix**
Add an exclusion condition to the interceptor for project validation, specifically to skip validation when a wildcard is used, ensuring that only explicit project validation is performed.


ronaveva (approved):
LGTM

PR #8665, opened 2023-11-06
Grails 6.1 upgrade

ahormazabal:


ahormazabal:
Implemented in https://github.com/rundeck/rundeck/pull/9068

PR #8685, opened 2023-11-15
RSE-886: Fix: Job option description too long

alexander-variacode:
### ❌ Issue:
Job option description is affected by an overflow due to long texts without spaces. This behaviour occurs on Job page and Edit Job page.

### ✅ Fix:
Adding css style on these pages to avoid the overflow from its DOM element.


DarwisNarvaezDev (approved):
LGTM

PR #8687, opened 2023-11-17
RUN-1852: Convert Edit Readme/MOTD Page to Vue

chrismcg14:


rundeckapp/grails-spa/packages/ui-trellis/src/app/components/readme-motd/EditProjectFile.vue
```diff
@@ -0,0 +1,161 @@
...
+  },
+  mounted() {
+    this.getFileText()
+  },
+  methods: {
+    async saveProjectFile() {
+      if(this.fileText === ''){
+        this.fileText = '#This is a test'
+      }
+      const resp = await rundeckClient.sendRequest({
```
smartinellibenedetti, line 101: nitpick (non-blocking): it would be good to move the sendRequest calls to its own service file, the benefits are:

- better separation of concerns, because then the saveProjectFile method focuses on what is relevant for the component, which is handling the error or success instead of the status of the response;
- makes it easier in the future to add unit tests;

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/readme-motd/EditProjectFile.vue
```diff
@@ -0,0 +1,161 @@
...
+      markdownSectionOpen: false,
+      errorMsg: ''
+    };
+  },
+  props: {
+    filename: '',
+    displayConfig: [],
+    project: '',
+    authAdmin: false,
+  },
```
smartinellibenedetti, line 87: issue: ideally for props we use the object syntax and type them, ie:

```
filename: {
   type: string,
   default: ''
}
```

The benefit of this approach is that you can add validation or mark a prop as required. Could you please change them for this format?

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/readme-motd/EditProjectFile.vue
```diff
@@ -0,0 +1,161 @@
...
+        method: 'GET'
+      }).then(response => {
+        if(response.status === 200){
+          this.fileText = response.parsedBody.contents
+        }
+        else{
+          if(response.status !== 200){
+            this.errorMsg = response.parsedBody.message
+            this.notifyError(this.errorMsg)
+          }
```
smartinellibenedetti, line 155: issue: When adding a message for the first time, this API is returning 404 as the file still doesn't exist. You might want to tweak this check to not output an error in this case.

smartinellibenedetti (commented):
Overall is looking great, there are just 2 minor issues. Would be nice to address the nitpick but as mentioned in the comment, it's non-blocking.

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/readme-motd/editProjectFileService.ts
```diff
@@ -0,0 +1,69 @@
...
+}
+
+function notifyWarning(message: string) {
+    Notification.notify({
+        type: "info",
+        title: "Warning",
+        content: message,
+        duration: 0
+    });
+}
```
smartinellibenedetti, line 30: The idea is not to move the notify calls to the service file, just the api call itself. So all methods in the service file return a response or an error.
Because then if one day other components need to use these API calls but do something else based on the response, it's possible to reuse these methods in the service

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/readme-motd/editProjectFileService.ts
```diff
@@ -0,0 +1,69 @@
...
+    });
+
+    if (resp.status === 200) {
+        notifySuccess("Success", "Saved Project File " + filename);
+    } else {
+        notifyError(resp.parsedBody.message);
+    }
+}
+export async function getFileText(project: string, filename: string) {
+    try {
```
smartinellibenedetti, line 48: would tweak this code to something more similar to this:

```
export async function getFileText () {
  return rundeckClient.sendRequest([data here]).then(response => {
        if(response.status === 200){

          return [response data or parsedBody]
        }
        else{
          // handle the error logic here, but instead of calling notifyWarning or notifyError, 
          // throwing an object that has the message and any other properties that might be helpful
          // for the component to decide how to handle this error
        }
}
```

smartinellibenedetti (commented):
As we spoke, just need to tweak a little bit the code on services as some of it should still stay in the component and it's g2g, the other changes are perfect!

PR #8690, opened 2023-11-20
RSE-892: Fix: Node wizard and resourceyaml sources cannot store +4K lines of node entries 

L2JE:
Ticket: https://pagerduty.atlassian.net/browse/RSE-744

#### Problem:

In 3.3.x it was possible to add a resource yaml node resource with more than 4000 lines and since 3.4.0, when db changes were moved to be managed with liquibase, servers with mysql started showing the following error message in the nodes page after saving a node source like so:

And the service log showing a stacktrace with this caused-by message:

`Caused by: org.rundeck.storage.api.StorageException: Failed to save content at path projects/P/etc/project.properties: validation error: Data too long for column 'data' at row 1`

This happens because after migrating to liquibase, the type for data was set to blob while in previous versions was longblob. A fragment of the Storage table changeset:

```
column( name: "data", type: '${bytearray.type}')
property name: "bytearray.type", value: "blob", dbms: "mysql,oracle,mariadb,h2"
```

#### Expected Outcome:
The node source should be added and the nodes loaded the same as in versions < 3.4

#### Proposed Solution:

* Add a new changeset for the Storage table to migrate the data column from blob to longblob as in previous versions.

* Add Storage exception errors to the gui error response payload

### DIFF
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/FrameworkController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/FrameworkController.groovy
@@ -70,6 +70,7 @@ import grails.converters.JSON
 import grails.converters.XML
 import grails.web.servlet.mvc.GrailsParameterMap
 import org.rundeck.core.projects.ProjectPluginListConfigurable
+import org.rundeck.storage.api.StorageException
 import org.springframework.context.ApplicationContext
 import org.springframework.context.ApplicationContextAware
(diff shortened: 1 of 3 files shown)


Jesus-Osuna-M (changes requested):
If possible try to add a unit test to these changes.

Jesus-Osuna-M (approved):
LGTM

PR #8695, opened 2023-11-21
RCLOUD-1066: Transactional write to database for AuthToken and Webhook

JoseOrtiz:
Enhancement for threaded write to the database
When using the gorm approach as a secondary write there was an issue with the session. By defining the writes as transactional the session will be respected

### DIFF
--- a/grails-webhooks/grails-app/services/webhooks/WebhookService.groovy
+++ b/grails-webhooks/grails-app/services/webhooks/WebhookService.groovy
@@ -50,7 +50,7 @@ class WebhookService {
     static final String TOPIC_RECENT_EVENTS = 'webhook:events:recent'
     static final String TOPIC_DEBUG_EVENTS = 'webhook:events:debug'
 
-    WebhookDataProvider webhookDataProvider;
+    WebhookDataProvider webhookDataProvider
 
     @Autowired
     EventStoreService eventStoreService
@@ -118,7 +118,7 @@ class WebhookService {
                 subsystem: "webhooks",
                 topic: "${eventTopic}:${webhook.uuid}"
         ))
-        return queryResult.totalCount;
+        return queryResult.totalCount
     }
 
     @PackageScope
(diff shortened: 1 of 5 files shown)


rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormWebhookDataProvider.groovy
```diff
@@ -46,65 +46,71 @@ class GormWebhookDataProvider implements WebhookDataProvider {
...
     @Override
     Integer countByNameAndProject(String name, String project) {
-        return Webhook.countByNameAndProject(name, project);
+        return Webhook.countByNameAndProject(name, project)
     }
 
     @Override
     void delete(Long id){
-        webhookDataService.delete(id);
+        Webhook.withTransaction {
```
jbrookspd, line 79: Shouldn't these be wrapped in a try/catch that throws DataAccessException like AuthToken above?

JoseOrtiz: I'll add those cases

grails-webhooks/grails-app/services/webhooks/WebhookService.groovy
```diff
@@ -183,7 +183,7 @@ class WebhookService {
         SaveWebhookRequest saveWebhookRequest = mapperSaveRequest(hookData)
         boolean shouldUpdate = false
         if(saveWebhookRequest.id) {
-            hook = webhookDataProvider.getWebhook(saveWebhookRequest.id)
+            hook = webhookDataProvider.getWebhookByUuid(saveWebhookRequest.uuid)
```
jbrookspd, line 186: Why is this changing? A lookup by id should be much faster.

JoseOrtiz: Because when fetching from both rba and gorm the id could be different, the only thing that it's constant is the uuid

jbrookspd (approved):

PR #8699, opened 2023-11-23
RSE- 897 Follow button not working on log view page

Jesus-Osuna-M:
Problem: After changes made to the Vue components in Rundeck 4.17.0, the "follow" button, which allows users to track outputs in the activity log output page, stopped working. Clicking on this button did not cause the page to follow the log output.

Expected Outcome: Restore the functionality of the "follow" button on the log output page.

Proposed Solution: In the log viewer.vue file, when instantiating the LogNodeChunk component, add the missing value for the "follow" button. Remove the @resize action from DynamicScroller, as it only activates the first time the component size is modified. In DynamicScrollerItem, add the emit-resize property that generates an event each time the item's size is modified. Add the @resize DynamicScrollerItem to receive the generated events and trigger the scrollToLine() method. This method evaluates the information of the "follow" button, and if it is true, the autoscroll action will be generated.


DarwisNarvaezDev:
We need tests in new test deck, maybe selenium tests will be appropriate.

DarwisNarvaezDev (changes requested):
We need tests for this, maybe new selenium testdeck tests will be appropriate.

DarwisNarvaezDev (approved):
LGTM

PR #8558, opened 2023-09-19
RUN-1346: remove db storage for "node filters"

gschueler:
**Is this a bugfix, or an enhancement? Please describe.**
Removes the db-backed storage for stored Node filters, replaces it with browser local storage

**Describe the solution you've implemented**

* Removes the db-backed storage for Node filters
* Replaces GSP UI for node filters with the Vue component
    * Commands page
    * Nodes page
    * Job edit > workflow edit > job reference
    * Job show page
* changes implementation for Node filters to use browser local storage

PR #8705, opened 2023-11-24
RUN-2069: Update default node filter indicator

smartinellibenedetti:
Follow up of two tickets, #8558 and #8688 , adding the possibility to save filters as default.

How to test:
- on nodes page, type the filter to search nodes and save it as a new filter (by clicking on the arrow on the right side of the word 'nodes');
- upon saving, you'll see under 'filters' tab in the node summary, ie:
Screenshot 2023-11-24 at 1 20 08 PM
- then click on the small arrow on the right side of the name and mark it as 'default filter'
** Important: at this point there's no visual feedback that it worked, but you can see that it's the new default filter by either looking at localstorage OR by refreshing the page OR by clicking again the same menu on the right side of the filter where you've clicked before**

### DIFF
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/resources/NodeCard.vue
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/resources/NodeCard.vue
@@ -139,55 +139,11 @@
               <h5 class="column-title text-uppercase text-strong">
                 {{ $t("filters") }}
               </h5>
-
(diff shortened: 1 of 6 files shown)


gschueler (approved):
the default filter is now loaded on page by default...
however, one small tweak would be: show the filter as the "selected" filter in the filter list.  i.e. if you select the filter manually, it shows the name in the dropdown in green.  however if you just load the page with a default filter, the filter is applied, but the name is not shown in green

PR #8717, opened 2023-11-29
RUN-2076: Add better error handling on node services

smartinellibenedetti:

image

With the vue conversion, the services were returning axiosErrors objects instead of the errors defined in the nodeServices.ts file. 

3 things:

- add validateStatus to axios requests in order to prevent the promise from being rejected (which was the reason why the axiosErrors were being returned);
- add try-catch to still allow server errors to be returned if they have a code bigger than 403;
-  move the HTML template for the error message out of the results tab, this way error can be seen no matter if user is in the results tab or no;

**Additional context**

After changes:
image


chrismcg14 (approved):
LGTM!

PR #8722, opened 2023-12-01
fix test inconsistency

ronaveva:
Instead of joining the strings it validates one by one.


carlosrfranco (changes requested):
Actually, the quote issue doesn´t seem to be related to the join method in the array

PR #8730, opened 2023-12-04
RSE-755 Fix: SCM Disabled in Cluster while Job Plugins Metadata Retrieval

DarwisNarvaezDev:
# RSE-755 Fix: SCM Disabled in Cluster while Job Plugins Metadata Retrieval
A exception is thrown in cluster when SCM try to get jobs plugin metadata (of no jobs).

## The Problem
No jobs to extract metadata from, so GORM complains with the wrong identifier exception.

## The Fix
Handling exception, and eventually the cluster will rearrange the metadata.


Jesus-Osuna-M (approved):
LGTM

PR #8731, opened 2023-12-05
RUN-2084: job import with JSON format with unsupported version should return expected error

gschueler:

Fix: if the Job import API is called with JSON job data, but incorrect API version, it returns an unexpected error

Return 400 error with "api.error.api-version.unsupported" error code


chrismcg14 (approved):
LGTM!

Issue #5205, opened 2019-09-04
Ability to Disable AutoComplete on password field of Rundeck login page


markp80:
Our server running Rundeck 3.0.24 is scanned by Qualys Vulnerability scanning and has detected the following vulnerability. Plesse could we have the ability to disable AutoComplete at the application level.

AutoComplete Attribute Not Disabled for Password in Form Based Authentication

THREAT: The Web server allows form based authentication without disabling the AutoComplete feature for the password field. Autocomplete should be turned off for any input that takes sensitive information such as credit card number, CVV2/CVC code, U.S. social security number, etc. 

IMPACT: If the browser is used in a shared computing environment where more than one person may use the browser, then "autocomplete" values may be retrieved or submitted by an unauthorized user. 

SOLUTION: Contact the vendor to have the AutoComplete attribute disabled for the password field in all forms. The AutoComplete attribute should also be disabled for the user ID field. Developers can add the following attribute to the form or input element: autocomplete="off" This attribute prevents the browser from prompting the user to save the populated form values for later reuse. Most browsers no longer honor autocomplete="off" for password input fields. These browsers include Chrome, Firefox, Microsoft Edge, IE, Opera However, there is still an ability to turn off autocomplete through the browser and that is recommended for a shared computing environment. Since the ability to turn autocomplete off for password inputs fields is controlled by the user it is highly recommended for application to enforce strong password rules.

Mark

runephilosof:
I would say Qualys is wrong in this recommendation. Which is an opinion shared by the browser vendors, since they don't honor the autocomplete="off".
Unique passwords and Password Managers is the best practice.
You shouldn't turn off autocompletion just because some users might use a shared os account (and have the browser save the password).
A shared account (such as a library computer) should instead have that browser feature turned off.

It would probably be a good idea to contact Qualys for arguments.

PR #8744, opened 2023-12-07
RUN-2127: Optimize HTML attributes for password manager autofill

EnguerrandDeclercq:
This change allows password managers (such as Dashlane) to autofill username and password (this is mostly a UX enhancement proposal).

**Additional context**

### DIFF
--- a/rundeckapp/grails-app/views/user/login.gsp
+++ b/rundeckapp/grails-app/views/user/login.gsp
@@ -110,7 +110,7 @@
               </div>
             </g:if>
             <div class="col-md-4 col-sm-6 col-md-offset-4 col-sm-offset-3">
-              <form action="${g.createLink(uri:"/j_security_check")}" method="post" class="form " role="form" onsubmit="return onLoginClicked()">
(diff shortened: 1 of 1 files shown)


fdevans:
@EnguerrandDeclercq - Can you help update this against the 'main' branch?  I've added it to our review queue.

gschueler:
`data-form-type` is not a standard mechanism for declaring login form types.  Possibly the `autocomplete` attribute should be used instead.

gschueler:
https://www.chromium.org/developers/design-documents/form-styles-that-chromium-understands/

gschueler:
fyi: the password `autocomplete` attribute was set to `off` for this reason https://github.com/rundeck/rundeck/issues/5205

gschueler:
I believe we should adopt the standard values for `autocomplete` attributes as described here https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill

If you would like you can update your PR to do that, e.g. `autocomplete="current-password"` and `autocomplete="username"`. 

I think we should avoid custom attributes for specific password managers if possible, e.g. Dashlane or 1Password, in favor of the correct semantics for the autocomplete attribute.

PR #8752, opened 2023-12-13
RUN-2100: Authenticate pipeline pagerduty

smartinellibenedetti:

Add registry information to allow circleci to connect with cloudsmith and install continuum.tokens package. 

**Important:**

Before running the installUiTrellis task locally, it's necessary to set an env variable for CLOUDSMITH_NPM_TOKEN or add it's value to the .npmrc file (replacing `${CLOUDSMIT_NPM_TOKEN}`)

**Additional context**
This is needed for future UI work.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -616,9 +616,6 @@ workflows:
     jobs:
       - build-rundeck:
           name: Build Release
-          context:
-            - Cloudsmith
-            - slack-secrets
           <<: *slack-defaults
       - packaging-publish:
           name: Package and Release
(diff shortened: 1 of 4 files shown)


smartinellibenedetti:
PagerDuty’s npm registry in Cloudsmith is private. While it’s completely doable to update the pipeline to authenticate as seen on the PR attached to the ticket, it turns out that this would prevent developers outside the company from being able to build and develop the open-source part of the project.

Upon discussion with the FE/AST team, it was decided that changes would be made in the publishing process of the continuum.tokens package (FEAST-609) to make it public and avoid the hassle of having to authenticate in Cloudsmith whenever new updates are released.

PR #8758, opened 2023-12-13
RUN-2109: Fix node step plugin filter

carlosrfranco:
Node step plugin filter is not working for the highlighted plugins

Add highlighted plugins to the stepDescriptionsAll list


gschueler:
it seems to correctly include the Local and Command steps in the search.. but for some reason it also includes the "Script" steps in the results?

gschueler (approved):

PR #8765, opened 2023-12-15
RCLOUD-1118: Project and Storage usage of unique value for fetch

JoseOrtiz:
Enhancement. 
As we moved away from the gorm implementation, a way to fetch project and storage withouth the id is needed. 
There is also a need to create transactions when writing to the database
For project `name` is a unique value, for storage is a combination of `namespace`, `dir` and `name`. So every use of `id` was changed for the entire entity so we have flexiblity for fetching.
`withTransaction` blocks where added when writing to the database


sjrd218 (changes requested):
When I check out this branch and create a new project, then try to save either a readme or message of the day it fails with a "no content" error message.

I don't see the same behavior on main.

sjrd218 (approved):

PR #8768, opened 2023-12-19
RSE-926: Fix: Active Connection Required, StackOverflow false upload error (s3) on LogFileStorage Service

L2JE:
Ticket: https://pagerduty.atlassian.net/browse/RSE-926

#### Problem:
When the server is under high load, some log storage requests can fail and if they reach the maximum store attempts, the following error is thrown:
`Exception in thread "LogFileStorageTask695" java.lang.IllegalArgumentException: Active Connection is required`

And even if clicking on re-queue buttons through GUI, the maximum of attempts is reached and the storage requests that initially failed keeps failing over and over 

Customers receive lots of messages like this due to [an internal issue with the grails-data-mapping library](https://github.com/grails/grails-data-mapping/issues/1234), which handles the DomainObject.withNewSession calls (which are used for managing the requests). 

Besides this, when the user clicks on a `requeue` button for a specific storage request, the server requeues the storage request and after that, it tries to build the response `json` content using a `request` object instead of a `serializable`, resulting `500` error on the browser and a `StackOverflow` error on the server.

#### Expected Outcome:
The server should not throw this `Active Connection is Required` error and the server should recover the failed log storage requests when the user clicks on the `requeue` or `requeue all` button on the Log Storage page.

#### Proposed Solution:

* This issue Active Connection is required can be avoided by just removing the `@Transactional` annotation and use `DomainObject.withNewSession` where needed in the `LogFileStorageService` (`dequeueIncompleteLogStorage` and `runStorageRequest` methods). This was taken from suggested changes by @sjrd218 on [this PR](https://github.com/rundeck/rundeck/pull/8519).

* Some log storage requests may fail because the local file doesn’t exist on the file system but was successfully stored in the remote storage, after this, when request is retried it fails because it tries to upload a not existing file. This is fixed by first checking if the file exists on the FS and the log stored in the remote storage for complete file logs.

* For the `StackOverflow` error the solution is to call the `exportRequestMap(storageRequest)` outside the contents closure, providing it with a previously calculated `serializable` object.

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/logging/StorageFile.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/logging/StorageFile.java
@@ -47,4 +47,11 @@ public interface StorageFile {
      * @return true if the file is complete
      */
     boolean isComplete();
+
+    /**
+     * @return true if the file is available for streaming
+     */
+    default boolean storageFileExists(){
+        return true;
+    }
 }
--- a/core/src/test/groovy/com/dtolabs/rundeck/core/logging/StorageFileTest.groovy
+++ b/core/src/test/groovy/com/dtolabs/rundeck/core/logging/StorageFileTest.groovy
@@ -1,0 +1,43 @@
+package com.dtolabs.rundeck.core.logging
+
+import spock.lang.Specification
+
+class StorageFileTest extends Specification {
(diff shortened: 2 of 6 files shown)


jtobard (changes requested):
Needs some unit test

jtobard (approved):

PR #8770, opened 2023-12-19
Remove tomcat tests from circleci workflow

ahormazabal:
Remove tomcat tests as is not longer supported since 5.0


smartinellibenedetti (approved):

chrismcg14:
@ahormazabal we should also delete the tomcat bash file if isn't being used anymore:
test/run-docker-tomcat-tests.sh

ahormazabal:
Yes @chrismcg14 i was thinking on removing the related files on a separate PR

PR #8786, opened 2024-01-04
RSE-952: Fix: Cluster not retrying execution on another node when owner shutdown

L2JE:
Ticket: https://pagerduty.atlassian.net/browse/RSE-952

**Problem:**
When the cluster member that is running an execution is shutdown, the execution is not being retried on any other cluster member (it is retried in the killed one, getting stuck in running state indefinitely) whereas in 4.8, the execution is retried on another cluster member.

**Expected Outcome:**
The running execution in the stopped cluster member gets retried on any other cluster member.

**Proposed Solution:**

Detect the application shutdown subscribing to `rdpro.shutdown` signal annotation to then force the `ConfigurationService` to toggle the execution mode in the current server so that it's not possible to choose it on any further reschedule operation during the shutdown process.

In order to make this work we must move the execution mode validation from `ExecutionService.retryExecuteJob` to the corresponding `JobSchedulerManager` (which is the one that finally schedules the execution):

* ProjobScheduleManager

* QuartzJobScheduleManagerService

_**Not Covered Case**_
In case the job is run through a runner, the execution is still not recovered. A different bug ticket was created to track that specific case.

### Additional Context (current workflow)

When a cluster member is killed during an execution, the execution is interrupted and after saving its state it is rescheduled again following the configured remote execution policy.

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/ConfigurationService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ConfigurationService.groovy
@@ -25,6 +25,7 @@ import org.rundeck.app.config.SysConfigProp
 import org.rundeck.util.Sizes
 import org.springframework.beans.factory.InitializingBean
 
+import javax.annotation.PreDestroy
 import java.util.concurrent.TimeUnit
 
 @GrailsCompileStatic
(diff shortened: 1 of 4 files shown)


ronaveva (changes requested):
it lacks tests for "interruptScheduleOnPassiveMode" = true

ronaveva (approved):
LGTM

PR #8787, opened 2024-01-05
RUN-36: Fix loading storage plugins to load plugin group settings

carlosrfranco:
Plugin group settings isn't loaded for storage plugins

Changed property resolver on storage tree factory to get framework property using the correct property prefix


chrismcg14 (approved):
LGTM

PR #8789, opened 2024-01-05
RSE-941: Unexpected Default Value Field for Options

DarwisNarvaezDev:
# RSE-941: Unexpected Default Value Field for Options
There is a wrong “Default Value“ field appearing in the job option section of job definition UI, when user write something in the “Storage“ input value and the remove it:
image

## Expected Outcome:
The “Default Value“ input must not be rendered if the secure option is selected.

## Proposed Solution:

In file “rundeckpro/rundeck/rundeckapp/grails-app/views/scheduledExecution/_optEdit.gsp“, aprox line: 270 there is an event handler that yields the behavior of the bug:

```groovy
<wdgt:eventHandler for="defaultStoragePath_${storagePathKey}" state="empty" inline="true" oneway="true" frequency="2" >
    <wdgt:action targetSelector=".opt_keystorage_disabled" visible="true"/> <-------- THIS LINE
    <wdgt:action targetSelector=".opt_keystorage_enabled" visible="false"/>
</wdgt:eventHandler>
```

One hypothesis is that this line, was mistakenly leaved there, so removing it will solve the issue.


alexander-variacode (approved):
LGTM

PR #8657, opened 2023-11-02
RUN-2126: Bug/fix behavior when remotevalues reload

muehan:
**Is this a bugfix, or an enhancement? Please describe.**
Bugfix: [[Bug report] Remote URL Job Options | Cascading Remote Options does not clear previous loaded values #8587](https://github.com/rundeck/rundeck/issues/8587)

**Describe the solution you've implemented**
cleanup if(remote()) function in jobOptions.
It now only showes the values loaded remote to not mixup values from previous remote calls

**Describe alternatives you've considered**
fix doublication check, but this would not fix the behavior that there are values from an old call.

**Additional context**
I dont wanted to push the 3000 files from the **/bin folders.
[...]

Issue #8587, opened 2023-10-03
[Bug report] Remote URL Job Options | Cascading Remote Options does not clear previous loaded values


pgmgb:
**Describe the bug**
We use [Cascading Remote Options](https://docs.rundeck.com/4.15.0/manual/job-options.html#cascading-remote-options) to load values from [remote url](https://docs.rundeck.com/4.15.0/manual/job-options.html#remote-url-job-options) depending on the previous selected values. When `Restrictions | None Any values can be used`, `Multi-valued | Yes` and `Multi-valued | Select All Values by Default` are set for the option, the previous loaded values will not be cleared when loading new values because of cascading.

We have a Rundeck job, which should be able to update users of an application. For that, we created the following job:

**Application API**
https://mocki.io/v1/1b95d58b-f2fc-49af-921a-f40cfad492ea
```json
[
	{
		"name": "Application 1",
		"value": "725d33c6-fd28-4af7-a4fa-c2764bed526b"
	},
	{
		"name": "Application 2",
		"value": "9bad66b1-80d8-428c-9a9b-2497bad80ba9"
	}
]
```

**User Application 1**
https://mocki.io/v1/725d33c6-fd28-4af7-a4fa-c2764bed526b
`["user1", "user2", "user3"]`

**Users Application 2**
https://mocki.io/v1/9bad66b1-80d8-428c-9a9b-2497bad80ba9
`["user1", "user4", "user5"]`

**Rundeck job definition**
```yml
- defaultTab: nodes
  description: ''
  executionEnabled: true
  id: f05ce6f9-9fba-4b08-a6ca-002415a54be6
  loglevel: INFO
  name: Cascading Remote Options
  nodeFilterEditable: false
  options:
  - enforced: true
    name: Application
    required: true
    sortValues: true
    valuesUrl: https://mocki.io/v1/1b95d58b-f2fc-49af-921a-f40cfad492ea
  - delimiter: ','
    multivalueAllSelected: true
    multivalued: true
    name: User
    required: true
    sortValues: true
    valuesUrl: https://mocki.io/v1/${option.Application.value}
  plugins:
    ExecutionLifecycle: null
  scheduleEnabled: true
  sequence:
    commands:
    - script: '#'
    keepgoing: false
    strategy: node-first
  uuid: f05ce6f9-9fba-4b08-a6ca-002415a54be6
```

When loading the job, application `Application 1` will be selected as default and the corresponding users were loaded.
[...]

PR #8793, opened 2024-01-08
RUN-2126: Bug/fix behavior when remotevalues reload

muehan:
I messed up this one: [https://github.com/rundeck/rundeck/pull/8657](https://github.com/rundeck/rundeck/pull/8657)
So here we go again.
------

Bugfix: https://github.com/rundeck/rundeck/issues/8587

cleanup if(remote()) function in jobOptions.
It now only showes the values loaded remote to not mixup values from previous remote calls

Describe alternatives you've considered
fix doublication check, but this would not fix the behavior that there are values from an old call.

Additional context
I dont wanted to push the 3000 files from the **/bin folders.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,7 @@
 *.iml
 /db/
 /lib/
-/bin/
+**/*/bin/
 **/node_modules
 **/build
 **/*/build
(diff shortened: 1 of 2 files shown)


gschueler (approved):

gschueler:
@muehan thank you!

PR #8800, opened 2024-01-08
RUN-2134: Attach change event listener on mount instead of watch

smartinellibenedetti:

AceEditorVue was only adding a change listener in case the initial content wasn't the same as the backupContent (an empty string). This works fine for scenarios where the editor already starts populated, but for readme and message of the day, where there's no content until the first save, this was a problem as the lack of listeners would prevent the content from being properly updated and saved.

Moved the call for adding the event listener from modelValue watcher to the mounted hook.  

**Additional context**

How to test:
* Create a project
* Navigate to Project Settings, either "Edit README" or "Edit Message of the Day" pages
* Attempt to save a message


chrismcg14 (approved):
LGTM

PR #8802, opened 2024-01-09
RUN-2136: Fix step and node step plugin translation

leandromqrs:
Bug fix when try to translate the step and node step plugins to another language.

When the messagePrefix is null, it puts the value of the property as `nullproperty.something.title`. After the fix, with the verification if is null, it puts with the correct value `property.something.title`

**Additional context**
Additional images to compare the before and after...

- Before
Screenshot from 2024-01-08 22-31-28
- After
Screenshot from 2024-01-08 22-41-52

- Before
Screenshot from 2024-01-08 22-42-31
- After
Screenshot from 2024-01-08 22-42-53


charles014:
I can't wait for it to be merged so I can contribute with my translations.

leandromqrs:
@ltamaster sorry for the question, but, why it is not merged yet ? It's a very simple fix 😅

ahormazabal:
Hi @leandromqrs, this has been merged here: https://github.com/rundeck/rundeck/pull/8961
Thanks for your contribution!

PR #8808, opened 2024-01-10
Fix new selenium testdeck

avelasquezr:


ronaveva (changes requested):
"bonigarcia" is not an official repo from selenium, please try using ChromeDriver from selenium with the new options first

ronaveva (approved):
LGTM, just make sure to run the circleci workflow at least 4 times to make it is stable

PR #8814, opened 2024-01-11
RUN-2148: UI:Next: jobs list SCM status fixes

gschueler:

Fix two issues with the SCM status shown in Jobs list (Next UI):
- Error message from invalid configuration or lack of access credentials should be shown in GUI
- SCM status shown for a job was incorrect in some cases

Fix one issue with Git plugin:
- if no ssh key or password path is configured, the GUI should not show an error about not having access to the path

---
test setup:

Create a source repo that can be used for both Import or Export:
```
# create src git repo "gitrepo"
mkdir gitrepo
(cd gitrepo && git init --bare )
mkdir gitcheckout
git clone $PWD/gitrepo gitcheckout
cd gitcheckout
git commit --allow-empty -m 'first'
git push origin -u
```

configure SCM export/import in the project:

* Base directory: (use a file path for scm export, such as `$RDECK_BASE/projects/$project/export)
* git URL: `file:///path/to/gitrepo`
* Base branch on: `master`
* branch: `master`

To test misconfigured key access:

under Authentication section, add a SSH Key Storage Path that doesn't exist.

In the Job list gui (next UI) there should be a warning icon with a message about the error.  (Compare to legacy UI)


rundeckapp/grails-spa/packages/ui-trellis/src/app/pages/job/browse/JobListScmStatus.vue
```diff
@@ -53,66 +69,64 @@ export default defineComponent({
...
+        },
+        importDisplayText() {
+          return this.$t(
+            `scm.import.status.${this.importState}.display.text`
+          );
+        },
+        exportDisplayText() {
+          return this.$t(
+            `scm.export.status.${this.exportState}.display.text`
+          );
```
smartinellibenedetti, line 129: tiny nit: when there is an error with import or export, both importDisplayText and exportDisplayText values will include undefined in the string, while this doesn't affect anything in the UI, I'd like to suggest returning an empty string in this case so that if one day the UI is changed to always show these strings, there's no risk of showing values that do not contain translations.


Screenshot 2024-02-13 at 5 15 08 PM
Screenshot 2024-02-13 at 5 10 01 PM

smartinellibenedetti (approved):
Fix confirmed! Thank you for the instructions regarding how to set up SCM, as I wasn't aware that the project had this functionality.

Noted one tiny nit, but it's an optional fix. 🚀

rundeckapp/grails-spa/packages/ui-trellis/src/app/pages/job/browse/JobListScmStatus.vue
gschueler: Thanks, makes sense to do that

PR #8815, opened 2024-01-13
RUN-2094:  UI:Next: Project list with lazy loading

smartinellibenedetti:
Convert project view (menu/home) to vue, replacing the table pagination with infinite scroll.

Functionality implemented:

- [x] Add 'Enable next UI' checkbox on /menu/home (accessible via the theme selector);
- [x] Add new gsp view (home.next.gsp) for rendering only vue UI;
- [x] project search;
- [x] project actions based on metadata permissions;
- [x] combine table pagination with lazy loading;
- [x] small UI fixes (search results message, accessibility);
- [x] move logic to render firstRun and welcome message to vue (without ui-sockets);

TLDR (tech):

- [x] Add metadata response to /projects endpoint;
- [x] Add new project metadata component (for readme and motd);
- [x] Add unit tests for new vue components;
- [x] Add basic tests in selenium;
- [x] Add new test in project controller;
- [x] Adjust jest config to handle alias in the components;
- [x] Remove redundant ProjectListPage (same functionality as HomePage) from functional tests;

What is not implemented in this PR:
- project activity column;
- project reports (enterprise);
- table view of only favorites (enterprise);

image

**Additional context**

Lighthouse report:
Before:
image

After:
image

Note: performance didn't have big gains as the main problem is some of the js files that haven't been deprecated yet.


rundeckapp/grails-spa/packages/ui-trellis/src/app/components/home/HomeActionsMenu.vue
```diff
@@ -0,0 +1,73 @@
+<template>
+  <dropdown menu-right append-to-body ref="dropdown">
+    <a role="button" class="as-block link-hover link-quiet link-block-padded text-inverse dropdown-toggle">
```
gschueler, line 3: Maybe changing this to a `<btn type="link">` would be better? currently i can't tab to this drop down toggle for some reason.

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/home/HomeActionsMenu.vue
```diff
@@ -0,0 +1,73 @@
...
+      {{ $t('button.Action') }}
+      <span class="caret"></span>
+    </a>
+    <template #dropdown>
+      <template v-for="(option, index) in availableOptions">
```
gschueler, line 8: Maybe an improvement to add: if the user has no admin or job create permissions, this drop down is blank (doesn't show anything when clicked on).  This matches the old behavior, but is maybe not the best behavior... Probably it should not be shown in that case

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/home/HomeBrowserItem.vue
```diff
@@ -0,0 +1,211 @@
+<template>
+  <div class="project_list_item" :key="`project${index}`" :data-project="project.name">
+    <div class="row row-hover row-border-top">
+      <div class="col-sm-6 col-md-8">
+        <ui-socket location="project-favorite" section="home" :socket-data="{projectName: project.name}" />
+        <a
+            :href="`/?project=${project.name}`"
```
gschueler, line 7: This absolute url will not work in cases where the app is running with a specific context path, such as `/rundeck`.  The URL should be built from the `getRundeckContext().rdBase` which is the base URL for the server,

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/home/HomeBrowserItem.vue
```diff
@@ -0,0 +1,211 @@
...
+            class="link-hover text-inverse project_list_item_link link-quiet"
+            :id="`project${index}`"
+        >
+          <p class="h5">
+            {{ project.label ? project.label : project.name }}
+          </p>
```
gschueler, line 13: the `<p>` tag causes the name to be in a block element, and then the schedule/execution disabled indicator icon gets cleared to be below it.  I think it will need to be changed to a `<span>`

Screenshot 2024-02-01 at 4 22 59 PM

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/home/HomeBrowserItem.vue
```diff
@@ -0,0 +1,211 @@
...
+            {{ project.description }}
+          </span>
+          </template>
+        </a>
+      </div>
+
+        <!--      TODO: ADJUST THIS SECTION ONCE ACTIVITY IS REFACTORED -->
+        <div class="col-sm-6 col-md-2 text-center">
+          <a
+              :href="`/project/${project.name}/activity`"
```
gschueler, line 48: this href must also be adjusted

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/home/HomeBrowserItem.vue
```diff
@@ -0,0 +1,211 @@
...
+            <span> {{ $tc("user", project.userCount) }}: </span>
+            <span>
+            {{ project.userSummary.join(", ") }}
+          </span>
+          </div>
+
+          <a
+              v-if="project.failedCount > 0"
+              class="text-warning"
+              :href="`/project/${project.name}/activity?statFilter=fail`"
```
gschueler, line 101: href

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/home/HomeCardList.vue
```diff
@@ -0,0 +1,217 @@
...
+
+      if(this.filterFavoritesOnly) {
+        this.filteredProjects = this.favoriteProjects.filter((project) => {
+          return project.name.match(regex) || project.label && project.label.match(regex);
+        });
+      } else {
+        this.filteredProjects = this.projects.filter((project) => {
+          return project.name.match(regex) || project.label && project.label.match(regex);
+        });
+      }
```
gschueler, line 130:
```suggestion
      let searchSet = this.filterFavoritesOnly ? this.favoriteProjects : this.projects
      this.filteredProjects = searchSet.filter((project) => {
        return project.name.match(regex) || project.label && project.label.match(regex);
      });
```

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/home/HomeCardList.vue
```diff
@@ -0,0 +1,217 @@
...
+
+      this.resultsPage = this.paginateResults(this.filteredProjects);
+
+      // update the pagination to reflect the total number of projects
+      this.pagination.total = this.searchResultsCount;
+
+      if (this.search !== "") {
+        this.showSearchResults = true;
+      }
```
gschueler, line 139: matches old behavior:
```suggestion
      if (this.search !== "" || this.filterFavoritesOnly) {
        this.showSearchResults = true;
      }else{
        this.showSearchResults = false;
      }
```

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/home/HomeCardList.vue
```diff
@@ -0,0 +1,217 @@
...
+      favoriteProjectNames: [] as Project[],
+      resultsPage: [] as Project[],
+      showSearchResults: false,
+      filterFavoritesOnly: false,
+      eventBus: getRundeckContext().eventBus,
+      pagination: {
+        offset: 0,
+        max: 30,
+        total: -1
+      }
```
gschueler, line 89: the home.next.gsp (and home.gsp) defines some paging parameters which could be used here, maybe just the `pagingMax` for now

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/home/services/homeServices.ts
```diff
@@ -25,5 +25,57 @@ export async function getSummary(): Promise<any> {
         // e.message in this case is the error message from the server response
         throw {message: "Error: " + e.message, response: e.response};
     }
+}
 
+export async function getProjectNames(): Promise<any> {
+    let appLinks = getAppLinks();
+    try {
+        const response = await axios
+            .get(appLinks.menuProjectNamesAjax, {
```
gschueler, line 34: let's avoid using the old "ajax" endpoints, we should only use API endpoints. for this purpose, maybe the endpoint to use is the project list, but specify no metadata params, to get the list without metadata.

rundeckapp/grails-app/views/menu/home.next.gsp
```diff
@@ -0,0 +1,109 @@
...
+          data="${[loaded: statsLoaded, execCount: execCount, totalFailedCount: totalFailedCount, recentUsers: recentUsers, recentProjects: recentProjects]}"
+          id="statsData"/>
+
+  <g:embedJSON data="${[
+          detailBatchMax  : params.getInt('detailBatchMax') ?: cfg.getInteger(config: 'gui.home.projectList.detailBatchMax', default: 15),
+          summaryRefresh  : cfg.getBoolean(config: 'gui.home.projectList.summaryRefresh', default: true),
+          refreshDelay    : cfg.getInteger(config: 'gui.home.projectList.summaryRefreshDelay', default: 30000),
+          detailBatchDelay: params.getInt('detailBatchDelay') ?: cfg.getInteger(config: 'gui.home.projectList.detailBatchDelay', default: 1000).toInteger(),
+          pagingEnabled   : params.getBoolean('pagingEnabled', cfg.getBoolean(config: 'gui.home.projectList.pagingEnabled', default: true)),
+          pagingMax       : params.getInt('pagingMax') ?: cfg.getInteger(config: 'gui.home.projectList.pagingMax', default: 30),
```
gschueler, line 56: this can be used in the paging

rundeckapp/grails-app/views/menu/home.gsp
```diff
@@ -25,391 +25,446 @@
...
-    <asset:javascript src="menu/home.js"/>
-
-    <!-- VUE CSS MODULES -->
-    <asset:stylesheet href="static/css/components/version-notification.css"/>
-    <!-- /VUE CSS MODULES -->
-    <asset:javascript src="static/pages/login.js"/>
-    <style type="text/css">
-    .project_list_item_link{
-        display:inline-block;
+	<meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>
```
gschueler, line 28: are the changes in this file, just to revert it to previous state before any vue conversion?

gschueler (changes requested):

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/home/services/homeServices.ts
smartinellibenedetti: Yes, I'm aware that shouldn't be using the old ajax endpoints but this endpoint had the same issue of the activity where it is relying on the session to populate the data, so didn't try to convert it:

```
def projectNamesAjax() {
        AuthContext authContext = rundeckAuthContextProcessor.getAuthContextForSubject(session.subject)
        def fprojects = frameworkService.refreshSessionProjects(authContext, session)

        render(contentType:'application/json',text:
                ([projectNames: fprojects] )as JSON
        )
    }
```

If it's fine to ignore the session logic, can make a new endpoint that just returns the project names but I'm suspecting that the session logic is exactly what is causing the endpoint to return data fast and that blindly converting might backfire. 

But I think there's a confusion here - the service call getProjectNames (that uses the URL  menuProjectNamesAjax) isn't fetching the metadata, it's the other endpoint (getProjects) that does it, and it's only fetching the necessary one (authz, config and message)

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/home/HomeBrowserItem.vue
smartinellibenedetti: Would like to avoid having it as a span as it has no semantic value alone and it causes issues for accessibility (IIRC screen readers ignore orphan spans). Can just make the p have a different display value.

rundeckapp/grails-app/views/menu/home.gsp
smartinellibenedetti: Yes, I had previously added nextUi test in this file to decide when to render the vue component for the header, so removed that logic as well since it all lives inside a vue page.

rundeckapp/grails-app/controllers/rundeck/controllers/ProjectController.groovy
```diff
@@ -582,6 +582,7 @@ class ProjectController extends ControllerBase{
      * @param vers api version requested
      */
     @PackageScope
+    @RdAuthorizeProject(RundeckAccess.General.AUTH_APP_READ)
```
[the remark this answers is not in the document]
smartinellibenedetti: Thank you for the detailed explanation!

rundeckapp/grails-app/controllers/rundeck/controllers/ProjectController.groovy
```diff
@@ -642,7 +655,10 @@ Authorization required: `read` for project resource
 *Since API version 26*: add the project `label` to the response
 
 *Since API version 33*: add the project `created` date to the response. This is based on the creation of the 
-`project.properties` file in the file system or in the DB storage.''',
+`project.properties` file in the file system or in the DB storage.
+
+*Since API version 46*: add the project `metadata` to the response. To retrieve this information, use the query param meta
```
gschueler, line 660:
```suggestion
*Since API version 47*: add the project `metadata` to the response. To retrieve this information, use the query param `meta`
```

gschueler: also, there should be a new `@Parameter` for the `meta` param for the OpenAPI spec that specifies the parameter details, see: 

annotated method param:
https://github.com/rundeck/rundeck/blob/4c390a5e2bbd48abdab9ff8f449a4ce973afccfc/rundeckapp/grails-app/controllers/rundeck/controllers/ScheduledExecutionController.groovy#L6171-L6176

or within the `@Operation`: https://github.com/rundeck/rundeck/blob/f8237b208edef4752f1907a170ea54d213cdea1f/rundeckapp/grails-app/controllers/rundeck/controllers/ProjectController.groovy#L962-L969

gschueler (changes requested):

gschueler (approved):

Issue #8273, opened 2023-04-21
LDAP: error code 12 - Unknown control 1.2.840.113556.1.4.319"


riverfell:
**Describe the bug**
got an error when login with ldap: 

`[LDAP: error code 12 - Unknown control 1.2.840.113556.1.4.319]`

**My Rundeck detail**

* Rundeck version: 4.11
* install type: rpm
* OS Name/version: centos 8
* DB Type/version: mysql

**Here is log**
`[2023-04-20T17:42:54,406] DEBUG jaas.JettyCachingLdapLoginModule - Cache Miss for xxx.
[2023-04-20T17:42:54,406] DEBUG jaas.JettyCachingLdapLoginModule - Searching for users with filter: '(&(objectClass={0})({1}={2}))' from base dn: ou=people,dc=xxx,dc=com
[2023-04-20T17:42:54,475] DEBUG jaas.JettyCachingLdapLoginModule - Found user?: true
[2023-04-20T17:42:54,476] INFO  jaas.JettyCachingLdapLoginModule - Attempting authentication: uid=xxx,ou=people,dc=xxx,dc=com
javax.naming.OperationNotSupportedException: [LDAP: error code 12 - Unknown control 1.2.840.113556.1.4.319]; remaining name 'ou=group,dc=xxx,dc=com'
        at com.sun.jndi.ldap.LdapCtx.mapErrorCode(LdapCtx.java:3321)
        at com.sun.jndi.ldap.LdapCtx.processReturnCode(LdapCtx.java:3207)
        at com.sun.jndi.ldap.LdapCtx.processReturnCode(LdapCtx.java:2998)
        at com.sun.jndi.ldap.LdapCtx.searchAux(LdapCtx.java:1874)
        at com.sun.jndi.ldap.LdapCtx.c_search(LdapCtx.java:1797)
        at com.sun.jndi.ldap.LdapCtx.c_search(LdapCtx.java:1814)
        at com.sun.jndi.toolkit.ctx.ComponentDirContext.p_search(ComponentDirContext.java:418)
        at com.sun.jndi.toolkit.ctx.PartialCompositeDirContext.search(PartialCompositeDirContext.java:396)
        at com.sun.jndi.toolkit.ctx.PartialCompositeDirContext.search(PartialCompositeDirContext.java:378)
        at javax.naming.directory.InitialDirContext.search(InitialDirContext.java:286)
[...]

PR #8822, opened 2024-01-18
RUN-2741: Add LDAP rolePagination config option for docker

palto42:
Some LDAP servers don't support pagination in the search queries so that pagination must be disabled to fix `error code 12 - Unavailable Critical Extension` in the LDAP search.

This PR addresses the issue described in #8273 by allowing the user to set the `RUNDECK_JAAS_LDAP_ROLEPAGINATION=false` in the docker based setup if needed.

### DIFF
--- a/docker/official/remco/templates/jaas-loginmodule.conf
+++ b/docker/official/remco/templates/jaas-loginmodule.conf
@@ -24,6 +24,7 @@
         roleMemberAttribute="{{ getv("/rundeck/jaas/ldap/rolememberattribute", "uniqueMember") }}"
         roleObjectClass="{{ getv("/rundeck/jaas/ldap/roleobjectclass", "groupOfUniqueNames") }}"
         rolePrefix="{{ getv("/rundeck/jaas/ldap/roleprefix", "") }}"
+        rolePagination="{{ getv("/rundeck/jaas/ldap/rolepagination", "true") }}"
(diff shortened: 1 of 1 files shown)


atsonkov:
I was about to open literally the same PR for slightly different reason. For us with pagination the LDAP conversation takes ~20 seconds to resolve the role<->member mapping (and thus causes slow login), while without pagination it takes less than a second, so we would also like to be able to disable it via remco.

I could not find details in CONTRIBUTING.md, but (looking at recently closed PRs) it seems that you should assign a milestone (maybe 5.6.0 at the moment is a reasonable choice) and a reviewer for this PR, otherwise the merge checks fail and nobody will check it.

atsonkov (approved):
Changes look good and were manually tested by passing "false" via environment variable.

palto42:
@atsonkov Thanks for your review. Ad far as I can see I can't add a milestone or users, assume that this can only be done by the project maintainers.

atsonkov:
> @atsonkov Thanks for your review. Ad far as I can see I can't add a milestone or users, assume that this can only be done by the project maintainers.

Thanks for trying! I was hoping that you as PR author will have permissions to do that.

@gschueler: Not sure what is the process for assigning a milestone and reviewing. Could you take a look?

gschueler:
Thanks for the contribution! It will be assigned to a release milestone when we are able to review it, but likely the next release since it is a small change.

gschueler (approved):

PR #8859, opened 2024-01-29
RUN-2173: NPE While Reading Null Default Values for Secure Options

DarwisNarvaezDev:
## RSE-954 Rollback and proper fix.

Discarded the UI changes and only ignored the default values for secure options if the key storage path is set.


rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
```diff
@@ -1859,8 +1859,10 @@ class ScheduledExecutionService implements ApplicationContextAware, Initializing
         def options = importedJob.job.options
         def secureOptionsWithDefaultValues = new ArrayList<Option>()
         options.each { option -> {
-            if( option.secureInput && (option.defaultValue != null || !option.defaultValue.isEmpty()) ){
-                secureOptionsWithDefaultValues << option
+            if (option.secureInput && !option?.defaultValue?.isEmpty()) {
+                if( option?.defaultValue != null ){
+                    secureOptionsWithDefaultValues << option
+                }
```
gschueler, line 1865: groovy treats non-null non-empty string as truthy, you can just check it with this
```suggestion
            if (option.secureInput && option.defaultValue) {
                  secureOptionsWithDefaultValues << option
```

DarwisNarvaezDev:
> I tested importing a job definition with a `value: ` entry for a secure option. The defaultValue is still imported to the job, and will appear in the job run GUI. However, it should be ignored on import.

Fixed that, i misunderstood the request, i ignored the value for executions but not for imports.

gschueler (approved):

PR #8863, opened 2024-01-31
RSE-934: Migrate API to New testdeck

avelasquezr:
A connection pool is added to control the stream when turning clusters on and off.
Also, option `--disable-blink-features=AutomationControlled` is added


ronaveva (approved):
LGTM

PR #8868, opened 2024-01-31
RUN-2175: fix: legacyXml flag not working

### LINKED ISSUE #8860: XML responses broken and API version mismatch (5.0.1)
**Describe the bug**

I have 2 major problems since the upgrade from 4.8 to 5.0.1.
I use the API version 41 in my calls.

1) Despite using undeck.feature.legacyXml.enabled=true in /etc/rundeck/rundeck-config.properties, I can't get an XML response on some routes. For example:

```
$ curl -X GET -H "Accept: application/xml" http://localhost:4440/api/41/execution/34771?authtoken=XXXXXXXXXX
<error>An internal server error occurred</error>
```

The error is:

```
==> /var/log/rundeck/service.log <==
[2024-01-30T12:30:12,268] WARN  server.HttpChannel - /api/41/execution/34771
javax.servlet.ServletException: Could not resolve view with name 'apiExecution' in servlet with name 'grailsDispatcherServlet'
        at org.springframework.web.servlet.DispatcherServlet.render(DispatcherServlet.java:1385) ~[spring-webmvc-5.3.27.jar!/:5.3.27]
...
```

It works with json:

```
$ curl -X GET -H "Accept: application/json" http://localhost:4440/api/41/execution/34771?authtoken=XXXXXXXXXXXX
{"id":34771, ...............
```

2) This problem can be the cause of my problem 1. It seems that the API version in the request is ignored (/api/41/)
If I make an incorrect call the the API, I can see the response coming from API version 46:

```
$ curl -X POST http://localhost:4440/api/41/tokens?authtoken=XXXXXXXXXXXX
{"errorCode":"api.error.invalid.request","apiversion":46,"error":true,"message":"Invalid API Request: Expected request content to be one of allowed formats: [xml, json], but was: null"}
```

**My Rundeck detail**

* Rundeck version: 5.0.1
* install type:deb
* OS Name/version: debian 11
* DB Type/version: mysql

**To Reproduce**

See the bug description.

**Expected behavior**

Working XML responses for all API calls with rundeck.feature.legacyXml.enabled=true and old API versions (41 in my examples).

N/A, related to the API.

**Additional context**

gschueler:

fixes #8860 

* fix check for allow XML inside withFormat closures
* use '*' to return json by default and whenever format is not identified

**Additional context**

### DIFF
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/ApiController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/ApiController.groovy
@@ -878,14 +878,14 @@ Since: v11
         }
 
         SystemInfoModel systemInfoModel = new SystemInfoModel(systemInfoMap)
-
+        def controller= this
         withFormat{
-            json{
+            '*' {
                 return apiService.renderSuccessJson(response){
                     systemInfoModel
                 }
             }
-            if (isAllowXml()) {
+            if (controller.isAllowXml()) {
                 xml{
                     return apiService.renderSuccessXml(request,response){
                         delegate.'system'{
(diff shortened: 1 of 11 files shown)


ltamaster (approved):
LGTM

PR #8871, opened 2024-02-01
RUN-2038: install primevue and continuum tokens

smartinellibenedetti:
Install primeVue and create a static CSS file with continuum tokens.

Important to note:
- Wired primeVue plugin on` ui/main.ts `to ensure that it will be available everywhere, to avoid having people to do so every time they use one of its components.
- Added tokens in the `base.gsp` file so that the css file lives in the same folder as other third party's css.

#### **- In `ui/main.ts`, the stylesheet for primevue is commented. The reason is that I need to rethink where to put this import to avoid causing style changes to existing elements (noticed changes on .btn-primary, but it's possible that more things got affected). In the original plan, the idea was to not rely on the themes from primeVue, but designers have opted for now to only apply continuum's colors and keep the other properties with the appearance they have in the figma file. The figma file uses the lara theme (as noted [here](https://primevue.org/uikit/)), therefore left the import commented for now.**

- Started to wrap the Message component to assert how long it takes to just tweak colors. This is an example how they look when the stylesheet isn't commented:

Screenshot 2024-01-31 at 4 14 43 PM

**Additional context**


gschueler:
tested, the theme seems to override the `--primary-color` and `--cta-color` variables, which causes the buttons to look different.  Perhaps we can make a copy of the lara-light theme and remove those overrides

gschueler (approved):

smartinellibenedetti:
> tested, the theme seems to override the `--primary-color` and `--cta-color` variables, which causes the buttons to look different. Perhaps we can make a copy of the lara-light theme and remove those overrides

PrimeVue has a big rewrite planned for how the tokens are defined (they will move out from sass into CSS variables) for this quarter, so I'm concerned that by making a copy now we might lose something important from this update—been thinking about either just adding a temporary CSS undoing the override for these 2 tokens OR testing how things would look with primeVue CSS being imported before theme.css.

PR #8882, opened 2024-02-02
RUN-2191: fix: feature flag defaultEnabled doesn't work

gschueler:
* fixes problem with defaultEnabled setting for feature flags
* cleanup


carlosrfranco (approved):
Tested

PR #8884, opened 2024-02-05
RSE-1007 Fix flaky tests on new Selenium Testdeck

Jesus-Osuna-M:
Problem:
Two tests were marked as flaky because they occasionally failed. The cause of the failure was that it was trying to go to the project that had been created and immediately clicking on an option that was not loaded, causing both tests to fail.

Solution:
Two validations were added. The first one checks that the page it is on is the project home, and the second one allows waiting for the URL to match the project name before selecting the JOBS option from the sidebar.


ronaveva (approved):
LGTM, make sure it runs a couple of times on circleci before merging it

PR #8890, opened 2024-02-06
Fix flaky tests

avelasquezr:
The flaky tests of:
`go to edit nodes`
`job workflow undo redo`
`job option simple redo`
`job option revert all`
`job option undo redo`
The old testdeck configuration for selenium is removed

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -505,61 +505,12 @@ jobs:
           command: << parameters.command >>
 
 
-  # Run Testdeck job on machine executor.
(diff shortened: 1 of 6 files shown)


ronaveva (changes requested):
This should not include a circleci config removing old testdeck config

avelasquezr:
The configuration that ran the old selenium test is removed, it is not currently executed.

ronaveva (approved):
LGTM

PR #8898, opened 2024-02-08
RUN-2135: Convert project node sources page to Vue

chrismcg14:
Convert project node sources page to vue, converting the current project configurable template to vue as well


rundeckapp/grails-spa/packages/ui-trellis/src/app/pages/project-nodes-config/ProjectConfigurableForm.vue
```diff
@@ -0,0 +1,167 @@
...
+      extraConfigSet: {},
+      isConfigSet: false,
+      projectConfigurableNames: [],
+      finalExtraConfig: {config: {}},
+      originalConfig: {},
+    };
+  },
+  props: ['category', 'categoryPrefix', 'helpCode', 'serviceName'],
+  methods: {
+    async loadConfig() {
```
smartinellibenedetti, line 65: suggestion: I am aware that you kept the name extraConfigSet from the original code in the gsp, but the code will become a little bit cleaner if you rename and tweak the data a little bit, because it's hard to infer without further reading, what is the difference between extraConfigSet, the finalExtraConfig, the originalConfig and which one is being mutated by user input.

So this is my suggestion, but feel free to name it differently:

extraConfigSet -> apiResponse or fetchedData
finalExtraConfig -> cache (It would be great if there wasn't the need for having a property config nested, but saw that it's how pluginConfig works)
originalConfig -> massagedConfig

Also, would consider to have originalConfig as a local variable instead of living in the data and to move lines 68-92 to its own method (or to add a comment before pointing out that you're massaging the data);

smartinellibenedetti (changes requested):
I think you've done a great job with this ticket and figuring out what was needed to get the PluginConfig working. 

I've labeled the comments to help to distinguish their intent and priority - chores and todo are blockers, the suggestions are not (but please consider doing them as well).

Also, I noticed that it's possible to save the form while having the cache delay field empty. Do you know if it was always like this? If yes, think that it would be a good improvement to address this, but don't think it has to be in this ticket if the codebase assumes that a lack of value means the cache delay is 30s.

Last but not least, if you'd like, can show you unit testing for vue with jest sometime next week. It's not a must do at this point, but it would be great if we start to add them before starting to work on the new UI.

Didn't check the selenium test yet, but will do first thing next week.

rundeckapp/grails-spa/packages/ui-trellis/src/app/pages/project-nodes-config/ProjectConfigurableForm.vue
```diff
@@ -0,0 +1,213 @@
...
+    },
+    async saveConfig() {
+      const configMap = this.transformConfigToMap(this.cacheConfig.config);
+      this.cacheConfig.config = {
+        ...this.cacheConfig.config,
+        ...configMap,
+      };
+      this.cacheConfig.config = this.convertMapNumbersToStrings(
+        this.cacheConfig.config,
+      );
```
smartinellibenedetti, line 148: since you're calling await this.loadConfig that will update cacheConfig, could you please change these lines to avoid mutating it twice?

```suggestion
      const configsToSave = this.convertMapNumbersToStrings(this.transformConfigToMap(this.finalExtraConfig.config)) 
```

and then update the setProjectConfigurable call to pass the configsToSave (or the name you prefer)

smartinellibenedetti (approved):
I left one last comment, but otherwise is g2g! Thanks for making these changes 🚀

PR #8900, opened 2024-02-09
RSE-934: Fixes Cluster Members

avelasquezr:
I add these validations to separate the cluster environment.


avelasquezr:
Wrong branch

PR #8904, opened 2024-02-09
enable ui-trellis jest tests

gschueler:
Enabling jest tests via gradle


smartinellibenedetti:
Would it be too much work to add the jest tests as a separate step inside the gradle tests? just would make it easier to find out what's wrong

smartinellibenedetti:
PR for fixing the old tests got merged now, g2g for a re-run

rundeckapp/grails-spa/packages/ui-trellis/package.json
```diff
@@ -4,7 +4,10 @@
   "description": "A collection of utilities used across various javascript applications and components.",
   "main": "./lib/index.js",
   "scripts": {
-    "test": "jest --no-cache",
+    "test": "npx jest --no-cache",
+    "test:unit": "npx jest",
```
smartinellibenedetti, line 8: since you're running npm install, it should be possible to remove the npx command.

Also, what do you think of adding --silent flag on test:unit?

PR #8916, opened 2024-02-13
update api version to 47 and setup cleanup when api version is changed

ronaveva:
Update the default api version to test to 47 and adds cleanup to set back version to default value


Jesus-Osuna-M (approved):
LGTM

PR #8923, opened 2024-02-16
RUN-2208: cleanup: prettier for pluginConfig

gschueler:
* fix a semantic error in pluginConfig.vue
* reformat with prettier


smartinellibenedetti (approved):
🚀

PR #8924, opened 2024-02-16
RSE-1024: Migrate Scm API TESTS to new testdeck

L2JE:
Ticket: https://pagerduty.atlassian.net/browse/RSE-1024

This includes the tests from the following scripts:

* `test-scm-plugin-status.sh`
* `test-scm-git-plugin-setup.sh`
* `test-scm-git-plugin-inputs.sh`
* `test-scm-plugin-config.sh`
* `test-scm-plugin-actions.sh`
* `test-scm-plugins-list.sh`

PR #8936, opened 2024-02-21
RSE-1024: Fixed test that was using a method that was moved to jobUtils

L2JE:
After merging this PR: https://github.com/rundeck/rundeck/pull/8924
a test started to fail because it was using a method that was moved to `JobUtils`


ronaveva (approved):
LGTM

PR #8948, opened 2024-02-23
RSE-1036: Fix: Scm tests are flaky on pro

L2JE:
ticket: https://pagerduty.atlassian.net/browse/RSE-1036

#### Solution
As the original tests from testdeck, exclude scm tests from pro test tasks


ronaveva (approved):
LGTM

PR #8949, opened 2024-02-23
Replaced webhook-notification-project.zip and scm tests project with directory

L2JE:
Replaced  `webhook-notification-project.zip` with its extracted content as required by project-import convention
Replaced  `scm/project-scm-export-one-job.rdproject.jar` with its extracted content as required by project-import convention


ronaveva (changes requested):
you could use "setupProjectArchiveDirectory" method directly

ronaveva (approved):
LGTM

ronaveva (approved):
LGTM

PR #8950, opened 2024-02-23
fixed overlapping label and button

jayas006:
Fixed the label and button which overlapping with the other UI element by adding `label` tag with classes and data-bind attributes for tooltip

**Additional context**


smartinellibenedetti (approved):
fix confirmed!

PR #8969, opened 2024-02-29
RSE-776 Fix: Server Error while Metrics Request

DarwisNarvaezDev:
## The Problem
When a execution is running the 'date_completed' data needed to make metrics calculations, is null.

## The Fix
If 'date_completed' is null, we exclude the calculation of that particular execution from the metrics.


alexander-variacode (approved):
LGTM

PR #8973, opened 2024-03-01
Rse 1042 move hold method from cluster base container

Jesus-Osuna-M:
Add extra waiting time to avoid NPE on "exportNeededStatus.commit.size()"


ronaveva (approved):
LGTM

PR #8974, opened 2024-03-02
RUN-2032: Percentage in number showing in progress bar instead of undefined%

jayas006:

"undefined%" message appears in the progress bar instead of percentage. Fixed it to make it shows the percentage in number. 

**Additional context**

### DIFF
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/app/components/activity/activityList.vue
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/app/components/activity/activityList.vue
@@ -288,7 +288,6 @@
               class="dateStarted date"
               colspan="2"
             >
-              <!--getting vue warning: Invalid prop: type check failed for prop "modelValue".Converted :value to :modelValue-->
               <progress-bar
                 v-if="exec.status === 'scheduled'"
                 :modelValue="100"
@@ -297,7 +296,6 @@
                 label
                 :label-text="
                   $t('job.execution.starting.0', [
-                    // if exec.dateStarted is null or undefined,it doesn't throw an error but instead returns undefined
                     runningStartedDisplay(exec.dateStarted?.date) || 'N/A',
                   ])
                 "
@@ -802,24 +800,6 @@ export default defineComponent({
       }
       return 0;
     },
-
-    // Below code makes the progress bar reach closer to 100%
-    /* jobDurationPercentage(exec: Execution) {
-      if (
-        exec.job &&
-        exec.job.averageDuration &&
-        exec.dateStarted &&
-        exec.dateStarted.date
-      ) {
-        const now = moment();
-        const startDate = moment(exec.dateStarted.date);
-        const diff = now.diff(startDate);
-        const percentage = Math.floor((diff / exec.job.averageDuration) * 100);
-        return Math.min(percentage, 100);
-      }
-      return 0;
-    },
-    */
 
     runningStartedDisplay(date: string) {
       if (!date) {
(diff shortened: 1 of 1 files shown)


rundeckapp/grails-spa/packages/ui-trellis/src/app/components/activity/activityList.vue
```diff
@@ -288,29 +288,31 @@
...
-                :value="100"
+                :modelValue="100"
                 striped
                 type="default"
                 label
                 :label-text="
                   $t('job.execution.starting.0', [
-                    runningStartedDisplay(exec.dateStarted.date),
+                    // if exec.dateStarted is null or undefined,it doesn't throw an error but instead returns undefined
+                    runningStartedDisplay(exec.dateStarted?.date) || 'N/A',
```
smartinellibenedetti, line 301: Question, what led you to the conclusion that this change was needed? Have you seen a scenario where this happens?

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/activity/activityList.vue
```diff
@@ -801,6 +802,25 @@ export default defineComponent({
...
+      if (
+        exec.job &&
+        exec.job.averageDuration &&
+        exec.dateStarted &&
+        exec.dateStarted.date
+      ) {
+        const now = moment();
+        const startDate = moment(exec.dateStarted.date);
+        const diff = now.diff(startDate);
+        const percentage = Math.floor((diff / exec.job.averageDuration) * 100);
```
smartinellibenedetti, line 817: One heads-up, if it's the first time running a job, `averageDuration` isn't available. We shouldn't be showing 0 in this case.

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/activity/activityList.vue
```diff
@@ -801,6 +802,25 @@ export default defineComponent({
       }
       return 0;
     },
+
+    // Below code makes the progress bar reach closer to 100%
```
smartinellibenedetti, line 806: ✂️

smartinellibenedetti (commented):
One heads-up, one common habit that engineers develop is to validate solutions before implementing them to avoid missing important aspects.

The architecture meeting on Mondays is an example of folks validating their ideas.

Whenever starting to work on a task that doesn't have steps regarding how to solve it, l'd like you to investigate the issue and send me or other folks a message with the plan of how this issue will be solved.

In this type of communication, it is also helpful to point out how you concluded that what you're solving is the problem in the first place.

This will help us to ensure you're on the right track, and that you have all the info you need, so that once you execute your plan, the PR is ready for review.

Currently, this will be moved back to in progress, and we can discuss it more tomorrow.

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/activity/activityList.vue
jayas006: > Question, what led you to the conclusion that this change was needed? Have you seen a scenario where this happens?

I was going through line by line code between progress bar (first I saw in vue dev tool to confirm which line of code I have to see to make it  right) and started understanding the functions. So the `runningStartedDispaly` requires the valid date to return the right output.

jayas006: > Question, what led you to the conclusion that this change was needed? Have you seen a scenario where this happens?

Ok so I fixed it now. It was just unnecessary change from my side. Ignore it

smartinellibenedetti (approved):
good job!

PR #9001, opened 2024-03-13
Fix failing test on circleCi "test_check_diff_after_actions"

Jesus-Osuna-M:
Problem:
Sometimes the method  callGetJobStatus to get the SCM status of a job via API get the value before is completely updated in Resulting on a NPE 

Solution:
Create a new method that calls the callGetJobStatus inside a cicle assuring that the SCM status is not empty before checking results.


ronaveva (approved):
LGTM

PR #9005, opened 2024-03-14
Wiz Scan Build Image Instead of Branch Image

chrismcg14:
Use build image instead of branch image to ensure it is for current build


gschueler (approved):

ahormazabal:
Wizcli is failing bc the image wasn't found. We need https://github.com/rundeck/rundeck/pull/9004 before being able to properly test this PR.

PR #9008, opened 2024-03-14
RUN-2248: Fix: don't include UI timezone data in schedules json

gschueler:
Fix a minor bug in the job editor Schedules UI: timezone data used by the GUI is included in the json data sent to the server.

Remove the timeZones from the json

**Additional context**
Screenshot 2024-03-14 at 3 10 29 PM


smartinellibenedetti (approved):
awesome catch :shipit:

PR #9015, opened 2024-03-19
RSE-846 & RSE-774 Fix: Users logged in rundeck from LDAP are duplicate 

Jesus-Osuna-M:
Related to: 
https://pagerduty.atlassian.net/browse/RSE-774

Problems:
When accessing Rundeck for the first time if the users had configured the rundeck.security.syncLdapUser=true property users were duplicated on the RDUSER table. 

The username on rundeck is case sensitive while usernames un LDAP are case insensitive this difference generated duplicated users on the RDUSER table since.

Solutions:
On the SetUserInterceptor remove code that was causing a duplicated register, this part of the code wasn't necessary since the user is already being created on the RundeckJaasAuthenticationSuccessEventListener.groovy.

The use of the method getUserByLoginOrCreate was replaced by findOrCreateUser since both methods had the same logic.

A new feature flag was added to the config.properties 
**rundeck.login.nameCaseSensitiveEnabled** 
By default this flag will be false meaning that username on login will case insensitive.

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/UserService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/UserService.groovy
@@ -41,7 +41,7 @@ class UserService {
     public static final String SESSION_ABANDONDED_MINUTES = 'userService.login.track.sessionAbandoned'
     public static final String SHOW_LOGIN_STATUS = 'gui.userSummaryShowLoginStatus'
     public static final String SHOW_LOGGED_USERS_DEFAULT = 'gui.userSummaryShowLoggedUsersDefault'
-    public static final String LDAP_SYNC_USERS_ENABLE = 'security.syncLdapUser'
+
 
     RdUser findOrCreateUser(String login) {
         return userDataProvider.findOrCreateUser(login)
@@ -246,7 +246,5 @@ class UserService {
         String lastname
     }
 
-    def isLdapSyncUserEnabled(){
-        configurationService.getBoolean(LDAP_SYNC_USERS_ENABLE, false)
-    }
+
 }
--- a/rundeckapp/grails-app/services/rundeck/services/UserService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/UserService.groovy
@@ -42,7 +42,6 @@ class UserService {
     public static final String SHOW_LOGIN_STATUS = 'gui.userSummaryShowLoginStatus'
     public static final String SHOW_LOGGED_USERS_DEFAULT = 'gui.userSummaryShowLoggedUsersDefault'
 
-
     RdUser findOrCreateUser(String login) {
         return userDataProvider.findOrCreateUser(login)
     }
@@ -245,6 +244,4 @@ class UserService {
         String firstname
         String lastname
     }
-
-
 }
--- a/rundeckapp/grails-app/services/rundeck/services/UserService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/UserService.groovy
@@ -41,6 +41,7 @@ class UserService {
     public static final String SESSION_ABANDONDED_MINUTES = 'userService.login.track.sessionAbandoned'
     public static final String SHOW_LOGIN_STATUS = 'gui.userSummaryShowLoginStatus'
     public static final String SHOW_LOGGED_USERS_DEFAULT = 'gui.userSummaryShowLoggedUsersDefault'
+    public static final String LDAP_SYNC_USERS_ENABLE = 'security.syncLdapUser'
 
     RdUser findOrCreateUser(String login) {
         return userDataProvider.findOrCreateUser(login)
@@ -243,5 +244,9 @@ class UserService {
         String email
         String firstname
         String lastname
+    }
+
+    def isLdapSyncUserEnabled(){
+        configurationService.getBoolean(LDAP_SYNC_USERS_ENABLE, false)
     }
 }
--- a/rundeckapp/grails-app/services/rundeck/services/events/UserActionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/events/UserActionService.groovy
@@ -24,7 +24,6 @@ class UserActionService implements LogoutHandler{
                 sessionId = event.getSource()?.details?.sessionId
             }
             userService.registerLogin(extractUsername(event.authentication), sessionId)
-
         }else{
             log.error("Null user name on handleAuthenticationSuccessEvent")
         }
--- a/rundeckapp/grails-app/services/rundeck/services/events/UserActionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/events/UserActionService.groovy
@@ -23,7 +23,10 @@ class UserActionService implements LogoutHandler{
             if(userService.isSessionIdRegisterEnabled()){
                 sessionId = event.getSource()?.details?.sessionId
             }
-            userService.registerLogin(extractUsername(event.authentication), sessionId)
+            if(!userService.isLdapSyncUserEnabled()){
+                userService.registerLogin(extractUsername(event.authentication), sessionId)
+            }
+
         }else{
             log.error("Null user name on handleAuthenticationSuccessEvent")
         }
--- a/rundeckapp/grails-app/services/rundeck/services/events/UserActionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/events/UserActionService.groovy
@@ -23,9 +23,7 @@ class UserActionService implements LogoutHandler{
             if(userService.isSessionIdRegisterEnabled()){
                 sessionId = event.getSource()?.details?.sessionId
             }
-            if(!userService.isLdapSyncUserEnabled()){
-                userService.registerLogin(extractUsername(event.authentication), sessionId)
-            }
+            userService.registerLogin(extractUsername(event.authentication), sessionId)
 
         }else{
             log.error("Null user name on handleAuthenticationSuccessEvent")
--- a/rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormUserDataProvider.groovy
+++ b/rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormUserDataProvider.groovy
@@ -234,7 +234,9 @@ class GormUserDataProvider implements UserDataProvider {
 
     @Override
     RdUser findByLogin(String login) {
-        return findUserByLoginCaseSensitivity(login)
+        User.withNewSession {
+            return findUserByLoginCaseSensitivity(login)
+        }
     }
 
     @Override
--- a/rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormUserDataProvider.groovy
+++ b/rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormUserDataProvider.groovy
@@ -219,7 +219,7 @@ class GormUserDataProvider implements UserDataProvider {
 
     @Override
     boolean validateUserExists(String username) {
-        return User.countByLoginIlike(username) > 0
+        return isLoginNameCaseSensitiveEnabled()? User.countByLogin(username) > 0 : User.countByLoginIlike(username) > 0
     }
 
     @Override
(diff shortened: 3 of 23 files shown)


rundeckapp/grails-app/services/rundeck/services/UserService.groovy
```diff
@@ -41,6 +41,7 @@ class UserService {
     public static final String SESSION_ABANDONDED_MINUTES = 'userService.login.track.sessionAbandoned'
     public static final String SHOW_LOGIN_STATUS = 'gui.userSummaryShowLoginStatus'
     public static final String SHOW_LOGGED_USERS_DEFAULT = 'gui.userSummaryShowLoggedUsersDefault'
+    public static final String LDAP_SYNC_USERS_ENABLE = 'security.syncLdapUser'
```
ronaveva, line 44: i can see this string being used in other places, could this be unified?
EG: RundeckJaasAuthenticationSuccessEventListener

rundeckapp/grails-app/services/rundeck/services/UserService.groovy
```diff
@@ -244,4 +245,8 @@ class UserService {
         String firstname
         String lastname
     }
+
+    def isLdapSyncUserEnabled(){
```
ronaveva, line 249: add javadoc to this new method

rundeckapp/grails-app/services/rundeck/services/events/UserActionService.groovy
```diff
@@ -24,6 +24,7 @@ class UserActionService implements LogoutHandler{
                 sessionId = event.getSource()?.details?.sessionId
             }
             userService.registerLogin(extractUsername(event.authentication), sessionId)
+
```
ronaveva, line 27: no need to include this change in the PR

rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormUserDataProvider.groovy
```diff
@@ -214,9 +209,10 @@ class GormUserDataProvider implements UserDataProvider {
             } as List<RdUser>
         }
         def response = new UserFilteredResponse()
-        response.setTotalRecords(totalRecords)
```
ronaveva, line 217: it looks like this change is not needed

rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormUserDataProvider.groovy
```diff
@@ -314,4 +310,9 @@ class GormUserDataProvider implements UserDataProvider {
     def getSessionIdRegisterMethod() {
         configurationService.getString(SESSION_ID_METHOD, 'hash')
     }
+
+    def isLoginNameCaseSensitiveEnabled(){
```
ronaveva, line 314: javadoc required

rundeckapp/src/test/groovy/org/rundeck/app/providers/GormUserDataProviderSpec.groovy
```diff
@@ -37,6 +37,37 @@ class GormUserDataProviderSpec extends RundeckHibernateSpec implements DataTest
...
+        User savedUser = new User(login: loginName)
+        savedUser.save()
+        when:
+        provider.findOrCreateUser(loginName.toUpperCase())
+        then:
+        User.findAll().size() == 1
+    }
+
+    @Unroll
+    def "Find or create User case sensitive  "() {
```
ronaveva, line 56: there is no need to have 2 separated tests for this, you should use "where" block to provide the necessary parameters and test both cases in a single test block

rundeckapp/src/test/groovy/org/rundeck/app/providers/GormUserDataProviderSpec.groovy
```diff
@@ -74,7 +105,7 @@ class GormUserDataProviderSpec extends RundeckHibernateSpec implements DataTest
     def "Should throw error on registerLogin with bad login"() {
         setup:
         provider.configurationService = Mock(ConfigurationService) {
-            1 * getBoolean(UserService.SESSION_ID_ENABLED, false) >> false
```
ronaveva, line 77: why is this being removed?

rundeckapp/src/test/groovy/org/rundeck/app/providers/GormUserDataProviderSpec.groovy
```diff
@@ -127,6 +158,48 @@ class GormUserDataProviderSpec extends RundeckHibernateSpec implements DataTest
...
+        savedUser.save()
+
+        when:
+        provider.updateUserProfile(login, lastName, firstName, email)
+        then:
+        provider.findAll().size() == 1
+
+        where:
+        login   | lastName  | firstName  | email
+        "USER1" | "last1"   | "first1"   | "email1@company.com"
```
ronaveva, line 174: there is no point on having "where" block for a single scenario

rundeckapp/src/test/groovy/org/rundeck/app/providers/GormUserDataProviderSpec.groovy
```diff
@@ -127,6 +158,48 @@ class GormUserDataProviderSpec extends RundeckHibernateSpec implements DataTest
...
+        firstUser.firstName == "otherFirstName"
+        firstUser.lastName == "otherLastName"
+        firstUser.email == "otherEmail@company.com"
+        def userCreatedAfterUpdate = User.findByLogin(login)
+        userCreatedAfterUpdate.firstName == firstName
+        userCreatedAfterUpdate.lastName == lastName
+        userCreatedAfterUpdate.email == email
+        provider.findAll().size() == 2
+
+        where:
```
ronaveva, line 198: there is no point on having "where" block for a single scenario

ronaveva (changes requested):
check observations please

ronaveva:
also, this requires for the new property to be added to the doc

rundeckapp/grails-app/services/rundeck/services/UserService.groovy
Jesus-Osuna-M: This constant needs to be removed

rundeckapp/grails-app/services/rundeck/services/UserService.groovy
Jesus-Osuna-M: This method needs to be removed.

rundeckapp/src/test/groovy/org/rundeck/app/providers/GormUserDataProviderSpec.groovy
Jesus-Osuna-M: After removing getUserByLoginOrCreate for findOrCreateUser this method was never called an the test was failing.

rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormUserDataProvider.groovy
```diff
@@ -39,6 +40,7 @@ class GormUserDataProvider implements UserDataProvider {
     public static final int DEFAULT_TIMEOUT = 30
     public static final String SESSION_ABANDONED_MINUTES = 'userService.login.track.sessionAbandoned'
     public static final String SHOW_LOGIN_STATUS = 'gui.userSummaryShowLoginStatus'
+    public static final String NAME_CASE_SENSITIVE_ENABLED = "login.nameCaseSensitiveEnabled"
```
gschueler, line 43: if you define a new system configuration key, you should use the `SystemConfigurable` mechanism

rundeckapp/src/main/groovy/org/rundeck/security/RundeckJaasAuthenticationSuccessEventListener.groovy
```diff
@@ -56,16 +57,12 @@ class RundeckJaasAuthenticationSuccessEventListener implements ApplicationListen
...
-                            firstName: firstNamePrincipal?.name,
-                            email: emailPrincipal?.name
-                        )
-                    )
+                userService.updateUserProfile(
+                        username,
+                        lastNamePrincipal?.name,
+                        firstNamePrincipal?.name,
+                        emailPrincipal?.name
+                )
```
gschueler, line 65: why change this from the event style? the event is also sent by RundeckPreauthSuccessEventHandler, is there a reason to do it differently here?

Jesus-Osuna-M: If the updateUserProfile is made here with the notify as it was before on the first login with LDAP a duplicate record will be generated on the RDUSER table since both transactions are happening at the same time, I  also had to add a Synchronized to the updateUserProfile method to avoid this behavior but wouldn't take effect if the notify is used here.

gschueler: then the same problem could happen with the Preauth handler?  I think that change only avoids the problem, perhaps we should add the unique constraint to the username

rundeckapp/src/test/groovy/org/rundeck/app/providers/GormUserDataProviderSpec.groovy
```diff
@@ -37,6 +37,32 @@ class GormUserDataProviderSpec extends RundeckHibernateSpec implements DataTest
...
+        User savedUser = new User(login: loginName)
+        savedUser.save()
+
+        and:
+        provider.configurationService = Mock(ConfigurationService) {
+            1 * getBoolean(provider.NAME_CASE_SENSITIVE_ENABLED, false) >> caseSensitiveParam
+        }
+
+        when:
+        provider.findOrCreateUser(caseSensitiveParam ? loginName : loginName.toUpperCase())
```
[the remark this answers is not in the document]
Jesus-Osuna-M: updated with suggestions

rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormUserDataProvider.groovy
```diff
@@ -47,8 +49,9 @@ class GormUserDataProvider implements UserDataProvider {
 
     @Override
     @Transactional
+    @Synchronized
     User findOrCreateUser(String login) throws DataAccessException {
-        User user = User.findByLogin(login)
+        User user = isLoginNameCaseSensitiveEnabled() ? User.findByLogin(login) : User.findByLoginIlike(login)
```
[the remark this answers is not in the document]
Jesus-Osuna-M: solved, new method added.

rundeckapp/src/main/groovy/org/rundeck/security/RundeckJaasAuthenticationSuccessEventListener.groovy
Jesus-Osuna-M: Yes, I was just looking for a way to prevent this behavior since adding the constraint will cause an error in clients that already have duplicated login names on RDUSER table. I tested the proposed changes and worked to avoid the creation of duplicated users without causing any error in DB with duplicated records. If we want to add this constraint we need to be sure that the user DB doesn't have already a duplicated login name.

rundeckapp/src/test/groovy/org/rundeck/app/providers/GormUserDataProviderSpec.groovy
```diff
@@ -50,17 +50,16 @@ class GormUserDataProviderSpec extends RundeckHibernateSpec implements DataTest
...
-        } else {
-            User.findAll().size() == 1
-        }
+        User.findAll().size() == userCountSpec
+        rdUser.login == userDataSpec
 
         where:
         caseSensitiveParam << [false, true]
+        userCountSpec << [1,2]
+        userDataSpec << ["loginName1","LOGINNAME1"]
```
gschueler, line 62: shouldn't this just be 
```suggestion
        userDataSpec = "LOGINNAME1"
```

Jesus-Osuna-M: I'm validating loginName1 because, on the first test, the search is case-insensitive, and even if I'm searching for the user as LOGINNAME1 the method shouldn't create a new user instead should find the existing one that is loginName1 on the other case where the search is case-sensitive since there is no coincidence with that user the method will create a user LOGINNAME1. That's why I also added the userCountSpec, to validate that on the first search, the user is being found and on the second search, the user is created since it doesn't exist.

PR #9030, opened 2024-03-25
RSE-1074 QAF Migration: Default File Copier

DarwisNarvaezDev:
In the PR:

1. check file copier list
2. scp file copier config
3. Sshj SCP File copier config
4. Stub file copier config
5. Script file copier config
6. WinRM file copier config
7. OpenSsh file copier config


ronaveva:
already covered

PR #9048, opened 2024-04-08
Run-2255 add unit tests for RundeckInfo component

jayas006:

This is an enhancement. I have created `RundeckInfo.spec.ts` to check if the` RundeckInfo `components of our project works right. This helps us make sure everything in `RundeckInfo` is working well.

In the `RundeckInfo.spec.ts` file, I wrote tests to check different parts of RundeckInfo. These tests help us find problems early and make sure we don't mess up the RundeckInfo when we change other things in the future.

**Additional context**


rundeckapp/grails-spa/packages/ui-trellis/src/library/components/widgets/rundeck-info/RundeckInfo.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+      logo: latestProps.logo,
+      number: latestProps.full,
+      tag: latestProps.tag,
+      logocss: latestProps.logocss,
+    });
+  });
+
+  it("renders the correct url", async () => {
+    const wrapper = await mountRundeckInfo();
+    expect(wrapper.vm.welcomeUrl()).toBe("mocked-url");
```
smartinellibenedetti, line 102: Instead of checking that the method returns "mocked-url", check that there's an <a> with the href of mocked-url

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/widgets/rundeck-info/RundeckInfo.spec.ts
```diff
@@ -0,0 +1,104 @@
+import { shallowMount } from "@vue/test-utils";
+import RundeckInfo from "./RundeckInfo.vue";
+
+jest.mock("../../../rundeckService", () => ({
+  getRundeckContext: jest.fn().mockReturnValue({ rdBase: "mocked-rdBase" }),
+  url: jest.fn().mockReturnValue("mocked-url"),
```
smartinellibenedetti, line 6:
```suggestion
  url: jest.fn().mockReturnValue("http://localhost"),
```

or another url

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/widgets/rundeck-info/RundeckInfo.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+}));
+
+const mountRundeckInfo = async (props?: Record<string, any>) => {
+  return shallowMount(RundeckInfo, {
```
smartinellibenedetti, line 10: Was there an issue that prevented you from using mount? So far from what I remember, tests seemed to be running fine with it.

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/widgets/rundeck-info/RundeckInfo.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+  });
+  it("renders RundeckVersions information correctly ", async () => {
+    const versionProps = {
+      logo: false,
+      number: "1.0.0",
+      tag: "stable",
+    };
+    const wrapper = await mountRundeckInfo({ version: versionProps });
+    const versionComponent = wrapper.findComponent({ name: "RundeckVersion" });
+    expect(versionComponent.exists()).toBe(true);
```
smartinellibenedetti, line 45: 2 things:

- the text you write in the "it" is meant to describe what the test does. In this case "renders information correctly". To check this, it's not enough to only assert if that component exists, especially since this component will actually generate the text it will display by combining the properties from the versionProps.

- secondly, you should check for the final dom elements, in this case, find the span inside versionComponent and check that the text generated is what it should be.

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/widgets/rundeck-info/RundeckInfo.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+      showId: true,
+    };
+    const wrapper = await mountRundeckInfo({ server: serverProps });
+    const serverComponent = wrapper.findComponent({ name: "ServerDisplay" });
+    expect(serverComponent.props()).toEqual({
+      name: serverProps.name,
+      glyphicon: serverProps.icon,
+      uuid: serverProps.uuid,
+      showId: serverProps.showId,
+    });
```
smartinellibenedetti, line 61:
```suggestion
    expect(serverComponent.props()).toEqual(serverProps);
```

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/widgets/rundeck-info/RundeckInfo.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+    const latestVersionComponent = wrapper.findComponent({
+      name: "RundeckVersion",
+    });
+
+    expect(latestVersionComponent.props()).toEqual({
+      logo: latestProps.logo,
+      number: latestProps.full,
+      tag: latestProps.tag,
+      logocss: latestProps.logocss,
+    });
```
smartinellibenedetti, line 97:
```suggestion
    expect(latestVersionComponent.props()).toEqual(latestProps);
```

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/widgets/rundeck-info/RundeckInfo.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+    const latestProps = {
+      logo: false,
+      full: "1.0.0",
+      tag: "GA",
+      logocss: "rdicon",
+    };
+    const wrapper = await mountRundeckInfo({
+      latest: latestProps,
+    });
+    const latestVersionComponent = wrapper.findComponent({
```
smartinellibenedetti, line 88: it's a good practice to name the const with the name of the component 
```suggestion
    const rundeckVersionComponent = wrapper.findComponent({
```

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/widgets/rundeck-info/RundeckInfo.spec.ts
smartinellibenedetti: Would like you to add another expect here, checking that the span of rundeckVersion has the expected text

smartinellibenedetti (commented):
It's a good start, left comments and pointers. Also, all the properties "versionProps", "serverProps", "VersionDisplay", "latestProps" can be moved to the mount method to make the tests cleaner.

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/widgets/rundeck-info/RundeckInfo.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+      full: "2.0.0",
+      tag: "GA",
+    };
+    const wrapper = await mountRundeckInfo({
+      latest: latestProps,
+    });
+
+    const latestSection = wrapper
+      .find(".rundeck-info-widget__latest")
+      .findComponent({ name: "RundeckVersion" });
```
smartinellibenedetti, line 96: You don't need to the .find here, the .findComponent method will find the component as long it's rendered. Changes to do:

- remove `.find(".rundeck-info-widget__latest")`
- rename latestSection to rundeckVersionComponent and update the rest of this test to use the new name

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/widgets/rundeck-info/RundeckInfo.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+    expect(latestSection.exists()).toBe(true);
+
+    expect(latestSection.props("number")).toEqual(latestProps.full);
+    expect(latestSection.props("tag")).toEqual(latestProps.tag);
+  });
+
+  it("should have an anchor with the correct href", async () => {
+    const wrapper = await mountRundeckInfo();
+    const expectedHref = "http://localhost";
+    const anchorElement = wrapper.find(`a[href="${expectedHref}"]`);
```
smartinellibenedetti, line 106: So ideally here you don't want to select the a by the property you're about to test, but by a data-test-id (it's more resilient. Changes to do:

- add a data-test-id for the `<a>` you're going to test in `RundeckInfo.vue`;
- update test to find this a by the data-test-id;

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/widgets/rundeck-info/RundeckInfo.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+  getRundeckContext: jest.fn().mockReturnValue({ rdBase: "mocked-rdBase" }),
+  url: jest.fn().mockReturnValue("http://localhost"),
+}));
+
+const mountRundeckInfo = async (props?: Record<string, any>) => {
+  return mount(RundeckInfo, {
+    props: {
+      appInfo: { title: "Rundeck", logocss: "some-css" },
+      ...props,
+    },
```
smartinellibenedetti, line 14: By my previous comment in this PR, I'd like to move the props here. Ie:

```suggestion
    props: {
      appInfo: { title: "Rundeck", logocss: "some-css" },
      version: { logo: false,
      logocss: "some-css",
      number: "1.0.0",
      tag: "stable",
      title: "Rundeck", },
      /// add the other props here
      ...props,
    },
```

The goal is to keep tests dry, but also mount the components with everything you need for the tests, while having the ability to override something if needed.

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/widgets/rundeck-info/RundeckInfo.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+      latest: latestProps,
+    });
+
+    const latestSection = wrapper
+      .find(".rundeck-info-widget__latest")
+      .findComponent({ name: "RundeckVersion" });
+    expect(latestSection.exists()).toBe(true);
+
+    expect(latestSection.props("number")).toEqual(latestProps.full);
+    expect(latestSection.props("tag")).toEqual(latestProps.tag);
```
smartinellibenedetti, line 100: Also, would like you to add another expect here, checking that the span of rundeckVersion has the expected text

smartinellibenedetti (commented):
There's still work to do, but you're on the right track. I've left some new pointers, let me know if you have any doubts.

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/widgets/rundeck-info/RundeckInfo.spec.ts
jayas006: > So ideally here you don't want to select the a by the property you're about to test, but by a data-test-id (it's more resilient. Changes to do:
> 
> * add a data-test-id for the `<a>` you're going to test in `RundeckInfo.vue`;
> * update test to find this a by the data-test-id;

Fixed

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/widgets/rundeck-info/RundeckInfo.spec.ts
jayas006: > Also, would like you to add another expect here, checking that the span of rundeckVersion has the expected text
I have question regarding latest release: 
In the latest release even though tag is included but the output it is giving "Rundeck and number". Then I checked the computed property and it seems like tag is not included  
```
computed: {
    cssclass(): string {
      return `${this.logocss} app-logo`;
    },
    text(): string {
      const { title, number, tag } = this;
      let text = "";
      text += `${title ? title + " " : ""}`;
      text += `${number}`;
      return text;
    },
  },
```
Two ways I am thinking to fix it. Either I include tag in the computed property or don't use tag in the test as the actual output it is asking `{ title: "Rundeck", full: "1.0.0" }`

I am sending fix. Please review and suggest. Thanks

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/widgets/rundeck-info/RundeckInfo.spec.ts
jayas006: > You don't need to the .find here, the .findComponent method will find the component as long it's rendered. Changes to do:
> 
> * remove `.find(".rundeck-info-widget__latest")`
> * rename latestSection to rundeckVersionComponent and update the rest of this test to use the new name

Done: 
 removed `.find(".rundeck-info-widget__latest")` 
 renamed latestSection to rundeckVersionComponent and update the rest of this test to use the new name

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/widgets/rundeck-info/RundeckInfo.spec.ts
```diff
@@ -120,6 +124,7 @@ describe("RundeckInfo", () => {
     expect(rundeckVersionComponent.exists()).toBe(true);
     expect(rundeckVersionComponent.props("title")).toEqual(latestProps.title);
     expect(rundeckVersionComponent.props("number")).toEqual(latestProps.full);
+    expect(rundeckVersionComponent.props("tag")).toEqual(latestProps.tag);
```
jayas006, line 127: included tag now here and also added one more expect for checking that the span of rundeckVersion has the expected text which is in previous commit. Please suggest If I am missing something.

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/widgets/rundeck-info/RundeckInfo.spec.ts
jayas006: > > Also, would like you to add another expect here, checking that the span of rundeckVersion has the expected text
> > I have question regarding latest release:
> > In the latest release even though tag is included but the output it is giving "Rundeck and number". Then I checked the computed property and it seems like tag is not included
> 
> ```
> computed: {
>     cssclass(): string {
>       return `${this.logocss} app-logo`;
>     },
>     text(): string {
>       const { title, number, tag } = this;
>       let text = "";
>       text += `${title ? title + " " : ""}`;
>       text += `${number}`;
>       return text;
>     },
>   },
> ```
> 
> Two ways I am thinking to fix it. Either I include tag in the computed property or don't use tag in the test as the actual output it is asking `{ title: "Rundeck", full: "1.0.0" }`
> 
> I am sending fix. Please review and suggest. Thanks

realized that span of rundeckVersion has the expected text of` title and number` and no need to test tag there. Added tag in my new commit which I removed in this commit due to confusion.

smartinellibenedetti (changes requested):
Some comments were not addressed so I will recap them here. As we spoke before, the reason to create a method on the top of the test suite to return the mounted component is to avoid repetition/having to pass the same props every time you write a test.

So unless there is a scenario where the component needs to be mounted with a different set of values (like on line 44), there's no need to declare versionProps, latestProps, etc.

Also, props required by a component, may not be a prop for their child and vice-versa. Therefore, it's important to check how these values are being defined to avoid the scenario of generating console warnings that mismatches are happening (by either passing extra props or lacking the required ones).

`ShowId` isn't a prop for `RundeckInfo`, but for `ServerDisplay` component, that when it's not passed, it has the default value of true. As there isn't anything in RundeckInfo that points out that it also requires this prop, we shouldn't pass this property when mounting this component.

Lastly, when asserting the value of an object, it's fine to pass a variable, but when asserting the value of a string, please write the literal value to highlight the intention.

jayas006:
Thank you @smartinellibenedetti , sending fix in few minutes

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/widgets/rundeck-info/RundeckInfo.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+    const wrapper = await mountRundeckInfo();
+
+    const rundeckVersionComponent = wrapper.findComponent({
+      name: "RundeckVersion",
+    });
+    expect(rundeckVersionComponent.exists()).toBe(true);
+
+    expect(rundeckVersionComponent.find("span").text()).toBe("Rundeck 1.0.0");
+  });
+  it("renders ServerDisplay correctly with default props", async () => {
```
smartinellibenedetti, line 66:
```suggestion
  it("renders the server's shortened uuid along with a title attribute", async () => {
```

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/widgets/rundeck-info/RundeckInfo.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+    const wrapper = await mountRundeckInfo();
+    const serverComponent = wrapper.findComponent({ name: "ServerDisplay" });
+    expect(serverComponent.exists()).toBe(true);
+
+    expect(
+      serverComponent.find('[data-test-id="server-title"]').attributes("title"),
+    ).toBe("paperclip-uu / uuid1");
```
smartinellibenedetti, line 73: Please add here another expect that checks that there's a span with the text equal to the short uuid

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/widgets/rundeck-info/RundeckInfo.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+  });
+
+  it("renders version display correctly", async () => {
```
smartinellibenedetti, line 76:
```suggestion
  it("renders Rundeck's version display name", async () => {
```

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/widgets/rundeck-info/RundeckInfo.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+    });
+    expect(versionDisplayComponent.props("text")).toBe("Erebus red glass");
+  });
+  it("renders the latest release information correctly", async () => {
+    const wrapper = await mountRundeckInfo();
+    const rundeckVersionComponent = wrapper.findComponent({
+      name: "RundeckVersion",
+    });
+    expect(rundeckVersionComponent.exists()).toBe(true);
+    expect(rundeckVersionComponent.find("span").text()).toBe("Rundeck 1.0.0");
```
smartinellibenedetti, line 89: RundeckInfo renders 2 RundeckVersion components - one that you've tested before, that will render "Rundeck 1.0.0", and one that renders the latest release info. 

This test should be checking the text of the second RundeckVersion. To do so, you can use the method [findAllComponents](https://test-utils.vuejs.org/api/#findAllComponents) as findComponent will always return the first occurrence that matches what you're searching.

smartinellibenedetti (changes requested):
Good progress! Left a few suggestions for the tests' descriptions that could be a little bit more clear. There are 2 small fixes to be done but afterwards, it's good to merge!

rundeckapp/grails-spa/package-lock.json
```diff
@@ -0,0 +1,1108 @@
+{
```
smartinellibenedetti, line 1: Also, could you please delete this package-lock?

jayas006:
@smartinellibenedetti thanks for the pointers. I've fixed it.

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/widgets/rundeck-info/RundeckInfo.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+      name: "RundeckVersion",
+    });
+    const latestReleaseComponent = rundeckVersionComponents[1];
+    expect(latestReleaseComponent.exists()).toBe(true);
+    expect(latestReleaseComponent.find("span").text()).toBe("v5.2.0-20240410");
+  });
+  it("renders the link with the correct href", async () => {
+    const wrapper = await mountRundeckInfo();
+    const anchorElement = wrapper.find('[data-test-id="welcome-link"]');
+    expect(anchorElement.exists()).toBe(true);
```
smartinellibenedetti, line 97: Don't need to change this but a note for the future: if an element doesn't have a v-if, you don't need to assert every time that the element exists

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/widgets/rundeck-info/RundeckInfo.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+  });
+  it("renders the server's shortened uuid along with a title attribute", async () => {
+    const wrapper = await mountRundeckInfo();
+    const serverComponent = wrapper.findComponent({ name: "ServerDisplay" });
+    expect(serverComponent.exists()).toBe(true);
+
+    expect(
+      serverComponent.find('[data-test-id="server-title"]').attributes("title")
+    ).toBe("paperclip-uu / uuid1");
+    expect(serverComponent.find("span").text()).toBe("uulocalHost");
```
smartinellibenedetti, line 75: Heads-up, this assertion is wrong. And furthemore, this span needs a data-test-id

smartinellibenedetti (changes requested):
Please delete the package-lock.json

smartinellibenedetti (approved):
good to go!

PR #9050, opened 2024-04-09
Fix hibernate user session on scm context

JoseOrtiz:

Bugfix for session not handled properly when context is Scm 

Use hibernate capabilities to attach a User query in the thread again

I didn't consider any other option
**Additional context**

Usage of `findByLogin` does not require further writes so no problem with return value not being attached to the context

### DIFF
--- a/rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormUserDataProvider.groovy
+++ b/rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormUserDataProvider.groovy
@@ -239,7 +239,9 @@ class GormUserDataProvider implements UserDataProvider {
 
     @Override
     RdUser findByLogin(String login) {
-        return User.findByLogin(login)
+        User.withNewSession {
+            return User.findByLogin(login)
+        }
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormUserDataProvider.groovy
```diff
@@ -239,7 +239,9 @@ class GormUserDataProvider implements UserDataProvider {
 
     @Override
     RdUser findByLogin(String login) {
-        return User.findByLogin(login)
+        User.withNewSession {
+            return User.findByLogin(login)
+        }
```
ocerda, line 244: I see that there is another method in the provider using `withNewSession` and that method is `getEmailWithNewSession`, which specifies that is using `withNewSession`. Question: is there any impact of starting a new session for every usage of `findByLogin`, or would it be better to add a method `findByLoginWithNewSession` just for this case?

JoseOrtiz: To be honest, I did that `getEmailWithNewSession` method not knowing that it was a specific thing of hibernate (the `withNewSession` stuff). I'd prefer to refactor that method that create another one .
Grails defines `withNewSession` as [this](http://docs.grails.org/latest/ref/Domain%20Classes/withNewSession.html), it will use the same transactional connection, and the found user will be attached on the session inside that block, but it will be deattached right away as the user is returned, also we don't trigger a flush on the database by making a query on the current session

ocerda (approved):

Issue #7501, opened 2022-01-26
JVM_MAX_RAM_FRACTION and JVM_MAX_RAM_PERCENTAGE are not in the same unit


ffe-enghouse:
**Describe the bug**

From the main README.md, we understand that
* `JVM_MAX_RAM_PERCENTAGE` is defined as an integer between `0` and `100` https://github.com/rundeck/rundeck/blob/578278f48d33fa380d1bb6d2cb9c28a4426a88ea/docker/official/README.md#L100
* `JVM_MAX_RAM_FRACTION` is defined as a float between `0` and `1` https://github.com/rundeck/rundeck/blob/578278f48d33fa380d1bb6d2cb9c28a4426a88ea/docker/official/README.md#L44

Moreover, by default, `JVM_MAX_PERCENTAGE` is set to `75`, unless if `JVM_MAX_RAM_FRACTION` is set:
https://github.com/rundeck/rundeck/blob/9d5c61581da74dc871c4bb06ee19f977c873729d/docker/official/lib/includes/100_defaults.sh#L7

But if I set `JVM_MAX_RAM_FRACTION` to `1`, I expect to have  `JVM_MAX_RAM_PERCENTAGE` set to `100`, but it is set to `1`.

So either modify:
* the README.md in order to tell that `JVM_MAX_RAM_FRACTION` is an integer between `0` and `100`,
* or modify 100_defaults.sh in order to x100 the `JVM_MAX_RAM_FRACTION` value before setting it to `JVM_MAX_RAM_PERCENTAGE`,
*  or remove `JVM_MAX_RAM_FRACTION` value in the README.md, keep only `JVM_MAX_RAM_PERCENTAGE`, and update the 100_defaults.sh (best option IMHO)

**My Rundeck detail**

* Rundeck version: 3.4.9
* install type: docker
* OS Name/version: ubuntu 20.04
* DB Type/version: postgres

**To Reproduce**
1. Set `JVM_MAX_RAM_FRACTION` to 1
2. Launch rundeck
3. get the launch command: `ps -ef` 
    You will obtain: `java -XX:MaxRAMPercentage=1 [...]`
4. Remove `JVM_MAX_RAM_FRACTION`
5. get the launch command: `ps -ef` 
    You will obtain: `java -XX:MaxRAMPercentage=75 [...]`

**Expected behavior**
[...]

PR #9052, opened 2024-04-09
RUN-2270: Fix docker image ignoring value of JVM_MAX_RAM_PERCENTAGE

ahormazabal:
The docker official image is ignoring the value set at the `JVM_MAX_RAM_PERCENTAGE` environment variable when setting the value for the `-XX:MaxRAMPercentage` argument, but instead is incorrectly used the value of the deprecated `JVM_MAX_RAM_FRACTION` env variable.

This could potentially mislead administrators into think the rundeck container is using less memory than what is using in reality, as the final value for `MaxRAMPercentage` will be 75% regardless of the value of `JVM_MAX_RAM_PERCENTAGE`

### DIFF
--- a/docker/official/lib/entry.sh
+++ b/docker/official/lib/entry.sh
@@ -71,7 +71,6 @@ if [[ -n "${EXEC_CMD}" ]] ; then
 fi
 
 exec java \
-    -XX:+UnlockExperimentalVMOptions \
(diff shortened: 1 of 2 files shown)


jbrookspd (approved):

ahormazabal:
Fixes https://github.com/rundeck/rundeck/issues/7501

PR #9059, opened 2024-04-12
RSE-774 Fix: Users logged from LDAP are duplicate

Jesus-Osuna-M:
Problem:
When accessing Rundeck for the first time if the users had configured the rundeck.security.syncLdapUser=true property users were duplicated on the RDUSER table.

Solution:
On the SetUserInterceptor remove code that was causing a duplicated register, this part of the code wasn't necessary since the user is already being created on the RundeckJaasAuthenticationSuccessEventListener.groovy.

The use of the method getUserByLoginOrCreate was replaced by findOrCreateUser since both methods had the same logic

### DIFF
--- a/rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormUserDataProvider.groovy
+++ b/rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormUserDataProvider.groovy
@@ -3,6 +3,7 @@ package org.rundeck.app.data.providers
 
 import grails.compiler.GrailsCompileStatic
 import grails.gorm.DetachedCriteria
+import groovy.transform.Synchronized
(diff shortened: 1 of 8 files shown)


rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormUserDataProvider.groovy
```diff
@@ -47,6 +48,7 @@ class GormUserDataProvider implements UserDataProvider {
 
     @Override
     @Transactional
+    @Synchronized
```
gschueler, line 51: can we remove this? I think it doesn't solve the issue, but may be misleading to someone who thinks it does solve it.

PR #9062, opened 2024-04-15
RSE-864 Fix: Case sensitive in users when used LDAP OSS

### LINKED ISSUE #6733: Usernames are case sensitive
**Describe the bug**
Usernames in the Rundeck database are case sensitive. This creates issues when interacting with LDAPuser stores that are case insensitive - we can create multiple user records for the same actual user when that happens.

**My Rundeck detail**

* Rundeck version: All of them, current as of 3.3.7+
* install type: All
* OS Name/version: all
* DB Type/version: all

**To Reproduce**

Log into Rundeck through the Rundeck login screen with LDAP auth enabled.

Log out.

Log in with the same user, but capitalized differently.

Log out.

Rundeck creates two internal user records based on the different capitalized usernames.

Jesus-Osuna-M:
Related to:
https://pagerduty.atlassian.net/browse/RSE-961

Problem:
The username on rundeck is case sensitive while usernames un LDAP are case insensitive this difference generated duplicated users on the RDUSER table since.

Solution:
The use of the method getUserByLoginOrCreate was replaced by findOrCreateUser since both methods had the same logic.

A new feature flag was added to the config.properties

    rundeck.feature.caseInsensitiveUsername.enabled=true

By default, this flag will be false meaning that username on login will be case sensitive.

Fixes #6733


gschueler (changes requested):

rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormUserDataProvider.groovy
```diff
@@ -309,4 +312,41 @@ class GormUserDataProvider implements UserDataProvider {
     def getSessionIdRegisterMethod() {
         configurationService.getString(SESSION_ID_METHOD, 'hash')
     }
+
+    /**
+     Checks if login name case sensitivity is enabled.
+     @return {@code true} if login name case sensitivity is enabled, {@code false} otherwise.
+     */
+    def isLoginNameCaseSensitiveEnabled(){
+        return configurationService?.getBoolean(NAME_CASE_INSENSITIVE_ENABLED,false)
```
gschueler, line 321: the feature check should use the `featureService.featurePresent` instead of direct check with the configurationService

core/src/main/java/com/dtolabs/rundeck/core/config/RundeckConfigBase.java
```diff
@@ -437,6 +437,7 @@ public static class RundeckFeatureConfig {
         Enabled legacyUi = new Enabled();
         Enabled legacyXml = new Enabled();
         Enabled apiProjectConfigValidation = new Enabled();
+        Enabled caseInsensitiveUsername = new Enabled();
```
gschueler, line 440: Also, please add an entry in the `Features` class

gschueler (changes requested):

rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormUserDataProvider.groovy
```diff
@@ -21,19 +23,22 @@ import rundeck.User
 import rundeck.services.ConfigurationService
 import rundeck.services.FrameworkService
 import rundeck.services.data.UserDataService
+import rundeck.services.feature.FeatureService
```
gschueler, line 26: you can use `com.dtolabs.rundeck.core.config.FeatureService` interface instead

rundeckapp/src/test/groovy/org/rundeck/app/providers/GormUserDataProviderSpec.groovy
```diff
@@ -37,6 +39,30 @@ class GormUserDataProviderSpec extends RundeckHibernateSpec implements DataTest
...
+        when:
+        provider.findOrCreateUser(loginName.toUpperCase())
+
+        then:
+        User.findAll().size() == userCountSpec
+
+        where:
+        caseSensitiveParam << [true, false]
+        userCountSpec << [1, 2]
+        userDataSpec << ["loginName1", "LOGINNAME1"]
```
gschueler, line 63: not used in test
```suggestion
```

gschueler (approved):

PR #9092, opened 2024-04-30
RPL-7: add TZ to scheduled jobs in tests

ronaveva:
minor improvement to allow TZ specific when creating scheduled jobs


Jesus-Osuna-M (approved):
LGTM

PR #9105, opened 2024-05-06
RSE-961 Login counting multiple times in user classes

Jesus-Osuna-M:
Related to:

Problem:
After solving the issue where users were duplicated when authenticating with LDAP, an error was thrown on user/profile page when authenticating with AD.

Solution:
Evaluate case sensitiveness on the validateUserExists method, this change avoid an error 500 that was generated on the user/profile page. It was only generated when authenticating from AD 

Extra info on :
https://pagerduty.atlassian.net/browse/RSE-961

### DIFF
--- a/rundeckapp/src/test/groovy/org/rundeck/app/providers/GormUserDataProviderSpec.groovy
+++ b/rundeckapp/src/test/groovy/org/rundeck/app/providers/GormUserDataProviderSpec.groovy
@@ -289,16 +289,19 @@ class GormUserDataProviderSpec extends RundeckHibernateSpec implements DataTest 
     @Unroll
     def "Should validate user exists"() {
         given:
-        User savedUser = new User(login: "saved")
+        User savedUser = new User(login: login)
         savedUser.save()
+        provider.featureService = Mock(FeatureService){
+            1 * featurePresent(Features.CASE_INSENSITIVE_USERNAME) >> caseSensitive
+        }
         when:
-        Boolean userExists = provider.validateUserExists(login)
+        Boolean userExists = provider.validateUserExists(login.toUpperCase())
         then:
         userExists == expect
         where:
-        login   | expect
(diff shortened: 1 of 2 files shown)


rundeckapp/src/test/groovy/org/rundeck/app/providers/GormUserDataProviderSpec.groovy
```diff
@@ -289,16 +289,19 @@ class GormUserDataProviderSpec extends RundeckHibernateSpec implements DataTest
...
+        Boolean userExists = provider.validateUserExists(login.toUpperCase())
         then:
         userExists == expect
         where:
-        login   | expect
-        "login" | false
-        "saved" | true
+        login   | expect   | caseSensitive
+        "user1" | false    | false
+        "user2" | true     | true
```
gschueler, line 304: should also keep a case where validateUserExists parameter is a missing user name

gschueler (changes requested):

rundeckapp/src/test/groovy/org/rundeck/app/providers/GormUserDataProviderSpec.groovy
```diff
@@ -289,16 +289,24 @@ class GormUserDataProviderSpec extends RundeckHibernateSpec implements DataTest
...
         when:
-        Boolean userExists = provider.validateUserExists(login)
+        Boolean userExists = provider.validateUserExists(login.toUpperCase())
         then:
         userExists == expect
         where:
-        login   | expect
-        "login" | false
-        "saved" | true
+        login   | expect   | caseSensitive | persitUser
```
gschueler, line 304: nitpick: `caseSensitive` should be called `caseInsensitive` since the meaning of the feature flag has been inverted.

gschueler (approved):

PR #9109, opened 2024-05-06
RCLOUD-1221: Remove Execution usages of GormExecReportDataProvider

ocerda:
Refactor of GormExecReportDataProvider

Update the GormExecReportDataProvider to remove usage of Execution and ReferencedExecution Domains

**Additional context**
There are non-destructive migrations in this PR, but deploying them could take a long time depending on the size of the tables, as we recently experienced.

### DIFF
--- a/rundeckapp/grails-app/migrations/core/BaseReportSpi.groovy
+++ b/rundeckapp/grails-app/migrations/core/BaseReportSpi.groovy
@@ -25,4 +25,32 @@ databaseChangeLog = {
                 "                  inner join execution ON execution.scheduled_execution_id = scheduled_execution.id\n" +
                 "         WHERE base_report.execution_id = execution.id)")
     }
+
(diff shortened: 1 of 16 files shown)


rundeckapp/grails-app/migrations/core/BaseReportSpi.groovy
```diff
@@ -25,4 +25,28 @@ databaseChangeLog = {
...
     }
+
+    changeSet(author: "rundeckdev", id: "add-execution-uuid-to-base-report") {
+        preConditions(onFail: "MARK_RAN") {
+            not {
+                columnExists(tableName: "base_report", columnName: 'execution_uuid')
+            }
+        }
+        addColumn(tableName: "base_report") {
+            column(name: 'execution_uuid', type: '${varchar255.type}')
```
sjrd218, line 36: You'll need an index on this column since it will be used for lookups.

PR #9111, opened 2024-05-07
Moving BaseNodeFilters and ExtNodeFilters to shared project

sjrd218:
Updating the location of BaseNodeFilters and ExtNodeFilters classes.


grails-rundeck-data-shared/src/main/groovy/rundeck/data/util/NodeFiltersUtil.groovy
```diff
@@ -0,0 +1,37 @@
...
+                result["nodeExclude${v}"]=params["exclude-"+k]
+                value = true
+            }
+        }
+        if(params.filter){
+            result.filter=params.filter
+            value=true
+        }
+        if (value && null!=params.'exclude-precedence') {
+            result.nodeExcludePrecedence = params['exclude-precedence'] == 'true'
```
JoseOrtiz, line 33: Is `params['exclude-precedence']` always a string?

sjrd218: yes

JoseOrtiz (approved):

PR #9115, opened 2024-05-08
RUN-2353-Refactor editProject test to improve readability and maintainability

jayas006:
This is an enhancement of the testing suite for the `EditProjectFile.vue` component.

I've refactored the tests for `EditProjectFile.vue` to improve readability and maintainability. The changes include better organization of test cases, clearer descriptions, and removal of unnecessary global mocks.

An alternative would have been to leave the tests as they were, but this would have made the test suite harder to maintain and understand in the long run.

**Additional context**

### DIFF
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/app/components/readme-motd/EditProjectFile.spec.ts
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/app/components/readme-motd/EditProjectFile.spec.ts
@@ -2,7 +2,17 @@ import { mount } from "@vue/test-utils";
 import EditProjectFile from "./EditProjectFile.vue";
 import * as editProjectFileService from "./editProjectFileService";
 
-jest.mock("@/app/components/readme-motd/editProjectFileService", () => ({
+jest.mock("@/library/rundeckService", () => ({
+  getRundeckContext: () => ({
+    rundeckClient: {},
+  }),
+  url: jest.fn().mockImplementation((path) => {
+    return {
+      href: `http://localhost:4440/${path}`,
+    };
+  }),
+}));
+jest.mock("./editProjectFileService", () => ({
   getFileText: jest.fn().mockResolvedValue({
     success: true,
     contents: "sample file content",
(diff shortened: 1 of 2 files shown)


smartinellibenedetti (commented):
there's still stuff to be cleaned up:
there are expects that are asserting exists, 
there is an override of wrapper.vm.fileText that should be removed
there is a test that isn't using the spy anymore (or that didn't get added before)

Also, you shouldn't have the template hardcoding that string inside the span - you can change the template to render the content of the getValue method (you don't need to add a new var inside data for this)

jayas006:
> there's still stuff to be cleaned up: there are expects that are asserting exists, there is an override of wrapper.vm.fileText that should be removed there is a test that isn't using the spy anymore (or that didn't get added before)
> 
> Also, you shouldn't have the template hardcoding that string inside the span - you can change the template to render the content of the getValue method (you don't need to add a new var inside data for this)

I have tried to fix according to the review pointers. Please let me know if there are any further changes required or if any aspect of the test needs to be updated.

mrdubr (approved):
LGTM

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/readme-motd/EditProjectFile.spec.ts
```diff
@@ -13,143 +23,103 @@ jest.mock("@/app/components/readme-motd/editProjectFileService", () => ({
...
 jest.mock("../../../library/components/utils/AceEditor.vue", () => ({
   name: "AceEditor",
-  functional: true,
-  template: '<span class="ace_text ace_xml"></span>',
+  props: ["value"],
+  template: '<span class="ace_text ace_xml">{{ getValue() }}</span>',
   methods: {
     getValue: jest.fn().mockReturnValue("sample file content"),
   },
 }));
```
smartinellibenedetti, line 33: There are problems with this approach: 

- this mock doesn't match any of the props that the component originally used, nor mock an existing method.  
- furthermore, the way it is implemented it's not possible to override the value nor assert that the component is rendering the data retrieved from the API.

To do:
- change props to "modelValue" instead of "value";
- change template to render just "modelValue";
- remove getValue method;

smartinellibenedetti (changes requested):
It's necessary to adjust the ace editor, otherwise it's not possible to assert that the component is rendering the results fetched from API. Please let me know in case of doubts.

smartinellibenedetti (approved):
🚀  ship it, looking good!

PR #9120, opened 2024-05-14
RUN-2453: Feature rundeck.feature.caseInsensitiveUsername.enabled doesn't show on System configuration page

Jesus-Osuna-M:
On bug bash found that rundeck.feature.caseInsensitiveUsername.enabled can be enabled/disabled on the rundeck-config.properties but not from the GUI on System configuration page.

### DIFF
--- a/rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormUserDataProvider.groovy
+++ b/rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormUserDataProvider.groovy
@@ -336,7 +336,7 @@ class GormUserDataProvider implements UserDataProvider, SystemConfigurable{
     List<SysConfigProp> getSystemConfigProps() {
         return [
                 SystemConfig.builder().with {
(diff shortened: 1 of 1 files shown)


rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormUserDataProvider.groovy
```diff
@@ -336,7 +336,7 @@ class GormUserDataProvider implements UserDataProvider, SystemConfigurable{
     List<SysConfigProp> getSystemConfigProps() {
         return [
                 SystemConfig.builder().with {
-                    key("rundeck."+Features.CASE_INSENSITIVE_USERNAME)
+                     key("rundeck.feature.caseInsensitiveUsername.enabled")
                     .datatype("Boolean")
                     .label("Enable case insensitive on login name")
```
gschueler, line 341: update for clarity
```suggestion
                    .label("Enable case insensitive login name")
```

gschueler (approved):

PR #9121, opened 2024-05-15
RUN-2193, RUN-2194, RUN-2195: :nextUI: convert forecast page

smartinellibenedetti:

This PR is part of the changes necessary for the vue conversion of the schedule forecast page (RUN-2193, RUN-2194 and RUN-2195).

Update the project select dropdown to allow multiple selections. 

ProjectSelect accepts now a mode prop, whose value will determine its behavior. If "single", projectSelect will render links that upon click will take the user to the project page, if "multi", ProjectSelect will emit an array with the names of the selected projects.

Also, ProjectSelectButton can render a label in accordance to the number of projects selected.

Adapt vue-multiselect, which was asked to change in prol of reusing the projectSelect. 

**Additional context**

Will add tests in a separate PR along with the tests for the full component behavior.

### DIFF
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/widgets/project-select/ProjectSelect.vue
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/widgets/project-select/ProjectSelect.vue
@@ -16,34 +16,59 @@
       <Skeleton :loading="!projectStore.loaded">
         <RecycleScroller
           ref="scroller"
-          :items="projectStore.search(searchTerm)"
+          :items="options"
           :item-size="25"
-          :key="projectStore.search(searchTerm).length"
+          :key="`${options.length}scroller`"
           v-slot:default="{ item }"
           key-field="name"
           class="scroller"
         >
           <a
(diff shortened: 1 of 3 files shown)


gschueler (commented):
changes to the API (adding averageDuration to the forecast info) should be put behind a new API version.

Additionally, perhaps the API change should be separated from the UI change here, so it can be tested separately

smartinellibenedetti:
> changes to the API (adding averageDuration to the forecast info) should be put behind a new API version.
> 
> Additionally, perhaps the API change should be separated from the UI change here, so it can be tested separately

Questions:

- should the API version be bumped upon any change on the APIs? Or the gist is, new params should bump the version, while fixes should not?
- happy to separate this PR into a second one, but this will make it harder to test the pro changes until at least one of them get merged. Is that okay?

gschueler:
@smartinellibenedetti a) yes, functional changes should increase API version so users of older API versions have old behavior. Bug fixes would not require that.  b) yes let's do 2 prs

gschueler (approved):

PR #9123, opened 2024-05-16
RUN-2369 Login as local user doesn't evaluate case sensitive feature

Jesus-Osuna-M:
Related to https://github.com/rundeckpro/rundeckpro/pull/3681
 
improve method findUserByLoginCaseSensitivity to avoid undesired logins


mrdubr (approved):
Approved, assuming the test suite has tests for this already

PR #9129, opened 2024-05-17
RCLOUD-1270: Hide version from footer for rba

ocerda:
Change of behavior for rba

Adding the option to hide the version number from the footer

**Additional context**
https://pagerduty.atlassian.net/browse/RCLOUD-1270

### DIFF
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/widgets/rundeck-info/RundeckInfo.spec.ts
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/widgets/rundeck-info/RundeckInfo.spec.ts
@@ -31,6 +31,7 @@ const mountRundeckInfo = async (props = {}) => {
         icon: "knight",
         edition: "Community",
       },
+      hideVersion: false,
       ...props,
     },
   });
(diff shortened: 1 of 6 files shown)


rundeckapp/grails-spa/packages/ui-trellis/src/library/services/featureConfig.ts
```diff
@@ -0,0 +1,6 @@
+import { api } from "./api";
+
+export async function getFeatureEnabled(featureName: string) {
+    const resp =  await api.get(`/feature/${featureName}`);
+    return !!resp.data.enabled;
```
JoseOrtiz, line 5: What happens if `resp` is `null`?

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/widgets/rundeck-info/RundeckInfo.spec.ts
```diff
@@ -108,4 +109,15 @@ describe("RundeckInfo", () => {
...
     expect(latestReleaseComponent.props().tag).toBe("GA");
   });
+  it("does not render release information when hideVersion is true", async () => {
+    const wrapper = await mountRundeckInfo({ hideVersion: true });
+    const rundeckVersionComponent = wrapper.findComponent({
+      name: "RundeckVersion",
+    });
+    const versionDisplayComponent = wrapper.findComponent({
+      name: "VersionDisplay",
+    });
```
JoseOrtiz, line 119: The idea of testing is to check that what is shown to the user is tested, so we need to change this to the right behavior, that some divs inside those components are displayed, so you should change that find elements and then check its existance
```suggestion
    const rundeckVersionComponent = wrapper.find('div.rundeck-version-display');
    const versionDisplayComponent = wrapper.find('span.rundeck-version-icon');
```

Please check if that works if you want to change it, I'm a bit rusty with those kind fo tests 😅

JoseOrtiz (commented):
Ideal approach is to use `data-test-id` for fetching, but is up to you

JoseOrtiz (approved):
Looks good, but could you add a test for the display of `div.rundeck-info-widget__group` element?

PR #9132, opened 2024-05-20
Fix buildUser call

JoseOrtiz:

Bugfix

An ambigous call was made by naming the argument on user controller, it was solved by deleting the `login: user` call

**Additional context**


gschueler (changes requested):
can you add a unit test for the controller method?

gschueler (approved):
👍

Issue #6433, opened 2020-09-04
Add ARM64 support to Rundeck


odidev:
Hi Team,

I am looking to add ARM64 jobs in Travis.
I have checked that the code is currently structured to support the AMD64 platform, hence Travis jobs for ARM64 are failing. Also, it seems that adding ARM64 support to Rundeck would need a good amount of changes in the code.

It will be really helpful if ARM64 support is added to the package. Do you have any plans for the same?

If you are interested, I can add ARM64 support to Rundeck and raise PR.

Also, it will be really helpful if you can share any blockers for this activity.

Please provide your opinion on this.

fdevans:
There are currently not active plans to support Rundeck on ARM64.  As we start to see more requests we'll review the effort to add this.

odidev:
@fdevans , thanks for your response. Please let me know whether you are interested in accepting any PR for adding rundeck support on arm64. I am happy to help on making rundeck available for aarch64.

odidev:
@fdevans I was trying to run travis build for arm64 but it was failing with below error:
standard_init_linux.go:211: exec user process caused "exec format error"

This package uses rundeckapp/testdeck:base-latest image which is not available for arm64. I have found the [github](https://github.com/rundeck/testdeck) link for this image, but I am unable to find its Dockerfile in it to add support for arm64.

Could you please point me to Dockerfile source, so that I can add arm64 support in it and raise PR?

foxhunt72:
Please i would second this request

DarwisNarvaezDev:
Hi, im currently having issues to set docker images of RD on my m1 mac. So i add my request too!

fdevans:
@odidev Yes, we do review community PRs.

agarwalpranab:
Would love this. +1

matttobes:
I would like to see this

calebcheng:
Got Apple Macbook Pro with M1 chip and really need this. I believe there will be many other people having the same issue

augustgerro:
I would like to see M silicon as well

florian-asche:
There is an existing java war file. That should work on arm64 or am i wrong?

c-geek:
I've successfully built Rundeck for ARM (here on an M1 Mac), I won't make a PR but you can build it too using this branch:

    git clone https://github.com/c-geek/rundeck.git -b fix/6433/arm
    cd rundeck
    ./gradlew :docker:officialBuild

And then run it:

[...]

PR #9141, opened 2024-05-23
[RUN-2508] Enables building arm64 Docker images locally

ChuckCrawford:
This is a small enhancement to enable building `rundeck` Docker images for an `arm64` architecture.

This change enables building `arm64` images locally as described in [README.md](https://github.com/rundeck/rundeck/blob/main/README.md). It is a prerequisite for official CI builds of arm64 images (as requested in https://github.com/rundeck/rundeck/issues/6433).

Adds the appropriate build of https://github.com/krallin/tini into the `ubuntu-base` image based on the Docker build's target architecture.

This seems like a practical first step?

**Additional context**
Before this enhancement, attempting to build and run the project from an Apple Silicon machine results in the following error when starting the container:

```
rosetta error: failed to open elf at /lib64/ld-linux-x86-64.so.2

      "Entrypoint": [
        "/tini",
        "--",
        "docker-lib/entry.sh"
      ],
```

### DIFF
--- a/docker/ubuntu-base/Dockerfile
+++ b/docker/ubuntu-base/Dockerfile
@@ -14,6 +14,8 @@ RUN go install github.com/HeavyHorst/remco/cmd/remco@$REMCO_VERSION
 ######################
 FROM ubuntu:22.04
 
+ARG TARGETARCH
+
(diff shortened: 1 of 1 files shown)


gschueler (approved):
Works fine still on intel

PR #9143, opened 2024-05-24
RUN-2193 RUN-2194 RUN-2195: update API, update tests and bump API version

smartinellibenedetti:
This PR is part of the changes necessary for the vue conversion of the schedule forecast page (RUN-2193, RUN-2194 and RUN-2195).

Update the job forecast endpoint to also return the average duration, to avoid having to make an extra request to an Ajax endpoint to gather this information.

To keep using the Ajax endpoint or make a new one.

**Additional context**


gschueler (approved):

gschueler:
It looks like some functional tests expect the old "latest" api version, we will need to fix those

smartinellibenedetti:
Docs updated here: https://github.com/rundeck/docs/pull/1487

gschueler (changes requested):
one additional test: if the api version < 48 the forecast endpoint would be expected not to return the averageDuration.

gschueler (approved):

PR #9162, opened 2024-06-03
RUN-2554: Restores Tomcat JDBC pool.

ChuckCrawford:
During the upgrade to Grails 6.1.2 (PR https://github.com/rundeck/rundeck/pull/9068) the `org.apache.tomcat:tomcat-jdbc` runtime dependency was removed by a developer thinking it was unused.

It turns out that the `tomcat-jdbc` pool was detected and configured by Spring's auto-configuration. The removal of this dependency changed the primary DataSource used by the job scheduler from `tomcat-jdbc` to `HikariCP` (which has a different default configuration).

This PR restores the `tomcat-jdbc` pool and upgrades the minor version.

Restores the `tomcat-jdbc` pool as the primary `DataSource` used by the application.

We considered rolling forward with the `HikariCP` connection pool (a change that might be nice in the future) but we would prefer to do that in a controlled way in the future.

**Additional context**
I have verified that this change restores the `org.apache.tomcat.jdbc.pool.DataSource` and fixes some issues we were observing with the primary connection pool.

### DIFF
--- a/rundeckapp/build.gradle
+++ b/rundeckapp/build.gradle
@@ -254,6 +254,7 @@ dependencies {
     runtimeOnly 'org.postgresql:postgresql:42.7.2'
     runtimeOnly 'org.rundeck.hibernate:rundeck-oracle-dialect:1.0.0'
 
+    runtimeOnly "org.apache.tomcat:tomcat-jdbc:9.0.89"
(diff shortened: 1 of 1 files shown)


ronaveva (approved):
LGTM

PR #9171, opened 2024-06-05
RUN-2540: Webpack fix - ensure that assets in css still use relative paths

smartinellibenedetti:

When deploying ui-trellis as part of enterprise, assetsDir needs to have the value "../../" and then there's a task that will ensure that the images and fonts are moved to the correct location, but while running `npm run dev`, the paths are not correct due to `assetsDir`. 

~~Add env variables to be able to setup the public path to the original value (before webpack update) during dev.~~

Adjust mini-extract-css plugin's publicPath property, forcing it to be "../../". Also, adjust the publicPath value back to what it was originally before webpack update.

As we want to avoid absolute URLs in the scenario where project will be run via docker with a contextPath. The solution needed to keep relative urls - I tried to move the assets inside CSS folder, but webpack would resolve them wrongly.

The positive aspect of this solution is that the URLs are stable and there's no need for moving assets.

**Additional context**

Symptoms before this PR:

by deleting everything from the grails-app/provided folder, and running the project with `core` gradle config + `npm run dev` in ui-trellis:

image

Project wouldn't load the images and even before login, it's possible to note that images are returning 404.

After changes, manual tests done:

Test 1: 
- deleted everything from the grails-app/provided folder;
- ran project with `core` gradle config 
outcome - background-images with path following this pattern:  `../../img/pagerdutylogo.svg`

Test 2: 
- deleted everything from the grails-app/provided folder;
- ran project with `enterprise` gradle config 
outcome - background-images with path following this pattern:  `../../img/pagerdutylogo.svg`

Test 3:
- deleted everything from the grails-app/provided folder;
- ran project with `core` gradle config + `npm run dev` in ui-trellis;
outcome - background-images with path following this pattern: `../../img/pagerdutylogo.svg`

Test 4: 
- deleted everything from the grails-app/provided folder;
- ran project with `enterprise` gradle config + `npm run dev` in ui-trellis
outcome - background-images with path following this pattern: `../../img/pagerdutylogo.svg`

Test 5:
- deleted everything from the grails-app/provided folder;
- ran project with `enterprise` gradle config + `npm run dev` in ui-trellis + `npm run build:dev:library`
- run the grails target `:plugins:rundeckpro-spa-ui:rundeckpro-ui-components:installUiTrellis`;
outcome - background-images with path following this pattern: `../../img/pagerdutylogo.svg`


gschueler (approved):
LGTM

PR #9173, opened 2024-06-06
fix flaky tests

ronaveva:
This provides stability to some flaky tests:

"run job modal should show validation error" excluded from PRO since PRO has additional icons in job list page
"create a job with referenced execution node step by name and run it successfully" now checks using the API for the execution to finish and then it checks for the specific UI component to be "SUCCEEDED"
"Select all json list options by default" since the options changes dynamically it is better to get the elements each time to avoid stale elements
"scheduled job" was failing due to being impossible to assign the crontab string using selenium, it now uses the API and checks the activity page to see if the job actually runs


functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/ScheduledJobSpec.groovy
```diff
@@ -17,41 +14,24 @@ class ScheduledJobSpec extends SeleniumBase{
...
+    def "scheduled job"(){
         given:
         def projectName = "scheduled-job-test"
         setupProject(projectName)
         LoginPage loginPage = page LoginPage
         ActivityPage activityPage = page ActivityPage
-        JobCreatePage jobCreatePage = page JobCreatePage
         JobListPage jobListPage = page(JobListPage)
         jobListPage.loadJobListForProject(projectName)
+        def jobXml = JobUtils.generateScheduledJobsXml("scheduledJob", "<time hour='*' seconds='*/5' minute='*' />")
```
Jesus-Osuna-M, line 25: Just to comment. this was a great idea, using the API instead of the GUI reduce allot of steps that can fail randomly.

Jesus-Osuna-M (approved):
LGTM

PR #9175, opened 2024-06-07
Add stability for tests

ronaveva:
scheduled job: sometimes fails on PRO due to not waiting enough for the executions to run
Select all json list options by default: it was not waiting for the options to be shown

- It increases the waiting time for scheduled execution to run
- It waits for the final options to be shown


Jesus-Osuna-M (approved):
LGTM

PR #9204, opened 2024-06-25
Run-2588 added unit test for nodes list

jayas006:
This is an enhancement.

I have implemented unit tests for the following Node components:

- NodeCard
- NodeTable
- NodeFilterLink
- NodeDetailsSimple

**Additional context**


smartinellibenedetti (changes requested):
There are a few minor changes, but overall, PR is looking good.

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/resources/NodeTable.vue
```diff
@@ -90,34 +94,42 @@
...
                     >
                       <i class="glyphicon glyphicon-circle-arrow-right"></i>
                     </node-filter-link>
 
                     <span class="nodedesc"></span>
+
                     <span class="text-strong">
                       <i
                         v-for="badge in glyphiconBadges(node.attributes)"
                         v-if="node.attributes['ui:badges']"
```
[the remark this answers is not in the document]
smartinellibenedetti: Adding here the reply I've sent on slack:

> there’s nothing wrong with using v-if and v-for in the same element since Vue 3 was released. Docs don’t recommend it because it’s not obvious what takes precedence just by reading the code, but the answer is that the v-if takes precedence. So the code first checks if the if passes, if it does, then executes the for loop.

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/resources/tests/NodeFilterLink.spec.ts
```diff
@@ -1,56 +1,52 @@
...
-    },
-    global: {
-      mocks: {
-        $t: (msg) => msg,
+import { EventBus } from "../../../../../library";
+import { RootStore } from "../../../../../library/stores/RootStore";
+import { RundeckToken } from "../../../../../library/interfaces/rundeckWindow";
+describe("NodeFilterLink Component", () => {
+  beforeAll(() => {
+    window._rundeck = {
```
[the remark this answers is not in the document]
smartinellibenedetti: Adding the reply sent on slack yesterday - yes, this is fine.

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/resources/tests/NodeDetailSimple.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+      mocks: {
+        $t: (msg) => msg,
+      },
+    },
+  });
+  await wrapper.vm.$nextTick();
+  return wrapper;
+};
+describe("NodeDetailsSimple Component", () => {
+  it("renders node attributes and checks for 'Test Node 1'", async () => {
```
smartinellibenedetti, line 41:
```suggestion
  it("renders node attributes", async () => {
```

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/resources/tests/NodeDetailSimple.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+    const wrapper = await mountNodeDetailsSimple();
+    const attributes = wrapper.findAll(".setting");
+    const attributeTexts = attributes.map((attr) => attr.text());
+    // This checks if any attribute text includes "Test Node 1"
+    const descriptionFound = attributeTexts.some((text) =>
+      text.includes("Test Node 1"),
+    );
```
smartinellibenedetti, line 48: blocker: please check all attributes, not just the description

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/resources/tests/NodeDetailSimple.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+    const wrapper = await mountNodeDetailsSimple();
+    const tags = wrapper.findAll(".label-muted");
+    expect(tags.length).toBe(2);
+    expect(tags.at(0).text()).toContain("Tag1");
+    expect(tags.at(1).text()).toContain("Tag2");
+  });
+  it("renders expandable attributes using class", async () => {
+    const wrapper = await mountNodeDetailsSimple();
+    const toggleButton = wrapper.find(".textbtn");
+    expect(toggleButton.exists()).toBe(true);
```
smartinellibenedetti, line 61:
```suggestion
```

smartinellibenedetti (changes requested):
Marked the comments as "blockers" for the ones that are really important to do. Also, please standardize the data-test ids by either having all of them with data-test-id or data-testid

smartinellibenedetti (commented):
I know that the jira ticket doesn't explicitly mention this in the acceptance criteria, but I realized a problem today during the review of the current implementation:

The tests done for nodeDetailsSimple do not cover the scenario found in NodeTable, where NodeDetailsSimple is used with useNamespace and nodeColumns set to true.

While useNamespace and nodeColumns were set as default to false, in fact everywhere in the codebase they are used with true, so we could even at a later stage refactor the component and remove the unused logic, as long as we have tests that cover the scenario being used. How would you prefer to address this? We can mark this ticket as concluded and then the new tests for NodeDetailsSimple will be added in a separate ticket, or they can be added to this PR. I'm fine with either option, but we should address this.

Please let me know what you prefer @jayas006 here or via slack.

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/resources/tests/NodeDetailSimple.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+    const wrapper = await mountNodeDetailsSimple();
+    const toggleButton = wrapper.find(".textbtn");
+    await toggleButton.trigger("click");
+    await wrapper.vm.$nextTick();
+    // Adjust the selector if necessary based on the actual HTML structure
+    const expandedAttributes = wrapper.findAll(".hover-action-holder");
+    expect(expandedAttributes.length).toBe(12);
+  });
+
+  it("renders correctly with useNamespace and nodeColumns set to true", async () => {
```
smartinellibenedetti, line 92: The name of the test doesn't match what is being asserted. The node filter links that received the data-testids `node-attribute-link-username` and `node-attribute-link-hostname` will be rendered when the prop `useDefaultColumns` is true, independently of the values of `useNamespace` and `nodeColumns`.

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/resources/tests/NodeDetailSimple.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+      "ns1:attr3": "value3",
+      "ns1:attr4": "value4",
+    };
+    Object.entries(expectedAttributes).forEach(([key, value]) => {
+      const attributeFound = attributeTexts.some((text) =>
+        text.includes(value),
+      );
+      expect(attributeFound).toBe(true);
+    });
+  });
```
smartinellibenedetti, line 74: Few things:

1) the test where the props `useNamespace` and `nodeColumns` have a false value shouldn't be removed (if we get to the point of refactoring the component, then that case can be removed, but we should still cover it until then);

2) When these props are false, the majority of the attributes are rendered within the same tbody, and therefore picking it up with the `setting` class was acceptable, even though this approach wasn't asserting the attributes grouped under description/status/operating system/tags (because they do not use that class); But this isn't the case when the props are true. This test needs to be rewritten and ideally reorganized, asserting things within the groups they belong to instead of having all values in one single constant. Suggestion of structure:

- test selects all the general attributes (the attributes that live in the first <tbody>, description/status/operating system/tags) and asserts them against the expectedGeneralAttributes;

- test selects all the attributes that don't have a namespace and asserts them against the expectedAttributesWithNoNamespaces and also checks if the right node-filter-link was rendered (if it's the one with the glyphicon-plus or the glyphicon-minus)

- And when useNamespace is true, the test selects all attributes with namespaces, asserts that the namespace.values.length rendered matches the number of nested attributes and then finally asserts the nested attributes against the expectedNestedAttributes;

This way, if one day something is broken, it's easier to identify what is going wrong/missing. The way the test currently is written, the error message only hints that an attribute wasn't found.

smartinellibenedetti (changes requested):
Further changes are required, please let me know if any clarification is needed.

smartinellibenedetti (approved):
I've left a suggestion in case you're curious how the test could be further refactored but this is a nit/curiosity, not a blocker. Good job!

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/resources/tests/NodeDetailSimple.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+        expect(link.classes()).toContain("glyphicon-plus");
+      });
+      // Check for .glyphicon-minus elements related to attributes without namespaces
+      const minusLinks = wrapper.findAll(".setting .glyphicon-minus");
+      expect(minusLinks.length).toBe(expectedMinusCount);
+      minusLinks.forEach((link) => {
+        expect(link.classes()).toContain("glyphicon-minus");
+      });
+      // Attributes with namespaces (only when useNamespace is true)
+      if (props.useNamespace) {
```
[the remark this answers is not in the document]
smartinellibenedetti: This is a totally valid case to add a if on a test 😄

jayas006:
> I've left a suggestion in case you're curious how the test could be further refactored but this is a nit/curiosity, not a blocker. Good job!

Thanks for the suggestion! I have refactored the test according to your feedback.

PR #9208, opened 2024-06-26
RCLOUD-1344: Fix API calls returning 500 where 400 is more appropriate

sjrd218:
Fix component errors return 500 on API job run action
Fix disabled job and project return 500 on API job run 
Fix failed user profile update returns 500 on API update

All of these API calls now return a 400 with a descriptive message.

### DIFF
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/UserController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/UserController.groovy
@@ -445,7 +445,7 @@ Since: v21''',
             if(!updateResponse.isSaved){
                 def errorMsg= updateResponse.errors.allErrors.collect { g.message(error:it) }.join(";")
                 return apiService.renderErrorFormat(response,[
-                        status: HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
+                        status: HttpServletResponse.SC_BAD_REQUEST,
                         message:errorMsg,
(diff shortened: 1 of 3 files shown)


rundeckapp/grails-app/controllers/rundeck/controllers/UserController.groovy
```diff
@@ -445,7 +445,7 @@ Since: v21''',
             if(!updateResponse.isSaved){
                 def errorMsg= updateResponse.errors.allErrors.collect { g.message(error:it) }.join(";")
                 return apiService.renderErrorFormat(response,[
-                        status: HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
+                        status: HttpServletResponse.SC_BAD_REQUEST,
```
jbrookspd, line 448: Should this be a 4XX error? it seems like this could legitimately be a 5XX depending on what happened.

sjrd218: If `updateResponse.isSaved` is being checked and returns false it's because of a validation error with the data. If there's a severe problem and an exception is thrown in the save process this code wouldn't be reached.

jbrookspd: Ahh, gotcha.

jbrookspd (approved):

PR #9213, opened 2024-07-01
RUN-2572: avoid multi executions for same schedule

ronaveva:

Changes the quartz schedule key to prevent scenarios in which certain combination of job deletions and import can cause multiple triggering of the same schedule.

Changes the quartz schedule key from using the DB id, project name and group, to the job UUID.

** Bonus Fix **
Also fixes a minor bug in which attempting to delete a job with two or more running executions results in a 500 error instead of the correct error message.

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
@@ -1176,7 +1176,7 @@ class ScheduledExecutionService implements ApplicationContextAware, Initializing
         boolean success = false
         Execution.withTransaction {
             //find any currently running executions for this job, and if so, throw exception
-            def found = Execution.createCriteria().get {
+            def found = Execution.createCriteria().list {
                 delegate.'scheduledExecution' {
                     eq('id', scheduledExecution.id)
                 }
--- a/rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
@@ -894,6 +894,7 @@ class ScheduledExecutionService implements ApplicationContextAware, Initializing
     def rescheduleJob(ScheduledExecution scheduledExecution, wasScheduled, oldJobName, oldJobGroup, boolean forceLocal, boolean remoteAssigned = false) {
         if (jobSchedulesService.shouldScheduleExecution(scheduledExecution.uuid) && shouldScheduleInThisProject(scheduledExecution.project)) {
             try {
+                deleteJob(scheduledExecution.generateJobScheduledName(), scheduledExecution.generateJobGroupName())
                 return scheduleJob(scheduledExecution, oldJobName, oldJobGroup, forceLocal, remoteAssigned)
             } catch (SchedulerException e) {
                 log.error("Unable to schedule job: ${scheduledExecution.extid}: ${e.message}")
--- a/rundeckapp/src/test/groovy/rundeck/ScheduledExecutionTest.groovy
+++ b/rundeckapp/src/test/groovy/rundeck/ScheduledExecutionTest.groovy
@@ -481,12 +481,12 @@ class ScheduledExecutionTest  {
 
     @Test void testGenerateJobGroupName() {
         def ScheduledExecution se = new ScheduledExecution()
-        se.properties = [jobName: 'TestName', project: "AFrameworkProject"]
-        assertEquals "incorrect group name: ${se.generateJobGroupName()}", "AFrameworkProject:TestName:", se.generateJobGroupName()
+        se.properties = [uuid: "a1", jobName: 'TestName', project: "AFrameworkProject"]
+        assertEquals "incorrect group name: ${se.generateJobGroupName()}", "AFrameworkProject", se.generateJobGroupName()
 
         se = new ScheduledExecution()
-        se.properties = [jobName: 'TestName', project: "AFrameworkProject", groupPath: 'The Group']
-        assertEquals "incorrect group name: ${se.generateJobGroupName()}", "AFrameworkProject:TestName:The Group", se.generateJobGroupName()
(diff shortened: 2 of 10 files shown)


rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
```diff
@@ -876,6 +876,7 @@ class ScheduledExecutionService implements ApplicationContextAware, Initializing
     def rescheduleJob(ScheduledExecution scheduledExecution, wasScheduled, oldJobName, oldJobGroup, boolean forceLocal, boolean remoteAssigned = false) {
         if (jobSchedulesService.shouldScheduleExecution(scheduledExecution.uuid) && shouldScheduleInThisProject(scheduledExecution.project)) {
             try {
+                deleteJob(scheduledExecution.generateJobScheduledName(), scheduledExecution.generateJobGroupName())
```
gschueler, line 879: I'm not sure why this "deleteJob" was added.  Eventually the `registerOnQuartz` method will be called with the job, and the old job will be removed via this line https://github.com/rundeck/rundeck/blob/d65c02cd001612f77f7aab541166384fc2046df2/rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy#L4510-L4510

rundeckapp/src/test/groovy/rundeck/services/RdJobServiceSpec.groovy
```diff
@@ -182,15 +181,15 @@ class RdJobServiceSpec extends Specification implements DataTest {
         def actual = service.detectJobChanges(se, job, event)
 
         then:
-        actual.scheduledJobName == expectedJobName
+        actual.scheduledJobName == uuid
         actual.renamed == renamed
-        event.changeinfo.origName == expectedJobName
+        event.changeinfo.origName == uuid
         if(renamed) event.changeinfo.rename
```
gschueler, line 187: !!FYI this line will never cause an assertion failure, due to the way spock only modifies the immediate statements in the block into assertions, and doesn't descend into if statements. 

This will need to be changed to:
```suggestion
        if(renamed){
            assert event.changeinfo.rename
        }
```

rundeckapp/src/test/groovy/rundeck/services/ScheduledExecutionServiceSpec.groovy
```diff
@@ -3408,21 +3431,27 @@ class ScheduledExecutionServiceSpec extends Specification implements ServiceUnit
...
 
         then:
-        result.jobname == jobname.replaceAll('_ID_', "$id").replaceAll('_JID_', "${job?.id}")
-        result.groupname == groupname.replaceAll('_ID_', "$id").replaceAll('_JID_', "${job?.id}")
+        if(!isJob && !jobScheduled && !executionType.equals("scheduled")){
+            result.jobname == jobname.replaceAll('_ID_', "$id").replaceAll('_JID_', "${job?.id}")
+            result.groupname == groupname.replaceAll('_ID_', "$id").replaceAll('_JID_', "${job?.id}")
+        }else{
+            result.jobname == jobname
+            result.groupname == groupname
```
gschueler, line 3439: These tests will all pass even if the statements return false, because they must be asserted!

This is a pitfall of using `if` statements inside the Spock then/expect sections!

```suggestion
            assert result.jobname == jobname.replaceAll('_ID_', "$id").replaceAll('_JID_', "${job?.id}")
            assert result.groupname == groupname.replaceAll('_ID_', "$id").replaceAll('_JID_', "${job?.id}")
        }else{
            assert result.jobname == jobname
            assert result.groupname == groupname
```

gschueler (changes requested):

rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
ahormazabal: We were aware of this but was still added as a preemptive measure.

rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
```diff
@@ -1175,7 +1176,7 @@ class ScheduledExecutionService implements ApplicationContextAware, Initializing
         boolean success = false
         Execution.withTransaction {
             //find any currently running executions for this job, and if so, throw exception
-            def found = Execution.createCriteria().get {
+            def found = Execution.createCriteria().list {
```
gschueler, line 1179: question: this was the bonus fix, is there a functional test that can be added to verify it?

rundeckapp/src/test/groovy/rundeck/ScheduledExecutionTest.groovy
```diff
@@ -481,12 +481,12 @@ class ScheduledExecutionTest  {
...
-        se.properties = [jobName: 'TestName', project: "AFrameworkProject"]
-        assertEquals "incorrect group name: ${se.generateJobGroupName()}", "AFrameworkProject:TestName:", se.generateJobGroupName()
+        se.properties = [uuid: "a1", jobName: 'TestName', project: "AFrameworkProject"]
+        assertEquals "incorrect group name: ${se.generateJobGroupName()}", "AFrameworkProject", se.generateJobGroupName()
 
         se = new ScheduledExecution()
-        se.properties = [jobName: 'TestName', project: "AFrameworkProject", groupPath: 'The Group']
-        assertEquals "incorrect group name: ${se.generateJobGroupName()}", "AFrameworkProject:TestName:The Group", se.generateJobGroupName()
+        se.properties = [uuid: "a1", jobName: 'TestName', project: "AFrameworkProject", groupPath: 'The Group']
+        assertEquals "incorrect group name: ${se.generateJobGroupName()}", "AFrameworkProject", se.generateJobGroupName()
```
gschueler, line 489: comment: FYI: since we want to try to have everything in Spock, I try to avoid modifying existing Junit tests and just convert them to Spock equivalents when I have to. No need to do that now but for future PRs we can try to make that change.

gschueler (approved):

ahormazabal:
As we agreed I'll improve the bonus fix and junit tests in a separate pr.

PR #9214, opened 2024-07-02
RUN-2603 Jobs scheduled at Project level continue to run when disabled by a Not Owner cluster member

Jesus-Osuna-M:
Problem:
Jobs scheduled at Project level continue to run when disabled by a Not Owner cluster member

Solution:
Add a condition to validate if scheduled executions are scheduled at project level.

STR:
1. Steps to reproduce the behavior:
2. Create a simple job (no schedule at job level)
3. Add as Project Schedule to this job (example, run every minute).
4. Disable the job schedule from the Jobs tab (Actions > Disable Schedule)
5. Wait to check that the job does not run and it has the NEVER next to it (expected)
6. Enable back the job schedule from the Jobs tab (Actions > Enable Schedule)
7. Log in to the other cluster member (not owner)
8. Disable the job schedule from the Jobs tab (Actions > Disable Schedule
9. Wait to check that the job still runs even if it has the NEVER next to it (NOT expected)


carlosrfranco (approved):
LGTM

PR #9217, opened 2024-07-03
RUN-537: Convert details tab to vue

smartinellibenedetti:

Convert details tab of job edit to Vue.

- considered to not have the json field (jobDetailsJSON); Might still end up removing it.

**Additional context**


rundeckapp/grails-app/views/scheduledExecution/_edit.gsp
```diff
@@ -59,7 +60,22 @@
   </div>
 
   <div class="tab-pane active" id="tab_details" data-ko-bind="jobeditor">
-  <section class="section-space-lg">
+  <g:if test="${uiType=='next'}">
+    <g:set var="allowHTML"
+           value="${!(cfg.getString(config: "gui.job.description.disableHTML") in [true,'true'])}"/>
```
gschueler, line 65: hmm, we will need a way to load necessary config metadata for the GUI...possibly a metadata component can be used

rundeckapp/grails-spa/packages/ui-trellis/src/app/pages/job/editor/DetailsEditorSection.vue
```diff
@@ -0,0 +1,51 @@
+<template>
+  <details-editor
+    v-if="updatedData"
+    v-model="updatedData"
+    :allow-html="allowHtml"
+  />
+  <json-embed :output-data="updatedData" field-name="jobDetailsJson" />
```
gschueler, line 7: this field `jobDetailsJson` still needs to be received on the backend when the form is submitted.  here: https://github.com/rundeck/rundeck/blob/e99997a50d5c17f551014e565b95ad9dd60075a5/rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy#L3278-L3278

See example for the options editor, it uses a `jobOptionsJson` field https://github.com/rundeck/rundeck/blob/e99997a50d5c17f551014e565b95ad9dd60075a5/rundeckapp/grails-spa/packages/ui-trellis/src/app/pages/job/editor/OptionsEditorSection.vue#L10-L10 and then uses the value here https://github.com/rundeck/rundeck/blob/e99997a50d5c17f551014e565b95ad9dd60075a5/rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy#L3030-L3042

for the `jobDetailsJson`, in the ScheduledExecutionService.jobDefinitionBasic method you will want to simply add the jobDetailsJson data after parsing it into the basicProps data that is used there.

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/details/DetailsGroupModal.vue
```diff
@@ -0,0 +1,146 @@
+<template>
+  <div
+    :id="modalId"
+    class="modal"
+    tabindex="-1"
+    role="dialog"
+    :aria-labelledby="`${modalId}_title`"
+    aria-hidden="true"
+  >
+    <div class="modal-dialog" :class="modalSize">
```
gschueler, line 10: why not use uiv modal?

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/common/ScheduledExecutionDetails.vue
```diff
@@ -0,0 +1,141 @@
...
+        return `/plugin/pluginFile?service=${this.service}&name=${this.name}`;
+      }
+      return "";
+    },
+    replTokens() {
+      // not relevant at the moment
+      return this.pluginBaseUrl.length > 0
+        ? { "plugin.url": this.pluginBaseUrl }
+        : null;
+    },
```
gschueler, line 108: is this used? service/name also doesn't seem to make sense in this component.

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/details/DetailsEditor.vue
```diff
@@ -0,0 +1,220 @@
...
+  watch: {
+    internalData: {
+      deep: true,
+      handler(newVal) {
+        this.$emit("update:modelValue", newVal);
+      },
+    },
+  },
+  mounted() {
+    this.onMounted();
```
gschueler, line 186: I think it would be cleaner to put the onMounted() logic inside the mounted() handler if that is all that it does

gschueler (commented):
some issues:
---
Job Name: if you leave it blank and blur the field, the Group field also gets a red error status:
Screenshot 2024-07-09 at 3 35 14 PM
---
The Group Choose button correctly shows the modal, but clicking a group in the modal does not fill the value into the Group field.
---
Description:
the `---` separator shows the Preview Readme tab, however the tab is empty.

I see a JS console error:

```
vue.global.js?compile=false:9450 Uncaught (in promise) TypeError: Cannot use 'in' operator to search for 'value' in null
```

rundeckapp/grails-spa/packages/ui-trellis/src/app/pages/job/editor/DetailsEditorSection.vue
gschueler: I also noticed, that the "jobDetailsJson" field doesn't get correctly updated when the field values are changed.

rundeckapp/grails-app/views/scheduledExecution/_edit.gsp
smartinellibenedetti: Anything in mind regarding how this component should look like?

gschueler: I'll make a design, this is ok for now

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/common/ScheduledExecutionDetails.vue
smartinellibenedetti: This component would be the equivalent of `scheduledExecution/_description.gsp` and the the replToken would be used in this line:

`<g:markdown><g:autoLink jobLinkId="${jobLinkId}" tokens="${replTokens}">${remainingLine}</g:autoLink></g:markdown>`

It's not relevant for editing/creating a job, but by searching for the template, I did see some places passing the service and name to it, so ended up adding the computed properties, but didn't finish the logic as I was unsure how it would look like with the VMarkdownView component

smartinellibenedetti: Can remove it for now if you'd like

gschueler: ok, I'd say remove it for now since it is not used in this PR

rundeckapp/grails-spa/packages/ui-trellis/src/app/pages/job/editor/DetailsEditorSection.vue
smartinellibenedetti: This is something I couldn't make a decision about:

the fields from the details tab were originally submitted as direct inputs. While the json is convenient to load the data, I wasn't sure if should move everything to the json and prevent the inputs from appearing in the form submission, if should keep both json and inputs updated OR if shouldn't use the json at all.

The PR as is, is using the json to load, but updating mainly the inputs only, even though the json is still being submitted. 

What do you recommend?

gschueler: Yes, let's remove the direct input field usage. We want to move to a pure json approach and won't be sending those fields, so we should not use them in the refactored UI.  In later phase we plan to migrate to use API directly, so the json data from this component will be part of that payload.  We will have to make the editor section components return js objects to include in the payload, so we might as well do it now.

smartinellibenedetti:
> some issues:
> 
> Job Name: if you leave it blank and blur the field, the Group field also gets a red error status: Screenshot 2024-07-09 at 3 35 14 PM
> 
> The Group Choose button correctly shows the modal, but clicking a group in the modal does not fill the value into the Group field.
> 
> Description: the `---` separator shows the Preview Readme tab, however the tab is empty.
> 
> I see a JS console error:
> 
> ```
> vue.global.js?compile=false:9450 Uncaught (in promise) TypeError: Cannot use 'in' operator to search for 'value' in null
> ```

The red board on both fields was the original behavior but I will adjust it then. As for the preview tab, I couldn't reproduce this error - do you remember what you had typed in the description field?

I pushed a few changes including the change in the ScheduledExecutionService, could you please take a look to see if that's what you were expecting? 

I'm wrapping up the change on the group modal to get it to update the values in the meantime.

rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
```diff
@@ -3277,6 +3277,15 @@ class ScheduledExecutionService implements ApplicationContextAware, Initializing
      */
     public void jobDefinitionBasic(ScheduledExecution scheduledExecution, ScheduledExecution input, Map params, UserAndRoles userAndRoles) {
         Map basicProps
+        if (params.jobDetailsJson) {
+            def detailsData = JSON.parse(params.jobDetailsJson.toString())
+
+            if(detailsData instanceof JSONObject) {
+                detailsData.keySet().forEach(detailKey -> {
+                    scheduledExecution[detailKey as String] = detailsData.get(detailKey)
```
gschueler, line 3285: this should add data to the basicProps Map, later the scheduledExecution is updated via the basicProps. 
since the basicProps is not initialized yet, you could move this after line 3313

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/common/ScheduledExecutionDetails.vue
```diff
@@ -0,0 +1,121 @@
+<template>
```
gschueler, line 1: It seems like this component is very similar to this one:
https://github.com/rundeck/rundeck/blob/main/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/PluginDetails.vue

probably we should consider how to consolidate to one component

smartinellibenedetti: I did see this while working on the other tab for plugins, but remember that there's some logic in the scheduledExecutionDetails that you asked me to remove it as it wasn't fully finished? I think it would be best to consolidate once it's clear that other piece won't be needed. Because if it is still needed, both components won't be so similar anymore.

gschueler: I looked into how those properties were used, I think they are basically not used. The PluginDetails.vue was the conversion of the original _description.gsp as it was used by the plugin properties gsps. It doesn't seem to support markdown, so since you added that feature here, let's keep this one, but we should merge them at some point and add the markdown capabilities

gschueler (approved):

PR #9224, opened 2024-07-10
RUN-2383: Fix node filter input text not showing in execute job modal

carlosrfranco:
It fix an issue where the node filter for the override node filter in execute job modal is not showing

The issue happesn because the ui-socket for the override node filter settings doesn't exist in the DOM when vue components are mounted in the job list page. To fix that it needs to use the uiSocketObserver in order to process the ui components after opening the execute job modal.


smartinellibenedetti (commented):
I see the node filter bar working, but the strings seems like they have are not working 100% for some reason:

Screenshot 2024-07-16 at 8 50 31 AM
And this is the job definition I used:
```
- defaultTab: nodes
  description: |-
    test1233456
    ------
    abc
  executionEnabled: true
  group: abc
  id: f78290c3-41b0-49ac-982b-e7e835edf5d0
  loglevel: INFO
  name: new-test223
  nodeFilterEditable: true
  nodefilters:
    dispatch:
      excludePrecedence: true
      keepgoing: false
      rankOrder: ascending
      successOnEmptyNodeFilter: false
      threadcount: '1'
    filter: .*
  nodesSelectedByDefault: true
  plugins:
    ExecutionLifecycle: {}
  runnerSelector:
    runnerFilterMode: LOCAL
    runnerFilterType: LOCAL_RUNNER
  scheduleEnabled: true
  schedules: []
  sequence:
    commands:
    - exec: echo oi
    keepgoing: false
    strategy: node-first
  tags: 'test,abc'
  uuid: f78290c3-41b0-49ac-982b-e7e835edf5d0
```

When looking at the console log, I'm getting an error that the nodeSummary isn't defined in the nodes.js file.

carlosrfranco:
@smartinellibenedetti  I rebased and check it again and maybe the issues with the string could be something on the build process. I tested locally using ci docker image and looks fine.

gif-ace8419c3a

smartinellibenedetti (approved):
Confirmed the fix on the current UI, and as we spoke, the issue I saw is happening in the nextUi which can be fixed in a later ticket.

PR #9231, opened 2024-07-15
RCLOUD-1351: Fix get reports

ocerda:

There was a problem when exporting reports. During the export, a conversion was taking place where the job ID from the report table was being updated to the job UUID, and that was persisted to the database. This caused missing activity because the query wouldn't return results, as the filter was set by job ID.

- Removed the conversion from job ID to job UUID, since there is already a `job_uuid` being exported.

- Updated the query so that databases that have already updated the column will fetch the reports correctly.

- Added some missing references to the job UUID and execution UUID, which were causing missing activities during import.

**Additional context**

Jira ticket: https://pagerduty.atlassian.net/browse/RCLOUD-1351

### DIFF
--- a/rundeckapp/src/test/groovy/rundeck/services/ProjectServiceSpec.groovy
+++ b/rundeckapp/src/test/groovy/rundeck/services/ProjectServiceSpec.groovy
@@ -2619,7 +2619,7 @@ class ProjectServiceSpec extends Specification implements ServiceUnitTest<Projec
   <dateStarted>1970-01-01T00:00:00Z</dateStarted>
   <dateCompleted>1970-01-01T01:00:00Z</dateCompleted>
   <executionId>123</executionId>
-  <jobId>test-job-uuid</jobId>
+  <jobId>1</jobId>
   <adhocExecution />
   <adhocScript />
   <abortedByUser />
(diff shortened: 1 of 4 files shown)


rundeckapp/src/test/groovy/rundeck/services/ProjectServiceSpec.groovy
```diff
@@ -2815,6 +2815,55 @@ class ProjectServiceSpec extends Specification implements ServiceUnitTest<Projec
...
     }
 
+    def testReportNotUpdatedAfterExport() {
+        def newJobId = 'test-job-uuid'
+        ScheduledExecution se = new ScheduledExecution(jobName: 'blue', project: 'AProject', adhocExecution: true,
+                uuid: newJobId,
+                adhocFilepath: '/this/is/a/path', groupPath: 'some/where', description: 'a job', argString: '-a b -c d',
+                workflow: new Workflow(keepgoing: true, commands: [new CommandExec([adhocRemoteString: 'test buddy', argString: '-delay 12 -monkey cheese -particle'])]),
+        )
+        def oldJobId=se.id
```
JoseOrtiz, line 2825: I think `se.id` is `null` as `se` is not saved

sjrd218 (approved):

PR #9232, opened 2024-07-16
RUN-2272: Improve validation of project archive import

ahormazabal:

Project archive import upload expects to receive a valid Jar file with the project contents, but doesn't do any validation on the validity of the received data.
This PR adds two validations to the archive processing:
- Will throw an error if the data received is an empty or either an invalid zip file.
- Will throw an error if the data is a valid zip file but no actual contents to import are found.

Also the file selector in the import UI now accepts only jar and zip files.

Jar files are just specifically formatted ZIP files. As the ZIP file format allows for arbitrary data to exists within a zip file along with the compressed contents (to support for example self extracting zip files), when the received data is not in zip format the java api considers it an empty zip file, not throwing any kind of error.

This PR adds a validation to the `ZipReader` class in order to throw an `IOException` if the first zip entry lookup returns null, which indicates there´re no zip contents in the file.

Also adds a validation at `ProjectService.importToProject` to throw an Exception if no contents to import are found within a valid jar file, and ensures the temp file deletion is executed even if an uncontrolled error is raised, by adding a `finally` block.

**Additional context**

To test:
- Try importing a file which is not a valid zip file into the "Project Settings -> Import Archive" UI. Before this fix, the UI responds with a success message. With this PR it should render an error.
- Repeat the prior test with a valid zip file which is not result of an archive export.
- Export a project archive using "Project Settings -> Import Archive" and validate it can be imported back successfully.


functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/project/ImportProjectSpec.groovy
```diff
@@ -1,34 +1,109 @@
...
+        loginPage.login(TEST_USER, TEST_PASS)
+        homePage.validatePage()
+        importPage.go()
+        File testArchive = new File(getClass().getResource("/projects-import/import-project-test/${archiveName}").getPath())
+        def fileInput = importPage.getFileInput()
+        fileInput.sendKeys(testArchive.getAbsolutePath())
+        importPage.getImportButton().click()
+
+        then:
+        importPage.getImportAlertMessage().getText().contains("There was some errors in the import process")
```
mrdubr, line 96: nit: 
I know it was not your code, but:
There *were* some errors <...>

mrdubr (approved):
LGTM

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/project/ImportProjectSpec.groovy
ahormazabal: fixed!

PR #9238, opened 2024-07-18
fixed and added test for glyphiconBadges

jayas006:
This update serves as both a bugfix and an enhancement. This is a bugfix related to the incorrect handling of `ui.attributes` while the enhancement includes the addition of tests to ensure the component behaves as expected.

I have modified the `glyphiconBadges` function to correctly parse and validate the badge names provided in the `attributes["ui:badges"]` string.
Additionally, I have added comprehensive tests to verify the function's behavior:
- Parsing well-formed input strings.
- Parsing input strings that do not end with numbers
- Filtering out invalid badge names.
- Handling empty input strings.
- Returning an empty array when no valid badges are provided.
- Handling null or undefined input gracefully.

**Additional context**

### DIFF
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/app/utilities/tests/nodeUi.spec.ts
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/app/utilities/tests/nodeUi.spec.ts
@@ -1,7 +1,7 @@
 import { glyphiconBadges } from "../nodeUi";
 
 describe("glyphiconBadges", () => {
-  test("parses well-formed input string correctly", () => {
+  it("parses well-formed input string correctly", () => {
     const input = "glyphicon-user1, fa-badge2, fab-badge1, glyphicon-lock-1";
     expect(glyphiconBadges({ "ui:badges": input })).toEqual([
       "glyphicon-user1",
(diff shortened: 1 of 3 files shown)


rundeckapp/grails-spa/packages/ui-trellis/src/app/utilities/tests/nodeUi.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+describe("glyphiconBadges", () => {
+  it("parses well-formed input string correctly", () => {
+    const input = "glyphicon-user1, fa-badge2, fab-badge1, glyphicon-lock-1";
+    expect(glyphiconBadges({ "ui:badges": input })).toEqual([
+      "glyphicon-user1",
+      "fa-badge2",
+      "fab-badge1",
+      "glyphicon-lock-1",
+    ]);
+  });
```
smartinellibenedetti, line 12: Please add a value in the input that does not end with a number, ie `fa-house-user glyphicon-download-alt`

they are both actual icons, and the glyphiconBadges should return them as well.

smartinellibenedetti (changes requested):
One small change required.

smartinellibenedetti (approved):
🎉

PR #9240, opened 2024-07-19
RUN-2689: fix: exitCode from script steps is not correct in context variable data

gschueler:
Fix: script steps that fail do not correctly export the "exec.exitCode" context variable

Fix the error where the script cleanup step exitCode was overriding the script's exit code.

**Additional context**


carlosrfranco (approved):
LGTM

PR #9247, opened 2024-07-26
RUN-2703 : improve functional test harness

mrdubr:
Improvement to the functional tests harness:
- Made Rundeck HTTP client timeouts in the RdContainer configurable
- Added handling for the unsuccessfully created job executions in the RdContainer that resulted in an infinite loop

Jira ticket: https://pagerduty.atlassian.net/browse/RUN-2703?atlOrigin=eyJpIjoiOGE5ZjhiYTM3MjllNGRlZDg1M2NjZGE4M2U1ZGViZmMiLCJwIjoiaiJ9

### DIFF
--- a/functional-test/src/test/groovy/org/rundeck/util/container/BaseContainer.groovy
+++ b/functional-test/src/test/groovy/org/rundeck/util/container/BaseContainer.groovy
@@ -13,7 +13,6 @@ import spock.lang.Specification
 import java.text.SimpleDateFormat
 import java.time.Duration
 import java.util.function.Consumer
-import java.util.function.Function
 import java.util.jar.JarEntry
 import java.util.jar.JarOutputStream
 import java.util.jar.Manifest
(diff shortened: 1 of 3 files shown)


ronaveva:
Hi @mrdubr perhaps each test task could set the timeout for the environment provided, i would say that just a single one for the 3 timeouts (read, write and connection) would be enough 🤔

mrdubr:
> Hi @mrdubr perhaps each test task could set the timeout for the environment provided, i would say that just a single one for the 3 timeouts (read, write and connection) would be enough 🤔

Hey @ronaveva, for some reason (could be `arm64`) multiple functional test suites timeout when run locally. Adjusting them individually does not appear to be productive. 

Should be noted that this change has zero effect on the existing tests and their performance (execution time) in circleci. If the previous limit (25) was insufficient, they would have been failing.

ronaveva (approved):
LGTM

PR #9254, opened 2024-07-31
Run-2589 Feat/unit tests for nodes filter components

jayas006:
This is an enhancement

- Added a test for  NodeFilterInput
- Added a test for  NodeDefaultFilterDropdown

**Additional context**


rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/resources/tests/NodeDefaultFilterDropdown.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+    const setDefaultLink = wrapper.find(
+      'a[data-testid="set-default-filter-link"]',
+    );
+    expect(setDefaultLink.exists()).toBe(true);
+    await setDefaultLink.trigger("click");
+    expect(emitSpy).toHaveBeenCalledWith(
+      "nodefilter:action:setDefault",
+      "filter1",
+    );
+  });
```
smartinellibenedetti, line 81: Blocker: Few things to adjust in this test and possibly in the mocks:

1. Mitt should be used when there's a need for an external eventBus to trigger an event that the component being tested reacts to.  This is defined by `this.eventBus.on([eventName], methodToExecuteOnceEventHappens)` ie:

```
 this.eventBus.on("plugin-configs-saved", (configs: any) =>
      this.refreshAll(),
    );
```

So the first thing to do is to remove mitt and replace it with the usual mock for the eventBus, as NodeDefaultFilterDropdown only emits events and doesn't react to them;

2. Line 71 should be removed;

3. Adding context before point 3: When a filter is set to default, its value is saved to the NodeSummary with the key `defaultFilter`, and the NodeCard component which is the parent of NodeDefaultFilterDropdown, will figure out if the selected filter is equal to `.*` and update the flag being passed to the prop `isDefaultFilter` (true if it's equal to ".*")

There's a decision to make here, that is if this test will start with a default filter or not. Both scenarios can happen, but then:

- option 1: If the decision is to start with a default filter, then nodeSummary mock needs to be updated to add the key `defaultFilter` with a string value (ideally with one of the filterNames in the mock, or ".*" - remembering that if the later, then `isDefaultFilter` needs to be set to true); test should assert what was rendered before and after the click happens, and the mock receives another value for `defaultFilter` as well (defaultFilter can become null if the test click to remove the default filter);
  
- option 2: If the decision is to start without a default filter, then it's not necessary to tweak nodeSummary, but the rest should be updated the same: the test should assert what was rendered before and after the click happens,  and the mock receives a value for `defaultFilter` as well;
  
Please let me know if you need more context. The ideal solution would be to cover both options and in the same test, remove the default filter.

smartinellibenedetti (changes requested):
I didn't review the components that are part of the other PR, to avoid requesting changes that were already addressed there.

Some minor things to clean up, and one of the tests of NodeDefaultFilterDropdown needs some adjustments. Mentioned what would be the ideal outcome, but the PR can be approved once one of the options is implemented.

Also, kudos to you for the `updates modelValue` test on the NodeFilterInput, was very happy to see both the blur and enter scenarios being covered.

Let me know if more context is needed!

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/resources/tests/NodeDefaultFilterDropdown.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+      });
+
+      const removeDefaultLink = wrapper.find(
+        'a[data-testid="remove-default-filter-link"]',
+      );
+      await removeDefaultLink.trigger("click");
+      expect(mockEventBus.emit).toHaveBeenCalledWith(
+        "nodefilter:action:removeDefault",
+      );
+    });
```
smartinellibenedetti, line 85: It would be good to add 2 more expects here:
- one expect before the click, asserting that the dropdown has an option with the text "remove.default.filter"
- one expect after the click, asserting that the dropdown has an option with the text "set.as.default.filter"

smartinellibenedetti (commented):
Changes are looking good, there's only one minor addition needed in one of the tests. Also, a heads-up, it is necessary to undo the changes in the NodeCard and NodeDetailsSimple that weren't meant to be part of this PR

smartinellibenedetti (approved):
LGTM - just need to wait 5.6 release to merge

PR #9255, opened 2024-08-01
Testdeck improvement scm

ronaveva:
**Issue:** Whenever changing from enable/disable, if the "toggleButton" was not found it was just ignoring it and just letting the process to continue, this way if the button is not found the test would fail after leaving no clue on what really happen

**Fix:** Now it waits for the toggleButton to exist and then clicking it. This way if the button for some reason does not show, it will be easier to identify the failure


ChuckCrawford (approved):
LGTM

PR #9257, opened 2024-08-01
RUN-2606: Look up execution by id if uuid not in report

gschueler:

Fix: executions created before version 4.17.0 are missing a uuid value.  The Activity list fails to display these executions.

If the uuid is not set in the report, look up the execution by the ID.

**Additional context**

### DIFF
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/ReportsController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/ReportsController.groovy
@@ -298,10 +298,16 @@ class ReportsController extends ControllerBase{
         results.reports=results?.reports.collect{
             def map=it.toMap()
             map.duration= (it.dateCompleted ?: new Date()).time - it.dateStarted.time
-            if(map.executionUuid){
+            def execution=null
+            if(map.executionUuid) {
+                execution = Execution.findByUuid(map.executionUuid)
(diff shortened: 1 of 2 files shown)


rundeckapp/grails-app/controllers/rundeck/controllers/ReportsController.groovy
```diff
@@ -298,10 +298,16 @@ class ReportsController extends ControllerBase{
         results.reports=results?.reports.collect{
             def map=it.toMap()
             map.duration= (it.dateCompleted ?: new Date()).time - it.dateStarted.time
-            if(map.executionUuid){
+            def execution=null
+            if(map.executionUuid) {
+                execution = Execution.findByUuid(map.executionUuid)
```
ChuckCrawford, line 303: Are we okay with moving these data-access operations outside of the try / catch block they were in before?

gschueler: Good catch! I will update it

ChuckCrawford (approved):
Probably a good idea to have @ahormazabal review as well but this looks good to me.

I also did a basic test to confirm that a job with 4.16 < and 4.16 >  activity properly display with this change.

ChuckCrawford (approved):

PR #9259, opened 2024-08-02
RUN-2606: fix report delete by execution ID

gschueler:

Fix: executions without UUID cannot be deleted

Support deleting by DB ID.

**Additional context**

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/ReportService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ReportService.groovy
@@ -24,8 +24,10 @@ import com.google.common.collect.Lists
 import grails.gorm.DetachedCriteria
 import grails.gorm.transactions.Transactional
 import org.rundeck.app.authorization.AppAuthContextEvaluator
+import org.rundeck.app.data.model.v1.execution.ExecutionData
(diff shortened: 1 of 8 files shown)


rundeckapp/grails-app/services/rundeck/services/ReportService.groovy
```diff
@@ -528,9 +530,18 @@ class ReportService  {
         return rundeckAuthContextEvaluator.authorizeProjectResources(authContext,resHS, constraints, project)
     }
 
-    def deleteByExecutionUuid(String uuid){
-        execReportDataProvider.deleteAllByExecutionUuid(uuid)
+    def deleteByExecution(ExecutionData e) {
+        if (e.uuid) {
+            execReportDataProvider.deleteAllByExecutionUuid(e.uuid)
+        } else if (execReportDataProvider instanceof DBExecReportSupport) {
```
ChuckCrawford, line 536: Do we actually need this extra interface?  Could we just do this:

```java
} else if (execReportDataProvider instanceof GormExecReportDataProvider) {
```

ChuckCrawford (approved):
This looks good to me.

I tested that deleting both old and new reports works correctly individually, in bulk, and on a schedule.

PR #9263, opened 2024-08-05
RUN-2730: Increase CI speed by splitting functional tests

ChuckCrawford:
This is an enhancement to increase the speed of the CI pipeline. This change splits the functional tests and executes them across multiple machines in parallel.

This solution splits test execution across a set of machines by:
* Using the `circleci` CLI to enumerate test source code files and split them across the set of running machines.
* Using Gradle to convert the paths of source code files into paths of compiled classes.
* Providing a custom [JUnit include function](https://docs.gradle.org/7.6.2/dsl/org.gradle.api.tasks.testing.Test.html#org.gradle.api.tasks.testing.Test:include(groovy.lang.Closure)) that matches against the set of compiled classes.

The alternative to this solution would be to write a Spock or JUnit extension that only executes a subset of the tests by partitioning them across the set of running machines. In an MVP of this approach we ran into some issues with accurately reporting test counts in CircleCI.

**Additional context**
I have done this but reviewer(s) should confirm that the number of tests executed before and after this change are the same.

### DIFF
--- a/functional-test/build.gradle
+++ b/functional-test/build.gradle
@@ -62,13 +62,47 @@ def checkWarFile(destination, fileName) {
     println "File copied from ${warFile} to ${destination + fileName}"
 }
 
+/** 
+ * Returns a function that includes only the test files specified in the `testFiles` parameter.
+ * The `testFiles` parameter is a string of paths to the source code of the tests to be run.
+ */
+def explicitTestIncluder(String testFiles) {
+    // Convert test file source code paths to compiled class paths.
+    // Example: `functional-test/src/test/groovy/org/rundeck/tests/functional/api/basic/BasicSpec.groovy`
+    // needs to become `org/rundeck/tests/functional/api/basic/BasicSpec.class`.
+    def compiledFilesToTest = testFiles
+        .split(System.lineSeparator())
+        .collect { testFilePath -> 
+            testFilePath
+            .trim()
+            .replaceFirst(".*\\/src\\/test\\/(groovy\\/|java\\/)", "")
+            .replaceFirst("(.groovy|.java)\$", ".class")
+        }
+
+    logger.info("Compiled files to test: {}", compiledFilesToTest)
+
+    return {
+        FileTreeElement e ->
+            if (e.isDirectory()) {
+                return true
+            }
+
+            compiledFilesToTest.contains(e.path)
+    }
+}
+
 task apiTest(type: Test){
     useJUnitPlatform()
     systemProperty('TEST_IMAGE', "rundeck/rundeck:SNAPSHOT")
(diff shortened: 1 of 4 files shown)


functional-test/build.gradle
```diff
@@ -62,13 +62,47 @@ def checkWarFile(destination, fileName) {
...
+
+    logger.info("Compiled files to test: {}", compiledFilesToTest)
+
+    return {
+        FileTreeElement e ->
+            if (e.isDirectory()) {
+                return true
+            }
+
+            compiledFilesToTest.any { e.getRelativePath().toString() == it }
```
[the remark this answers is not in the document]
ChuckCrawford: Nice. Thanks.

ronaveva (changes requested):
Test file pattern should have "Spec" before "{.groovy,.java}". This way we will ignore not Spec files

ChuckCrawford:
> Test file pattern should have "Spec" before "{.groovy,.java}". This way we will ignore not Spec files

The current path we are using does not ignore spec files? I was trying to use as loose of a match as possible to avoid the potential for missing tests in the future.

functional-test/build.gradle
```diff
@@ -62,13 +62,47 @@ def checkWarFile(destination, fileName) {
...
 }
 
+/** 
+ * Returns a function that includes only the test files specified in the `testFiles` parameter.
+ * The `testFiles` parameter is a string of paths to the source code of the tests to be run.
+ */
+def explicitTestIncluder(String testFiles) {
+    // Convert test file source code paths to compiled class paths.
+    // Example: `functional-test/src/test/groovy/org/rundeck/tests/functional/api/basic/BasicSpec.groovy`
+    // needs to become `org/rundeck/tests/functional/api/basic/BasicSpec.class`.
```
ahormazabal, line 72: I presume it was tried but i must ask: I wonder if there's a simpler way to get this information from gradle.

ahormazabal:
Please create a complementary `rundeckpro` PR pointed to this branch to confirm the changes don't affect its pipeline.

functional-test/build.gradle
ChuckCrawford: As far as I can tell this is not possible while hooking into the [include](https://docs.gradle.org/7.6.2/javadoc/org/gradle/api/tasks/testing/Test.html#include-groovy.lang.Closure-) function of Gradle test tasks. The [FileTreeElements](https://docs.gradle.org/7.6.2/javadoc/org/gradle/api/file/FileTreeElement.html) point to the compiled test classes rather than source code.

ChuckCrawford: I guess an alternative approach might have been to modify Gradle's source sets but that might have other challenges?

ronaveva (approved):
LGTM

PR #9264, opened 2024-08-06
RUN-575: Initial work for job workflow tab

smartinellibenedetti:
Initial work for workflow tab, hidden behind alphaUi flag + new API version


rundeckapp/grails-app/controllers/rundeck/controllers/PluginController.groovy
```diff
@@ -225,101 +233,158 @@ Since: v33
...
+                mediaType = MediaType.APPLICATION_JSON,
+                schema = @Schema(implementation = ApiPluginProviderDetail)
+            )
+        )
+    )
+    def apiPluginDetail(
+        @Parameter(in = ParameterIn.PATH, required = true) String service,
+        @Parameter(in = ParameterIn.PATH, required = true) String provider
+    ) {
+        if (!apiService.requireApi(request, response, ApiVersions.V47)) {
```
gschueler, line 261: will need to increment API version to v49, and specify that here.

rundeckapp/grails-app/controllers/rundeck/controllers/PluginController.groovy
```diff
@@ -225,101 +233,158 @@ Since: v33
...
+            provider,
             'plugin.description',
             desc.description,
             RequestContextUtils.getLocale(request)
         )
+        if(!apiRequest){
+            detail.desc = detail.description
+            detail.description=null
+        }
         if(service != "UI") {
```
gschueler, line 334:
```suggestion
        if(service != ServiceNameConstants.UI) {
```

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobsSpec.groovy
```diff
@@ -413,6 +413,34 @@ class JobsSpec extends SeleniumBase {
...
+            jobCreatePage.workflowAlphaUiContainer.isDisplayed()
+        cleanup:
+            def cleanUpConfig = [
+                    [
+                            "key": "rundeck.feature.alphaUi.enabled",
+                            "value": false
+                    ],
+            ]
+            enableFeatureFlag(cleanUpConfig)
+    }
```
smartinellibenedetti, line 442: I'm unsure if this file is the best place for this test. The same question also applies to the enable/wait for flag methods in the baseContainer. 

I wanted to add a toggle method in the JobCreatePage to do this but then it wouldn't see the methods as it doesn't extend it (it extends basePage), but moving them there also seemed wrong as it would need to bring in the client logic.

gschueler (approved):

PR #9270, opened 2024-08-07
RUN-2744: Fix pagination calculation by casting string into number

smartinellibenedetti:

With the recent work done with runners, the node table got updated to fix some type errors happening on Pro. With the typing, the page number got concatenated instead of being added one.

Move the logic to a method, cast the string into an integer, and add one.

**Additional context**

After:
image


smartinellibenedetti:
Closed in favor of https://github.com/rundeck/rundeck/pull/9271

PR #9276, opened 2024-08-09
RUN-2623: Improvement on the Resource Model plugins configuration using runners

ltamaster:

Improvement on the Resource Model plugins configuration using runners:
- mark StubResourceModelSource as a local plugin

**Additional context**


alexander-variacode (approved):
LGTM

PR #9284, opened 2024-08-14
RUN-2705: Fix notification timeout results in hanging jobs

chrismcg14:
Improves the stability of job execution when long-running or unreliable notification destinations are used.

Refactors Execution completion so that the transaction responsible for updating the Execution record commits before notifications are processing. With this change an Execution should always end up in a "completed" status; regardless of what happens while sending job notifications.

There are fancier versions of this refactoring that involve making notification handling more asynchronous. This first change is limited to improving stability with minimal refactoring.

**Additional context**


rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -3103,8 +3102,12 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
      */
     @CompileStatic
     def saveExecutionState( schedId, exId, Map props, AsyncStarted execmap, Map retryContext){
+        Map result;
```
ChuckCrawford, line 3105: Is it possible to write this like so?
```
 Map result =  Execution.withNewTransaction {
            saveExecutionState_currentTransaction(schedId, exId, props, execmap, retryContext)
        }
```

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -3204,6 +3205,7 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
         }
         if(execSaved) {
             //summarize node success
+            def execInfo
```
ChuckCrawford, line 3208: Is this still necessary?

ChuckCrawford (commented):
nitpicky things

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -3102,10 +3064,12 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
      * @return
      */
     @CompileStatic
-    def saveExecutionState( schedId, exId, Map props, AsyncStarted execmap, Map retryContext){
-        Execution.withNewTransaction {
+    def saveCompletedExecution_currentTransaction( schedId, exId, Map props, AsyncStarted execmap, Map retryContext){
```
ChuckCrawford, line 3067: The name of this function is a little strange since this function does more than just saving the execution and it actually starts new transactions. What do you think about naming this function:  `completeExecution` or leaving it as `saveExecutionState` and instead renaming the inner function: `private ExecutionCompleteEvent saveExecutionState_currentTransaction(` to something like `private ExecutionCompleteEvent saveCompletedExecution_currentTransaction(` ?

ChuckCrawford:
This is looking good. We should look to add some tests either in this PR or in a quick follow-up. 

One way of testing this might be to assert the job success / failed webhook payloads contain the fields that now show up after the transactions are properly ordered. The `execution.date-ended` field being one such example: 

```json
{
    "trigger": "success",
    "status": "succeeded",
    "executionId": 3,
    "execution": {
        "id": 3,
        "href": "http://localhost:4440/project/test/execution/show/3",
        "permalink": null,
        "status": "succeeded",
        "project": "test",
        "executionType": "user",
        "user": "admin",
        "date-started": {
            "unixtime": 1723732225284,
            "date": "2024-08-15T14:30:25Z"
        },
        "date-ended": {
            "unixtime": 1723732225629,
            "date": "2024-08-15T14:30:25Z"
        },
        "job": {
            "id": "83ba0f8f-ad9d-498c-9491-c4a7f1abb433",
            "averageDuration": 946,
            "name": "Broken Notify",
            "group": "",
            "project": "test",
            "description": "",
            "href": "http://localhost:4440/api/48/job/83ba0f8f-ad9d-498c-9491-c4a7f1abb433",
            "permalink": "http://localhost:4440/project/test/job/show/83ba0f8f-ad9d-498c-9491-c4a7f1abb433"
        },
        "description": "Plugin[exec-command, nodeStep: true]",
        "argstring": null,
        "serverUUID": "51ff5ee6-0b74-4b8b-9259-d4f3bb2e1376",
        "successfulNodes": [
            "localhost"
        ]
    }
}
```

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -842,23 +842,25 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
     }
 
     /**
-     * calls {@link #saveExecutionState_currentTransaction(java.lang.Object, java.lang.Object, java.util.Map, rundeck.services.ExecutionService.AsyncStarted, java.util.Map)}
+     * calls {@link #saveCompletedExecution_currentTransaction(java.lang.Object, java.lang.Object, java.util.Map, rundeck.services.ExecutionService.AsyncStarted, java.util.Map)}
      * @param e execution
      * @param status
      */
     private void cleanupExecution_currentTransaction(Execution e, String status = null) {
```
ChuckCrawford, line 849: The `cleanupExecution_currentTransaction` function name is a little strange now that we are changing the code to create inner transactions. Once we do this it essentially becomes the same as the other `cleanupRunningJobs` methods in this service. 

What do you think about the following:

* Delete this function.
* Delete [the only caller of this function](https://github.com/rundeck/rundeck/blob/a7dfef49f6c6c89f8747289b42e4ca3587425cf5/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy#L821-L827) and [the only caller of that function](https://github.com/rundeck/rundeck/blob/a7dfef49f6c6c89f8747289b42e4ca3587425cf5/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy#L798-L804)
* Change the code in `rundeckpro` to instead call [the other cleanupRunningJobs method](https://github.com/rundeck/rundeck/blob/a7dfef49f6c6c89f8747289b42e4ca3587425cf5/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy#L795-L797).

functional-test/src/test/groovy/org/rundeck/tests/functional/api/job/JobExecutionSpec.groovy
```diff
@@ -1205,6 +1206,77 @@ class JobExecutionSpec extends BaseContainer {
...
+        def readJobSucceeded = JobUtils.waitForExecutionToBe(
+                ExecutionStatus.SUCCEEDED.state,
+                readJobRunResponse.id as String,
+                mapper,
+                client,
+                WaitingTime.LOW.milliSeconds,
+                WaitingTime.MODERATE.milliSeconds / 1000 as int
+        )
+        assert readJobSucceeded.status == ExecutionStatus.SUCCEEDED.state
+        def execOutputResponse = client.doGetAcceptAll("/execution/$readJobRunResponse.id")
```
ChuckCrawford, line 1269: We need to take a look at this test as well. (It passes even before we apply the fix.) It looks like here you are calling the API rather than looking at the webhook payload...

Since these are "expensive" functional tests; I also think it makes sense to just modify the existing test rather than adding a whole new one.

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -795,13 +795,6 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
     def cleanupRunningJobs(String serverUUID = null, String status = null, Date before = new Date()) {
         cleanupRunningJobs(findRunningExecutions(serverUUID, before), status)
     }
-    /**
-     * Set the result status to FAIL for any Executions that are not complete, does not create a new transaction
-     * @param serverUUID if not null, only match executions assigned to the given serverUUID
-     */
-    def cleanupRunningJobs_currentTransaction(String serverUUID = null, String status = null, Date before = new Date()) {
```
ChuckCrawford, line 802: Note: these functions are removed since updating them to produce the desired behavior (committing `execution` records in a separate inner transaction) would make them functionally equivalent to `def saveExecutionState(schedId, exId, Map props, AsyncStarted execmap, Map retryContext)`

ronaveva (changes requested):
It looks ok overall, it's only missing unit tests for saveCompletedExecution_currentTransaction and triggerJobCompleteNotifications

ronaveva (approved):
LGTM

PR #9283, opened 2024-08-14
add additional functional tests for Command executions

mrdubr:
Additional selenium functional tests for the Command execution flows per
https://pagerduty.atlassian.net/browse/RUN-2610?atlOrigin=eyJpIjoiNTJmY2NjYjg2Y2NiNDE3OTg4NWI1MjhlMDk0MjZiMTAiLCJwIjoiaiJ9

PR #9285, opened 2024-08-14
Refactor execution waiting helper for a more clear api.

ahormazabal:

Enhancement to the functional api, to improve the execution waiting api.

Changed the api to use java.time and consistent measurements, to avoid confusion between seconds and milliseconds.

### DIFF
--- a/functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
+++ b/functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
@@ -5,8 +5,10 @@ import org.rundeck.util.api.responses.execution.Execution
 import org.rundeck.util.api.responses.jobs.Job
 import org.rundeck.util.api.scm.GitScmApiClient
 import org.rundeck.util.api.scm.httpbody.ScmJobStatusResponse
+import org.rundeck.util.common.WaitingTime
 import org.rundeck.util.container.RdClient
 
+import java.time.Duration
 import java.time.ZoneId
 import java.util.concurrent.TimeUnit
 
@@ -116,26 +118,38 @@ class JobUtils {
         return jobs[0]
     }
 
+
     static Execution waitForExecutionToBe(
             String state,
             String executionId,
             ObjectMapper mapper,
             RdClient client,
-            int iterationGap,
-            int timeout
+            WaitingTime iterationGap,
+            WaitingTime timeout
+    ) {
+        return waitForExecutionToBe(state, executionId, mapper, client, iterationGap.duration, timeout.duration)
+    }
+
+    static Execution waitForExecutionToBe(
+            String state,
+            String executionId,
+            ObjectMapper mapper,
+            RdClient client,
+            Duration iterationGap,
+            Duration timeout
     ){
         Execution executionStatus
         def execDetail = client.doGet("/execution/${executionId}")
         executionStatus = mapper.readValue(execDetail.body().string(), Execution.class)
         long initTime = System.currentTimeMillis()
         while(executionStatus.status != state){
-            if ((System.currentTimeMillis() - initTime) >= TimeUnit.SECONDS.toMillis(timeout)) {
-                throw new InterruptedException("Timeout reached (${timeout} seconds), the execution had \"${executionStatus.status}\" state.")
+            if ((System.currentTimeMillis() - initTime) >= timeout.toMillis()) {
+                throw new InterruptedException("Timeout reached (${timeout.toSeconds()} seconds), the execution had \"${executionStatus.status}\" state.")
             }
             def transientExecutionResponse = client.doGet("/execution/${executionId}")
             executionStatus = mapper.readValue(transientExecutionResponse.body().string(), Execution.class)
             if( executionStatus.status == state ) break
-            Thread.sleep(iterationGap)
+            Thread.sleep(iterationGap.toMillis())
         }
         return executionStatus
     }
--- a/functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
+++ b/functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
@@ -119,6 +119,18 @@ class JobUtils {
     }
 
 
+    /**
+     * Waits for the execution to be in the expected state.
+     *
+     * @param expectedStates The expected state.
+     * @param executionId The execution ID to query.
+     * @param mapper The ObjectMapper instance to parse the response.
+     * @param client The RdClient instance to perform the HTTP request.
+     * @param iterationGap The duration to wait between each check.
+     * @param timeout The maximum duration to wait for the execution to reach the expected state.
+     * @return The Execution object representing the execution status.
+     * @throws InterruptedException if the timeout is reached before the execution reaches the expected state.
+     */
     static Execution waitForExecutionToBe(
             String state,
             String executionId,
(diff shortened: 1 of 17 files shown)


gschueler (approved):

functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
```diff
@@ -116,26 +118,38 @@ class JobUtils {
...
             RdClient client,
-            int iterationGap,
-            int timeout
+            WaitingTime iterationGap,
+            WaitingTime timeout
+    ) {
+        return waitForExecutionToBe(state, executionId, mapper, client, iterationGap.duration, timeout.duration)
+    }
+
+    static Execution waitForExecutionToBe(
```
mrdubr, line 133: Looks like we have a duplicate method in BaseContainer that does very similar thing.
I refactored it yesterday, not being aware that  you're doing this:
https://github.com/rundeck/rundeck/pull/9283/files#diff-389e36f39f7e4b74b8e3e0c5e9996dafa659960d77d217cd06626c4bbb9bdaf4R371

functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
```diff
@@ -116,26 +118,38 @@ class JobUtils {
...
+            String state,
+            String executionId,
```
mrdubr, line 135:
```suggestion
            int executionId,
```
Poor man's URL injection protection :-)

functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
```diff
@@ -116,26 +118,38 @@ class JobUtils {
...
-            int iterationGap,
-            int timeout
+            WaitingTime iterationGap,
+            WaitingTime timeout
+    ) {
+        return waitForExecutionToBe(state, executionId, mapper, client, iterationGap.duration, timeout.duration)
+    }
+
+    static Execution waitForExecutionToBe(
+            String state,
```
mrdubr, line 134: I'd recommend making this a `List<String>` for flexibility.
A common use case is to wait for the job to "complete" -- either successfully or not -- represented by a multitude of statuses: `[         'aborted',             'failed',             'succeeded',            'timedout',             'other'])`

functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
```diff
@@ -116,26 +118,38 @@ class JobUtils {
...
 
+
     static Execution waitForExecutionToBe(
             String state,
             String executionId,
             ObjectMapper mapper,
             RdClient client,
-            int iterationGap,
-            int timeout
+            WaitingTime iterationGap,
```
mrdubr, line 127: I'd recommend providing default values for the `WaitingTime`s

functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
mrdubr, line 133: Should we make this private? If it is too inflexible -- provide a comment saying that the method taking `WaitingTime` should be preferred.

functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
```diff
@@ -116,26 +118,38 @@ class JobUtils {
...
-            if ((System.currentTimeMillis() - initTime) >= TimeUnit.SECONDS.toMillis(timeout)) {
-                throw new InterruptedException("Timeout reached (${timeout} seconds), the execution had \"${executionStatus.status}\" state.")
+            if ((System.currentTimeMillis() - initTime) >= timeout.toMillis()) {
+                throw new InterruptedException("Timeout reached (${timeout.toSeconds()} seconds), the execution had \"${executionStatus.status}\" state.")
             }
             def transientExecutionResponse = client.doGet("/execution/${executionId}")
             executionStatus = mapper.readValue(transientExecutionResponse.body().string(), Execution.class)
             if( executionStatus.status == state ) break
-            Thread.sleep(iterationGap)
+            Thread.sleep(iterationGap.toMillis())
```
mrdubr, line 152: Technically, it should sleep of min(iterationGap, timeout) covering the `iterationGap > timeout` scenario

mrdubr (changes requested):
Please see the minor recommendations for improvements

functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
ahormazabal: What i see is you're adding a similar method in BaseContainer in the other PR, i think you should use the method in JobUtils which is the one being used by all the other tests.

This method already existed, i'm just refactoring only for the purpose of having time units more explicit, and avoid all those `/ 1000 as int` in the tests.

functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
ahormazabal: I'm good with this change, but can we do it in a different PR? Honestly i don't want to go again over the 116 callers to check if they expect an int or a String.

functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
ahormazabal: Similar as the above comment. I like this idea, but its not in the scope of what i'm doing here. Let's make a different PR for this.

functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
ahormazabal: My original idea was to just use Duration and delete WaitingTime. I kept at the end it because i found it was nice to have a way to modify a lot of tests waiting time without going all over them. But i still want to provide the option of just providing a Duration instead of having to modify the enum in case some tests need a more fine tuning.

functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
ahormazabal: Sounds good!

functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
ahormazabal: I'm not sure if i want devs to be allowed to use this method without providing an explicit waiting time. I think it would make the tests code more clear to the ones that will be reading that code for the first time. Maybe would be better to just check for nulls?

mrdubr: I'm worried devs will pick uneccessary long durations because "not sure exactly, what I need, but very_long_duration seems to work locally". This will result in unnecessary slowdowns. I'd recommend defaulting to something short so that overriding would be an explicit choice. Regardless, this is a `nit`. Your call.

mrdubr (approved):
With a few minor `nit`s

functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
ahormazabal: As discussed, we'll implement this in a next phase.

functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
ahormazabal: I did implement it :D

smartinellibenedetti:
Thank you for this refactor, it's way cleaner now! 😄

PR #9305, opened 2024-08-22
RCLOUD-1427: Improve validation of JobId parameter

andrecloutier-pd:
Instead of attempting to identify problematic characters, this PR adjusts the validation to only accept valid UUIDs.

### DIFF
--- a/grails-rundeck-data-shared/src/main/groovy/rundeck/data/util/JobTakeoverQueryBuilder.groovy
+++ b/grails-rundeck-data-shared/src/main/groovy/rundeck/data/util/JobTakeoverQueryBuilder.groovy
@@ -1,12 +1,10 @@
 package rundeck.data.util
 
-import groovy.sql.Sql
-
 class JobTakeoverQueryBuilder {
     static String buildTakeoverQuery(String toServerUUID, String fromServerUUID, boolean selectAll, String projectFilter, List<String> jobids, ignoreInnerScheduled = false) {
         String executionQueryPart = createJobTakeoverExecutionQueryPart(selectAll,fromServerUUID, toServerUUID, projectFilter)
         String useScheduledFlagQueryPart = !ignoreInnerScheduled ? "se.scheduled = true" : null
-        String jobUuidQueryPart = jobids ? "se.uuid in (${safeConvertJobIds(jobids)})" : null
+        String jobUuidQueryPart = jobids ? filteredJobIdsPart(jobids) : null
         String projectPart = projectFilter ? "se.project = :projectFilter" : null
         String serverNodePart = createServerNodeQueryPart(selectAll, fromServerUUID, toServerUUID)
         String sePart = [useScheduledFlagQueryPart,jobUuidQueryPart,projectPart, serverNodePart].findAll{it}.join(" AND ")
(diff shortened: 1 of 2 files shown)


grails-rundeck-data-shared/src/main/groovy/rundeck/data/util/JobTakeoverQueryBuilder.groovy
```diff
@@ -1,22 +1,25 @@
...
     }
 
-    static String safeConvertJobIds(List<String> jobids) {
-        //remove invalid characters in jobids and make sure they are 36 characters long
-        return jobids.findAll{jobid -> jobid?.replace("'","")?.replace(";","")?.length() == 36}.collect{"'${it}'"}.join(",")
+    static String filteredJobIdsPart(List<String> jobids) {
+        var joinedValidJobIds = jobids.findAll{jobid -> isValidUUID(jobid)}.collect{"'${it}'"}.join(",")
+        if (joinedValidJobIds.isEmpty()) {
+            // This will force this part of the condition to be false
+            return "1=0"
```
sjrd218, line 19: Could return null from here. Null parts get removed by the `findAll{it}` on line 10.

andrecloutier-pd: I think that would change the behavior of this part of the query. The intent is to filter based on a given set of Job Ids, however, in this case this due to the filtering, it will become an empty set. Meaning this part of the expression will match nothing. If we return null and omit this condition, it will potentially match everything rather than nothing. If you're sketched out by the use of `1=0` to signal false, it's [not that uncommon](https://stackoverflow.com/questions/9140606/why-would-you-use-where-1-0-statement-in-sql).

grails-rundeck-data-shared/src/main/groovy/rundeck/data/util/JobTakeoverQueryBuilder.groovy
```diff
@@ -42,4 +45,10 @@ class JobTakeoverQueryBuilder {
         return qry
     }
 
+    static boolean isValidUUID(String val) {
+        if (null == val) return false;
+        try { return null!= UUID.fromString(val) } catch (IllegalArgumentException e) {
+            return false
+        }
+    }
```
gschueler, line 53: fyi, `org.rundeck.utils.UUIDs.isValid(String)` does a more extensive validation. `1-1-1-1-1` is a valid UUID, but we expect them to be 36 characters in length

grails-rundeck-data-shared/src/main/groovy/rundeck/data/util/JobTakeoverQueryBuilder.groovy
sjrd218: The criteria version of the query leaves out the jobid in check if no jobids are provided. https://github.com/rundeck/rundeck/blob/8b965bee57f333bf219dd50bd124dfa9441f0d6f/rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy#L4513
Returning null would replicate that behavior.

andrecloutier-pd:
Closing in favour of https://github.com/rundeck/rundeck/pull/9313 .

PR #9314, opened 2024-08-22
RUN-2801: Testdeck refactorings for improved reliability

ahormazabal:
This PR introduces several performance enhancements on the testdeck functional test framework:

- Refactor the `waitForExecutionToBe` method:

This method has been changed to use `java.time` library and introduces a default wait timeout 5 seconds, with a 1 second check period. Also its arguments have been rearranged to allow callers to be able to not specify timeout or period and use the default, or be able to just set the wait timeout. This to make it easier for developers to do waiting in the tests, and to prevent high waiting times to spread when authors rely on copying and pasting other tests.
Also most of the callers have been adjusted to check every one second instead of five, reducing the amount of time lost waiting.

- Switches rundeck database from H2 to Mysql 8: All tests were using H2 as rundeck database. This has been switched to Mysql 8.0 in order to provide a more realistic scenario.

- Upgrades to underlying docker support:
  - Docker compose support has been upgraded to use docker's `docker compose V2` instead of the old `docker-compose` tool, to support newer docker-compose features.
  - Compose files now use "local mode" by default, meaning that testdeck will now use the local `docker compose` command present in the host instead of a containerized version. A flag to revert to the previous mode has been added.

- Improvements in networking:

Many tests were incorrectly communicating to rundeck through the container network instead of the host exposed service and vice-versa, requiring some of them to hack around the rundeck container configuration or the client URLs.
The networking settings and access hosts have been streamlined so the tests (which run in the host) uses the host exposed service, and the operations that ran within containers would access through the container network.


functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
```diff
@@ -155,35 +152,33 @@ class JobUtils {
...
         RdClient client,
-        Duration iterationGap = WaitingTime.LOW,
-        Duration timeout = WaitingTime.MODERATE
+        Duration timeout = WaitingTime.MODERATE,
+        Duration checkPeriod = WaitingTime.LOW
     ) {
         def execDetail = client.doGet("/execution/${executionId}")
-        Execution executionStatus = mapper.readValue(execDetail.body().string(), Execution.class)
+        Execution executionStatus = OBJECT_MAPPER.readValue(execDetail.body().string(), Execution.class)
         long initTime = System.currentTimeMillis()
```
mrdubr, line 170: in 2010, it was strongly discouraged to implement your own thread pool.
in 2020s, it should be discouraged to to implement your own retry logic.
I'm sure there multiple java libs that will turn this while loop into something similar to:
```
Retry.timeout(15).retryEvery(5) { code_to_retry() }
```

functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
mrdubr, line 1: Waiting for an execution status to be "blah" is a special case of waiting for:
 `func(T)` to return true for `T`.

We can have a very generic `waitFor` that takes a closure that does the wait check.
This function should be implemented at a higher level than the JobUtils.

I would use that `waitFor` function right away. I need to wait for the backend to change the Job property after a async UI change.

functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
ahormazabal: I agree with this, but for the moment i don't want to add a new dependency. I couldn't find an existing api that would leverage this within our stack, but probably we could explore spring retry: https://www.baeldung.com/spring-retry

functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
ahormazabal: I'm not sure if i'll have time to refactor all waiting methods, but i could work on a basic `waitFor` as a first step to both consolidate the waiting methods into one, and then change the generic implementation for a suitable library.

functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
ahormazabal: Since the scope of this PR has grown too big, i'll be implementing these improvements in a different PR

functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
ahormazabal: Since the scope of this PR has grown too big, i'll be implementing these improvements in a different PR

functional-test/src/test/groovy/org/rundeck/tests/functional/api/job/JobSpec.groovy
```diff
@@ -24,7 +24,7 @@ class JobSpec extends BaseContainer {
             assert r.successful
             Execution execution = MAPPER.readValue(r.body().string(), Execution.class)
 
-            waitForExecutionStatus(execution.id as String)
+            waitForExecutionFinish(execution.id as String)
```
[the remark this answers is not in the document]
ChuckCrawford: Looks like we can remove this cast since you changed the type?

functional-test/src/test/groovy/org/rundeck/util/container/BaseContainer.groovy
```diff
@@ -364,7 +364,7 @@ abstract class BaseContainer extends Specification implements ClientProvider {
         }
 
         Execution execution = MAPPER.readValue(r.body().string(), Execution.class)
-        waitForExecutionStatus(execution.id as String, WaitingTime.EXCESSIVE)
+        waitForExecutionFinish(execution.id as String, WaitingTime.EXCESSIVE)
```
ChuckCrawford, line 367: We need to change the name of the wait function in this method's comments: https://github.com/rundeck/rundeck/pull/9314/files#diff-389e36f39f7e4b74b8e3e0c5e9996dafa659960d77d217cd06626c4bbb9bdaf4L353-L359.

functional-test/src/test/groovy/org/rundeck/util/container/RdContainer.groovy
```diff
@@ -2,6 +2,7 @@ package org.rundeck.util.container
 
 import groovy.transform.CompileStatic
 import groovy.util.logging.Slf4j
+import org.testcontainers.containers.ComposeContainer
 import org.testcontainers.containers.DockerComposeContainer
```
ChuckCrawford, line 6:
```suggestion
```

functional-test/src/test/groovy/org/rundeck/util/container/RdContainer.groovy
```diff
@@ -10,7 +11,8 @@ import java.time.Duration
 
 @CompileStatic
 @Slf4j
-class RdContainer extends DockerComposeContainer<RdContainer> implements ClientProvider {
+class RdContainer extends ComposeContainer implements ClientProvider {
+//class RdContainer extends DockerComposeContainer<RdContainer> implements ClientProvider {
```
ChuckCrawford, line 15:
```suggestion
```

ChuckCrawford (approved):
A few nitpicky comments to clean up but this looks good to me!

functional-test/src/test/groovy/org/rundeck/tests/functional/api/job/JobExecutionSpec.groovy
```diff
@@ -731,45 +727,35 @@ class JobExecutionSpec extends BaseContainer {
         def execId = parsedExecutionsResponse.id
 
         then:
-        execId > 0
+        !execId?.isBlank()
```
ronaveva, line 730: shouldn't the execId always be not null?
```suggestion
        !execId.isBlank()
```

ronaveva: this applies to every execId validation

ahormazabal: Nice catch, this would made the test pass if execId is null. Fixed.

PR #9326, opened 2024-08-28
remove nextUI test failing in pro

mrdubr:
Removed the test failing in `pro`.


smartinellibenedetti:
Just adding more context to the discussion:
- the issue that is causing this test to fail is a known issue that surfaced during another bug fix in the past;
- the test will be uncommented once the issue is solved;

PR #9336, opened 2024-08-30
RUN-2833 : Introduce wait and poll functions and rewrite a few functional tests to use them

mrdubr:
Jira: https://pagerduty.atlassian.net/browse/RUN-2833

- Introduction of conditional sleep functionality
- deprecation of unconditional sleep functions
- conversion of selected functional tests from unconditional to conditional sleeps

## Perf Improvements

See the test suite execution timing

### Before -> After

Screenshot 2024-09-05 at 2 25 26 PM

Screenshot 2024-09-05 at 2 27 10 PM

Screenshot 2024-09-05 at 2 28 46 PM

Screenshot 2024-09-05 at 4 34 33 PM

### DIFF
--- a/functional-test/src/test/groovy/org/rundeck/tests/functional/api/execution/ExecutionSpec.groovy
+++ b/functional-test/src/test/groovy/org/rundeck/tests/functional/api/execution/ExecutionSpec.groovy
@@ -2,7 +2,9 @@ package org.rundeck.tests.functional.api.execution
 
 import com.fasterxml.jackson.databind.ObjectMapper
 import org.rundeck.util.annotations.APITest
+import org.rundeck.util.common.WaitingTime
 import org.rundeck.util.common.execution.ExecutionStatus
+import org.rundeck.util.common.execution.ExecutionUtils
 import org.rundeck.util.common.jobs.JobUtils
 import org.rundeck.util.common.projects.ProjectUtils
 import org.rundeck.util.container.BaseContainer
@@ -128,17 +130,11 @@ class ExecutionSpec extends BaseContainer {
                 def json2 = client.jsonValue(response.body(), Map)
                 json2.executions.size() == 6
             }
-            // wait for executions to finish
-            sleep 60000
-            // get executions
-            def responseClean = client.doGet("/project/${projectName}/executions")
-            // verify if executions were cleaned
-            verifyAll {
-                responseClean.successful
-                responseClean.code() == 200
-                def json3 = jsonValue(responseClean.body())
-                json3.executions.size() == 0
-            }
+
+            // Waits for all executions to get cleaned
+            assert waitFor(ExecutionUtils.Retrievers.executionsForProject(client, projectName),
+                    {it.isEmpty()},
+            WaitingTime.EXCESSIVE).isEmpty()
             deleteProject(projectName)
         cleanup:
             tmpjar.delete()
@@ -164,8 +160,9 @@ class ExecutionSpec extends BaseContainer {
     def "execution state OK"() {
         when:
             def runCommand = post("/project/${PROJECT_NAME}/run/command?exec=echo+testing+execution+api", null, Map)
-            def idExec = runCommand.execution.id
-            sleep 5000
+            def idExec = runCommand.execution.id as String
+            // Waits for all executions to finish
+            waitFor(ExecutionUtils.Retrievers.executionById(client, idExec), ExecutionUtils.Verifiers.executionFinished())
             def response = doGet("/execution/${idExec}/state")
         then:
             verifyAll {
@@ -351,13 +348,19 @@ class ExecutionSpec extends BaseContainer {
             }
         cleanup:
             (2..4).each {
-                updateConfigurationProject("${projectNameSuffix}-${it}", [
+                def projectName = "${projectNameSuffix}-${it}"
+                updateConfigurationProject(projectName, [
                     "project.disable.schedule": "true",
                     "project.later.schedule.enable": "false",
                     "project.disable.executions": "true"
                 ])
-                hold 10 //Wait until the executions stop
-                deleteProject("${projectNameSuffix}-${it}")
+
+                // Waits for all executions to finish
+                waitFor(ExecutionUtils.Retrievers.executionsForProject(client, projectName),
+                        verifyForAll(ExecutionUtils.Verifiers.executionFinished()),
+                        WaitingTime.EXCESSIVE )
+
+                deleteProject(projectName)
             }
     }
 
--- a/functional-test/src/test/groovy/org/rundeck/tests/functional/api/execution/ExecutionSpec.groovy
+++ b/functional-test/src/test/groovy/org/rundeck/tests/functional/api/execution/ExecutionSpec.groovy
@@ -130,13 +130,13 @@ class ExecutionSpec extends BaseContainer {
                 def json2 = client.jsonValue(response.body(), Map)
                 json2.executions.size() == 6
             }
-
-            // Waits for all executions to get cleaned
-            assert waitFor(ExecutionUtils.Retrievers.executionsForProject(client, projectName),
-                    {it.isEmpty()},
-            WaitingTime.EXCESSIVE).isEmpty()
-            deleteProject(projectName)
         cleanup:
+            // Waits for all executions to get cleaned
+            waitFor(ExecutionUtils.Retrievers.executionsForProject(client, projectName),
+                {it.isEmpty()},
+                WaitingTime.EXCESSIVE)
+            deleteProject(projectName)
+
             tmpjar.delete()
         where:
             version | projectName
--- a/functional-test/src/test/groovy/org/rundeck/tests/functional/util/WaitUtilsSpec.groovy
+++ b/functional-test/src/test/groovy/org/rundeck/tests/functional/util/WaitUtilsSpec.groovy
@@ -1,0 +1,88 @@
+package org.rundeck.tests.functional.util
+
+import org.rundeck.util.annotations.APITest
+import org.rundeck.util.common.ResourceAcceptanceTimeoutException
+import org.rundeck.util.common.WaitUtils
+import spock.lang.Specification
+
+@APITest
+class WaitUtilsSpec extends Specification {
+
+    static final Map<String, String> TEST_VALUES = ["1": "one", "2": "two"].asImmutable()
+
+    def testWaitForThatNeverWaits() {
+        given:
+        Closure<String> retriever = { return TEST_VALUES[it] }
+
+        when:
+        String result = WaitUtils.waitForResource("1", retriever)
+
+        then:
+        result == "one"
+    }
+
+    def testWaitForThatWaitsAndFailsToGetAccepted() {
+        given:
+        Closure<String> retriever = { return "123" }
+
+        Closure<Boolean> resourceAcceptanceEvaluator = { false }
+
+        Closure<String> acceptanceFailureOutputProducer = { String key ->
+            return "Custom output $key".toString()
+        }
+
+        when:
+        WaitUtils.waitForResource("1", retriever, resourceAcceptanceEvaluator, acceptanceFailureOutputProducer)
+
+        then:
+        ResourceAcceptanceTimeoutException e = thrown(ResourceAcceptanceTimeoutException)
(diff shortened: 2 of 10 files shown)


functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
```diff
@@ -197,6 +195,74 @@ class JobUtils {
...
+     * Waits for the resources to be in the expected state.
+     * @param resourceIdentifiers collection of resource identifiers. Each identifier should be passable into the resourceRetriever as an input argument.
+     * @param resourceRetriever function that takes the RID resourceId and returns the current state of the R resource.
+     * @param resourceAcceptanceEvaluator function that evaluates the state of the resource by taking a R resource and returning a boolean.
+     * @param acceptanceFailureOutputProducer function that produces a string output when the resource acceptance evaluator returns false. Takes RID and R as two params.
+     * @param timeout max duration to wait for the resource to reach the expected state.
+     * @param checkPeriod  time to wait between each check.
+     * @return a map of resource identifiers and their corresponding resources for the resources that met the acceptance criteria. An identifier maps to null for resources that did not meet the acceptance criteria.
+     */
+    static <RID, R> Map<RID, R> waitFor(Collection<RID> resourceIdentifiers,
```
mrdubr, line 208: Ignore this function. It's untested WIP.

functional-test/src/test/groovy/org/rundeck/util/common/WaitUtils.groovy
```diff
@@ -0,0 +1,78 @@
...
+
+import groovy.util.logging.Slf4j
+
+import java.time.Duration
+
+@Slf4j
+class WaitUtils {
+
+    /**
+     * Waits for the resources to be in the expected state.
```
mrdubr, line 11: This one is an untested work-in-progress. Ignore it for now.

functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
```diff
@@ -165,22 +165,20 @@ class JobUtils {
...
-                throw new InterruptedException("Timeout reached (${timeout.toSeconds()} seconds), the execution had \"${executionStatus.status}\" state. Execution output was: \n${execOutput}\n")
-            }
-            def transientExecutionResponse = client.doGet("/execution/${executionId}")
-            executionStatus = OBJECT_MAPPER.readValue(transientExecutionResponse.body().string(), Execution.class)
-            if (expectedStates.contains(executionStatus.status)) {
-                break
-            }
-            Thread.sleep(Math.min(checkPeriod.toMillis(), timeout.toMillis()))
+
+        Closure<Execution> resourceRetriever = {String id ->
```
mrdubr, line 169: An example of refactoring of an existing method to use the WaitUtils

functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
```diff
@@ -165,22 +165,20 @@ class JobUtils {
         Duration timeout = WaitingTime.MODERATE,
         Duration checkPeriod = WaitingTime.LOW
     ) {
-        def execDetail = client.doGet("/execution/${executionId}")
```
mrdubr, line 168: An example of refactoring of an existing method to use the WaitUtils

functional-test/src/test/groovy/org/rundeck/tests/functional/util/WaitUtilsSpec.groovy
```diff
@@ -0,0 +1,88 @@
...
+
+    def testWaitForThatNeverWaits() {
+        given:
+        Closure<String> retriever = { return TEST_VALUES[it] }
+
+        when:
+        String result = WaitUtils.waitForResource("1", retriever)
+
+        then:
+        result == "one"
```
ahormazabal, line 21: How do we make sure that the code (or future changes) never actually waited?

functional-test/src/test/groovy/org/rundeck/util/common/execution/ExecutionUtils.groovy
```diff
@@ -0,0 +1,89 @@
...
+
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper()
+    static final List<String> EXECUTION_FINISH_STATUSES = [
+            'aborted',
+            'failed',
+            'succeeded',
+            'timedout',
+            'other'].asImmutable()
+
+    static class Verifiers {
```
ahormazabal, line 18: I like this idea of a class with common Verifiers and Retrievers, but maybe they would be better as independent classes with many kinds of each as this api expands, instead of having different classes of them for each domain.
I.e: `Verifiers.executionFinished()` and `Verifiers.jobUpdated()` so they are easier to find by the uninitiated dev.
OR maybe: `Verifiers.Executions.executionFinished()` would be a good approach too.

functional-test/src/test/groovy/org/rundeck/util/common/execution/ExecutionUtils.groovy
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    static class Retrievers {
+        /**
+         * Returns a closure that retrieves an execution by its id.
+         * @param client
+         * @param executionId
+         * @return
+         */
+        static final Closure<Execution> executionById(RdClient client, String executionId) {
```
ahormazabal, line 51: This is not a request for change but just a thought: I'm realizing that these methods would end up being very handy outside of the waiting scope. Maybe we could explore a way for these to return the plain object instead of a Closure, while still being usable by the waiting api. Or maybe i'm dreaming too high :)

functional-test/src/test/groovy/org/rundeck/util/container/BaseContainer.groovy
```diff
@@ -279,15 +280,16 @@ abstract class BaseContainer extends Specification implements ClientProvider {
...
+        waitFor(
+                {
+                    //force refresh project
+                    client.doPutWithJsonBody("/project/$project/config/time", ["time": System.currentTimeMillis()])
+                    response = client.doGet("/project/$project/resources")
+                    safelyMap(response, Map.class, { [:] })
+                },
+                { it.get(nodename) != null },
+                WaitingTime.EXCESSIVE
+        )
```
ahormazabal, line 292: This is what i dreaded about the high reliance on closures, and one of the things i hated the most when i first started working in the rundeck codebase.
For a dev that first stumbles into this and its not deep into groovy or functional programming, this would be difficult to understand, and will require them to go study closures, the waitFor code and/or ask for an explanation in order to get what it does.
A builder pattern would be more clear as you get explicit info on what is expected from each closure, while still allowing their use.

functional-test/src/test/groovy/org/rundeck/util/common/WaitUtils.groovy
```diff
@@ -0,0 +1,108 @@
...
+     * @param resourceAcceptanceEvaluator evaluates the state of the resource and and returns true if the resource is accepted.
+     * @param timeout max duration to wait for the resource to reach the expected state.
+     * @param checkPeriod time to wait between each check.
+     * @return the resource that met the acceptance criteria.
+     * @throws ResourceAcceptanceTimeoutException if the timeout is reached waiting for the resource to reach the expected state.
+     */
+    @TypeChecked
+    static <R> R waitFor(
+            Closure<R> retryableResourceRetriever,
+            @ClosureParams(value = FromString, options = ["R"]) Closure<Boolean> resourceAcceptanceEvaluator = { R r -> !!r },
```
ahormazabal, line 23: - What does the `!!` operator exactly does? It's the second time i see it somewhere but can't find any info on it. Is groovy? Is java 11?
- This also sums to my concerns specifically about groovy closures, as they strength of being too flexible is also their weakness as you're never sure what they actually expect without IDE introspection. Maybe using `java.util.Function` would be better for these cases?

mrdubr: - It has to do with groovy truthy and falsy. Consider: 
```
def foo = "bar"
println foo
println !foo
println !!foo

Output:
bar
false
true
```
In this case, returning an `r` violates typing, but `!!r` results in a `Boolean` (with compiler autoboxing)

functional-test/src/test/groovy/org/rundeck/tests/functional/api/execution/ExecutionSpec.groovy
```diff
@@ -128,17 +130,11 @@ class ExecutionSpec extends BaseContainer {
...
-                responseClean.successful
-                responseClean.code() == 200
-                def json3 = jsonValue(responseClean.body())
-                json3.executions.size() == 0
-            }
+
+            // Waits for all executions to get cleaned
+            assert waitFor(ExecutionUtils.Retrievers.executionsForProject(client, projectName),
+                    {it.isEmpty()},
+            WaitingTime.EXCESSIVE).isEmpty()
```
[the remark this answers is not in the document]
mrdubr: The problem here is that this statement appears in the `then:` block, thus is being asserted by the testing framework. The `wait` returns `[]`, which is considered a failure by the framework, thus the workaround is to check for emptiness to ensure the statement evaluates to `true`. Let me refactor this...

functional-test/src/test/groovy/org/rundeck/tests/functional/util/WaitUtilsSpec.groovy
mrdubr: Changed to wait for `0`.

functional-test/src/test/groovy/org/rundeck/util/common/execution/ExecutionUtils.groovy
mrdubr: There are pros and cons to both approaches. Fortunately, such refactoring is easy with IDEs.
I'd propose we push in what we have and see how it picks up prior to premature refactoring.

functional-test/src/test/groovy/org/rundeck/util/common/execution/ExecutionUtils.groovy
mrdubr: Hmm... I think `waitForExecution` in the JobUtils does that. It returns an execution object and relies on this closure for behaviour implementation.

functional-test/src/test/groovy/org/rundeck/util/container/BaseContainer.groovy
mrdubr: You're making an excellent point. As a team, we need to decide what is our appetite for newer language features (like closures and FP for example). While I do agree that the Builder pattern might aid with comprehension in this particular scenario it comes with complexity that I'd like to avoid.

functional-test/src/test/groovy/org/rundeck/util/common/WaitUtils.groovy
ahormazabal: - Kinda hacky IMO. why don't just return true or let it throw an exception? 
- Using Function in practice should be mostly the same, but instead of declaring `@ClosureParams(value = FromString, options = ["R"]) Closure<Boolean> resourceAcceptanceEvaluator` you can just `Function<R, Boolean> resourceAcceptanceEvaluator`.

functional-test/src/test/groovy/org/rundeck/util/container/BaseContainer.groovy
ahormazabal: Made a small PoC of what i'm talking about: [branch](https://github.com/rundeck/rundeck/compare/avg.poc-wait-for-resource...avg.poc-wait-for-resource-builder?expand=1)

functional-test/src/test/groovy/org/rundeck/util/common/execution/ExecutionUtils.groovy
ahormazabal: My only objective with this point it to prevent proliferation of classes when all could be managed by one or two, but as you comment we can improve that in a later step. 👍🏽

ahormazabal (approved):
Even though there're few nits i don't agree with, i think this PR is a huge improvement so i'm approving it.

PR #9340, opened 2024-08-30
Fix logic on method waitForModal

Jesus-Osuna-M:
Jira ticket: https://pagerduty.atlassian.net/browse/RUN-2822

Context: Modifications were made to the waitForModal method in the RunnerResourceModelSpec class to enhance its flexibility and eliminate redundant usage.

Changes Made:

Refactoring of waitForModal Method:

The original waitForModal method took an expected parameter and waited for the modal to contain exactly that number of elements before proceeding. If the condition wasn’t met within a specified timeout, a TimeoutException was thrown.
The method was modified to include a second optional parameter, modalFieldCssSelector, which allows specifying a custom CSS selector to identify the modal. This enhances the method's flexibility by enabling it to work with different modals on the same page.
Removal of Redundant waitForModal Calls:

Several redundant calls to waitForModal(0) were removed. These calls were previously checking that the modal was not present before performing certain actions. By optimizing these checks elsewhere in the code, the need for these specific calls was eliminated, reducing code complexity and test execution time.

### DIFF
--- a/functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/appadmin/AccessControlSpec.groovy
+++ b/functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/appadmin/AccessControlSpec.groovy
@@ -1,5 +1,6 @@
 package org.rundeck.tests.functional.selenium.appadmin
 
+import org.openqa.selenium.By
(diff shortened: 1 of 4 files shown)


ronaveva (approved):
LGTM

PR #9342, opened 2024-09-03
RUN-2376 api/job/id/executions to return referenced executions

carlosrfranco:

This PR fixes two issues related to referenced executions:
1- The activity tab at job page wasn't showing referenced executions as previous versions did.
2- The api endpoint api/jjob/{id}/executions to return the same result as activity tab at job page, including referenced executions

For the first issue listed above, it fixes the gorm data provider to collect execution uuid (instead of the job uuid).
For the second case, the criteria to build query to list executions should check also the referenced executions table. For the api calls, it is included by default but can be used the parameter includeJobRef as false if it is not desired to get the referenced executions

### DIFF
--- a/functional-test/src/test/groovy/org/rundeck/tests/functional/api/job/JobExecutionSpec.groovy
+++ b/functional-test/src/test/groovy/org/rundeck/tests/functional/api/job/JobExecutionSpec.groovy
@@ -1900,6 +1900,127 @@ class JobExecutionSpec extends BaseContainer {
         client.jsonValue(client.doGetAcceptAll("/execution/$jobExecutionStatus.id/state").body(), Map).nodes.values()[0].every({val -> val.executionState == "SUCCEEDED"} )
     }
 
+    def "test /api/job/id/executions returns referenced executions"() {
+        given:
+        def projectName = "test-referenced-executions"
+        def client = getClient()
+        Object projectJsonMap = [
+                "name": projectName.toString(),
+                "description": "test-referenced-executions",
+                "config": [
+                        "test.property": "test value",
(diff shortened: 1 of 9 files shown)


functional-test/src/test/groovy/org/rundeck/tests/functional/api/job/JobExecutionSpec.groovy
```diff
@@ -1900,6 +1900,122 @@ class JobExecutionSpec extends BaseContainer {
...
+                ExecutionStatus.SUCCEEDED.state,
+                exec.id as String,
+                client,
+                WaitingTime.EXCESSIVE)
+
+        then:
+        jobExecutionStatus.status == ExecutionStatus.SUCCEEDED.state
+
+        when: "get executions"
+        sleep 5000
```
ahormazabal, line 1998: Why is this sleep necessary?

ahormazabal (changes requested):
LGTM and works ok. But i'd extend the scope of the tests

functional-test/src/test/groovy/org/rundeck/tests/functional/api/job/JobExecutionSpec.groovy
carlosrfranco: I removed this sleep. I dont think it will be needed since it`s already calling `waitForExecution` method

ahormazabal (approved):

gschueler (changes requested):
Documentation should be updated in the API OpenAPI spec for the updated endpoints

gschueler (approved):

PR #9343, opened 2024-09-04
RUN-2242: Plugins control security issue

smartinellibenedetti:

Security issue: the plugins control allows access by users with read permission to a project.

Upon investigating, turns out that the plugins control page actually is meant to be accessed by users with read permission to a project. Therefore, adjust the following things:
- navBarProjectSettingsData.gsp to allow the project-config menu items have the enabled property;
- update _navData.gsp, so that it can consider the read permissions when deciding to render the menu item for project settings  (if there are project-config menu items to be displayed, consider the read project permissions as well, otherwise keep the original permissions)
- update navBar vue component to not display the settings items that are not enabled;

Considered to change the access of the page, so that it can only be accessed by users with editing permissions. But as the information regarding which plugins are available was deemed not concerning, followed the team's guidance to display the menu item instead.

**Additional context**


rundeckapp/grails-app/i18n/messages.properties
```diff
@@ -1915,7 +1915,7 @@ end=End
 job.editor.workflow.unsavedchanges.warning=Some changes to the workflow have not been completed.
 scheduledExecution.property.executionLifecyclePluginConfig.help.text=Selected Plugins will be enabled for this Job.
 scheduledExecution.executionLifecyclePlugins.pluginTypeNotFound.message=Execution Life Cycle plugin type "{0}" was not found or could not be loaded
-scheduledExecution.executionLifecyclePlugins.invalidPlugin.message=Invalid Configuration for Execution Life Cycle plugin: {0} - {1}
+scheduledExecution.executionLifecyclePlugins.invalidPlugin.message=Invalid Configuration for Execution Life Cycle plugin: {0}
```
smartinellibenedetti, line 1918: Included an improvement related to another change in this release. If this is a problem I can split the PR.

carlosrfranco (approved):
LGTM

PR #9361, opened 2024-09-12
RUN-2612: Functional Tests for Job Listing API

mrdubr:
Additional functional tests to test the [project's jobs](https://docs.rundeck.com/docs/api/#listing-jobs)  API.

### DIFF
--- a/functional-test/src/test/groovy/org/rundeck/tests/functional/api/job/JobsSpec.groovy
+++ b/functional-test/src/test/groovy/org/rundeck/tests/functional/api/job/JobsSpec.groovy
@@ -6,28 +6,69 @@ import org.rundeck.util.annotations.APITest
 import org.rundeck.util.common.jobs.JobUtils
 import org.rundeck.util.container.BaseContainer
 
+/**
+ * Tests for the jobs API endpoint
+ */
 @APITest
 class JobsSpec extends BaseContainer {
 
+    static final def newProject = UUID.randomUUID().toString()
+    static List<String> createdJobIds
+
     def setupSpec() {
         startEnvironment()
-        setupProject()
+        setupProject(newProject)
+
+        createdJobIds = createJobs()
     }
 
-    def "Listing RunDeck Jobs for project test"() {
+    def cleanupSpec() {
+        deleteProject(newProject)
+    }
+
+    def "Listing jobs for a project that has no jobs"() {
         when:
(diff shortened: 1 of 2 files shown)


functional-test/src/test/groovy/org/rundeck/tests/functional/api/job/JobsSpec.groovy
```diff
@@ -6,28 +6,71 @@ import org.rundeck.util.annotations.APITest
...
+ */
 @APITest
 class JobsSpec extends BaseContainer {
 
+    static final def newProject = UUID.randomUUID().toString()
+    static List<String> createdJobIds
+
     def setupSpec() {
         startEnvironment()
         setupProject()
```
smartinellibenedetti, line 20:
```suggestion
```

functional-test/src/test/groovy/org/rundeck/tests/functional/api/job/JobsSpec.groovy
```diff
@@ -6,28 +6,71 @@ import org.rundeck.util.annotations.APITest
...
-    def "Listing RunDeck Jobs for project test"() {
+    def cleanupSpec() {
+        deleteProject(newProject)
+    }
+
+    def "Listing jobs for a project that has no jobs"() {
         when:
-            def response = doGet("/project/${PROJECT_NAME}/jobs")
+        def aProject = UUID.randomUUID().toString()
+        setupProject(aProject)
```
smartinellibenedetti, line 34: It shouldn't be necessary to call setupProject here again

smartinellibenedetti (approved):
Looks good to me, the only change necessary is to remove the extra setupProject calls. I would suggest adding a column to the table with a description of what is being tested to make it easier to identify the cases in the future, but that's totally optional.

functional-test/src/test/groovy/org/rundeck/tests/functional/api/job/JobsSpec.groovy
mrdubr: That's another project being set up in addition to the one in setupSpec()

mrdubr:
> Looks good to me, the only change necessary is to remove the extra setupProject calls. I would suggest adding a column to the table with a description of what is being tested to make it easier to identify the cases in the future, but that's totally optional.

That's an excellent suggestion. I will keep this in mind for the next time. Thanks for the suggestion.

PR #9364, opened 2024-09-16
RUN-2857: navbar fix for "more" items

smartinellibenedetti:
Bug introduced by RUN-2242 changes, where items in the submenus won't be displayed if they don't have enabled property. 

Fix adds enabled property to navbar items in case they don't have one.

**Additional context**


jayas006 (approved):
LGTM

PR #9366, opened 2024-09-16
RUN-2847: Flaky - rundeck: T Selenium > disable schedules at project level

smartinellibenedetti:
https://app.circleci.com/pipelines/github/rundeck/rundeck?branch=RUN-2847-ace
Sometimes the test fails as the setting that needs to be disabled isn't set in the project config.

Add a try-catch, so that if setting isn't available, it will be added anyway at the end of the config.

**Additional context**

### DIFF
--- a/functional-test/src/test/groovy/org/rundeck/util/gui/pages/project/ProjectEditPage.groovy
+++ b/functional-test/src/test/groovy/org/rundeck/util/gui/pages/project/ProjectEditPage.groovy
@@ -216,7 +216,11 @@ class ProjectEditPage extends BasePage {
      * @param replacement
      */
     def replaceConfiguration(String original, String replacement){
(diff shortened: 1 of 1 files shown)


functional-test/src/test/groovy/org/rundeck/util/gui/pages/project/ProjectEditPage.groovy
```diff
@@ -216,7 +216,11 @@ class ProjectEditPage extends BasePage {
      * @param replacement
      */
     def replaceConfiguration(String original, String replacement){
-        ((JavascriptExecutor) context.driver).executeScript("ace.edit('_id0').session.replace(ace.edit('_id0').find('${original}',{wrap: true, wholeWord: true }), '${replacement}');")
+        try {
```
mrdubr, line 219: Looks pretty fragile and hacky but my lack of knowledge in this area prevents me from offering an alternative solution. If you believe there is no better alternative, I will approve it, but I would urge to give it another thought.

mrdubr (approved):
Approved, with a nit

PR #9369, opened 2024-09-18
Wait improve

mrdubr:

**Additional context**


smartinellibenedetti:
Suggestion/question:

IMO this is a situation where the refactoring might come at the expense of readability. This isn't a big loss but I'd prefer to keep the readability, so I'd suggest doing ie `waitForAnyResource` that as `waitForAllResources` calls under the hood `waitForResource`, with the proposed resourceAcceptanceEvaluator change of this draft. 

So that it's clear when reading the code, that the test is waiting for at least one resource to be found while reading `waitForResource` without the acceptance evaluator, might require devs to read the context to figure out if the test is waiting for a single one or for N to appear.

mrdubr:
> IMO this is a situation where the refactoring might come at the expense of readability. This isn't a big loss but I'd prefer to keep the readability, 

The patter of `waiting` for truthy is pretty popular in groovy. 
Ex: https://www.gebish.org/manual/current/api/geb/waiting/Wait.html#waitFor(Closure)
Currently, we are waiting for a non-null value -- that's very `java`ish. 

> so I'd suggest doing ie `waitForAnyResource` that as `waitForAllResources` calls under the hood `waitForResource`, with the proposed resourceAcceptanceEvaluator change of this draft.

I can't find `waitForAnyResource`, but `waitForAllResources` uses the newly proposed resourceAcceptanceEvaluator

> So that it's clear when reading the code, that the test is waiting for at least one resource to be found

The wait is not for one, at least one or many. 
The wait is for the `resourceRetriever` to return groovy trurthy value: https://groovy-lang.org/semantics.html#the-groovy-truth which depends on the `resoruceRetrier` return type.

smartinellibenedetti:
> > IMO this is a situation where the refactoring might come at the expense of readability. This isn't a big loss but I'd prefer to keep the readability,
> 
> The patter of `waiting` for truthy is pretty popular in groovy. Ex: https://www.gebish.org/manual/current/api/geb/waiting/Wait.html#waitFor(Closure) Currently, we are waiting for a non-null value -- that's very `java`ish.
> 
> > so I'd suggest doing ie `waitForAnyResource` that as `waitForAllResources` calls under the hood `waitForResource`, with the proposed resourceAcceptanceEvaluator change of this draft.
> 
> Yes, both `waitForResource` and `waitForAllResources` use the newly proposed resourceAcceptanceEvaluator: `{ it ? true : false }`
> 
> > So that it's clear when reading the code, that the test is waiting for at least one resource to be found
> 
> The wait is not for one, at least one or many. The wait is for the `resourceRetriever` to return groovy trurthy value: https://groovy-lang.org/semantics.html#the-groovy-truth which depends on the `resoruceRetrier` return type.

There has been a misunderstanding, my suggestion had nothing to do in regards of using only one evaluator or another. My suggestion was to create a new method, ie `waitForAnyResource` (that exactly waits for at least one or many). Please read my statement again:

> So that it's clear when reading the code, that the test is waiting for **at least one** resource to be found

PR #9371, opened 2024-09-19
RUN-2948: refactor jsch dependencies

gschueler:

* moves jsch dependencies to the correct place (jsch-plugin)
* refactor SSHAgentProcess to no longer use jsch type, and deprecate old class
  * add SSHAgentUtil to replace the behavior

**Additional context**


ahormazabal (changes requested):
LGTM but a functional test case should be added to `JschNodeExecutorSpec` using an ssh-agent.

gschueler:
@ahormazabal can you try this again?

PR #9374, opened 2024-09-23
RUN-2241: Provider to add cache control headers to http responses

ahormazabal:

Security enhancement to add cache control headers to app responses, preventing local caching of sensitive data and browsing back once logged out.

Added a new `SecurityHeaderProvider` at the `grails-securityheaders` plugin to add the following headers to responses by default:
```
 Cache-Control: no-cache, no-store, max-age=0, must-revalidate
 Pragma: no-cache
 Expires: 0
 ```

- Using [Grails cache headers plugin](http://grails-plugins.github.io/cache-headers/latest/), but would be overshoot for the scope of this fix. 
- Adding spring security filters to the filter chain, which would had the same effect as reusing the existing `RundeckSecurityHeadersFilter` but with more complex and unclear configuration.

**Additional context**
- The implementation of the provider follows the same method used by Spring Security's [CacheControlHeadersWriter](https://github.com/spring-projects/spring-security/blob/main/web/src/main/java/org/springframework/security/web/header/writers/CacheControlHeadersWriter.java)

**Testing**

By default you should:
- Be able to see the specified headers in most of the application http responses.
- You shouldn't be able to navigate back to previous pages in browser history once logged out.

**Configuration**

To disable:
`rundeck.security.httpHeaders.provider.cache-control.enabled=false`


mrdubr (approved):
LGTM

PR #9382, opened 2024-10-03
RUN-2419 skip interceptor if request to an error page

carlosrfranco:
When Rundeck for some reason loses the connection with database, it causes to print a lot of errors to the log file, making its size to increase very fast

If a request to an error page reaches the interceptor, it should skip the interceptor validations and let the request to proceed. Then it avoid the request to be stuck in a loop that is the cause of the issue

Firstly, I tried to wrap part of the before method in a try-catch block to return an error status and error message to interrupt the request

**Additional context**

### DIFF
--- a/rundeckapp/grails-app/controllers/rundeck/interceptors/ProjectSelectInterceptor.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/interceptors/ProjectSelectInterceptor.groovy
@@ -53,6 +53,7 @@ class ProjectSelectInterceptor {
 
     ProjectSelectInterceptor() {
         matchAll().excludes(controller: 'framework', action:'(createProject(Post)?|selectProject|projectSelect|noProjectAccess|(create|save|check|edit|view)ResourceModelConfig)')
+                  .excludes(controller: 'error', action: 'fiveHundred')
     }
 
 
@@ -69,9 +70,6 @@ class ProjectSelectInterceptor {
             return true
         }
         if(controllerName=='menu' && ( actionName in ['home'] )){
-            return true
-        }
-        if(controllerName=='error' && ( actionName in ['fiveHundred'] )){
             return true
         }
         if (session && session.user && session.subject && params.project) {
--- a/rundeckapp/grails-app/controllers/rundeck/interceptors/ProjectSelectInterceptor.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/interceptors/ProjectSelectInterceptor.groovy
@@ -18,6 +18,7 @@ package rundeck.interceptors
 
 import com.dtolabs.rundeck.core.authorization.AuthContext
 import com.dtolabs.rundeck.core.config.Features
+import org.hibernate.exception.JDBCConnectionException
 import org.rundeck.app.access.InterceptorHelper
 import org.rundeck.app.authorization.AppAuthContextEvaluator
 import org.rundeck.core.auth.AuthConstants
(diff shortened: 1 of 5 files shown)


mrdubr (commented):
Based on your description:

> In an interceptor, it checks if the project is disabled and try to access the database. If the jdbc doesnt have a connection with the database it will fail but the request will not return an error. So, to fix that, the code in the interceptor was wrapped up in a try-catch clause to get the jdbc exception and return an error

I'm not sure I fully understand why the [repro](https://pagerduty.atlassian.net/browse/RUN-2419?atlOrigin=eyJpIjoiZTY5NmJlOWE1YjQ1NDE1ZDkzNjIwZWEyMGQzMjIwMmYiLCJwIjoiaiJ9) steps need scheduled jobs?

IIUC, you're fixing an API request path. Can you please explain.

carlosrfranco:
> Based on your description:
> 
> > In an interceptor, it checks if the project is disabled and try to access the database. If the jdbc doesnt have a connection with the database it will fail but the request will not return an error. So, to fix that, the code in the interceptor was wrapped up in a try-catch clause to get the jdbc exception and return an error
> 
> I'm not sure I fully understand why the [repro](https://pagerduty.atlassian.net/browse/RUN-2419?atlOrigin=eyJpIjoiZTY5NmJlOWE1YjQ1NDE1ZDkzNjIwZWEyMGQzMjIwMmYiLCJwIjoiaiJ9) steps need scheduled jobs?
> 
> IIUC, you're fixing an API request path. Can you please explain.

If you go through the steps you can see that after disconnecting the database step, you should try to navigate on the GUI (step 6). So, from what I could reproduce, the error is caused by the requests performed while we are navigating on the GUI pages that and are intercepted by the interceptor

mrdubr:
Looking at the `service.log`, there are multiple scenarios that produce error logs when a DB is inaccessible. This is not unexpected. You are only addressing one of them in this PR. 
I cannot figure out what makes this scenario produce so many error logs? Why short-circuiting it would help?

carlosrfranco:
Yes, the errors is expected and this PR will not avoid the errors to be printed in the log. But, for the interceptor, when it throws an error that is not handled so that the request respond an error status, the request appears to be called like a loop and it produces so many logs that makes the log file to increase very fast.

mrdubr:
> the request appears to be called like a loop and it produces so many logs that makes the log file to increase very fast.

And that's the pickle! :-) 
Who/what is causing these API calls being made in a tight loop? Do we know?
Excessive logging is the side affect of this behaviour that I'm concerned about.

mrdubr (dismissed):
The proposed solution swallows the exception hiding the wealth of information that is exposed to us by the underlying frameworks. In this particular scenario (db is inaccessible), such information is not useful since we know what's going on. This might not be the case all the time (when we know the reason) and will result in us flying blind.

IMO, we need to understand what causes the spike of requests and fix the source of the problem. Swallowing exceptions should be the last resort.

carlosrfranco:
@mrdubr I agree with you when you said about fixing the issue in a broader way and not strictly related to the DB connection issue. Maybe we could catch a more generic and not so exclusive exception (catch Exception instead of JDBCException?).

About fix the issue by checking the where and how Rundeck make the requests on the GUI: from what I've seen in my tests and while debugging this issue, the Rundeck GUI is not actually causing the problem, I mean, the application is not triggering multiple requests. In fact, Rundeck makes only one request that reaches the interceptor (multiple requests match with this interceptor). However, in the "before()" method is expected to return true or false for each request so that it can proceed or not with the request. When an exception is thrown in the interceptor, no boolean value is returned and the request gets stuck in a loop inside the interceptor. This seems to me to be a problem with the way Grails handles the interceptor (I even found some Grails issues reported on the internet that are related to this, but only in older versions of Grails).

Anyway, it is perfectly expected that the interceptor returns an error if any step of the interceptor throws an exception, and I believe that this solution would not be suppressing any information to the user about what is happening.

mrdubr:
> <..> the Rundeck GUI is not actually causing the problem, I mean, the application is not triggering multiple requests. In fact, Rundeck makes only one request that reaches the interceptor (multiple requests match with this interceptor).

This is great! It means we can reproduce this easily!

> However, in the "before()" method is expected to return true or false for each request so that it can proceed or not with the request. When an exception is thrown in the interceptor, no boolean value is returned and the request gets stuck in a loop inside the interceptor. 

This is a very surprising behaviour. I would expect an exception in a filter/plugin/interceptor to materialize as a 5xx response.
If this is the framework behaviour -- your solution makes perfect sense.
However, can you please point me to the framework code that does this -- loops on an exception. 

> Anyway, it is perfectly expected that the interceptor returns an error if any step of the interceptor throws an exception, and I believe that this solution would not be suppressing any information to the user about what is happening.

https://github.com/rundeck/rundeck/pull/9382/files#diff-ff3331f0376891dc9e0b56ecca9b02953a4b44447ea2dead86c24e8f04ad853fR131-R139
You've suppressed logging of the exception (and its causes).
`e.message` provides very little value.

carlosrfranco:
@mrdubr 
That is what I believe is happening:

On the class GrailsInterceptorHandlerInterceptorAdapter.groovy:78, it go through a interceptors list and for those that matches with the request, it call the before() method.

```
boolean preHandle(HttpServletRequest request, HttpServletResponse response, Object handler) throws Exception {
        if(!interceptors.isEmpty()) {
            List<Interceptor> matchInterceptors = []
            request.setAttribute(ATTRIBUTE_MATCHED_INTERCEPTORS, matchInterceptors)
            for(i in interceptors) {
                if(i.doesMatch(request)) {
                    matchInterceptors.add(i)
                    if( !i.before() ) {
                        return false
                    }
                }
            }
        }
        return true
    }

```
When it calls the method from projectSelectInterceptor and it throws an exception, the execution will reach the spring-webmvc line code DispatcherServlet.java:1089

´processDispatchResult(processedRequest, response, mappedHandler, mv, dispatchException)´


```
			catch (Exception ex) {
				dispatchException = ex;
			}
			catch (Throwable err) {
				// As of 4.3, we're processing Errors thrown from handler methods as well,
				// making them available for @ExceptionHandler methods and other scenarios.
				dispatchException = new NestedServletException("Handler dispatch failed", err);
			}
			processDispatchResult(processedRequest, response, mappedHandler, mv, dispatchException);

```

Finally, the process will call some exception resolvers and one of them is GrailsExceptionResolver.java that has this code to resolve the exception:
```
@Override
    public ModelAndView resolveException(HttpServletRequest request, HttpServletResponse response,
                                         Object handler, Exception ex) {
        // don't reuse cached controller attribute 
        request.removeAttribute(GrailsApplicationAttributes.GRAILS_CONTROLLER_CLASS_AVAILABLE);        
        
        ex = findWrappedException(ex);

        filterStackTrace(ex);

        ModelAndView mv = super.resolveException(request, response, handler, ex);

        setStatus(request, response, mv, ex);

        logStackTrace(ex, request);

        UrlMappingsHolder urlMappings = lookupUrlMappings();
        if (urlMappings != null) {
            mv = resolveViewOrForward(ex, urlMappings, request, response, mv);
        }

        return mv;
    }
```

So, when it tries to resolve view or forward, it is forwarding to the same url that makes the same request and causes the loop issue. You can check this code at GrailsExceptionResolver:223

```
protected String determineUri(HttpServletRequest request) {
        String uri = (String)request.getAttribute(WebUtils.FORWARD_REQUEST_URI_ATTRIBUTE);
        if (uri == null) {
            uri = request.getRequestURI();
        }
        return uri;
    }
```

And in the comments for WebUtils.FORWARD_REQUEST_URI_ATTRIBUTE we will see this description:

```
/**
	 * Standard Servlet 2.4+ spec request attribute for forward request URI.
	 * <p>If forwarded to via a RequestDispatcher, the current resource will see its
	 * own request URI. The originating request URI is exposed as a request attribute.
	 */
public static final String FORWARD_REQUEST_URI_ATTRIBUTE = "javax.servlet.forward.request_uri";
```

And, from what I understood, it is fowarded via RequestDispatcher (please, correct me if I´m wrong here) so it will get its own request URI

So, that was my conclusion and maybe I should improve the error message to be more descriptive. So, if you have any other idea just let me know and we can discuss or I can change this solution. IMO, that is a simple way to fix this issue that is not related to the root cause itself (db disconnect) but the way that this error is handled by the application.

mrdubr:
> So, when it tries to resolve view or forward, it is forwarding to the same url that makes the same request and causes the loop issue

Nice, now we are getting to the bottom of it. It appears that the problem is that we have UrlMappings (or interceptors) misconfigured which results in an infinite loop. We need to address that. I'll be happy to pair if you want to take a look at this together.

rundeckapp/grails-app/controllers/rundeck/interceptors/ProjectSelectInterceptor.groovy
```diff
@@ -70,11 +71,13 @@ class ProjectSelectInterceptor {
         if(controllerName=='menu' && ( actionName in ['home'] )){
             return true
         }
+        if(controllerName=='error' && ( actionName in ['fiveHundred'] )){
```
mrdubr, line 74: Instead of adding this hack, can we exclude the interceptor from processing this controller-action tuple? It seems to be logical, but I don't know about the implications.

carlosrfranco: We can add the controller+action to the excluded actions here: https://github.com/rundeck/rundeck/blob/fc2e59be849479b33864dd1b6b12d0c4f9d18e66/rundeckapp/grails-app/controllers/rundeck/interceptors/ProjectSelectInterceptor.groovy#L55

Honestly, we would be doing exactly the same thing with the same result.

carlosrfranco: I addressed the changes requested to exclude the interceptor if error request

gschueler (changes requested):
I tested this and it seems another interceptor also needs the `.excludes(controller: 'error', action: 'fiveHundred')` added: AuditInterceptor.  after disconnecting the database it seems this interceptor also repeatedly causes exceptions to be dumped in the log

gschueler (approved):
after the changes, the DB connection messages are all related to specific internal scheduled tasks. Only one message is logged when reloading a page that produces a 500 error.

PR #9385, opened 2024-10-07
RUN-2879: Simplifies node versioning

ChuckCrawford:
Removes a few places where we specify the Node version that are not actually used.

* Removes the version from `config.yaml` since the `circleci/node` orb will use `.nvmrc`. 
* Removes version from Gradle since it is configured to use the globally installed version.

**Additional context**


.nvmrc
```diff
@@ -1 +1 @@
-v18.20.4
+v18.20.3
```
ChuckCrawford, line 1: This is just for testing. Will revert before merging.

ChuckCrawford:
[This build](https://app.circleci.com/pipelines/github/rundeck/rundeck/16197/workflows/aa30a7dd-f3f8-481a-a105-cebd90a2684e) was run with `v18.20.3` and you can see it being used in the appropriate places.

ahormazabal (approved):

PR #9389, opened 2024-10-09
RUN-2929: Remove build dependency on local Groovy SDK

ahormazabal:
This PR provides some improvements on the build and pipeline:

* Removes dependency on a local groovy sdk, using gradle directly to run groovy scripts.
* Improve encapsulation of some dependency installation functions.
* Remove unnecessary node and awscli installation from many jobs.
* Explicitely set all circleci job images to latest ubuntu 22.

### DIFF
--- a/scripts/circleci/dependencies-functions.sh
+++ b/scripts/circleci/dependencies-functions.sh
@@ -29,21 +29,6 @@ dependencies_install_groovy() {
 
 }
 
-# Install nodejs
-dependencies_install_nodejs() {
-  node_version=$(cat .nvmrc)
-  curl -sSl -o /tmp/node.tar.xz "https://nodejs.org/dist/${node_version}/node-${node_version}-linux-x64.tar.xz" && \
-  echo '592eb35c352c7c0c8c4b2ecf9c19d615e78de68c20b660eb74bd85f8c8395063 /tmp/node.tar.xz' | sha256sum --check && \
-  sudo mkdir -p /usr/local/nodejs && \
-  sudo tar --strip 1 -xvf /tmp/node.tar.xz --directory /usr/local/nodejs && \
-  sudo ln --symbolic /usr/local/nodejs/bin/node /usr/local/bin/node && \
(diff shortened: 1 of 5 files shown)


scripts/circleci/dependencies-functions.sh
```diff
@@ -26,24 +27,35 @@ dependencies_build_setup() {
...
         sudo ln --symbolic "/usr/local/groovy/bin/groovyc" /usr/local/bin/groovyc
 
-    mkdir -p "$HOME/.gradle"
 }
 
+# Install nodejs
+dependencies_install_nodejs() {
+  node_version=$(cat .nvmrc)
+  curl -sSl -o /tmp/node.tar.xz "https://nodejs.org/dist/${node_version}/node-${node_version}-linux-x64.tar.xz" && \
+  echo '592eb35c352c7c0c8c4b2ecf9c19d615e78de68c20b660eb74bd85f8c8395063 /tmp/node.tar.xz' | sha256sum --check && \
```
ChuckCrawford, line 36: Won't this sha change per version? Kind of strange to read from `.nvmrc`  but then have this hard-coded?

ahormazabal: Yes i'm undecided on this, on one side i want to base on nvmrc to not having multiple node version placeholders, from other side i'd like to have the version sha hardcoded to prevent supply chain attacks.

Moved this change to another PR: https://github.com/rundeckpro/rundeckpro/pull/3981

scripts/circleci/dependencies-functions.sh
```diff
@@ -1,20 +1,21 @@
...
-    sudo apt-get update
-    sudo apt-get -y --no-install-recommends install zulu11-jdk-headless
+      sudo apt-get update
+      sudo apt-get -y --no-install-recommends install zulu11-jdk-headless
 
-    # Node and aws installed by orb.
+}
 
+# Install groovy SDK
+dependencies_install_groovy() {
```
gschueler, line 18: can we delete this since it is no longer used?

gschueler (approved):

PR #9395, opened 2024-10-11
RUN-2898: fix: flaky test, timing might cause failure

gschueler:
The test should should test for all executions to finish, otherwise it's possible it will proceed when the running execution has not yet completed, because it only verifies if *any* execution has finished.

this change aligns with the following test case "disable schedules at project level UI" which uses "verifyForAll"

Fix a flaky test

**Additional context**

### DIFF
--- a/functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/project/ExecutionModeSpec.groovy
+++ b/functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/project/ExecutionModeSpec.groovy
@@ -207,7 +207,7 @@ class ExecutionModeSpec extends SeleniumBase{
         waitFor(ExecutionUtils.Retrievers.executionsForProject(client, projectName),
                 { List<Execution> execs ->  execs.any(ExecutionUtils.Verifiers.executionRunning()) },
                 WaitingTime.EXCESSIVE)
-        // Waits for at least one execution to start running
+        // Waits for all executions to finish
         waitFor(ExecutionUtils.Retrievers.executionsForProject(client, projectName),
                 verifyForAll(ExecutionUtils.Verifiers.executionFinished()),
                 WaitingTime.EXCESSIVE)
--- a/functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/project/ExecutionModeSpec.groovy
+++ b/functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/project/ExecutionModeSpec.groovy
@@ -209,7 +209,7 @@ class ExecutionModeSpec extends SeleniumBase{
                 WaitingTime.EXCESSIVE)
         // Waits for at least one execution to start running
         waitFor(ExecutionUtils.Retrievers.executionsForProject(client, projectName),
-                { List<Execution> execs ->  execs.any(ExecutionUtils.Verifiers.executionFinished()) },
+                verifyForAll(ExecutionUtils.Verifiers.executionFinished()),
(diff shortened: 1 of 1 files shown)


functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/project/ExecutionModeSpec.groovy
```diff
@@ -209,7 +209,7 @@ class ExecutionModeSpec extends SeleniumBase{
                 WaitingTime.EXCESSIVE)
         // Waits for at least one execution to start running
```
ChuckCrawford, line 210:
```suggestion
        // Waits for all executions to finish
```

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/project/ExecutionModeSpec.groovy
```diff
@@ -207,9 +207,9 @@ class ExecutionModeSpec extends SeleniumBase{
         waitFor(ExecutionUtils.Retrievers.executionsForProject(client, projectName),
                 { List<Execution> execs ->  execs.any(ExecutionUtils.Verifiers.executionRunning()) },
                 WaitingTime.EXCESSIVE)
-        // Waits for at least one execution to start running
+        // Waits for all executions to finish
         waitFor(ExecutionUtils.Retrievers.executionsForProject(client, projectName),
-                { List<Execution> execs ->  execs.any(ExecutionUtils.Verifiers.executionFinished()) },
+                verifyForAll(ExecutionUtils.Verifiers.executionFinished()),
```
ChuckCrawford, line 212: I can see how this fix would make this test _less flaky_ but i'm still a little concerned that this job will be producing new executions the longer this portion of the test runs.

Ideally it seems like this test needs a way to do something closer to: "wait for at least one new execution to start and complete"?

cc: @ronaveva.

ChuckCrawford (approved):
Approving this since it is better than it was before. If we continue to see flakiness in this test we might want to come back and address https://github.com/rundeck/rundeck/pull/9395/files#r1801779431.

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/project/ExecutionModeSpec.groovy
gschueler: I think it's fine if it produces new executions too, I believe the flakiness issue was due to it passing this waitFor check *before* the later expected condition was true (number of completed executions is greater than it was before on line 216). As long as it waits for 1 execution to complete, that should satisfy the later check.

gschueler: if someone wants to refactor this test case (and the one after it) to be more exact, that's worth a pull request

ChuckCrawford: > I believe the flakiness issue was due to it passing this waitFor check before the later expected condition was true (number of completed executions is greater than it was before on line 216). 

Yes I see what you are saying and that makes sense.

> As long as it waits for 1 execution to complete, that should satisfy the later check.

Yes that is what I am trying to say in my comment above as well. My concern is that the job is starting a new execution every `*/2` seconds and this assertion code is waiting for _all_ executions to be finished. This feels like another potential race condition where we could tighten up the code.

Let's merge it and see how it goes though.

PR #9397, opened 2024-10-14
RPL-82: Plugin Blocklist Enhancement

ChuckCrawford:
This is an enhancement to [the plugin blocklist](https://docs.rundeck.com/docs/administration/security/blocklist.html) when blocking specific plugins by provider name. In the current state, when plugins for certain types of providers (e.g. `WorkflowNodeStep`) were blocked, they would still show up in various UIs and APIs. With this change, a blocked plugin will be more cleanly removed from these UIs and APIs.

**Context**
The `PluginBlocklist` / `RundeckPluginBlocklist` manages the list of blocked plugins and plugin containers (files). A few services are aware of this blocklist but the PluginRegistry is the primary one. Flows that obtain plugin information from the `PluginService` and / or the `PluginRegistry` properly filter out blocked plugins.

A number of other flows including the `/plugins/list` API and the "Installed Plugins" UI are based on the data provided by [PluginApiService.listPluginsDetailed](https://github.com/rundeck/rundeck/blob/c9436262299372134d160d7aedd1dcd9bde7bb06/rundeckapp/grails-app/services/rundeck/services/PluginApiService.groovy#L60). For some service providers, this function loads up plugins via the `PluginService` (which properly applies the blocklist). For others providers, the plugins are loaded using the `PluginManagerService` (which is not aware of the blocklist).

A few other flows [including the creating / editing a workflow] also load plugins by directly calling listDescriptions() on particular services. After loading the plugins, these flows make use of `PluginControlServiceImpl` to filter out disabled plugins. However, `PluginControlServiceImpl` is currently only aware of plugins that are disabled at a Project-level. It is not aware of globally disabled plugins.

This solution updates the `PluginManagerService` such that it removes any plugins present in the `PluginBlocklist` before providing results.

See: https://github.com/rundeck/rundeck/pull/9432.

**Additional context**
This PR also updates a few services that require the `PluginBlocklist` so that they depend on the interface rather than the concrete `RundeckPluginBlocklist` class.

### DIFF
--- a/rundeckapp/src/main/groovy/com/dtolabs/rundeck/server/plugins/RundeckEmbeddedPluginExtractor.groovy
+++ b/rundeckapp/src/main/groovy/com/dtolabs/rundeck/server/plugins/RundeckEmbeddedPluginExtractor.groovy
@@ -16,17 +16,14 @@
 
 package com.dtolabs.rundeck.server.plugins
 
+import com.dtolabs.rundeck.core.plugins.PluginBlocklist
 import com.dtolabs.rundeck.server.plugins.loader.ApplicationContextPluginFileSource
 import com.dtolabs.rundeck.server.plugins.loader.PluginFileManifest
 import com.dtolabs.rundeck.server.plugins.loader.PluginFileSource
 import com.dtolabs.utils.Streams
-import grails.spring.BeanBuilder
-import org.rundeck.security.RundeckPluginBlocklist
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.InitializingBean
-import org.springframework.beans.factory.NoSuchBeanDefinitionException
-import org.springframework.beans.factory.annotation.Autowire
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.context.ApplicationContext
 import org.springframework.context.ApplicationContextAware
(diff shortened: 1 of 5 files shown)


rundeckapp/src/main/groovy/com/dtolabs/rundeck/server/plugins/RundeckEmbeddedPluginExtractor.groovy
```diff
@@ -40,11 +37,11 @@ class RundeckEmbeddedPluginExtractor implements ApplicationContextAware, Initial
 
     ApplicationContext applicationContext
     File pluginTargetDir
-    RundeckPluginRegistry rundeckPluginRegistry
```
ChuckCrawford, line 43: This is not used and also created a circular dependency with `RundeckPluginRegistry`.

rundeckapp/src/main/groovy/com/dtolabs/rundeck/server/plugins/AppExecutionPluginLoader.groovy
```diff
@@ -51,7 +48,6 @@ class AppExecutionPluginLoader implements IExecutionProviders, ApplicationContex
     @Autowired PluginConfigureService pluginService
     @Autowired IFrameworkNodes rundeckNodeSupport
     @Autowired NodeProviderName nodeProviderName
-    @Autowired RundeckPluginBlocklist rundeckPluginBlocklist
```
ChuckCrawford, line 54: This also does not seem to be used but I would appreciate someone triple-checking me here...

alexander-variacode: In RundeckproRunnerGrailsPlugin class the property RundeckPluginBlocklist is not loaded in line 214 so probably it would always be null.
Another reference is in rundeckapp/grails-app/conf/spring/resources.groovy line 346 and it is not loaded either.

alexander-variacode (approved):
LGTM

alexander-variacode (approved):
LGTM

PR #9399, opened 2024-10-15
Functional Test improvement: validate job import response

gschueler:

Validate response to job import does not indicate any failure

**Additional context**

### DIFF
--- a/functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
+++ b/functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
@@ -82,7 +82,7 @@ class JobUtils {
             final String jobDefinition,
             RdClient client,
             String contentType = 'application/xml',
-            Consumer<CreateJobResponse> failedJobsHandler = { List<Object> failedJobs -> throw new Exception("Some jobs failed on import: " + failedJobs) } ) {
+            Consumer<CreateJobResponse> failedJobsHandler = { List<Object> failedJobs -> throw new IllegalArgumentException("Some jobs failed on import: " + failedJobs) } ) {
         final String CREATE_JOB_ENDPOINT = "/project/${project}/jobs/import"
         Response responseImport = client.doPostWithRawText(CREATE_JOB_ENDPOINT, contentType, jobDefinition)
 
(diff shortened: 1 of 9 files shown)


functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
```diff
@@ -358,59 +377,60 @@ class JobUtils {
...
-        String projectName,
-        String xmlResourcePath,
-        RdClient client,
-        String dupeOption = DUPE_OPTION_DEFAULT,
-        String contentType = CONTENT_TYPE_DEFAULT
-    ) {
-        URL resourceUrl = getClass().getResource(xmlResourcePath)
-        def pathXmlFile = resourceUrl != null ? resourceUrl.getPath() : xmlResourcePath
-        return jobImportFile(projectName, new File(pathXmlFile), client, dupeOption, contentType)
+    static def jobImportFile = (String projectName, String resourcePath, RdClient client, String dupeOption = DUPE_OPTION_DEFAULT, String contentType = CONTENT_TYPE_DEFAULT) -> {
```
gschueler, line 405: I had reverted the closure/lambda into a basic method definition, which is more idiomatic.  Is there a reason to use a lambda?

functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
```diff
@@ -358,59 +377,60 @@ class JobUtils {
...
+
+    /**
+     * Executes the handler if at least one job import fails
+     * @param data
+     * @param failedJobsHandler
+     */
+    private static void validateJobsImportAllSuccess(data,
+                                                     Consumer<CreateJobResponse> failedJobsHandler = { List<Object> failedJobs -> throw new Exception("Some jobs failed on import: " + failedJobs) } ) {
+        if(data?.failed?.size()>0) {
+            failedJobsHandler.accept(data?.failed)
```
gschueler, line 389: maybe a cleaner way is to use an override with one parameter, and pass the default handler to 2nd method definition

functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
mrdubr: Yes, the behaviour of `getClass().getResource(...)` differs when the code executes in a named module (as per your refactoring) and a Closure (original version). Multiple tests started to fail due to the inability to locate files on a file system after the change to a simple function.

Per the docs:  
> If this class is in a named Module then this method will attempt to find the resource in the module. This is done by delegating to the module's class loader findResource(String,String) method, invoking it with the module name and the absolute name of the resource. Resources in named modules are subject to the rules for encapsulation specified in the Module getResourceAsStream method and so this method returns null when the resource is a non-".class" resource in a package that is not open to the caller's module.

> Otherwise, if this class is not in a named module then the rules for searching resources associated with a given class are implemented by the defining class loader of the class. This method delegates to this object's class loader. If this object was loaded by the bootstrap class loader, the method delegates to ClassLoader. getSystemResource.
Before delegation, an absolute resource name is constructed from the given resource name using this algorithm:

mrdubr (approved):
LGTM

PR #9408, opened 2024-10-24
RUN-2853: Fix race condition between the quartz scheduler and the web server thread that creates ScheduledExecution

mrdubr:
This is a bugfix (mitigation) for the race condition between the quartz scheduler thread that gets ScheduledExecution ID from the memory store and the web server thread that puts the ID into the memory store AND creates the ScheduledExecution object in the DB. The issue occurs when the quartz thread processes the associated job and attempts to fetch the newly (almost) created ScheduledExecution object from the DB **BEFORE** the DB transaction is committed.

The problem is mitigated by retrying the DB fetch operation multiple times (with a timeout) giving the web server thread additional time to commit.

The proper alternative would be to wait for the txn to commit prior to scheduling the corresponding job in Quartz.
However, the code to CRUD ScheduledExecution is very intermingled with job scheduling and refactoring it would require significant effort.

**Additional context**
Fixes the "flaky" ScheduledJobSpec` test

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/utils/WaitUtils.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/utils/WaitUtils.java
@@ -1,0 +1,38 @@
+package com.dtolabs.rundeck.core.utils;
+
+import java.time.Duration;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+public class WaitUtils {
+
+    /**
+     * Waits for the resource to be in the expected state by retrieving it and evaluating its acceptance as many times as needed.
+     *
+     * @param retryableResourceRetriever executed multiple times to retrieve the current state of the R resource.
+     * @param resourceAcceptanceEvaluator evaluates the state of the resource and returns true if the resource is accepted.
+     * @param timeout max duration to wait for the resource to reach the expected state.
+     * @param checkPeriod time to wait between each check.
+     * @return the resource that met the acceptance criteria.
+     * @throws ResourceAcceptanceTimeoutException if the timeout is reached waiting for the resource to reach the expected state.
+     * @throws InterruptedException if the thread sleep was interrupted.
+     */
+    public static <R> R waitFor(
+            Supplier<R> retryableResourceRetriever,
+            Function<R, Boolean> resourceAcceptanceEvaluator,
+            Duration timeout,
+            Duration checkPeriod) throws InterruptedException {
+        R r = retryableResourceRetriever.get();
+        Boolean acceptanceResult = resourceAcceptanceEvaluator.apply(r);
+        long initTime = System.currentTimeMillis();
+        while (!acceptanceResult) {
+            Thread.sleep(Math.min(checkPeriod.toMillis(), timeout.toMillis()));
+            if ((System.currentTimeMillis() - initTime) >= timeout.toMillis()) {
+                throw new ResourceAcceptanceTimeoutException("Timeout reached (" + timeout.toMillis() + "ms) waiting for value: " + r + " to reach the desired state");
+            }
+            r = retryableResourceRetriever.get();
+            acceptanceResult = resourceAcceptanceEvaluator.apply(r);
+        }
+        return r;
+    }
+}
--- a/core/src/main/java/com/dtolabs/rundeck/core/utils/WaitUtils.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/utils/WaitUtils.java
@@ -26,6 +26,11 @@ public class WaitUtils {
         Boolean acceptanceResult = resourceAcceptanceEvaluator.apply(r);
         long initTime = System.currentTimeMillis();
         while (!acceptanceResult) {
+            // Check timeout prior to sleep to avoid sleep that is not necessary
+            if ((System.currentTimeMillis() - initTime) >= timeout.toMillis()) {
+                throw new ResourceAcceptanceTimeoutException("Timeout reached (" + timeout.toMillis() + "ms) waiting for value: " + r + " to reach the desired state");
+            }
+            // Check timeout right after sleep to short circuit execution on timeout expiration
             Thread.sleep(Math.min(checkPeriod.toMillis(), timeout.toMillis()));
             if ((System.currentTimeMillis() - initTime) >= timeout.toMillis()) {
                 throw new ResourceAcceptanceTimeoutException("Timeout reached (" + timeout.toMillis() + "ms) waiting for value: " + r + " to reach the desired state");
--- a/functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/ScheduledJobSpec.groovy
+++ b/functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/ScheduledJobSpec.groovy
@@ -29,22 +29,27 @@ class ScheduledJobSpec extends SeleniumBase{
         setupProject(projectName)
         ActivityPage activityPage = page ActivityPage
         JobListPage jobListPage = page(JobListPage)
-        jobListPage.loadJobListForProject(projectName)
-        def jobXml = JobUtils.generateScheduledJobsXml("scheduledJob", "<time hour='*' seconds='*/5' minute='*' />")
+        def jobXml = JobUtils.generateScheduledJobsXml("scheduledJob", "<time hour='*' seconds='*/2' minute='*' />")
         JobUtils.createJob(projectName, jobXml, client)
+
         when:
+        jobListPage.loadJobListForProject(projectName)
         jobListPage.go()
         jobListPage.validatePage()
-        waitFor(ExecutionUtils.Retrievers.executionsForProject(client, projectName), {it.size() >= 1}, WaitingTime.EXCESSIVE )
+
         activityPage.loadActivityPageForProject(projectName)
         activityPage.go()
 
+        def refreshActivityPageAndGetExecutionCount = {
+            activityPage.refresh()
+            Integer.parseInt(activityPage.executionCount.text)
+        }
+
         then:
-        Integer.parseInt(activityPage.executionCount.text) > 0
+        assert waitFor(refreshActivityPageAndGetExecutionCount, { Integer executionCount -> executionCount  > 0 }, WaitingTime.EXCESSIVE) > 0
 
         cleanup:
         deleteProject(projectName)
-
     }
 
     /**
@@ -76,7 +81,6 @@ class ScheduledJobSpec extends SeleniumBase{
         then:
         jobShowPage.els(jobShowPage.scheduleTimeBy).size() == 1
         jobShowPage.el(jobShowPage.scheduleTimeBy).getText().contains("in")
-        jobShowPage.el(jobShowPage.scheduleTimeBy).getText().contains("on")
 
         cleanup:
         deleteProject(projectName)
--- a/rundeckapp/grails-app/jobs/rundeck/quartzjobs/ExecutionJob.groovy
+++ b/rundeckapp/grails-app/jobs/rundeck/quartzjobs/ExecutionJob.groovy
@@ -27,6 +27,8 @@ import com.dtolabs.rundeck.core.execution.WorkflowExecutionServiceThread
 import com.dtolabs.rundeck.core.execution.workflow.WorkflowExecutionResult
 import com.dtolabs.rundeck.core.execution.workflow.steps.node.NodeStepResult
 import com.dtolabs.rundeck.core.schedule.JobScheduleManager
+import com.dtolabs.rundeck.core.utils.ResourceAcceptanceTimeoutException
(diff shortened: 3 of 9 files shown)


functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/ScheduledJobSpec.groovy
```diff
@@ -76,7 +81,6 @@ class ScheduledJobSpec extends SeleniumBase{
         then:
         jobShowPage.els(jobShowPage.scheduleTimeBy).size() == 1
         jobShowPage.el(jobShowPage.scheduleTimeBy).getText().contains("in")
-        jobShowPage.el(jobShowPage.scheduleTimeBy).getText().contains("on")
```
mrdubr, line 79: The "on" condition does not occur when executed locally

rundeckapp/grails-app/jobs/rundeck/quartzjobs/ExecutionJob.groovy
```diff
@@ -126,12 +129,10 @@ class ExecutionJob implements InterruptableJob {
                 log.error("Unable to start Job execution: ${es.message ?: 'no message'}")
                 return
             } else {
-                log.error("Unable to start Job execution: ${es.message ?: 'no message'}", es)
-                throw es
+                throw new ExecutionServiceException("Unable to start Job execution", es)
```
mrdubr, line 132: This exception will get logged 3 times by the quartz code. No need for the 4th.

rundeckapp/grails-app/jobs/rundeck/quartzjobs/ExecutionJob.groovy
```diff
@@ -126,12 +129,10 @@ class ExecutionJob implements InterruptableJob {
...
             }
-        }catch(ScheduledExecutionDeletedException sede){
-            log.error("ScheduledExecution not found on DB: ${sede.message?sede.message:'no message'}",sede)
-            throw sede
+        }catch(MissingScheduledExecutionException e){
+            throw e
```
mrdubr, line 135: Logging and rethrowing is an anti-pattern.

rundeckapp/grails-app/jobs/rundeck/quartzjobs/ExecutionJob.groovy
```diff
@@ -658,24 +659,37 @@ class ExecutionJob implements InterruptableJob {
...
-            context.getScheduler().deleteJob(context.jobDetail.key)
-            throw new RuntimeException("ScheduledExecution found but it does not match the original project name to schedule. Project name was : ${projectName} , Project name now : ${se.project}")
-        }
+        // There is a possibility of race condition between this quartz thread that gets ScheduledExecution ID from memory and
+        // the web server thread that creates the ScheduledExecution object in the DB. This issue occurs when this thread
+        // attempts to fetch the newly created ScheduledExecution object from the DB before it's commited.
+        // The fetch retry logic is introduced to mitigate this problem.
+        try {
+            ScheduledExecution se = WaitUtils.<ScheduledExecution>waitFor(
+                    {ScheduledExecution.findByUUID(seid).find()},
```
mrdubr, line 669: It's worth noting that the `findByUUID` is configured to bypass cache

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/ScheduledJobSpec.groovy
```diff
@@ -29,22 +29,27 @@ class ScheduledJobSpec extends SeleniumBase{
...
         when:
+        jobListPage.loadJobListForProject(projectName)
         jobListPage.go()
         jobListPage.validatePage()
-        waitFor(ExecutionUtils.Retrievers.executionsForProject(client, projectName), {it.size() >= 1}, WaitingTime.EXCESSIVE )
+
         activityPage.loadActivityPageForProject(projectName)
         activityPage.go()
 
+        def refreshActivityPageAndGetExecutionCount = {
```
ChuckCrawford, line 43: This may be your intention but it looks like this is also changing what this test is testing?

The `ActivityPage` has that auto-refresh functionality built into it where it is loading new executions in the background. With this change the test itself will be performing the refresh rather than letting the application do so...

mrdubr: That's a good point. It helps to understand what exactly are we testing here.
Per the test description:
> Schedules a job and checks if with time passing, executions reflects in activity section.

I've never treated this as a test of the auto-refresh functionality. 
Only as a test of "executions will show up in UI eventually".
IMO, this alligns with the pre-refactoring behaviour -- waiting for the backend to have executions (line 38) prior to loading the UI. I don't think the refresh was tested prior to the refactoring.

core/src/main/java/com/dtolabs/rundeck/core/utils/WaitUtils.java
```diff
@@ -0,0 +1,38 @@
...
+            Duration timeout,
+            Duration checkPeriod) throws InterruptedException {
+        R r = retryableResourceRetriever.get();
+        Boolean acceptanceResult = resourceAcceptanceEvaluator.apply(r);
+        long initTime = System.currentTimeMillis();
+        while (!acceptanceResult) {
+            Thread.sleep(Math.min(checkPeriod.toMillis(), timeout.toMillis()));
+            if ((System.currentTimeMillis() - initTime) >= timeout.toMillis()) {
+                throw new ResourceAcceptanceTimeoutException("Timeout reached (" + timeout.toMillis() + "ms) waiting for value: " + r + " to reach the desired state");
+            }
```
gschueler, line 32: it seems the logic here will result in: resource acceptance check->false, wait, throw.  So there seems to be an unnecessary wait before the throw.  i.e. it would be better to have a loop like:
- initial check
- wait
    - check again,
    - if true, pass
    - if false,  
        - check timeout expired, if true throw
        - otherwise loop again

mrdubr: Indeed, this implementation does not guarantee to timeout at `timeout`. In fact, it cannot do it even in theory if the resource retrieval takes over `timeout`, but let's ignore that possibility for now.

The current implementation attempts to timeout between the `timeout` and `timeout + checkPeriod`.

I'm not sure I fully understand your proposed algo, but I think it still suffers from the possibility of running over the `timeout`. Also, I think it would be logically equivalent to adding another timeout check before the sleep?:
```
            if ((System.currentTimeMillis() - initTime) >= timeout.toMillis()) {
                throw new ResourceAcceptanceTimeoutException("Timeout reached (" + timeout.toMillis() + "ms) waiting for value: " + r + " to reach the desired state");
            }
            Thread.sleep(Math.min(checkPeriod.toMillis(), timeout.toMillis()));
            if ((System.currentTimeMillis() - initTime) >= timeout.toMillis()) {
                throw new ResourceAcceptanceTimeoutException("Timeout reached (" + timeout.toMillis() + "ms) waiting for value: " + r + " to reach the desired state");
            }
```

Does it make sense?

core/src/main/java/com/dtolabs/rundeck/core/utils/WaitUtils.java
```diff
@@ -0,0 +1,38 @@
...
+     *
+     * @param retryableResourceRetriever executed multiple times to retrieve the current state of the R resource.
+     * @param resourceAcceptanceEvaluator evaluates the state of the resource and returns true if the resource is accepted.
+     * @param timeout max duration to wait for the resource to reach the expected state.
+     * @param checkPeriod time to wait between each check.
+     * @return the resource that met the acceptance criteria.
+     * @throws ResourceAcceptanceTimeoutException if the timeout is reached waiting for the resource to reach the expected state.
+     * @throws InterruptedException if the thread sleep was interrupted.
+     */
+    public static <R> R waitFor(
```
ChuckCrawford, line 20: This does seem like a useful helper to have. There seem to be a number of places in the code-base that individually implement a retry with backoff.

We definitely don't have to do it in this implementation but a future improvement might be to add a growing backoff and a jitter around the `checkPeriod` duration.

core/src/main/java/com/dtolabs/rundeck/core/utils/WaitUtils.java
ChuckCrawford: Would a "compromise improvement" be to just move the timeout check before the sleep so that it looks something like:

```
        while (!acceptanceResult) {
            if ((System.currentTimeMillis() - initTime) >= timeout.toMillis()) {
                throw new ResourceAcceptanceTimeoutException("Timeout reached (" + timeout.toMillis() + "ms) waiting for value: " + r + " to reach the desired state");
            }
            
            Thread.sleep(Math.min(checkPeriod.toMillis(), timeout.toMillis()));

            r = retryableResourceRetriever.get();
            acceptanceResult = resourceAcceptanceEvaluator.apply(r);
        }
```

core/src/main/java/com/dtolabs/rundeck/core/utils/WaitUtils.java
mrdubr: IMO, we should be using a proper retry library with all the bells and whistles instead of trying to implement our own. However, this is a starting point with minimal effort.

core/src/main/java/com/dtolabs/rundeck/core/utils/WaitUtils.java
mrdubr: This can work as well, but there is an increased probability of doing a retreival after the timeout.
Surrounding the `sleep` with checks appears to be a compromise.

rundeckapp/grails-app/jobs/rundeck/quartzjobs/ExecutionJob.groovy
```diff
@@ -658,24 +659,40 @@ class ExecutionJob implements InterruptableJob {
...
-            throw new RuntimeException("JobDataMap contained invalid ScheduledExecution type: " + se.getClass().getName())
+            if(se instanceof ScheduledExecution){
+                se.refreshOptions() //force fetch options and option values before return object
+            } else {
+                // Legacy logic. Is this even possible?
+                throw new RuntimeException("JobDataMap contained invalid ScheduledExecution type: " + se.getClass().getName())
+            }
+
+            if(!projectName.equals(se.project)){
+                context.getScheduler().deleteJob(context.jobDetail.key)
```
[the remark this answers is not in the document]
mrdubr: I'll add it to the description of the exception thrown on the next line...

ChuckCrawford (approved):
I left a few small comments but I think this is a good first step towards fixing a nasty bug. 

I was able to confirm the bug you are referring to between `rundeck.services.ScheduledExecutionService#_dosave` and the quartz thread `rundeck.quartzjobs.ExecutionJob#execute`. 

This fix does a good job of reducing the likelihood of that race condition without totally refactoring this core code. It also introduces a nice `waitFor` library function that looks to be useful in other places as well.

core/src/main/java/com/dtolabs/rundeck/core/utils/WaitUtils.java
gschueler: The code you suggest @ChuckCrawford  is what I was attempting to describe. I was just trying to avoid the unnecessary wait. Yes it could check before and after the wait if you want to be tidy, but really it's not very important for the solution if we overrun the timeout. Either of those options would avoid having an unneeded wait 👍

core/src/main/java/com/dtolabs/rundeck/core/utils/WaitUtils.java
gschueler: I'm sure there's half a dozen other hand-rolled versions elsewhere in the code 😄

rundeckapp/grails-app/jobs/rundeck/quartzjobs/ExecutionJob.groovy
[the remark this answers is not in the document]
ChuckCrawford: But does that exception still get logged somewhere? I see [you removed that in some places](https://github.com/rundeck/rundeck/pull/9408/files#r1815719553)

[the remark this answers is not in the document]
mrdubr: Yes, the quartz itself logs it a few times. Verified.
However, IMO, in the codebase we should assume that someone will log it downstack -- controller error handler, thread pool error handler, etc...
It's simply not sustainable to log it "just in case".

mrdubr:
> My only other suggestion might be to reduce the `timeout` of the wait to 5s and the `checkPeriod` to 500ms as a first step.

In my unscientific load test experiment, the 10s timeout resulted in 100% pass rate (20 executions in total). 5 sec timeout had one failure.  I would keep it at 10 for now.

PR #9411, opened 2024-10-28
RUN-2634: add functional test for sshj with pty enabled/disabled

ronaveva:
This PR is intended to add functional tests for sshj when having pty enabled/disabled from the PR listed bellow. These tests will fail until having this changes in a new release and set the new sshj plugin version

https://github.com/rundeck-plugins/sshj-plugin/pull/66

EDIT: This also updates the sshj version to 0.1.16

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -7,4 +7,4 @@ rundeck:
   - "com.github.rundeck-plugins:openssh-node-execution:2.0.3@zip"
   - "com.github.rundeck-plugins:multiline-regex-datacapture-filter:1.1.0"
   - "com.github.rundeck-plugins:attribute-match-node-enhancer:v0.1.5"
-  - "com.github.rundeck-plugins:sshj-plugin:v0.1.15"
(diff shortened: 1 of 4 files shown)


Jesus-Osuna-M (commented):
I believe it would make more sense to include this test within SshjNodeExecutorSpec.groovy

https://github.com/rundeck/rundeck/blob/8a4c31bc72f46cbaa59165672475d7b3d95ece23/functional-test/src/test/groovy/org/rundeck/tests/functional/api/executor/SshjNodeExecutorSpec.groovy#L11

ronaveva:
> I believe it would make more sense to include this test within SshjNodeExecutorSpec.groovy
> 
> https://github.com/rundeck/rundeck/blob/8a4c31bc72f46cbaa59165672475d7b3d95ece23/functional-test/src/test/groovy/org/rundeck/tests/functional/api/executor/SshjNodeExecutorSpec.groovy#L11

this can be done in another pr

PR #9420, opened 2024-10-29
RUN-2932: Introduce wait and retry to the flaky functional CommandOnMultipleNodesSpec tests

mrdubr:
Flaky test fix

There is a race condition between the output log lines produced by the system and the tests verifying the log line count. The wait and retry logic introduced gives the system more time to produce the expected number of log lines.


ahormazabal (approved):
LGTM

PR #9426, opened 2024-10-31
Test out a refactor.

ChuckCrawford:

**Additional context**

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/common/PluginControlServiceImpl.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/common/PluginControlServiceImpl.java
@@ -1,5 +1,6 @@
 package com.dtolabs.rundeck.core.common;
 
+import com.dtolabs.rundeck.core.plugins.PluginRegistry;
 import com.dtolabs.rundeck.core.plugins.configuration.Description;
 
 import java.util.*;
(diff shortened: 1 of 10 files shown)


core/src/main/java/com/dtolabs/rundeck/core/common/PluginControlServiceImpl.java
```diff
@@ -15,16 +16,20 @@ public class PluginControlServiceImpl implements PluginControlService {
     private HashSet<String> disabledPlugins;
     private final IFramework framework;
     private final String project;
+    private final Optional<PluginRegistry> pluginRegistry;
```
ChuckCrawford, line 19: This is temporary while I incrementally add-in the registry.

core/src/main/java/com/dtolabs/rundeck/core/common/PluginControlServiceImpl.java
```diff
@@ -15,16 +16,21 @@ public class PluginControlServiceImpl implements PluginControlService {
     private HashSet<String> disabledPlugins;
     private final IFramework framework;
     private final String project;
+    private final PluginRegistry pluginRegistry;
```
ChuckCrawford, line 19: TODO: If we actually take this approach, we would need to go through the rest of the methods and apply registry-level blocking.

PR #9428, opened 2024-11-01
RUN-2949: UI: Project Dashboard: Fix: activity summary should not be shown for 0 counts

gschueler:

- Fix: if userCount or failedCount was 0 it should not show the related text. The Vue "v-if" was used incorrectly.
- Fix i18n
- Add unit tests for activitySummary.vue

**Additional context**


smartinellibenedetti (approved):
I almost had a heart attack at first thinking "don't tell me I was the one who did that v-if wrong" but git blame said I'm clear 🤣

smartinellibenedetti:
The only thing is that the tests counting on the localization are failing - we've been just checking that the right key was used for the translation to avoid redoing the whole translation functionality

gschueler:
Fixed tests.

PR #9429, opened 2024-11-02
RUN-2947: Creates baseStore 

smartinellibenedetti:
As per https://docs.google.com/document/d/1s2k4BaLpBkxaFEqs-53nkvEgACugR_175ealC-j2HpU/edit?tab=t.xa3sd6k9kpe0#heading=h.7ieekuzay2dl

The goal of this PR is to create a boilerplate that can be used for creating stores and performing state management in the front-end.  The link to the doc that explains things in more detail is here: https://docs.google.com/document/d/1qNcdQ8hmDH-nPsypLAS0AHdHBK96T37HSdvCGZXnh0o/edit?tab=t.0, 
but I also added an initial readme that can be considered a TLDR; of sorts.

Create a baseStore based on the flux pattern, that can cache and save data in the localStorage.

Considered to also add commits and dispatchers to the store so that people wouldn't trigger mutations (setters) or actions directly, but thought that it would be overengineering at this point.

**Additional context**

Pagination and search should be optimized but will do so as soon as will implement the first store.


smartinellibenedetti:
Closing in favor of https://github.com/rundeck/rundeck/pull/9494

PR #9432, opened 2024-11-07
RPL-82: Alternate Plugin Blocklist Enhancement

ChuckCrawford:
This is an enhancement to [the plugin blocklist](https://docs.rundeck.com/docs/administration/security/blocklist.html) when blocking specific plugins by provider name. In the current state, when plugins for certain types of providers (e.g. WorkflowNodeStep) were blocked, they would still show up in various UIs and APIs. With this change, a blocked plugin will be more cleanly removed from these UIs and APIs.

**Context**
The `PluginBlocklist` / `RundeckPluginBlocklist` manages the list of blocked plugins and plugin containers (files). A few services are aware of this blocklist but the PluginRegistry is the primary one. Flows that obtain plugin information from the `PluginService` and / or the `PluginRegistry` properly filter out blocked plugins.

A number of other flows including the `/plugins/list` API and the "Installed Plugins" UI are based on the data provided by [PluginApiService.listPluginsDetailed](https://github.com/rundeck/rundeck/blob/c9436262299372134d160d7aedd1dcd9bde7bb06/rundeckapp/grails-app/services/rundeck/services/PluginApiService.groovy#L60). For some service providers, this function loads up plugins via the `PluginService` (which properly applies the blocklist). For others providers, the plugins are loaded using the `PluginManagerService` (which is not aware of the blocklist).

A few other flows [including the creating / editing a workflow] also load plugins by directly calling listDescriptions() on particular services. After loading the plugins, these flows make use of `PluginControlServiceImpl` to filter out disabled plugins. However, `PluginControlServiceImpl` is currently only aware of plugins that are disabled at a Project-level. It is not aware of globally disabled plugins.

This solution more clearly filters out blocked plugins by making changes in two areas:
* Updating the [PluginApiService.listPluginsDetailed()](https://github.com/rundeck/rundeck/blob/6f8d475276bb76d49f27270ee92f542d42eb26e1/rundeckapp/grails-app/services/rundeck/services/PluginApiService.groovy#L66-L77) method to load more plugin types via `PluginService.listPlugins` instead of via the `DescribableService` itself.
* Updating the [PluginControlServiceImpl](https://github.com/rundeck/rundeck/blob/6f8d475276bb76d49f27270ee92f542d42eb26e1/core/src/main/java/com/dtolabs/rundeck/core/common/PluginControlServiceImpl.java#L10) such that the service is aware of plugins blocked globally and plugins blocked at the project-level.

This solution is cleaner than the one that was introduced in https://github.com/rundeck/rundeck/pull/9397 because it makes use of or enhances plugin-related services that were already [at least partially] aware of the blocklist: `PluginControlServiceImpl`, `PluginService`, `RundeckPluginRegistry`, etc. while leaving the `PluginManagerService` as a lower-level service responsible for actually loading and providing plugins.

See: https://github.com/rundeck/rundeck/pull/9397

**Additional context**
This change stops short of updating all flows that use `PluginControlServiceImpl` due to the complexity involved with getting the `PluginRegistry` into `ExecutionContextImpl`.

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/common/PluginControlServiceImpl.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/common/PluginControlServiceImpl.java
@@ -1,9 +1,12 @@
 package com.dtolabs.rundeck.core.common;
 
+import com.dtolabs.rundeck.core.plugins.PluginRegistry;
+
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 import java.util.function.Predicate;
 
@@ -12,16 +15,30 @@ public class PluginControlServiceImpl implements PluginControlService {
     public static final String DISABLED_PLUGINS = "disabled.plugins";
     private HashSet<String> disabledPlugins;
     private final IFramework framework;
+    private final PluginRegistry pluginRegistry;
     private final String project;
 
-    private PluginControlServiceImpl(IFramework framework, String project) {
+    private PluginControlServiceImpl(IFramework framework, PluginRegistry pluginRegistry, String project) {
         this.framework = framework;
+        this.pluginRegistry = pluginRegistry;
         this.project = project;
     }
 
 
+    /**
+     * Creates a PluginControlService that only checks for disabled plugins at the project level. This version of the
+     * service is deprecated and the PluginRegistry aware version of the service should be preferred.
+     */
+    @Deprecated
     public static PluginControlService forProject(IFramework framework, String project) {
-        return new PluginControlServiceImpl(framework, project);
+        return new PluginControlServiceImpl(framework, null, project);
+    }
+
+    /**
+     * Creates a PluginControlService that checks for disabled plugins at both the global and project levels.
+     */
+    public static PluginControlService forProject(IFramework framework, PluginRegistry pluginRegistry, String project) {
+        return new PluginControlServiceImpl(framework, pluginRegistry, project);
     }
 
     private Set<String> getDisabledPlugins() {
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/dtolabs/rundeck/core/common/PluginControlServiceImpl.java
```diff
@@ -1,27 +1,31 @@
...
 
-
+    @Deprecated
     public static PluginControlService forProject(IFramework framework, String project) {
-        return new PluginControlServiceImpl(framework, project);
+        return new PluginControlServiceImpl(framework, null, project);
+    }
+
+    public static PluginControlService forProject(IFramework framework, PluginRegistry pluginRegistry, String project) {
+        return new PluginControlServiceImpl(framework, pluginRegistry, project);
```
ChuckCrawford, line 28: If we decide we like this approach, there is still a small TODO to update the `public Predicate<String> enabledPredicateForService(final String serviceName)` method so that it accounts for the `pluginRegistry` when available.

rundeckapp/grails-app/services/rundeck/services/PluginApiService.groovy
```diff
@@ -62,24 +71,34 @@ class PluginApiService {
...
-                ResourceModelSourceFactory,
-                framework.getResourceModelSourceService()
-        ).findAll { it.value.description }.collect {
-            it.value.description
-        }.sort { a, b -> a.name <=> b.name }
+        Map<String, List<Description>> pluginDescs = [:]
+
+        StepExecutionService ses = framework.getStepExecutionService()
+        pluginDescs[ses.name] = pluginService.listPlugins(StepExecutor, ses)
+                .findAll { it.value.description != null }
```
ChuckCrawford, line 78: Note the are still a few plugin services that contain plugins descriptors (`DescribedPlugin`) without a description property. The previous `service.listDescriptions()` approach filtered these out. The  `.findAll` is present here to compensate.

alexander-variacode (approved):
LGTM

PR #9439, opened 2024-11-11
Hw2024 step disabler

Jesus-Osuna-M:

**Additional context**

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,40 @@
+Rundeck Step Disabler - PD Hackweek '23 Project
+========
+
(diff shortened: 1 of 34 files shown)


smartinellibenedetti (commented):
Looking good for a hackathon project, I'd suggest to add extra logic to avoid loading the new UI when it's a new job and improving a bit the error handling.

The UI logic could be simplified by either having the config flag available in the original model or fetching it once the page initializes, but don't see it as a major blocker.

PS: it would also be fantastic to move the CSS outside of style tags.

PR #9446, opened 2024-11-15
RUN-2973: Cleanup: remove unused "ajax nowrunning" endpoint

gschueler:

Remove an unused endpoint previously used by the GUI, and  update activityList.vue to use a service function for API requests

**Additional context**


smartinellibenedetti (approved):
LGTM - the only thing that this change breaks the legacy forecast page. Should we pull the trigger and remove that one as well? (Forecast method in project-dashboard.js and _LegacyForecast.gsp)

PR #9448, opened 2024-11-15
[Hackweek] Vite build alternative

JoseOrtiz:

Enhacement to allow a user to freely choose between webpack or vite to build the frontend

A taglib was created to inject vite entrypoints alongside webpack generated files for the frontend, plus the addition of typescript to its build

**Additional context**

To enable vite: `RUNDECK_SPA_VITE_ENABLED` env variable should be set to true


JoseOrtiz:
> Is there anything that I can help with this @JoseOrtiz ? Upgrading to webpack 5 was meant to be a temporary step, and do think we'd benefit by moving to vite

First we need find a way to run the selenium test with vite (right now is using webpack) and then to find a way to use with rundeckpro. I think that they could work in a mixed way, but we need to upgrade i18n version to 10. I didn't have time to test rundeckpro

smartinellibenedetti:
> > Is there anything that I can help with this @JoseOrtiz ? Upgrading to webpack 5 was meant to be a temporary step, and do think we'd benefit by moving to vite
> 
> First we need find a way to run the selenium test with vite (right now is using webpack) and then to find a way to use with rundeckpro. I think that they could work in a mixed way, but we need to upgrade i18n version to 10. I didn't have time to test rundeckpro

Do you have bandwidth/plans to keep working on it? if you don't mind, I can do the i18n update as we'll possibly need to do it anyway, and then check in with you for the selenium changes.

JoseOrtiz:
> > > Is there anything that I can help with this @JoseOrtiz ? Upgrading to webpack 5 was meant to be a temporary step, and do think we'd benefit by moving to vite
> > 
> > 
> > First we need find a way to run the selenium test with vite (right now is using webpack) and then to find a way to use with rundeckpro. I think that they could work in a mixed way, but we need to upgrade i18n version to 10. I didn't have time to test rundeckpro
> 
> Do you have bandwidth/plans to keep working on it? if you don't mind, I can do the i18n update as we'll possibly need to do it anyway, and then check in with you for the selenium changes.

To be honest, I don't have too much bandwith but I could help if we prioritize it

smartinellibenedetti:
> > > > Is there anything that I can help with this @JoseOrtiz ? Upgrading to webpack 5 was meant to be a temporary step, and do think we'd benefit by moving to vite
> > > 
> > > 
> > > First we need find a way to run the selenium test with vite (right now is using webpack) and then to find a way to use with rundeckpro. I think that they could work in a mixed way, but we need to upgrade i18n version to 10. I didn't have time to test rundeckpro
> > 
> > 
> > Do you have bandwidth/plans to keep working on it? if you don't mind, I can do the i18n update as we'll possibly need to do it anyway, and then check in with you for the selenium changes.
> 
> To be honest, I don't have too much bandwith but I could help if we prioritize it

Understood, definitely it's not a priority/urgent so don't want to take your time, I'll start to work on it and see how far I'll manage to go. I'll keep you posted

PR #9455, opened 2024-11-18
RUN-2440 multiselect with all values selected should use values if job referenced or via api

carlosrfranco:
When a job has options that uses multivalues and all values is selected by default, the option value is empty when called by parent job or via api

It add default option values if it is multivalued type and all values is selected by default

**Additional context**


smartinellibenedetti (approved):
Confirmed that the fix is working both for running the job via API and via UI

PR #9457, opened 2024-11-20
RUN-2722: wrap up ui actions

smartinellibenedetti:
Converts the following things related to workflow steps to vue:
	- Drag and drop
	- Undo/redo stack
	- Numbering
	- validate all values when user “Saves” the step changes
	- adds step search

Also:
- starts refactor of OptionsEditor into a more reusable component by bringing its logic into commonUndoRedoDraggableList;
- fixes an issue in the pluginStore, where some plugins wouldn't be stored in case there's another plugin with the same name but different type;
- finishes globalLogFilter logic and styling;
- start a global mock file to start refactoring some tests;

Not included:
- auto complete logic;
- job ref workflow steps;
- step highlights;


gschueler (changes requested):
Step > Add Log Filter> The log filter can get added to the wrong step sometimes

reproduce: add step 1, add step 2, drag and drop to reorder 2 above 1. Add log filter to step 1 -> the log filter is added to step 2

Undo/Redo:
* there seems to be incorrect log filter application.  Sometimes the undo/redo puts it on the wrong step

gschueler:
Dragging a step has a transparent BG, which is slightly hard to see.  Reference: dragging options or original workflow steps have a white or gray background, which has slight alpha transparency, instead of fully transparent.
Screenshot 2024-11-21 at 3 28 55 PM

gschueler (approved):

PR #9458, opened 2024-11-20
replaced hardcoded UUID to random UUID

jayas006:

**Additional context**

### DIFF
--- a/functional-test/src/test/groovy/org/rundeck/tests/functional/api/job/JobScheduledSpec.groovy
+++ b/functional-test/src/test/groovy/org/rundeck/tests/functional/api/job/JobScheduledSpec.groovy
@@ -28,13 +28,14 @@ class JobScheduledSpec extends BaseContainer {
             def nh = upDate.hour
             def nm = upDate.minute
             def ns = upDate.second
+            def randomUUID = UUID.randomUUID().toString()
 
             def xml = """
                 <joblist>
                    <job>
                       <name>scheduled job</name>
                       <group>api-test/job-run-scheduled</group>
-                      <uuid>api-test-job-run-scheduled</uuid>
+                      <uuid>${randomUUID}</uuid>
                       <description></description>
                       <loglevel>INFO</loglevel>
                       <context>
--- a/rundeckapp/grails-spa/package-lock.json
+++ b/rundeckapp/grails-spa/package-lock.json
@@ -1,0 +1,1111 @@
+{
+  "name": "grails-spa",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "grails-spa",
+      "devDependencies": {
+        "concurrently": "^5.2.0"
+      }
+    },
+    "node_modules/@babel/runtime": {
+      "version": "7.26.0",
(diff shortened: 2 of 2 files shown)


functional-test/src/test/groovy/org/rundeck/tests/functional/api/job/JobScheduledSpec.groovy
```diff
@@ -28,13 +28,14 @@ class JobScheduledSpec extends BaseContainer {
...
             def ns = upDate.second
+            def randomUUID = UUID.randomUUID().toString()
 
             def xml = """
                 <joblist>
                    <job>
                       <name>scheduled job</name>
                       <group>api-test/job-run-scheduled</group>
-                      <uuid>api-test-job-run-scheduled</uuid>
+                      <uuid>${randomUUID}</uuid>
```
jayas006, line 38: @gschueler @mrdubr,  I am working on this JobSchedule groovy test (mostly practicing/ learning), and initially it had a hard coded UUID. When I ran the test, it worked the first time but failed on subsequent runs, showing an error that the same UUID cannot be used again. To resolve this, I replaced the hard coded UUID  with a random UUID. After making this changes, the tests runs successfully every time as the UUID is unique for each run.I also confirmed that the tests are passing in the localhost. 
Since this is my first time working on this, I wanted to check if there is any cache or underlying mechanism that I should clear or reset when running these tests? Or Is there anything else I might be missing that could cause conflicts during subsequent test runs?

Sorry accidentally I have pushed package-lock.json. I will delete this file

mrdubr (changes requested):
Switching to UUID is a good idea. I do it myself when refactoring tests and encourage others to do so -- poor man's solution to test isolation.  👍 

As for the current state of this PR:
To be valuable, justify its existence and time investment, a PR needs to solve a problem. This PR does not solve a problem -- it's a "nice to have" refactoring exercise. Make it a part of a bigger task - I'll gladly approve it. By itself -- poor precedent for the appropriate PR size.

As for your env issues:
You are running tests against your local rundeck instance. 
If the instance has a job with the same uuid already, you won't be able to create a new one due to the uniqueness constraints. You can delete jobs from your instance or use the uuid approach you've implemented (encouraged).

jayas006:
> Switching to UUID is a good idea. I do it myself when refactoring tests and encourage others to do so -- poor man's solution to test isolation. 👍
> 
> As for the current state of this PR: To be valuable, justify its existence and time investment, a PR needs to solve a problem. This PR does not solve a problem -- it's a "nice to have" refactoring exercise. Make it a part of a bigger change - I'll gladly approve it. By itself -- poor precedent for the appropriate PR size.
> 
> As for your env issues: You are running tests against your local rundeck instance. If the instance has a job with the same uuid already, you won't be able to create a new one due to the uniqueness constraints. You can delete jobs from your instance or use the uuid approach you've implemented (encouraged).

Thank you Alex for your feedback and clarifying the best approach. I understand now that while using a random UUID is a good idea for test isolation.
To align with your suggestion, I can close this PR and create a new ticket to refactor these tests as part of a larger effort in the future. For now, I made this change primarily for learning and understanding, and it helped me grasp how test isolation works and how to avoid environmental conflicts.
Let me know if you have any other thoughts.

PR #9459, opened 2024-11-21
RUN-2810: Fix execution modal when nextUi mode is enabled

smartinellibenedetti:

Fixes the problem where the execution modal would be broken if the UI currently uses the nextUi flag enabled, as it would fail to load the UI for overriding nodes for the job.

The problem consisted of 2 separate matters:

- with the nextUi enabled, the UI loads a different gsp page than the default one, which was missing 2 imports for correctly loading the UI for overriding nodes and an embedJson called pageParams;
- besides that, the UI for the execution modal was allowing users to override the nodes as long as the initial filter wasn't a dynamic one (ie `name: ${options.name}`), but wasn't initiating the required scripts for this functionality if the job was set to not allow node changes (filter overrides). 

To solve both matters, the following was done:

- add missing imports and embedJson;
- disable the checkbox that would allow the override when job is set to not allow filter overrides;

plus: 
- added a small padding on the top of the nodes section in the execution modal;
- removed an event bus parameter that was triggering an warning in the console;

I've considered to tweak the scripts initialization instead of blocking the checkbox, but upon speaking with the team, the conclusion was that it was better to prioritize honoring the flag for filter overrides being disabled.

**Additional context**

Screenshot of results:
Screenshot 2024-11-21 at 12 46 12 PM

Screenshot 2024-11-21 at 12 46 25 PM

Screenshot 2024-11-21 at 12 46 33 PM

Screenshot 2024-11-21 at 12 46 49 PM

Screenshot 2024-11-21 at 12 47 33 PM

### DIFF
--- a/functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/BulkJobActionsSpec.groovy
+++ b/functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/BulkJobActionsSpec.groovy
@@ -187,31 +187,28 @@ class BulkJobActionsSpec extends SeleniumBase {
         ExecutionShowPage executionShowPage = page ExecutionShowPage
         executionShowPage.validatePage()
     }
+    
+    def "job can be executed from the Job List Page modal with nextUI"() {
+        given:
+        String jobToExecute = generateJob(["opt1-required": "false", "opt2-required": "false"])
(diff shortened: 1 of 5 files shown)


mrdubr (approved):
So good to see that the test, once uncommented, was able to provide value.

PR #9464, opened 2024-11-22
RUN-2725 : Expose additional plugin properties in `*/plugin/list` API responses

mrdubr:
The enhancement to extend the responses of the `*/plugin/list API` to include `isHighlighted` and `highlightedOrder` properties.

**QA**

***Sample Response:***
```
GET http://localhost:4440/api/50/plugin/list
Accept:*/*
X-Rundeck-Auth-token:<...>
```

```
[
  {
    "iconUrl": "http://localhost:4440/plugin/icon/WorkflowStep/WinRMCheck",
    "title": "WinRM Check Step",
    "builtin": false,
    "id": "59ed572534b2",
    "providerMetadata": null,
    "service": "WorkflowStep",
    "highlightedOrder": 0,
    "artifactName": "py-winrm-plugin",
    "pluginVersion": "2.1.3",
    "description": "Check the connection with a remote node using winrm-python",
    "isHighlighted": false,
    "name": "WinRMCheck",
    "author": "© 2022, Pagerduty, Inc."
  },
  {
    "iconUrl": "http://localhost:4440/plugin/icon/WorkflowNodeStep/script-inline",
    "title": "Script",
    "builtin": false,
    "id": "4ca0a4eec11b",
    "providerMetadata": {
      "NodeStepPluginAdapter.DeprecatedConfigurationMode": "skip"
    },
    "service": "WorkflowNodeStep",
    "highlightedOrder": 1,
    "artifactName": "Script Node Step Plugin",
    "pluginVersion": "5.8.0-SNAPSHOT",
    "description": "Execute an inline script",
    "isHighlighted": true,
    "name": "script-inline",
    "author": "Rundeck, Inc."
  },
  {
    "iconUrl": null,
    "title": "Data Node Step",
    "builtin": false,
    "id": "44d92ebf4a8f",
    "providerMetadata": {
      "faicon": "database",
      "EnvironmentType": "localRunner"
    }
]
```

### DIFF
--- a/rundeckapp/src/test/groovy/rundeck/controllers/PluginControllerSpec.groovy
+++ b/rundeckapp/src/test/groovy/rundeck/controllers/PluginControllerSpec.groovy
@@ -265,6 +265,48 @@ class PluginControllerSpec extends Specification implements ControllerUnitTest<P
             'WebhookEvent'   | 2
     }
 
+    def "plugin list returns isHighlighted and highlightedOrder properties"() {
+        given:
+        controller.pluginService = Mock(PluginService)
+        controller.pluginApiService = Mock(PluginApiService)
+        controller.uiPluginService = Mock(UiPluginService)
+        params.service = null
+
+        def plugins = [
+                [
+                        service  : 'WebhookEvent',
+                        providers: [
+                                [isHighlighted   : true,
+                                 highlightedOrder: 123
+                                ],
+                                [:]],
+                ],
+                [
+                        service  : 'Foo',
+                        providers: [
+                                [isHighlighted   : Boolean.FALSE,
+                                 highlightedOrder: Integer.valueOf(321)
+                                ]
+                        ]
+                ]
+        ]
+
+        when:
+        controller.listPlugins()
+        then:
+        1 * controller.pluginApiService.listPlugins() >> plugins
+        response.json.size() == 3
+
+        response.json[0].isHighlighted == true
+        response.json[0].highlightedOrder == 123
+
+        !response.json[1].isHighlighted
+        !response.json[1].highlightedOrder
+
+        response.json[2].isHighlighted == false
+        response.json[2].highlightedOrder == 321
+    }
+
     def "plugin service descriptions"() {
         given:
         controller.pluginService = Mock(PluginService)
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/PluginController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/PluginController.groovy
@@ -199,6 +199,8 @@ Since: v33
                         pluginVersion: p.pluginVersion,
                         title: p.title,
                         description: p.description,
+                        isHighlighted: p.isHighlighted,
+                        highlightedOrder: p.highlightedOrder,
                         author: p.pluginAuthor,
                         iconUrl: p.iconUrl,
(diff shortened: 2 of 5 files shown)


rundeckapp/src/test/groovy/rundeck/controllers/PluginControllerSpec.groovy
```diff
@@ -265,6 +265,48 @@ class PluginControllerSpec extends Specification implements ControllerUnitTest<P
...
+                                [isHighlighted   : true,
+                                 highlightedOrder: 123
+                                ],
+                                [:]],
+                ],
+                [
+                        service  : 'Foo',
+                        providers: [
+                                [isHighlighted   : Boolean.FALSE,
+                                 highlightedOrder: Integer.valueOf(321)
```
smartinellibenedetti, line 288: Tiny nit: I'd suggest to remove the highlightedOrder here to represent the plugins that do not have that set (so that the default value would be returned)

smartinellibenedetti (approved):
LGTM :shipit:

rundeckapp/src/test/groovy/rundeck/controllers/PluginControllerSpec.groovy
mrdubr: IMO, it's a sub-standard practice to deal with data inconsistencies in the presentation layer (API response layer). Instead, this layer should reflect the state of the model -- a single source of truth. Model invariant violations should be checked and enforced in the model.

smartinellibenedetti: But that's exactly why I suggested to leave a plugin without setting an order value, to reflect the state of the model as not all plugins have an order value defined (if they are not highlighted, the code would return the default value). Take a look on the AzurePluginGroup for example.

mrdubr: What problem are we trying to solve here? This is the mock data we are looking at here?

smartinellibenedetti: Context: We have more than 100 plugins, and only 4-5 are actually highlighted and have a value for order other than the default value.

The "problem": I'm suggesting to change the mock data to match the scenarios that the API will face, that is: 
- a plugin is highlighted, and has an order value;
- a plugin isn't highlighted, and therefore there's no order value defined, so API will be returning the default value for order;

smartinellibenedetti: Marking as resolved as I missed the `[:]` in the first array

mrdubr: Both scenarios are tested:

> a plugin is highlighted, and has an order value;
https://github.com/rundeck/rundeck/pull/9464/files#diff-8b683baffd417ea324face5542fe52d1e39d676918e24d2e703660b6e96cf8ecR279

> a plugin isn't highlighted, and therefore there's no order value defined, so API will be returning the default value for order
https://github.com/rundeck/rundeck/pull/9464/files#diff-8b683baffd417ea324face5542fe52d1e39d676918e24d2e703660b6e96cf8ecR282

PR #9466, opened 2024-11-25
 Run-2644: Functional tests for job activity history

jayas006:

This is an enhancement. 

	I have added Selenium functional tests to check job activity history on the following pages:
             **Job List Page:**
	•	Ensures job activity history is retrieved correctly.
               **Job List Page with Saved Filters:**
	•	Confirms filters are saved.
                 **Activity Page:**
	•	Checks activity logs are displayed properly

Additionally, I implemented a helper method to simplify the tests by reducing repetitive code. This method handles common actions like finding elements and waiting for them to load, making the tests easier to read and maintain
 
**Additional context**


functional-test/src/test/groovy/org/rundeck/tests/functional/api/job/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,54 @@
...
+@APITest
+class JobActivityHistorySpec extends BaseContainer {
+    def setupSpec() {
+        startEnvironment()
+
+    }
+
+    def "Retrieve job Activity history from Job List Page "() {
+        given: "A Job exists in the project with execution history"
+        def jobId = "3957f735-5269-4c85-bc57-f123598886f6"
```
mrdubr, line 15: Why would this job exist?

functional-test/src/test/groovy/org/rundeck/tests/functional/api/job/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,54 @@
...
+
+        when: "The activity history for the job is requested"
+
+        def response = doGet("/job/${jobId}/executions")
```
mrdubr, line 19: Use org.rundeck.util.common.jobs.JobUtils functions to retrieve job data. This is the layer where the REST API calls are meant to be isolated. Avoid calling the API directly from the tests.

functional-test/src/test/groovy/org/rundeck/tests/functional/api/job/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,54 @@
...
+            response.code() == 200
+            def json = jsonValue(response.body(), Map)
+            json.executions.every { execution ->
+                def executionMap = execution as Map
+                println "Execution Details: ${executionMap}"
+                expectedKeys.every { key ->
+                    executionMap.containsKey(key)
+
+                }
+            }
```
mrdubr, line 35: Keep in mind the AC of your tickets.
https://pagerduty.atlassian.net/browse/RUN-2644?atlOrigin=eyJpIjoiN2YwM2VkYzcxMWM5NGVhNDg4NWUwNTUzNzk0NjBiMGIiLCJwIjoiaiJ9

You're writing (UI) "selenium" test that very "job history"

mrdubr (changes requested):
Good start!
Please, see my comments.
Keep on working on this PR!

functional-test/src/test/groovy/org/rundeck/tests/functional/api/job/JobActivityHistorySpec.groovy
jayas006: > Why would this job exist?

I have used this jobID to represent a job that already exists in the project. This makes it easier to test if the API can get the job's details.  Do I need to create a job first before checking its activity?

functional-test/src/test/groovy/org/rundeck/tests/functional/api/job/JobActivityHistorySpec.groovy
jayas006: > Keep in mind the AC of your tickets. https://pagerduty.atlassian.net/browse/RUN-2644?atlOrigin=eyJpIjoiN2YwM2VkYzcxMWM5NGVhNDg4NWUwNTUzNzk0NjBiMGIiLCJwIjoiaiJ9
> 
> You're writing (UI) "selenium" test that very "job history"

Thanks, I understand that I need to follow selenium based test structure for this. I am reviewing the existing selenium tests for guidance and update my code accordingly. It's great to see the browser interaction during the tests.

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,82 @@
...
+class JobActivityHistorySpec extends SeleniumBase {
+    def setupSpec() {
+        setupProject(SELENIUM_BASIC_PROJECT)
+    }
+    def setup() {
+        def loginPage = go LoginPage
+        loginPage.login(TEST_USER, TEST_PASS)
+        println "Login successful"
+    }
+    def "create job has basic fields"() {
```
mrdubr, line 29: this test is not related to your ticket AC

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,82 @@
...
+        println "job created successfully"
+    }
+    def "run job and validate execution"() {
+        def executionShowPage = page ExecutionShowPage
+        def jobShowPage = go JobShowPage, SELENIUM_BASIC_PROJECT
+        when: "Navigate to the Job Show Page and run the job"
+        println "Navigating to Job Show Page"
+        jobShowPage.validatePage()
+        println "Job Show Page validated"
+        def runJobButton = driver.findElement(By.cssSelector(".btn.btn-success.btn-simple.btn-hover.btn-xs.act_execute_job"))
```
mrdubr, line 46: Isolate css selectors to the `JobShowPage`.
Use "helpers" from the `JobShowPage` e.g. `runJobLink`

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,82 @@
...
+        println "Run Job button located"
+        runJobButton.click()
+        println "Run Job button clicked"
+        def requestOptionField = driver.findElement(By.cssSelector("input[name='extra.option.reqOpt1']"))
+        requestOptionField.sendKeys("test-value")
+        println "Filled required option field"
+
+        def createJobButton = driver.findElement(By.id("execFormRunButton"))
```
mrdubr, line 54: is it `create` or `run`?

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,82 @@
...
+        when:
+        def jobCreatePage = go JobCreatePage, SELENIUM_BASIC_PROJECT
+        then:
+        jobCreatePage.validatePage()
+        jobCreatePage.jobNameInput
+        jobCreatePage.groupPathInput
+        jobCreatePage.descriptionTextarea
+        println "job created successfully"
+    }
+    def "run job and validate execution"() {
```
mrdubr, line 39: this test is not related to your ticket AC

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,82 @@
...
+import org.rundeck.util.gui.pages.jobs.JobCreatePage
+import org.rundeck.util.gui.pages.execution.ExecutionShowPage
+import org.rundeck.util.gui.pages.execution.HtmlRenderedOutputPage
+import org.rundeck.util.gui.pages.jobs.JobListPage
+import org.rundeck.util.gui.pages.activity.ActivityPage
+import org.rundeck.util.gui.pages.jobs.JobTab
+import org.rundeck.util.gui.pages.jobs.JobShowPage
+import spock.lang.Stepwise
+@SeleniumCoreTest
+@Stepwise
```
mrdubr, line 19: You should avoid introducing an execution sequence to the tests as it is generally considered a bad practice in unit testing. Rewrite your tests by factoring out shared logic to "setup" and "setupspec".

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,82 @@
...
+        println "Clicked Run Job Now"
+        then: "Wait for the execution to complete and validate success"
+        println "Waiting for execution to complete"
+        executionShowPage.waitForElementVisible(executionShowPage.executionStateDisplayLabel)
+        println "Execution state display label visible"
+        println "Execution completed"
+        executionShowPage.validateStatus('SUCCEEDED')
+        println "Job executed successfully"
+    }
+    def "validate activity history from Job List Page"() {
```
mrdubr, line 65: I've added screenshots to the ticket. 
Please, ensure you cover all scenarios in the AC.

mrdubr (changes requested):
Please, see my comments.

functional-test/src/test/groovy/org/rundeck/util/gui/pages/jobs/JobShowPage.groovy
```diff
@@ -48,6 +48,8 @@ class JobShowPage extends BasePage implements ActivityListTrait {
     By jobLinkTitleBy = By.xpath("//a[contains(@class, 'job-header-link')]")
     By autocompleteJobStepDefinitionBy = By.cssSelector("#wfitem_0 > span > div > div > span > span > span.text-success")
     By runFormBy = By.cssSelector("#execDiv #exec_options_form #formbuttons #execFormRunButton")
+    By runJobButtonBy = By.cssSelector(".btn.btn-success.btn-simple.btn-hover.btn-xs.act_execute_job")
+    By executionStateLabelBy = By.id("executionStateDisplayLabel")
```
jayas006, line 52: @mrdubr  before using it in my test I wanted to make sure if I am implementing it correctly?  Do I need to use these selector with the helper method clickRunJobButton and waitForExecutionState

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
jayas006: > this test is not related to your ticket AC

Thank you for pointing that out. I initially thought the test needed to include the flow of creating a job and then executing it. I have removed it

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
jayas006: > Isolate css selectors to the `JobShowPage`. Use "helpers" from the `JobShowPage` e.g. `runJobLink`

I have not  implemented  this yet as I wanted to confirm it first.

functional-test/src/test/groovy/org/rundeck/util/gui/pages/jobs/JobShowPage.groovy
mrdubr: IIRC the ticket AC, your tests are not related to the JobShow page, thus I'm not sure you'd need this.
It's much simpler (and more maintainable) to setup (prepare) data for your tests using backend API.
There are numerous tests that create jobs and wait for executions to complete.
I'd recommend following that pattern.

jayas006: > IIRC the ticket AC, your tests are not related to the JobShow page, thus I'm not sure you'd need this. It's much simpler (and more maintainable) to setup (prepare) data for your tests using backend API. There are numerous tests that create jobs and wait for executions to complete. I'd recommend following that pattern.

I tried to refactor the tests by looking other tests. Though it's failing but I am keep working on it. It  doesn't look complicated but I may be missing something.

functional-test/src/test/groovy/org/rundeck/tests/functional/api/job/JobActivityHistorySpec.groovy
mrdubr: Your tests should be independent. Given a test suite TS with (optional) setup and tear down functions and multiple tests t1, t2, ..., tn in that suite. You should be able to run `tx` at any time and in any order. 
If you deviate from this rule, there should be a very good reason.

Thus, if your test relies on an job guide, it should be created by the test itself (or setup functionality).

functional-test/src/test/groovy/org/rundeck/util/gui/pages/jobs/JobShowPage.groovy
mrdubr: Feel free to zoom me if you need some guidance on this

jayas006:
@mrdubr I have refactored the tests and now each test is independent. The tests look smaller and more concise. I have declared the CSS selector in the test file temporarily to check if they works. Since I have confirmed they are  working, I will move them to the Execution and Job page as  helper methods. I understand that adding them as helper method is Better for clarity and reusability.

functional-test/src/test/groovy/org/rundeck/util/gui/pages/jobs/JobListPage.groovy
```diff
@@ -52,7 +52,49 @@ class JobListPage extends BasePage implements ActivityListTrait {
...
     By dashboardTabBy = By.id("joblisttab")
-
+    static final By SAVE_FILTER_BUTTON = By.cssSelector("button[data-test-id='save-filter-button']")
+    static final By SAVE_FILTER_INPUT = By.cssSelector("input.form-control[data-action='auto-focus']")
+    static final By OK_BUTTON = By.xpath("/html/body/div[6]/div/div/div[3]/button[2]")
+    static final By FILTER_DROPDOWN = By.cssSelector("span.dropdown-toggle.btn.btn-secondary.btn-sm.text-secondary")
+    static final By SAVED_FILTER_LINK = By.cssSelector("a[data-test='filter-link']")
+    static final By ACTIVITY_ROWS = By.cssSelector("table.activity-list-table tbody tr.link.activity_row")
+    static final By EVENT_ICON = By.cssSelector("td.eventicon")
+/** Click Save Filter button */
```
mrdubr, line 62: We tend to stick to the javadoc commenting style: https://www.oracle.com/ca-en/technical-resources/articles/java/javadoc-tool.html

In particular, in this case, the indentation is unconventional.

functional-test/src/test/groovy/org/rundeck/util/gui/pages/jobs/JobShowPage.groovy
```diff
@@ -136,10 +136,12 @@ class JobShowPage extends BasePage implements ActivityListTrait {
         }
     }
 
+
```
mrdubr, line 139: Please, avoid introducing unnecessary formatting changes. It adds noise to a PR w/o providing value.

functional-test/src/test/groovy/org/rundeck/util/gui/pages/jobs/JobListPage.groovy
```diff
@@ -52,7 +52,49 @@ class JobListPage extends BasePage implements ActivityListTrait {
...
+    JobListPage clickSaveFilterButton() {
+        el(SAVE_FILTER_BUTTON).click()
+        return this
+    }
+/** Enter filter name */
```
mrdubr, line 67: Also, notice how in the multiple methods above the comment itself does not add any value since it's equivalent to the method name. This makes the comment unnecessary.

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,106 @@
...
+class JobActivityHistorySpec extends SeleniumBase {
+    WebDriverWait wait
+    String jobId
+
+    def setupSpec() {
+        setupProject(SELENIUM_BASIC_PROJECT)
+    }
+
+    def setup() {
+        wait = new WebDriverWait(getDriver(), Duration.ofSeconds(30))
```
mrdubr, line 25: Is there a reason you're configuring/reconfiguring `wait`. If so, please add a comment explaining that.

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,106 @@
...
+        // Login
```
mrdubr, line 26: Useless comment

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,106 @@
...
+        def loginPage = go LoginPage
+        loginPage.login(TEST_USER, TEST_PASS)
+        // Create a job
+        def jobDefinition = JobUtils.generateScheduledExecutionXml("TestJobActivityHistory")
+        def client = getClient() // Assuming a method exists to get RdClient
```
mrdubr, line 31: Does this comment add value?

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,106 @@
...
+        def response = JobUtils.createJob(SELENIUM_BASIC_PROJECT, jobDefinition, client)
+        assert response.succeeded.size() > 0: "Failed to create job"
+        jobId = response.succeeded[0].id
+        println "Job created with ID: ${jobId}"
```
mrdubr, line 35: Please, ensure that debugging statements and any temporary code are removed from your code before a code review is started.

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,106 @@
...
+        // Run the job
+        def runResponse = JobUtils.executeJob(jobId, client)
+        assert runResponse.code() == 200: "Failed to execute job"
+        println "Job executed with response code: ${runResponse.code()}"
+    }
+
+    def cleanup() {
+        getDriver().quit()
```
mrdubr, line 43: Is there a reason you need this? If so, please add a comment explaining that.

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    /**
+     Test 1: Review Job Activity History from Job List Page
```
mrdubr, line 47: Your comment is equivalent to the test name and does not provide any additional value.

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,106 @@
...
+        then: "Validate job execution is listed in Activity History"
+        activityPage.validatePage()
+        List<WebElement> activityList = activityPage.getActivityRows()
+        println "Activity List Size: ${activityList.size()}"
+        assert activityList.size() > 0: "No activity rows found"
+        WebElement firstActivityRow = activityList.get(0)
+        WebElement statusIcon = ExecutionShowPage.getActivityExecStatusIcon(firstActivityRow)
+        String status = statusIcon.getAttribute("data-statusstring")
+        println "First Activity Row Status: ${status}"
+        assert status.equalsIgnoreCase("SUCCEEDED"): "Expected 'SUCCEEDED', but found: '${status}'"
```
mrdubr, line 62: What will ensure that the job has finished executing by the time this assertion is checked?

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
jayas006: > Is there a reason you're configuring/reconfiguring `wait`. If so, please add a comment explaining that.

I have explained it in the comment, I added it because the test was failing in CircleCI due to element took longer and it was passing locally. Increasing the wait duration resolved the issue.

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
jayas006: > Comment that does not provide additional value

Thanks, I have removed all the redundant comments.

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
jayas006: > Please, ensure that debugging statements and any temporary code are removed from your code before a code review is started.

Yes I have checked few times and removed the debugging statements.

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
jayas006: > Is there a reason you need this? If so, please add a comment explaining that.

It was needed earlier when my previous script caused test failures in CircleCI but since updating the script, I have removed it as well

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,94 @@
...
+        when: "Navigate to the Job List Page and apply a saved filter"
+        def jobListPage = go JobListPage, SELENIUM_BASIC_PROJECT
+        jobListPage.validatePage()
+        jobListPage.clickSaveFilterButton()
+                .enterFilterName("MySavedFilter")
+                .confirmFilterSave()
+        jobListPage.openFilterDropdown()
+                .selectSavedFilter()
+        then: "Validate that the saved filter is applied and results are filtered"
+        def filteredRows = jobListPage.getActivityRows()
```
mrdubr, line 72: The saved/selected filter is the same as the default 'any time' filter, thus this test does not verify that filtering applies.

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,94 @@
...
+        def activityPage = go ActivityPage, SELENIUM_BASIC_PROJECT
+        activityPage.loadActivityPageForProject(SELENIUM_BASIC_PROJECT)
+        then: "Validate job execution is listed in Activity History"
+        activityPage.validatePage()
+        def activityList = activityPage.getActivityRows()
+        assert activityList.size() > 0: "No activity rows found on the Activity Page"
+        def firstActivityRow = activityList.get(0)
+        def firstActivityStatusIcon = ExecutionShowPage.getActivityExecStatusIcon(firstActivityRow)
+        def firstActivityStatus = firstActivityStatusIcon.getAttribute("data-statusstring")
+        assert firstActivityStatus.equalsIgnoreCase("SUCCEEDED"):
```
mrdubr, line 91: Ditto

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,94 @@
...
+        jobListPage.clickSaveFilterButton()
+                .enterFilterName("MySavedFilter")
+                .confirmFilterSave()
+        jobListPage.openFilterDropdown()
+                .selectSavedFilter()
+        then: "Validate that the saved filter is applied and results are filtered"
+        def filteredRows = jobListPage.getActivityRows()
+        assert filteredRows.size() > 0: "No filtered results found"
+        def firstRowStatus = jobListPage.getFirstRowStatus()
+        assert firstRowStatus.equalsIgnoreCase('SUCCEEDED'):
```
mrdubr, line 75: Ditto

mrdubr (changes requested):
Great progress!
Please, see my comments for the additional info on the improvement recommendations.

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
mrdubr: I see... You're using this to wait for the job to get into the `SUCCEEDED` status.

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
mrdubr: I see now that the `wait.until` on line 56 will ensure this! 👍

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
jayas006: > The saved/selected filter is the same as the default 'any time' filter, thus this test does not verify that filtering applies.

I didn't understand this. Just for clarification,  I need to update the test to ensure it applies specific filter criteria ("In Last day/ In last week") when saving the filter and validate that all the rows in the filtered results meet the criteria?

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,94 @@
...
+        setupProject(SELENIUM_BASIC_PROJECT)
+    }
+
+    /**
+     * Setup the test environment.
+     */
+    def setup() {
+        // Set up WebDriverWait to wait up to 30 seconds for elements that take time to appear or change
+        wait = new WebDriverWait(getDriver(), Duration.ofSeconds(30))
+        def loginPage = go LoginPage
```
mrdubr, line 33: nit:

Note: My `nit` comments are FYIs, subjective and will not stop the PR from getting approved! :-) 

In general, it's considered too noisy to introduce a var to use it *once* right away.
This could be compressed to something like `(go LoginPage).login(TEST_USER, TEST_PASS)`

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,94 @@
...
+
+
+    def "review job activity history from Job List Page"() {
+        when: "Navigate to the Job List Page"
+        def activityPage = go ActivityPage, SELENIUM_BASIC_PROJECT
+        activityPage.loadActivityPageForProject(SELENIUM_BASIC_PROJECT)
+        then: "Validate job execution is listed in Activity History"
+        activityPage.validatePage()
+        List<WebElement> activityList = activityPage.getActivityRows()
+        assert activityList.size() > 0: "No activity rows found"
```
mrdubr, line 53:
```suggestion
        !activityList.empty()
```
In general, avoid using `size()` to check for emptiness for perf reasons

Also, in a Spock specification, you do not necessarily need to use `assert` statements in the `then` section. Spock's then block is designed for assertions, and it implicitly treats any boolean expression as an assertion.

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,94 @@
...
+        WebElement firstActivityRow = activityList.get(0)
+        WebElement statusIcon = ExecutionShowPage.getActivityExecStatusIcon(firstActivityRow)
+        wait.until{statusIcon.getAttribute("data-statusstring").equalsIgnoreCase("SUCCEEDED")}
```
mrdubr, line 56: You're using UI state to wait for the backend job to succeed. Technically, it's correct but slightly unconventional. I'd add a comment explaining this to remove ambiguity.

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,94 @@
...
+        String status = statusIcon.getAttribute("data-statusstring")
+        assert status.equalsIgnoreCase("SUCCEEDED"): "Expected 'SUCCEEDED', but found: '${status}'"
```
mrdubr, line 58: I'd assume `wait.until` ensure this is true. I don't think we need these lines.

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
mrdubr: You need to determine if verifying **correctness** of a filter application is in scope for your work. 
If not, state this explicitly. If it's not possible to do so (e.g. due to test data provisioning constraints) -- state that explicitly. 

Otherwise, it looks like this test is testing filtering correctness, but, in reality, it only tests the `*` (any time) fitler.

mrdubr (changes requested):
You're on the write path! 👍 
Just need a few tweaks to make this test suite resilient!

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
jayas006: > You need to determine if verifying **correctness** of a filter application is in scope for your work. If not, state this explicitly. If it's not possible to do so (e.g. due to test data provisioning constraints) -- state that explicitly.
> 
> Otherwise, it looks like this test is testing filtering correctness, but, in reality, it only tests the `*` (any time (aka lack of filtering)) filter.

Thank you for your feedback. I have made changes `in the last week` instead of` anytime` to improve this test. using anytime filter, I am now testing with a specific time filter. I appreciate your guidance, as it is helping me make my tests more accurate. I am learning a lot and striving to perfect my work.

jayas006:
> You're on the write path! 👍 Just need a few tweaks to make this test suite resilient!

@mrdubr, Thank you so much for your feedback. I believe I have addressed all your comments and made improvements to the test. There might be few recent changes left to review. I plan to go over it a few more times to ensure nothing is missing. Once that’s done, I think the test will be ready for review. I truly appreciate your help in making this test as perfect as possible.

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,97 @@
...
+        // Set up WebDriverWait to wait up to 30 seconds for elements that take time to appear or change
+        wait = new WebDriverWait(getDriver(), Duration.ofSeconds(30))
+        (go LoginPage).login(TEST_USER, TEST_PASS)
+        def jobDefinition = JobUtils.generateScheduledExecutionXml("TestJobActivityHistory")
+        def client = getClient()
+        def response = JobUtils.createJob(SELENIUM_BASIC_PROJECT, jobDefinition, client)
+        assert response.succeeded.size() > 0: "Failed to create job"
+        jobId = response.succeeded[0].id
+        def runResponse = JobUtils.executeJob(jobId, client)
+        assert runResponse.code() == 200: "Failed to execute job"
```
mrdubr, line 36: Using assert within the setup method is generally not recommended because setup is intended for initializing the test environment rather than for performing assertions. Assertions should be placed within test methods to validate conditions.

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,97 @@
...
+
+    def "review job activity history from Job List Page"() {
+        when: "Navigate to the Job List Page"
+        def activityPage = go ActivityPage, SELENIUM_BASIC_PROJECT
+        activityPage.loadActivityPageForProject(SELENIUM_BASIC_PROJECT)
+
+        then: "Validate job execution is listed in Activity History"
+        activityPage.validatePage()
+        wait.until {
+            def activityList = activityPage.getActivityRows()
```
mrdubr, line 49: nit:
ditto
`!activityPage.getActivityRows().isEmpty()`

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,97 @@
...
+            !activityList.isEmpty()
+        }
+
+        def firstActivityRow = activityPage.getActivityRows().get(0)
```
mrdubr, line 53: Unnecessary validation:
Your test does not need to check if the execution has succeeded. 
You are testing "activity", not job execution status. Seeing a row is sufficient.
Contrary to the initial gut feel, "less is more" with unit testing. Keep them focused.

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,97 @@
...
+        wait.until {
+            statusIcon.getAttribute("data-statusstring").equalsIgnoreCase("SUCCEEDED")
+        }
+    }
+
+
+    def "validate activity history from Job List Page Saved Filters"() {
+        when: "Navigate to the Job List Page and apply a saved filter"
+        def jobListPage = go JobListPage, SELENIUM_BASIC_PROJECT
+        jobListPage.clickAnyTimeButton()
```
mrdubr, line 64: This code block is so elegant! 👍
Compact, declarative, to the point! 🥇

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,97 @@
...
+                .clickSaveFilterButton()
+                .enterFilterName("testFilter")
+                .confirmFilterSave()
+                .openFilterDropdown()
+                .selectSavedFilter()
+
+        then: "Wait for the job to complete and validate the saved filter is applied"
+        wait.until {
+            def status = jobListPage.getFirstRowStatus()
+            status.equalsIgnoreCase('SUCCEEDED')
```
mrdubr, line 75: ditto

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        and: "Validate that the results are filtered"
```
mrdubr, line 78: I'm not sure I understand how this confirms that the results are filtered. What value does this assertion add to the one above?

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,97 @@
...
+        wait.until {
+            def filteredRows = jobListPage.getActivityRows()
+            println "Number of filtered rows: ${filteredRows.size()}"
+            filteredRows.size() > 0
+        }
+    }
+
+    def "review job activity history from Activity Page"() {
+        when: "Navigate to the Activity Page"
```
mrdubr, line 87: don't you validate the same things in the `"review job activity history from Job List Page"` test already?

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,97 @@
...
+        def activityPage = go ActivityPage, SELENIUM_BASIC_PROJECT
+        activityPage.validatePage()
+        activityPage.loadActivityPageForProject(SELENIUM_BASIC_PROJECT)
+
+        then: "Validate job execution is listed in Activity History"
+        activityPage.validatePage()
```
mrdubr, line 93: Already validated above

functional-test/src/test/groovy/org/rundeck/util/gui/pages/activity/ActivityPage.groovy
```diff
@@ -44,4 +45,11 @@ class ActivityPage extends BasePage implements ActivityListTrait{
         els By.xpath("//*[contains(text(),'${jobName}')]")
     }
 
+    static void validateFirstActivityRow(List<WebElement> activityList) {
+        assert !activityList.isEmpty() : "Activity list is empty, no rows to validate."
```
mrdubr, line 49: Assertions are for tests, but this is the "model".
You're using groovy assertions here (that are different from Spec assertions) that are generally used to asserts invariants -- e.g. null checks.

functional-test/src/test/groovy/org/rundeck/util/gui/pages/activity/ActivityPage.groovy
```diff
@@ -44,4 +45,11 @@ class ActivityPage extends BasePage implements ActivityListTrait{
...
+        WebElement firstActivityRow = activityList.get(0)
+        WebElement statusIcon = ExecutionShowPage.getActivityExecStatusIcon(firstActivityRow)
+        String status = statusIcon.getAttribute("data-statusstring")
+        assert status.equalsIgnoreCase("SUCCEEDED") : "Expected 'SUCCEEDED', but found: '${status}'"
```
mrdubr, line 53: ditto: activity is different from "success".

functional-test/src/test/groovy/org/rundeck/util/gui/pages/jobs/JobListPage.groovy
```diff
@@ -52,7 +52,61 @@ class JobListPage extends BasePage implements ActivityListTrait {
...
+    static final By IN_THE_LAST_WEEK_BUTTON = By.cssSelector("a[data-test-id='period-option']")
 
+    JobListPage clickAnyTimeButton() {
+        waitForElementVisible(ANY_TIME_BUTTON).click()
+        return this
+    }
+
+    JobListPage clickLastWeekButton() {
+        waitForElementVisible(IN_THE_LAST_WEEK_BUTTON).click()
+        return this
```
mrdubr, line 72: I don't think you need the `return` statement.

functional-test/src/test/groovy/org/rundeck/util/gui/pages/jobs/JobListPage.groovy
```diff
@@ -52,7 +52,61 @@ class JobListPage extends BasePage implements ActivityListTrait {
...
+    }
+    JobListPage clickSaveFilterButton() {
+        waitForElementVisible(SAVE_FILTER_BUTTON).click()
+        return this
```
mrdubr, line 76: ditto

mrdubr (changes requested):
Good progress! 👍 
Please see my comments.

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
jayas006: > nit: Using assert within the setup method is generally not recommended because setup is intended for initializing the test environment rather than for performing assertions. Assertions should be placed within test methods to validate conditions.

So  I remove the assertions and use throw exceptions if initialization fails?

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
jayas006: > I'm not sure I understand how this confirms that the results are filtered. What value does this assertion add to the one above?

The goal  of this check is to make sure the saved filter is actually applied. While the previous check ensuring the job status is succeeded. I thought by adding this it ensures that the filtered results are not empty and match the filter criteria.  Do you have suggestions for a better way to confirm the filter is working? If I missing anything please let me know, I will be happy to revise it

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
jayas006: > don't you validate the same things in the `"review job activity history from Job List Page"` test already?

Yes you are correct. This test and review job activity history from job list page test do validate similar functionality. The main intent behind this test was to ensure that activity history is accessible and validated directly from the Activity Page. I will revise this test and remove it if redundant. Thank you for pointing this out

jayas006: > > don't you validate the same things in the `"review job activity history from Job List Page"` test already?
> 
> Yes you are correct. This test and review job activity history from job list page test do validate similar functionality. The main intent behind this test was to ensure that activity history is accessible and validated directly from the Activity Page. I will revise this test and remove it if redundant. Thank you for pointing this out

Also it required according to the ticket `Review Job Activity History From Activity Page.`

functional-test/src/test/groovy/org/rundeck/util/gui/pages/jobs/JobListPage.groovy
jayas006: > I don't think you need the `return` statement.

the return this statements are necessary to support method chaining in the test casesWithout the return this statement, chaining multiple actions like 
clickAnyTimeButton()
             ```
   .clickLastWeekButton()
                .clickSaveFilterButton()
                .enterFilterName("testFilter")
                .confirmFilterSave()
                .openFilterDropdown()
                .selectSavedFilter()
```
                would not work. 
                If I remove the return this then I have to refactor my script like this 
              ```
  jobListPage.clickAnyTimeButton()
                jobListPage .clickLastWeekButton()
                jobListPage  .clickSaveFilterButton()
```
        and test is passing

jayas006: > > I don't think you need the `return` statement.
> 
> the return this statements are necessary to support method chaining in the test casesWithout the return this statement, chaining multiple actions like
> 
> ```
> clickAnyTimeButton()
>    .clickLastWeekButton()
>                 .clickSaveFilterButton()
>                 .enterFilterName("testFilter")
>                 .confirmFilterSave()
>                 .openFilterDropdown()
>                 .selectSavedFilter()
> ```

But after removing return this I need to refactor test like this 
              ```
  jobListPage.clickAnyTimeButton()
                jobListPage .clickLastWeekButton()
                jobListPage  .clickSaveFilterButton()

```

File: functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy [hunk 40]
```diff
@@ -30,14 +29,16 @@ class JobActivityHistorySpec extends SeleniumBase {
         def jobDefinition = JobUtils.generateScheduledExecutionXml("TestJobActivityHistory")
         def client = getClient()
         def response = JobUtils.createJob(SELENIUM_BASIC_PROJECT, jobDefinition, client)
-        assert response.succeeded.size() > 0: "Failed to create job"
+        if (response.succeeded.size() == 0) {
+            throw new RuntimeException("Failed to create job")
+        }
```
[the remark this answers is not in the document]
mrdubr: There is a fine line between defensive programming and unnecessary code. 
Are you sure you need to check the `response`? What's the error behaviour of the helper function you're calling? That's what my comment was alluding to. 

Groovy assertion (don't confuse them with Spec assertions) are perfectly appropriate **when needed** and are definitely prefered over `if(error_condition) { throw }` due to compactness and intention clarity.

functional-test/src/test/groovy/org/rundeck/util/gui/pages/activity/ActivityPage.groovy
```diff
@@ -44,4 +45,10 @@ class ActivityPage extends BasePage implements ActivityListTrait{
         els By.xpath("//*[contains(text(),'${jobName}')]")
     }
 
+    static void validateFirstActivityRow(List<WebElement> activityList) {
```
mrdubr, line 48: the term `validate` in the name of this function does not convey much meaning. What does it mean for the first activity raw to be valid or invalid? It depends on the context.

Also, `Exception`s (albeit overused in java) are reserved for something exceptional.
Having no activity is a perfectly valid state in our business logic. A boolean would make much more sense.

E.g. function along the lines of `boolean hasActivity(List<WebElement> activityList)`

Now, we can simplify this further employing groovy truthy logic
```
!!activityList
```

So, do we even need this function?

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,82 @@
...
+        if (runResponse.code() != 200) {
+            throw new RuntimeException("Failed to execute job")
+        }
+    }
+
+    def "review job activity history from Job List Page"() {
+        when: "Navigate to the Job List Page"
+        def activityPage = go ActivityPage, SELENIUM_BASIC_PROJECT
+        activityPage.loadActivityPageForProject(SELENIUM_BASIC_PROJECT)
+        println "Navigated to Activity Page"
```
mrdubr, line 46: ditto

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
mrdubr: Please, take a look at the screenshots in: https://pagerduty.atlassian.net/browse/RUN-2644?atlOrigin=eyJpIjoiZDgxYzVjZjAyMmFmNDQ1M2I4NTZhNThhMDgwZmFjN2YiLCJwIjoiaiJ9

The "job activity" appears on **two** different pages.
It seems that the tests are only testing one of the pages.

functional-test/src/test/groovy/org/rundeck/util/gui/pages/jobs/JobListPage.groovy
mrdubr: Good point! Thanks for the clarification!

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,82 @@
...
+    WebDriverWait wait
+    String jobId
+
+    def setupSpec() {
+        setupProject(SELENIUM_BASIC_PROJECT)
+    }
+
+    /**
+     * Setup the test environment.
+     */
```
mrdubr, line 24: nit: adds no value
```suggestion
```

mrdubr (changes requested):
Almost there! 👍 
A few more things to take care of!

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
jayas006: Yes both pages are validating to activity instead one of them validate to Job page where JobLists are. So for the first test it should show only to Job Page and third test for activity page. Thanks for pointing that out

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,73 @@
...
+                .confirmFilterSave()
+                .openFilterDropdown()
+                .selectSavedFilter()
+        then: "Validate the saved filter is applied"
+        wait.until{
+            jobListPage.getAppliedFilterName() == "TestFilter"
+        }
+
+        }
+
```
jayas006, line 62: I changed the validation to check if the saved filter is applied by comparing it to the filter name TestFilter. This is different from the first test, which checks if job activity rows are present. For this second test, the goal is to make sure the saved filter is correctly applied when viewing job activity history

jayas006:
> Almost there! 👍 A few more things to take care of!

Thanks for your feedback. I have tried to applied all the pointers. Double checked if I left any debugging statements.

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,73 @@
...
+                .enterFilterName("TestFilter")
+                .confirmFilterSave()
+                .openFilterDropdown()
+                .selectSavedFilter()
+        then: "Validate the saved filter is applied"
+        wait.until{
+            jobListPage.getAppliedFilterName() == "TestFilter"
+        }
+
+        }
```
mrdubr, line 61: white spacing

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
mrdubr: Are you validating that the filtering logic is correct -- only activity matching the updated filter shows up? If not, the test name does not make much sense: `"validate activity history from Job List Page Saved Filters"`

functional-test/src/test/groovy/org/rundeck/util/gui/pages/activity/ActivityPage.groovy
```diff
@@ -44,4 +45,8 @@ class ActivityPage extends BasePage implements ActivityListTrait{
         els By.xpath("//*[contains(text(),'${jobName}')]")
     }
 
+    static boolean hasActivity(List<WebElement> activityList) {
```
mrdubr, line 48: you do not need this static function anymore in this class anymore

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -34,16 +34,16 @@ class JobActivityHistorySpec extends SeleniumBase {
     def "review job activity history from Job List Page"() {
         when: "Navigate to the Job List Page"
         def JobListPage = go JobListPage, SELENIUM_BASIC_PROJECT
-        JobListPage.loadJobListForProject(SELENIUM_BASIC_PROJECT)
```
jayas006: I removed the `loadJobListForProject` method because the `go JobListPage` method already handling navigation and set up the page context. I believe  loadJobListForProject was not adding any additional value and the test works consistently without it.

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,74 @@
...
+
+    def "review job activity history from Job List Page"() {
+        when: "Navigate to the Job List Page"
+        def JobListPage = go JobListPage, SELENIUM_BASIC_PROJECT
+
+        then: "Validate job execution is listed in Activity History"
+        JobListPage.validatePage()
+        wait.until {
+        !JobListPage.getActivityRows().isEmpty()
+
```
mrdubr, line 42: indentation

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,74 @@
...
+        def client = getClient()
+        def response = JobUtils.createJob(SELENIUM_BASIC_PROJECT, jobDefinition, client)
+        jobId = response.succeeded[0].id
+        assert JobUtils.executeJob(jobId, client).isSuccessful()
+    }
+
+
+    def "review job activity history from Job List Page"() {
+        when: "Navigate to the Job List Page"
+        def JobListPage = go JobListPage, SELENIUM_BASIC_PROJECT
```
mrdubr, line 36: Variable name.
Are you using IntelliJ? It should be giving you warning for case violations.
Code review is not the appropriate time to detect such errors.

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,74 @@
...
+
+
+    def "review job activity history from Activity Page"() {
+        when: "Navigate to the Activity Page"
+        def activityPage = go ActivityPage, SELENIUM_BASIC_PROJECT
+        activityPage.validatePage()
+        activityPage.loadActivityPageForProject(SELENIUM_BASIC_PROJECT)
+
+        then: "Validate job execution is listed in Activity History"
+        !activityPage.getActivityRows().isEmpty()
```
mrdubr, line 71: How can we be sure that the job has started (has some activity) prior to running this check?

functional-test/src/test/groovy/org/rundeck/util/gui/pages/activity/ActivityPage.groovy
```diff
@@ -5,6 +5,7 @@ import org.openqa.selenium.By
 import org.openqa.selenium.WebElement
 import org.rundeck.util.container.SeleniumContext
 import org.rundeck.util.gui.pages.BasePage
+import org.rundeck.util.gui.pages.execution.ExecutionShowPage
```
mrdubr, line 8: do we need this import?

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
jayas006: > Variable name. Are you using IntelliJ? It should be giving you warning for case violations. Code review is not the appropriate time to catch such errors.

Yes right, I know what may happened. When I copied the class name and used it as a variable then missed to use the defined variable name. But yes I didn't get any warning in intelliJ

functional-test/src/test/groovy/org/rundeck/util/gui/pages/execution/ExecutionShowPage.groovy
```diff
@@ -105,6 +107,11 @@ class ExecutionShowPage extends BasePage  implements ActivityListTrait{
         el optionValueSelected
     }
 
+    static WebElement getActivityExecStatusIcon(WebElement row) {
```
mrdubr, line 110: This static method is not safe and hard to reuse.
It takes a generic WebElement and returns a generic WebElement.
What stops someone from passing in a WebElement that has nothing to do with the row?

functional-test/src/test/groovy/org/rundeck/util/gui/pages/execution/ExecutionShowPage.groovy
```diff
@@ -2,6 +2,7 @@ package org.rundeck.util.gui.pages.execution
 
 import groovy.transform.CompileStatic
 import org.openqa.selenium.By
+
```
mrdubr, line 5: don't introduce unneeded formatting changes

functional-test/src/test/groovy/org/rundeck/util/gui/pages/jobs/JobShowPage.groovy
mrdubr: Revert these unneeded changes

rundeckapp/grails-spa/packages/ui-trellis/src/app/utilities/locales/en_US.js
```diff
@@ -937,6 +937,8 @@ const messages = {
   "workflow.search.help.string15": "property:title=value",
   "workflow.search.help.string16": "property name:",
   "workflow.search.help.string17": "property:name=value",
+  "node.step.plugin.plural": "{0} Node Step Plugins",
```
mrdubr, line 940: I does not look like these changes are related

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/ChoosePluginModal.vue
```diff
@@ -17,24 +17,39 @@
         :title="tabTitle(service.service, i)"
       >
         <div class="list-group">
-          <button
-            v-for="prov in service.providers"
-            class="list-group-item"
-            @click.prevent="chooseProviderAdd(service.service, prov.name)"
-            data-testid="provider-button"
+          <template
```
mrdubr, line 20: I does not look like these changes are related

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/tests/ChoosePluginModal.spec.ts
```diff
@@ -1,79 +1,118 @@
-import { shallowMount, VueWrapper, flushPromises } from "@vue/test-utils";
+import { mount, VueWrapper, flushPromises } from "@vue/test-utils";
 import ChoosePluginModal from "../ChoosePluginModal.vue";
-import { Modal, Btn, Tabs, Tab } from "uiv";
```
mrdubr, line 3: I does not look like these changes are related

rundeckapp/grails-spa/packages/ui-trellis/src/library/stores/Plugins.ts
```diff
@@ -18,7 +18,7 @@ export class PluginStore {
   async load(service: string): Promise<void> {
     if (this.pluginsByService[service]) return void 0;
     const plugins = await this.client.apiRequest({
-      pathTemplate: "api/40/plugin/list",
```
mrdubr, line 21: I does not look like these changes are related

rundeckapp/grails-spa/packages/ui-trellis/src/app/utilities/locales/en_US.js
jayas006: > I does not look like these changes are related

This may be from the conflict files.

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
jayas006: Need to use wait.until ?

jayas006:
> Please, see my comments.
> 
> Also, please review the changes you're submitting for the review. It appears that some of the changes are unrelated.

Thanks for the feedback. I have removed all unrelated files, corrected variable names, and indented all files properly. Thank you for your patience.

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,75 @@
...
+
+    def setupSpec() {
+        setupProject(SELENIUM_BASIC_PROJECT)
+    }
+
+    def setup() {
+        // Set up WebDriverWait to wait up to 30 seconds for elements that take time to appear or change
+        wait = new WebDriverWait(getDriver(), Duration.ofSeconds(30))
+        (go LoginPage).login(TEST_USER, TEST_PASS)
+        def jobDefinition = JobUtils.generateScheduledExecutionXml("TestJobActivityHistory")
```
mrdubr, line 26: Do you need to generate and execute a job for every test or can this be done once?
If so, this will result in a serious perf gain.

mrdubr (changes requested):
The code itself looks great! 👍 
Let's see if we can address the perf issue commented on and we should be good to go.

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -0,0 +1,75 @@
...
+    WebDriverWait wait
+    String jobId
+
+    def setupSpec() {
+        setupProject(SELENIUM_BASIC_PROJECT)
+    }
+
+    def setup() {
+        // Set up WebDriverWait to wait up to 30 seconds for elements that take time to appear or change
+        wait = new WebDriverWait(getDriver(), Duration.ofSeconds(30))
```
mrdubr, line 24: do you need to do this for every test or just once?

jayas006:
> The code itself looks great! 👍 
> 
> Let's see if we can address the perf issue commented on and we should be good to go.

Thanks Alex,I will fix this so that it's ready for 5.9 release.

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -55,11 +58,10 @@ class JobActivityHistorySpec extends SeleniumBase {
 
         then: "Validate the saved filter is applied"
         wait.until {
-            jobListPage.getAppliedFilterName()=="TestFilter"
+            jobListPage.getAppliedFilterName() == "TestFilter"
```
jayas006, line 61: The code is auto-formatted by the IDE

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
jayas006: > Do you need to generate and execute a job for every test or can this be done once? If so, this will result in a serious perf gain.

I have tried fixing it, can you please check if its look ok? Thanks

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
mrdubr: Looks good!

mrdubr (approved):
Looks great!
Love how compact and elegant your final solution is!
👍

PR #9472, opened 2024-11-27
RUN-2989: Add promoted/highlighted steps in the vue converted pieces for workflow

smartinellibenedetti:
Follow-up of RUN-2725, consuming the exposed parameters in the /plugins/list to properly categorize the step plugins as highlighted/non-highlighted 

Update ChoosePluginModal to be able to handle sorting of the plugins by isHighlighted parameter and then by highlightedOrder, and finally by name.

Also, update the divider's title to match the number of non-highlighted plugins that matches the search query. 

**Additional context**

To see the changes, enable the alphaUi flag in the system config's UI (rundeck.feature.alphaUi.enabled)and the nextUi flag in the theme selector (bottom right corner).

After this PR:

Screenshot 2024-11-27 at 2 31 36 PM


carlosrfranco (approved):
Tested locally and could see the highlighted plugins on the modal:
image

PR #9478, opened 2024-11-29
RUN-3015: Update happy-dom test dependency

smartinellibenedetti:

Fixes https://github.com/rundeck/rundeck/security/dependabot/416

Update tests as needed to comply with update.

**Additional context**

### DIFF
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/resources/tests/NodeDetailSimple.spec.ts
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/resources/tests/NodeDetailSimple.spec.ts
@@ -81,34 +81,27 @@ describe("NodeDetailsSimple Component", () => {
         nodeColumns,
         showExcludeFilterLinks: true,
       });
-      // General attributes
-      const generalAttributes = {
-        description: "Test Node 1",
-        "ui:status:text": "Healthy",
-        "ui:status:icon": "fa-check",
-        osFamily: "Linux",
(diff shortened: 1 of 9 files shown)


gschueler:
note: the happy-dom dependency is used by jest at test time, so this is not a security issue in rundeck server

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/resources/tests/NodeDetailSimple.spec.ts
```diff
@@ -81,34 +81,27 @@ describe("NodeDetailsSimple Component", () => {
...
+      const icon = wrapper.find(`i.fa-check.node-status-icon`);
+      expect(icon.exists()).toBeTruthy();
+
+      const osAttributes = ["Linux", "Ubuntu", "20.04", "x86_64"];
+      const osElements = wrapper.findAll("td[data-test='os-data'] > a");
+      expect(osElements.length).toBe(4);
+      osElements.forEach((osElement) =>
+        expect(osAttributes).toContain(osElement.text()),
+      );
+
```
gschueler, line 100: it seems the attribute value tests 'key3' through 'key6' have not been converted, so the new test does not check for the presence of those values

gschueler: oops, i didn't read enough, looks good

PR #9492, opened 2024-12-06
RUN-2619: Added Docker Commands and Rundeck Restart to BaseContainer Tests

trodriguezpd:
…st the jobqueue when the instance restarts

This is an enhancement, to allow docker restarts on functional tests, specifically for the job queue ones implemented in this ticket:

https://pagerduty.atlassian.net/browse/RUN-2619

I created a restar rundeck instance method in the test containers, and moved the dockerNodeAction to the BaseContainer class so they're shared between the different test suites. This allows to test behaviours like the job queue continues after an instance restart.  

Using the test containers provided methods to restart containers, but I couldn't find ones for the compose test suite

**Additional context**

### DIFF
--- a/functional-test/src/test/groovy/org/rundeck/util/api/responses/execution/Execution.groovy
+++ b/functional-test/src/test/groovy/org/rundeck/util/api/responses/execution/Execution.groovy
@@ -4,7 +4,6 @@ import com.fasterxml.jackson.annotation.JsonAutoDetect
 import com.fasterxml.jackson.annotation.JsonFormat
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties
 import com.fasterxml.jackson.annotation.JsonProperty
-import org.openqa.selenium.json.Json
 import org.rundeck.util.api.responses.jobs.Job
 
 @JsonIgnoreProperties(ignoreUnknown = true)


functional-test/src/test/groovy/org/rundeck/util/container/BaseContainer.groovy
```diff
@@ -518,4 +522,48 @@ abstract class BaseContainer extends Specification implements ClientProvider, Wa
...
+        return RUNDECK_CONTAINER.containerId
+    }
+
+    void waitForRundeckUp(RdClient client){
+        int count = 0
+        def response = null
+        try {
+            response = client.doGet("/system/info")
+        } catch (InvokerInvocationException) {
+            //Here we just ignore the exception since i couldn't find the root cause
```
mrdubr, line 548: In general, swallowing exceptions is a bad practice.
You should `log` them with the appropriate level.
Otherwise, how would we ever know why the `"Rundeck did not start"`

mrdubr (changes requested):
Please, see my comments.

I totally understand that the "container" modelling you have to deal with is rather confusing. However, let's try keep it as clean and improve what we can.

functional-test/src/test/groovy/org/rundeck/util/container/RdContainer.groovy
```diff
@@ -53,6 +54,15 @@ class RdContainer extends ComposeContainer implements ClientProvider {
 
     }
```
mrdubr, line 56: As mentioned above, let's RdContainer to RdComposeContainer.
There will be some ripple effect, but I checked and it won't be that bad.

functional-test/src/test/groovy/org/rundeck/util/container/BaseContainer.groovy
```diff
@@ -31,6 +31,8 @@ abstract class BaseContainer extends Specification implements ClientProvider, Wa
     protected static final String TEST_RUNDECK_URL = System.getenv("TEST_RUNDECK_URL") ?: System.getProperty("TEST_RUNDECK_URL")
     protected static final String TEST_RUNDECK_TOKEN = System.getenv("TEST_RUNDECK_TOKEN") ?: System.getProperty("TEST_RUNDECK_TOKEN", "admintoken")
     protected static final ObjectMapper MAPPER = new ObjectMapper()
+    private static RdDockerContainer RUNDECK_CONTAINER
+    private static RdContainer RUNDECK_COMPOSE_CONTAINER
```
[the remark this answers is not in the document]
trodriguezpd: RdContainer was updated into RdComposeContainer

functional-test/src/test/groovy/org/rundeck/util/container/BaseContainer.groovy
```diff
@@ -76,6 +79,7 @@ abstract class BaseContainer extends Specification implements ClientProvider, Wa
                 )
                 rundeckContainer.start()
                 CLIENT_PROVIDER = rundeckContainer
+                RUNDECK_COMPOSE_CONTAINER = rundeckContainer
```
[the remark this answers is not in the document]
trodriguezpd: Both vars were removed, I replaced them with just the plain container id, which is the only thing we need from the container currently.

Also to note, the ComposeContainer does not provide or have RdDockerContainer instances, we can only get a ContainerState info by searching by service name, but we have no way of returning a RdDockerContainer instance from the compose as they're not actually related as far as my understanding goes.

functional-test/src/test/groovy/org/rundeck/util/container/BaseContainer.groovy
```diff
@@ -518,4 +522,48 @@ abstract class BaseContainer extends Specification implements ClientProvider, Wa
...
+        def process = "docker ${action} ${containerId}".execute()
+        def code = process.waitFor()
+        if (code != 0) {
+            println(process.getText())
+            throw new RuntimeException("Failed to ${action} container ${containerId}")
+        }
+    }
+
+    String getRundeckContainerId(){
+        if(RUNDECK_COMPOSE_CONTAINER){
```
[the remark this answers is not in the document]
trodriguezpd: Thanks for the docs! It triggers me that it looks similar to js ternary 😂, I learned something new. Either way, since I now directly store the container id this method was removed

functional-test/src/test/groovy/org/rundeck/util/container/BaseContainer.groovy
```diff
@@ -518,4 +522,48 @@ abstract class BaseContainer extends Specification implements ClientProvider, Wa
         response.successful ? MAPPER.readValue(response.body().string(), valueType) :  unsuccessfulResponseHandler(response)
     }
 
+
+    void dockerNodeAction(String action, String containerId) {
```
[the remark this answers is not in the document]
trodriguezpd: This was updated and now gets executed with the params of the instance instead of receiving them, also it was renamed

functional-test/src/test/groovy/org/rundeck/util/container/BaseContainer.groovy
```diff
@@ -518,4 +522,48 @@ abstract class BaseContainer extends Specification implements ClientProvider, Wa
...
+    }
+
+    String getRundeckContainerId(){
+        if(RUNDECK_COMPOSE_CONTAINER){
+            return RUNDECK_COMPOSE_CONTAINER.getRundeckContainerId()
+        }
+        return RUNDECK_CONTAINER.containerId
+    }
+
+    void waitForRundeckUp(RdClient client){
```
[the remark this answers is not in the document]
trodriguezpd: Same as on top, it was renamed and it does use the client of the current instance.

functional-test/src/test/groovy/org/rundeck/util/container/BaseContainer.groovy
```diff
@@ -518,4 +522,48 @@ abstract class BaseContainer extends Specification implements ClientProvider, Wa
...
+        int count = 0
```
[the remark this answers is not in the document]
trodriguezpd: This was updated to reuse the wait with retry utils

functional-test/src/test/groovy/org/rundeck/util/container/BaseContainer.groovy
```diff
@@ -518,4 +522,48 @@ abstract class BaseContainer extends Specification implements ClientProvider, Wa
         response.successful ? MAPPER.readValue(response.body().string(), valueType) :  unsuccessfulResponseHandler(response)
     }
 
+
+    void dockerNodeAction(String action, String containerId) {
+        def process = "docker ${action} ${containerId}".execute()
+        def code = process.waitFor()
+        if (code != 0) {
+            println(process.getText())
```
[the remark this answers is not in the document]
trodriguezpd: This was a debugging leftover by accident, but I took the opportunity add both a stderr capture and the process exit code in the exception itself

functional-test/src/test/groovy/org/rundeck/util/container/BaseContainer.groovy
```diff
@@ -518,4 +522,48 @@ abstract class BaseContainer extends Specification implements ClientProvider, Wa
...
+            throw new RuntimeException("Failed to ${action} container ${containerId}")
```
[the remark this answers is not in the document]
trodriguezpd: both the code and an stderr capture were added to the exception text

functional-test/src/test/groovy/org/rundeck/util/container/BaseContainer.groovy
```diff
@@ -518,4 +522,48 @@ abstract class BaseContainer extends Specification implements ClientProvider, Wa
         response.successful ? MAPPER.readValue(response.body().string(), valueType) :  unsuccessfulResponseHandler(response)
     }
 
+
```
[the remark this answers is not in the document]
trodriguezpd: This was caused basically because I was told to move the dockerNodeAction from the Cluster setup into the BaseContainer and it mantained the naming, but now It has a proper naming and works with the instance instead of receiving container params

functional-test/src/test/groovy/org/rundeck/util/container/BaseContainer.groovy
trodriguezpd: This was updated with the wait and retry logic, I tried using the log var which gets defined by the @Slf4j annotation, but due to some lifecycle issues, considering we are restarting containers with a direct process instead of using TestContainers( which does not provide methods for this ) it caused a weird Exception. 

Now I log them by instantiating a logger directly in the catch clause. Also tried the List<Exception> accumulator and logging them after the waitfor, but the issue still popped up.

functional-test/src/test/groovy/org/rundeck/util/container/RdContainer.groovy
trodriguezpd: renamed!

functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
```diff
@@ -361,7 +361,9 @@ class JobUtils {
                 "ode-filter-exclude": "",
                 "dispatch-rank-order": "ascending",
                 "opt1-required": "true",
-                "opt2-required": "true"
+                "opt2-required": "true",
```
mrdubr, line 364: Do you need to document the ability to override these values in the function javadoc?

functional-test/src/test/groovy/org/rundeck/util/container/BaseContainer.groovy
```diff
@@ -518,4 +525,49 @@ abstract class BaseContainer extends Specification implements ClientProvider, Wa
...
+
+    void waitForRundeckAppToBeResponsive(){
+
+        def checkIsRundeckApiResponding = {
+            try {
+                def response = client.doGet("/system/info")
+                return (response != null && response.code() == 200)
+            } catch (Exception e) {
+                LoggerFactory.getLogger(BaseContainer.class).error(e.getMessage())
```
mrdubr, line 551: You do expect to see a few of these log lines in a "normal" flow, right?
In that case, the `error` level is a bit too harsh. 

```suggestion
                LoggerFactory.getLogger(BaseContainer.class).info(e.getMessage())
```

mrdubr (approved):
A few minor concerns. 
See my comments.
Once fixed, no need for another review.

functional-test/src/test/groovy/org/rundeck/util/common/jobs/JobUtils.groovy
trodriguezpd: Both queue-enabled and queue-size are now documented

functional-test/src/test/groovy/org/rundeck/util/common/execution/ExecutionUtils.groovy
```diff
@@ -40,6 +40,18 @@ class ExecutionUtils {
             }
         }
 
+        /**
+         * Returns a boolean that verifies if the executions where executed sequentially,
```
mrdubr, line 44: nit:
In javadoc, you'd specify the return type in the `@return. Also, the best practices prescribe avoiding implementation details (e.g. start-date, end-date) and favour buinsiess meaning.
E.g.
```
Checks if executions were executed sequentially, meaning the execution at a position `n` has ended before the execution at position `n+1` has started.
```

functional-test/src/test/groovy/org/rundeck/util/common/execution/ExecutionUtils.groovy
```diff
@@ -40,6 +40,18 @@ class ExecutionUtils {
...
+         * meaning the first execution date-ended is lesser than the second's start-date.
+         *
+         * @param executions is the list of executions to verify sorted from first executed to last executed.
+         * @return Boolean
+         */
+        static final haveExecutedSequentially(List<Execution> executions) {
```
mrdubr, line 50: The `org.rundeck.util.api.responses.execution.Execution` type does not provide `date-started` or `date-ended` properties. You're relying on the "hidden" jackson deserialization artifacts that are not guaranteed.

mrdubr (changes requested):
Please, see my comments. Most are `nit`s. Some are not.

functional-test/src/test/groovy/org/rundeck/util/common/execution/ExecutionUtils.groovy
```diff
@@ -40,6 +40,18 @@ class ExecutionUtils {
...
+            executions.collate(2, 1, false)
+                    .every { pair -> pair[1]["date-started"]["unixtime"] >= pair[0]["date-ended"]["unixtime"] }
```
mrdubr, line 52: nit:
Only if you could say:
`pair[1].enderBeforeOrAt(pair[0])`
Leaving the knowledge of the start and end dates to the `Execution` class where it belongs ;-)

mrdubr (approved):
LGTM

PR #9477, opened 2024-11-29
RUN-2947: Install pinia

smartinellibenedetti:
**Is this a bugfix, or an enhancement? Please describe.**
Add pinia to the project

**Describe the solution you've implemented**

**Describe alternatives you've considered**

**Additional context**

PR #9494, opened 2024-12-09
RUN-3019: Add pinia, jobs store and refactor editor main

smartinellibenedetti:
Install pinia, add a store for fetching job definition and refactor editor main file so that it's easier to extend it.

**Additional context**


gschueler (approved):

smartinellibenedetti:
> does this obviate #9477 ?

yup, it does. I had broken it into 2 tasks as wasn't sure if we'd get to do it now or just during phase 2, but in the end, I realized that with the store implemented earlier, we can add the context info for autocomplete directly there 😅

PR #9499, opened 2024-12-12
Upgraded netty common version to resolve CVE-2024-47535

trodriguezpd:
Package upgrade to fix netty-common CVE

https://pagerduty.atlassian.net/browse/RUN-2996

Parameterized and upgraded the netty-common package version


ahormazabal (changes requested):
Since netty-common is not part of the core build, i don't think we need to add it here to solve the CVEs in the rundeckpro part. The other PR should be enough

PR #9502, opened 2024-12-16
RPL-246: Added modification to ExpandedJobGroupsSpec

alexander-variacode:
## ⛔️ Ticket
https://pagerduty.atlassian.net/browse/RPL-246

## ✅ Fix description
The name of the main group has been modified to avoid any other test with the same name interfering with the subgroups of the jobs


functional-test/src/test/resources/projects-import/expanded-job-groups/rundeck-ExpandedJobGroupsSpec/jobs/job-02ac10de-459a-405c-bdbf-4b908ea1ddb4.xml
```diff
@@ -3,7 +3,7 @@
     <defaultTab>nodes</defaultTab>
     <description></description>
     <executionEnabled>true</executionEnabled>
-    <group>Parent</group>
+    <group>ParentGroup</group>
```
gschueler, line 6: this shouldn't be necessary to change the parent group name, there is only one test for this spec using the project name "ExpandedJobGroupsSpec" so it shouldn't conflict.  I think the original error is not due to any conflict in the job names or project name.

gschueler (changes requested):

gschueler (approved):
I think you can revert the changed

PR #9511, opened 2024-12-20
RUN-2993: autocomplete

jayas006:

enhancement

**Additional context**

### DIFF
--- a/rundeckapp/grails-app/assets/stylesheets/tokens-rundeck.css
+++ b/rundeckapp/grails-app/assets/stylesheets/tokens-rundeck.css
@@ -1,14 +1,12 @@
 :root {
-    --colors-white: #FFFFFF;
-    --colors-black: #18181B;
     --colors-blue-100:  #DCE6FB;
     --colors-blue-500: #285ED2;
     --colors-blue-600:  #1248BD;
     --colors-gray-100: #FAFAFA;
     --colors-gray-300: #E4E4E7;
-    --colors-grey-500: #A1A1AA;
-    --colors-grey-600: #71717A;
-    --colors-grey-900:  #18181B;
+    --colors-gray-500: #A1A1AA;
+    --colors-gray-600: #71717A;
+    --colors-gray-900:  #18181B;
     --colors-green-100: #D6E4DB;
     --colors-green-600: #149343;
     --colors-orange-100: #FEE8D8;
--- a/rundeckapp/grails-app/assets/stylesheets/tokens-rundeck.css
+++ b/rundeckapp/grails-app/assets/stylesheets/tokens-rundeck.css
@@ -1,12 +1,14 @@
 :root {
+    --colors-white: #FFFFFF;
+    --colors-black: #18181B;
     --colors-blue-100:  #DCE6FB;
     --colors-blue-500: #285ED2;
     --colors-blue-600:  #1248BD;
     --colors-gray-100: #FAFAFA;
     --colors-gray-300: #E4E4E7;
-    --colors-gray-500: #A1A1AA;
-    --colors-gray-600: #71717A;
-    --colors-black:  #18181B;
+    --colors-grey-500: #A1A1AA;
+    --colors-grey-600: #71717A;
+    --colors-grey-900:  #18181B;
     --colors-green-100: #D6E4DB;
     --colors-green-600: #149343;
     --colors-orange-100: #FEE8D8;
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/primeVue/PtAutoComplete/PtAutoComplete.vue
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/primeVue/PtAutoComplete/PtAutoComplete.vue
@@ -1,14 +1,18 @@
 <template>
   <AutoComplete
-    v-model="value"
-    :suggestions="suggestions"
-    @complete="search"
+      v-model="value"
+      :suggestions="suggestions"
+      :optionLabel="optionLabel"
+      :name="name"
+      @complete="onComplete"
+      @change="onChange"
+      @input="updateValue"
   ></AutoComplete>
 </template>
 
 <script lang="ts">
 import { defineComponent } from "vue";
-import AutoComplete, { AutoCompleteCompleteEvent } from "primevue/autocomplete";
+import AutoComplete, { AutoCompleteCompleteEvent, AutoCompleteChangeEvent } from "primevue/autocomplete";
 
 export default defineComponent({
   name: "PtAutoComplete",
@@ -23,11 +27,23 @@ export default defineComponent({
       type: Array,
       default: () => [],
     },
+    defaultValue:{
+      type: String,
+      default: "",
+    },
+    name: {
+      type: String,
+      default: "",
+    },
+    optionLabel: {
+      type: String,
+      default: "label",
+    },
   },
-  emits: ["search"],
+  emits: ["update:modelValue","onChange","onComplete","search"],
   data() {
     return {
-      value: this.modelValue,
+      value: this.modelValue || this.defaultValue,
     };
   },
   watch: {
@@ -36,9 +52,16 @@ export default defineComponent({
     },
   },
   methods: {
-    search(event: AutoCompleteCompleteEvent) {
-      this.$emit("search", event);
+    onComplete(event: AutoCompleteCompleteEvent) {
+      this.$emit("onComplete", event);
     },
+    onChange(event: AutoCompleteChangeEvent) {
+      this.$emit("onChange", event);
+    },
+    updateValue() {
+      this.$emit("update:modelValue", this.value);
+    },
+
   },
 });
 </script>
@@ -46,5 +69,7 @@ export default defineComponent({
 <style lang="scss">
 .p-inputtext {
   background: var(--color-white);
+  //border: 1px solid var(--color-gray-500);
+  //color: var(--color-gray-900);
 }
 </style>
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/primeVue/PtAutoComplete/PtAutoComplete.vue
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/primeVue/PtAutoComplete/PtAutoComplete.vue
@@ -67,9 +67,26 @@ export default defineComponent({
 </script>
 
 <style lang="scss">
-.p-inputtext {
-  background: var(--color-white);
-  //border: 1px solid var(--color-gray-500);
-  //color: var(--color-gray-900);
+.p-autocomplete {
+  /* Default state */
+  .p-inputtext {
+    background-color: var(--colors-white);
+    border: 1px solid var(--colors-gray-500);
+    color: var(--colors-gray-900);
+
+    &::placeholder {
+      color: var(--colors-gray-600);
+    }
+
+    /* Selected/Hover */
+    &:focus {
+      border-color: var(--colors-blue-500);
+    }
+
+    /* Error State */
+    &.p-invalid {
+      border-color: var(--colors-red-500);
+    }
+  }
 }
 </style>
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/primeVue/PtAutoComplete/PtAutoComplete.vue
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/primeVue/PtAutoComplete/PtAutoComplete.vue
@@ -1,18 +1,21 @@
 <template>
   <AutoComplete
-      v-model="value"
-      :suggestions="suggestions"
-      :optionLabel="optionLabel"
-      :name="name"
-      @complete="onComplete"
-      @change="onChange"
-      @input="updateValue"
+    v-model="value"
+    :suggestions="suggestions"
+    :optionLabel="optionLabel"
+    :name="name"
+    @complete="onComplete"
+    @change="onChange"
+    @input="updateValue"
   ></AutoComplete>
 </template>
 
 <script lang="ts">
 import { defineComponent } from "vue";
-import AutoComplete, { AutoCompleteCompleteEvent, AutoCompleteChangeEvent } from "primevue/autocomplete";
+import AutoComplete, {
+  AutoCompleteCompleteEvent,
+  AutoCompleteChangeEvent,
+} from "primevue/autocomplete";
 
 export default defineComponent({
   name: "PtAutoComplete",
@@ -27,7 +30,7 @@ export default defineComponent({
       type: Array,
       default: () => [],
     },
-    defaultValue:{
+    defaultValue: {
       type: String,
       default: "",
     },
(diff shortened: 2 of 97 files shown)


smartinellibenedetti:
@jayas006 could you please rebase this PR? I see that the rundeck-tokens.css is an outdated version

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/primeVue/PtAutoComplete/tests/PtAutoComplete.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+                            <input
+                                class='p-inputtext'
+                                @input='$emit("update:modelValue", $event.target.value)'
+                                @change='$emit("onChange", $event)'
+                                @complete='$emit("complete", "test suggestion")'
+                            />
+                        </div>
+                    `,
+                },
+            },
```
smartinellibenedetti, line 27:
```suggestion
```

Once I pushed the fix for the jest.config, it was just a matter of running `npm install` for that error to disappear. It's safe to remove this stub now.

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/primeVue/PtAutoComplete/tests/PtAutoComplete.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+    it("renders the component and passes props correctly", async () => {
+        const wrapper = await createWrapper();
+        expect(wrapper.props("modelValue")).toBe("initial value");
+        expect(wrapper.props("suggestions")).toEqual([
+            { label: "Option 1", value: 1 },
+            { label: "Option 2", value: 2 },
+        ]);
+        expect(wrapper.props("name")).toBe("test-autocomplete");
+        expect(wrapper.props("optionLabel")).toBe("label");
+    });
```
smartinellibenedetti, line 43: blocker: this test needs to be adjusted to expect that autocomplete is rendering things correctly based on the props received by checking the rendered HTML, as discussed before

smartinellibenedetti (changes requested):
Please ensure that the component's appearance matches what was defined in figma.

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/primeVue/PtAutoComplete/tests/PtAutoComplete.spec.ts
```diff
@@ -0,0 +1,77 @@
+import { mount, config } from "@vue/test-utils";
+import PtAutoComplete from "../PtAutoComplete.vue";
+import AutoComplete from "primevue/autocomplete";
+
+config.global.mocks = {
+  $primevue: {
+    config: {}
+    }
+  };
```
smartinellibenedetti, line 9: question: did you undo the changes in the setupTest because it didn't work with the file being referenced only on setupFiles hook? 

If that was the case, please undo this change, and bring back the changes you had in setupTests to the file jest.global-mocks.js. It was beneficial to load primeVue as a global configuration for the team.

smartinellibenedetti (changes requested):
The component still does not reflect the correct colors as defined in figma. Please provide a screenshot from storybook in this PR once you've done the fixes, making sure that both CSS files with the name tokens-rundeck match their content.

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/primeVue/PtAutoComplete/tests/PtAutoComplete.spec.ts
jayas006: > question: did you undo the changes in the setupTest because it didn't work with the file being referenced only on setupFiles hook?
> 
> If that was the case, please undo this change, and bring back the changes you had in setupTests to the file jest.global-mocks.js. It was beneficial to load primeVue as a global configuration for the team.

I didn't undo your changes. I first made my changes in the setup file. However, after reviewing it I thought I wasn't supposed to modify jest.global-mocks.js or jest.config.js, so I reverted my changes and left the files as it was before. So what's currently in the file is the original one. If you would like to review the changes I made before reverting them, I would happy to do them so that you can compare

jayas006:
> The component still does not reflect the correct colors as defined in figma. Please provide a screenshot from storybook in this PR once you've done the fixes, making sure that both CSS files with the name tokens-rundeck match their content.

Thanks, I will share the screenshot. I have asked this question before that why component is not reflecting the colors defined in Figma. But I will review it one more time and share the screenshot

smartinellibenedetti:
> Thanks, I will share the screenshot. I have asked this question before that why component is not reflecting the colors defined in Figma. But I will review it one more time and share the screenshot

If you're referring to the question asked in slack, I quote my own answer:

>  there are 2 files with the same name tokens-rundeck.css that are currently in different locations (this is temporary for now until we nail the final structure). One of them you've added new tokens in one of the previous tasks you've worked on, but need to check that the other file also has those tokens. Let me know if it doesn't make sense
> the second file was added as part of the PR you're working on right now, so it wasn't possible to have added the original tokens in both places at once

At the time, you had replied that this made sense. So as we discussed previously, please ensure that both files contain the same content and provide a screenshot once corrected, to facilitate the review.

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/primeVue/PtAutoComplete/tests/PtAutoComplete.spec.ts
smartinellibenedetti: Let me clarify my comment as it seems that there's a misunderstanding. My changes weren't affected, I am talking about the changes in this commit: https://github.com/rundeck/rundeck/pull/9511/commits/e5b1ccb1a054d3f1cf34b700d1eece2ab0c79a98

These changes were beneficial for the team, as they would prevent the need for doing additional configuration in each test file.

What I'm recommending is to bring them back, but instead of having them living in the file setupTest.js file, to have them in the jest.global-mocks.js file.

jayas006:
> > Thanks, I will share the screenshot. I have asked this question before that why component is not reflecting the colors defined in Figma. But I will review it one more time and share the screenshot
> 
> If you're referring to the question asked in slack, I quote my own answer:
> 
> > there are 2 files with the same name tokens-rundeck.css that are currently in different locations (this is temporary for now until we nail the final structure). One of them you've added new tokens in one of the previous tasks you've worked on, but need to check that the other file also has those tokens. Let me know if it doesn't make sense
> > the second file was added as part of the PR you're working on right now, so it wasn't possible to have added the original tokens in both places at once
> 
> At the time, you had replied that this made sense. So as we discussed previously, please ensure that both files contain the same content and provide a screenshot once corrected, to facilitate the review.

Thanks for the clarification, I will make sure to check both files and provide a screenshot once corrected

jayas006:
tokens-rundeck.css
tokens-rundeck css

stylesheet/tokens-rundeck.css
Screenshot 2025-02-12 at 6 22 57 PM

jayas006:
@smartinellibenedetti , I have added the screenshot and also updated the test. Please let me know if anything else I need to do. Thank you!

jayas006:
> @smartinellibenedetti ,I have added the screenshots. The only difference I noticed is the spelling of "gray" and "grey." I have changed it to "grey" as it is in Figma. Also, not all content from the stylesheet tokens-rundeck.css is in tokens-rundeck.css. I don't remember the reason for this, but other than that, all content matches. I have also updated the jest.global-mocks file. Please let me know if there is anything else I need to do. Thank you!

smartinellibenedetti:
> > The component still does not reflect the correct colors as defined in figma. Please provide a screenshot from storybook in this PR once you've done the fixes, making sure that both CSS files with the name tokens-rundeck match their content.
> 
> Thanks, I will share the screenshot. I have asked this question before that why component is not reflecting the colors defined in Figma. But I will review it one more time and share the screenshot

Please as discussed before, the screenshot was meant to be of the styled component in storybook, not from the code. Also, I'd like to suggest to review the screenshots of the CSS files already attached, to find discrepancies.

jayas006:
@smartinellibenedetti , Thanks for your patience, Attaching the screenshots from storybook 
Screenshot 2025-02-13 at 1 26 29 PM
Screenshot 2025-02-13 at 1 25 33 PM

rundeckapp/grails-app/assets/stylesheets/tokens-rundeck.css
```diff
@@ -1,12 +1,14 @@
 :root {
+    --colors-white: #FFFFFF;
+    --colors-black: #18181B;
```
smartinellibenedetti, line 3: Please remove these 2 lines as they won't be needed (one of them will be redundant)

rundeckapp/grails-app/assets/stylesheets/tokens-rundeck.css
```diff
@@ -1,12 +1,14 @@
...
     --colors-blue-500: #285ED2;
     --colors-blue-600:  #1248BD;
     --colors-gray-100: #FAFAFA;
     --colors-gray-300: #E4E4E7;
-    --colors-gray-500: #A1A1AA;
-    --colors-gray-600: #71717A;
-    --colors-black:  #18181B;
+    --colors-grey-500: #A1A1AA;
+    --colors-grey-600: #71717A;
+    --colors-grey-900:  #18181B;
```
smartinellibenedetti, line 11: these should have been kept as the original names (gray and black). I know that the figma has "grey" instead of "gray", but we need to keep the names matching what was originally defined by continuum, which is "gray" (theme/tokens.css)

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/primeVue/PtAutoComplete/PtAutoComplete.vue
```diff
@@ -36,15 +55,39 @@ export default defineComponent({
...
+    color: var(--colors-grey-900);
+
+    &::placeholder {
+      color: var(--colors-grey-600);
+    }
+
+    &:hover{
+      border-color: var(--colors-blue-500);
+    }
+
```
smartinellibenedetti, line 86: the border color of blue-500  should also be applied to `:enabled:focus`, along with `box-shadow: 0 0 0 0.2rem var(--colors-blue-100);`.

The box shadow wasn't described within that part of the figma, but during the review I noticed the green outline, that can be addressed within the same selector.

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/primeVue/PtAutoComplete/PtAutoComplete.vue
```diff
@@ -23,11 +30,23 @@ export default defineComponent({
...
+      default: "",
+    },
+    name: {
+      type: String,
+      default: "",
+    },
+    optionLabel: {
+      type: String,
+      default: "label",
+    },
```
smartinellibenedetti, line 44: They weren't in the ticket but noticed during the review that 2 more props are needed: `invalid` which is a boolean, and `placeholder` which is a string. Could you please add them?

rundeckapp/grails-spa/packages/ui-trellis/src/library/theme/tokens-rundeck.css
```diff
@@ -0,0 +1,19 @@
+:root {
```
smartinellibenedetti, line 1: Please make sure that the changes done in the other CSS file are also reflected here.

smartinellibenedetti (changes requested):
The reason screenshots were asked was to allow both the author and the reviewer to validate that the component had the correct appearance. 

For example, upon trying to validate that selecting the input shows the right colors, this is the outcome:
image

In the same way, to see how the component looks when invalid, it was necessary to pass the prop with the same name. While the ticket missed this info and the placeholder props, validating the implemented solution would have caught that.

Lastly, this PR is missing the CSS rules for when the input is focused (it should show a blue border while the user types).  

Please add these changes: the CSS for the focused state and the props invalid and placeholder.

jayas006:
> The reason screenshots were asked was to allow both the author and the reviewer to validate that the component had the correct appearance.
> 
> For example, upon trying to validate that selecting the input shows the right colors, this is the outcome: image
> 
> In the same way, to see how the component looks when invalid, it was necessary to pass the prop with the same name. While the ticket missed this info and the placeholder props, validating the implemented solution would have caught that.
> 
> Lastly, this PR is missing the CSS rules for when the input is focused (it should show a blue border while the user types).
> 
> Please add these changes: the CSS for the focused state and the props invalid and placeholder.

Thanks Sarah, It's so nice to know about it. Thank you for explaining

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/primeVue/PtAutoComplete/PtAutoComplete.vue
jayas006: > They weren't in the ticket but noticed during the review that 2 more props are needed: `invalid` which is a boolean, and `placeholder` which is a string. Could you please add them?

 I am adding it, And I have added placeholder and invalid prop (for invalid it's a :class binding) inside Autocomplete component.

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/primeVue/PtAutoComplete/PtAutoComplete.vue
```diff
@@ -2,13 +2,22 @@
   <AutoComplete
     v-model="value"
     :suggestions="suggestions"
-    @complete="search"
+    :optionLabel="optionLabel"
+    :name="name"
+    :placeholder="placeholder"
+    :class="{ 'p-invalid': invalid }"
```
smartinellibenedetti, line 8:
```suggestion
    :invalid="invalid"
```

The component already has this logic inside of it to add the "p-invalid" class when invalid is true, so all it's required is to pass the prop. If it didn't have this logic, the statement you had added would be completely correct.

smartinellibenedetti (commented):
One tiny fix and it's good to go!

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/primeVue/PtAutoComplete/PtAutoComplete.vue
jayas006: > The component already has this logic inside of it to add the "p-invalid" class when invalid is true, so all it's required is to pass the prop. If it didn't have this logic, the statement you had added would be completely correct.

For learning and understanding, So If invalid is true then simply passing the prop is enough, and no logic is needed. However, if invalid is false then the logic should be added accordingly?

smartinellibenedetti: Not exactly. I'm saying that the component inside primevue already has the logic to add the class "p-invalid" when the prop invalid is true, so all it is required is to pass the prop to it. 

If we had a special case and let's say we wanted to add a completely different class, ie "rundeck-autocomplete" when invalid is true, then the logic you had added would be correct (adjusting the class name of course):
`:class="{ 'rundeck-autocomplete': invalid }"`.

Whenever in doubts if the component has the behaviour already, the primevue documentation is super handy: 
https://primevue.org/autocomplete

jayas006:
> One tiny fix and it's good to go!

Done 👍

PR #9513, opened 2025-01-02
RUN-3016 Fix scm import when file name has quotes

carlosrfranco:

On SCM import plugin, the job is not imported when user tries to import from GUI a file added on source with quotes in file name 

The value attribute on the html page is not built correctly when the file name has quotes. To fix it, the quotes should be replaced by `&quot;`

**Additional context**


rundeckapp/grails-app/views/scm/performAction.gsp
```diff
@@ -255,7 +255,7 @@
                                         <div class="checkbox flex-none"  style="margin-top: 0;margin-left: 20px;">
                                                 <g:checkBox name="chosenTrackedItem"
                                                             id="chosenTrackedItem${trackedItem.id}"
-                                                            value="${trackedItem.id}"
+                                                            value="${trackedItem.id?.replaceAll("\"", '&quot;')}"
```
ahormazabal, line 258: I think we should find a more generic way to encode this as a html-safe string.
This would fix it for quotes, but what if there's another html artifact? It could even lead to a vulnerability.

ahormazabal (changes requested):

PR #9515, opened 2025-01-03
RUN-3049: fixes the nodes targeted with having editable filter false

ronaveva:
This is a bugfix for https://pagerduty.atlassian.net/browse/RUN-3049
When having "Editable Filter" false at job level, rundeck would ignore the selected nodes and it would run the steps against all of the nodes matching the node filter

Rollback change https://github.com/rundeck/rundeck/pull/9459

**Additional context**
"Editable Filter" should not prevent users select/unselect the nodes matches by the node filter, it should only prevent the user from changing the node filter


Jesus-Osuna-M:
Pro branch for testing purpose https://github.com/rundeckpro/rundeckpro/pull/4074

smartinellibenedetti (approved):

PR #9522, opened 2025-01-14
RCLOUD-1557: Reclaim and schedule in smaller transactions

ChuckCrawford:
This PR contains a few required pieces for implementing an improved execution takeover for rundeckpro running in cluster mode. More detail to come soon.

* Adds a new takeover function that commits claimed executions before scheduling them to run on the server.
* The `rundeck.scheduledExecutionService.startup.rescheduleMode` configuration property actually works now.

**Additional context**

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
@@ -1054,6 +1054,7 @@ class ScheduledExecutionService implements ApplicationContextAware, Initializing
      * scheduled execution one-by-one to avoid a number of potential race conditions that can occur when scheduling
      * the job before committing the corresponding ScheduledExecution updates.
      */
+    @NotTransactional
     def reclaimAndScheduleJobByJob() {
         String toServerUuid = frameworkService.getServerUUID()
         Map claimed = [:]
@@ -1072,6 +1073,8 @@ class ScheduledExecutionService implements ApplicationContextAware, Initializing
                 def claimResult = ScheduledExecution.withNewTransaction { status ->
                     return claimScheduledJob(se, toServerUuid)
                 }
+                // Note that the ScheduledExecution in se does not contain the updates from the previous transaction.
+                // Use with care.
 
                 claimed[se.extid] = [
                         success         : claimResult.claimed,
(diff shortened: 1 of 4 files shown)


rundeckapp/grails-app/services/rundeck/services/ScheduledExecutionService.groovy
```diff
@@ -1049,6 +1059,68 @@ class ScheduledExecutionService implements ApplicationContextAware, Initializing
         claimed
     }
 
+    def reclaimAndScheduleJobByJob() {
```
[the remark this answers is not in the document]
ronaveva: maybe as part of the takeoverservice

gschueler, line 1062: by default since the class is marked `@Transactional`, calling this method will create a new outer transaction.  Do you want to mark it `@NotTransactional`?

ChuckCrawford: Marking this method as `@NotTransactional` causes this function to explode with `org.springframework.dao.DataAccessResourceFailureException: Could not obtain current Hibernate Session; nested exception is org.hibernate.HibernateException: No Session found for current thread` when we attempt to `.refresh()` the `ScheduledExecution` between the first and second nested transactions.

I would kind of prefer to leave this as-is since we have done a lot of testing here. Do you see any issue with that?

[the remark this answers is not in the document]
ChuckCrawford: I agree that it is a little strange but the method requires private methods inside of this class in order to work properly. If you would like to try and refactor it I would love to see the ideas.

ChuckCrawford: Just a couple of final notes here: the previous implementation worked as expected since `withNewTransaction` _always_ creates a separate, totally independent transaction using `PROPAGATION_REQUIRES_NEW`. However using `@NotTransactional` as you suggested should use slightly fewer resources so I went ahead and made that change.  

(See https://docs.spring.io/spring-framework/reference/data-access/transaction/declarative/tx-propagation.html#tx-propagation-requires_new for more detail.)

gschueler (approved):

PR #9525, opened 2025-01-16
remove params from resource-level runner acl at system level

hiawvp:

This is a bugfix for [RPL-81](https://pagerduty.atlassian.net/browse/RPL-81): Remove "Params" from resource-level Runner ACL

This PR removes `username` and `roles` params from resource-level runner ACL when creating rules at system level as neither of these are applicable to Runners.

Replace the list containing the params that don't apply with an empty list in the `AuthResource` map used by `AclManagerController`. 

**Additional context**
Runners / Enterprise related change

[RPL-81]: https://pagerduty.atlassian.net/browse/RPL-81?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


alexander-variacode (approved):
LGTM

Issue #9440, opened 2024-11-12
job edit screen is not properly displayed in certain cases


alekunlp:
_Note: Bug Reports not following this template may be closed without feedback._

**Getting Help**
Please use this issue template for reporting identified or suspected **bugs** *only*. For support on configuration or troubleshooting please see [Getting Help](https://docs.rundeck.com/docs/manual/02-getting-help.html#enterprise-support) for the appropriate channels.

**Describe the bug**
We have observed this problem in rundeck 5.5.0 and still reproducible in the latest version we use (v5.7.0-20241021)

When editing jobs where the steps contains long lines (in a global variable "value" field, or a job step's "Arguments" field) the job edition page won't be properly displayed.

The whole screen is offset to the right and you can't just see part of the screen.
Horizontal scrolling is not available.

Zooming out enough (30%, 50%, depending how long the longest line is) will eventually display the whole job.

I have a job with a Global Variable step which "Value" field is 487 characters long where this symptom is particularly annoying.

Removing these long lines will fix the issue.
We tested in chrome, edge, firefox with the same results. 

**My Rundeck detail**

* Rundeck version: [v5.7.0-20241021]
* install type: [choose: rpm]
* OS Name/version: [Oracle Linux Server release 7.9]
* DB Type/version: [mysql: 8.0.40-1.1.el7]

**To Reproduce**
1. Edit a job in the rundeck GUI
2. Go to workflow
3. Observe the offset in the screen moved to the right

**Expected behavior**
There should be a horizontal bar to scroll left/right or the options should be wrapped (rather than single-line)

**Screenshots**
![image](https://github.com/user-attachments/assets/520a54d0-fbe6-4968-b525-ae11de27aebc)

**Desktop (please complete the following information):**
 - OS: [MacOS 14.7 (23H124)]
 - Browser [Firefox, Edge]
 - Version [Firefox: 132.0.1. Edge: 130.0.2849.80]
[reproduce_bug.json](https://github.com/user-attachments/files/17712541/reproduce_bug.json)

**Additional context**
Attaching JSON example of a minimal job showing the problem

MegaDrive68k:
Confirmed on 5.7.0. Thanks for your feedback @alekunlp!

gaddman:
We had a similar issue previously, fixed in the 5.0.0 release: see #8685.

fdevans:
[...]

PR #9529, opened 2025-01-17
RUN-2988: Add job reference to the converted workflow tab

smartinellibenedetti:

Convert the job reference step to Vue and add it to the already converted workflow tab components.

As the job reference step isn't an actual plugin, hardcoded it as a new form component and as a plugin option whenever fetching WorkflowNodeSteps and WorkflowSteps.

The form component connects to the node-filter-input and to a newly created pinia store, that fetches and filters nodes. Connects to the job-list-page, to display the jobs available within the selected project. And updates the later component to use buttons instead of links depending on the props passed, besides emitting via eventBus, when there are no jobs available to display.

Also fix an issue with ui-socket, that was preventing the visible filter from being correctly applied.

A few alternatives:
- Originally, the goal of this ticket was also to consume a different API for fetching nodes, but unfortunately, that API was missing attributes (isLocal, and attributes) that would require changing how the resource format works for nodes, or doing big changes within node-list-embed and its children to handle it;
- Considered to add a new flag to EditPluginModal to disable fetching the provider, in order to have the logic of job reference step inside of it. Discarded due to concern of how big the component would be;
- Considered to try to format the form for job reference in a way that would be compatible with plugin-config. Discarded due to complication of integrating with the node search bar;
- Considered to not create the NodeStorePinia, calling instead the service without storing data and leveraging the existing nodeFilterStore. Discarded due to the intention that the pinia store will make it easier in the long run to update other parts of the project that rely on it;

Also:
- Considered to not have the hardcoded logic of the job ref inside the plugins store. Discarded to avoid sorting the data twice. Open to other suggestions;

**Additional context**


rundeckapp/grails-app/views/scheduledExecution/edit.gsp
```diff
@@ -6,6 +6,10 @@
     <meta name="tabpage" content="jobs"/>
     <meta name="layout" content="base"/>
     <asset:javascript src="static/pages/dynamic-form.js" defer="defer"/>
+    <asset:javascript src="static/pages/job/browse.js" defer="defer"/>
+    <asset:stylesheet href="static/css/pages/job/browse.css" />
+
+    <g:embedJSON id="jobTreeUiMeta" data="[hideActions: true, hideHeader: true]"/>
```
gschueler, line 12: this is missing from `create.gsp`.  however you could just move it to `_jobEditCommon.gsp` instead

gschueler: Job browser doesn't work on "Create" page

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/workflow/JobRefForm.vue
```diff
@@ -0,0 +1,820 @@
...
+    <ui-socket
+      section="job-list-page"
+      location="tree-browser"
+      :socket-data="{
+        showBulkEdit: false,
+        showCreateButton: false,
+        projectToDisplay: editModel.jobref.project,
+        key: editModel.jobref.project,
+      }"
+    />
```
gschueler, line 610: question: why not use the job-list-page component directly?

gschueler (changes requested):
The Job Browser selector has another issue: it shows the Folder icon for groups, and if you click it, it changes the URL without navigating away, and then focuses on the selected group without a way to navigate back.  It should be disabled, or that behavior fixed.

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/workflow/JobRefForm.vue
smartinellibenedetti: That was the intention, but the moment I've used it, I realized that it would require parts of the logic that currently live in the main.ts. I didn't want to duplicate code, but also didn't have bandwidth to extract that logic into a new component, therefore opted to keep using the ui-socket.

smartinellibenedetti:
@gschueler Fixes done:

- disabled folder navigation in the browserGroupItem when allowFolderNavigation is false;
- fixed issue where the job selection wasn't working during job creation (was missing import);
- added a message when the number of results for nodes is bigger than 20;
- node overrides section will be loaded open when editing a step that has this info set (either a filter OR set nodeIntersection value, which matches original behavior);
- added truncation to the list of arguments (https://github.com/rundeck/rundeck/issues/9440);
- improved labels for finding the job via name or uuid (moved the description to inside the label);
- fixed edge case for search: if the user erases the filter previously set, the UI will show back the message that the nodes originally set by the job will be used;

Also, about this comment:

> It seems we have to turn on the "rundeck.feature.alphaUi.enabled=true" to try the new Workflow Editor
> 
> However, do we also have to use nextUi=true in the URL? The checkbox in the Cog menu has been removed to set the Cookie.
> 
> If nextUi is not set, then the Workflow Editor is not shown.
> 
> Originally, I wanted to allow the "next" UI changes to be tested by users with the checkbox. (such as the Options editor). Perhaps we can move Options out from behind the flag. In that case, maybe "alphaUi" should be used only for the Workflow Editor.

I had adjusted in #9535 to have everything behind the nextUI + alpha - both to be able to consolidate the info and also because I didn't see this comment before. I'd like to ask if we can have the workflow tab just behind the nextUI flag, as it is so close to the conclusion.

gschueler (commented):
request: the only change I think would be good is to make the modal for selecting step plugin/job reference wider.  Even with a wide screen, the modal is very narrow

gschueler (approved):

PR #9535, opened 2025-01-24
Run 2726 form validation

smartinellibenedetti:

**Additional context**

Will add info in a bit (might add extra tests as well)


rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/ChoosePluginModal.vue
```diff
@@ -214,6 +215,20 @@ export default defineComponent({
       }
       return "";
     },
+    dateStepType(service: string, name: string) {
```
gschueler, line 218: meant to be "dataStepType"?

smartinellibenedetti: yes, thanks for the catch!

gschueler (approved):

Issue #9544, opened 2025-01-30
When enable-sync is true, call sync with the filepath so that it only syncs that specific file


spaced0gg:
/bin/sh: /var/lib/rundeck/rundeckTmpfs/[2773691-37737343-localhost-dispatch-script.tmp.sh](http://2773691-37737343-localhost-dispatch-script.tmp.sh/): /bin/bash: bad interpreter: Text file busy^

We encountered this issue a few years ago, on various rundeck 3 versions.  Current testing is being performed on 4.14.2.  We are on RHEL7 and use mariadb as our database.  RPM rundeck installation.

We were seeing this issue intermittently on jobs that used Node Step Inline Script steps, that were set to run on the local server node.  These are used to allow us to run a step on the rundeck server per node with access to the node parameters.

The step would randomly fail with the following error message:

/tmp/[2773691-37737343-localhost-dispatch-script.tmp.sh](http://2773691-37737343-localhost-dispatch-script.tmp.sh/): /bin/bash: bad interpreter: Text file busy^

After a lot of back and forth with rundeck support we eventually confirmed that setting enable-sync=true on the localhost node resolved the issue.

After a recent hardware upgrade and mandated security software upgrade on our servers, we have noticed that using enable-sync=true on the rundeck localhost node is causing massive performance issues.  For example. we have a job that runs across around 20 nodes takes around 6 minutes with enable-sync=false, but takes around an hour with enable-sync=true.

After looking at the source code:

https://github.com/rundeck/rundeck/blob/main/core/src/main/java/com/dtolabs/rundeck/core/execution/workflow/steps/node/impl/DefaultScriptFileNodeStepUtils.java

we have been testing with enable-sync=false and file-busy-err-retry=true in hopes that this would prevent the text file busy issue while also disabling the sync command.  This did not wok.

The error message that file-busy-err-retry is looking for is:

    public static final String MESSAGE_ERROR_FILE_BUSY_PATTERN = "Cannot run program.+: error=26.*";

which is similar, but different from our error message:

/bin/sh: /var/lib/rundeck/rundeckTmpfs/[2773691-37737343-localhost-dispatch-script.tmp.sh](http://2773691-37737343-localhost-dispatch-script.tmp.sh/): /bin/bash: bad interpreter: Text file busy^

I also tried setting
   file-copy-destination-dir="/var/lib/rundeck/rundeckTmpfs"

[...]

PR #9546, opened 2025-01-30
RUN-3747: call sync with the filepath to only sync that file when enable sync is set to true

spaced0gg:
Enhancement.  

After a recent hardware upgrade and mandated security software upgrade on our servers, we have noticed that using enable-sync=true on the rundeck localhost node is causing massive performance issues.  For example. we have a job that runs across around 20 nodes takes around 6 minutes with enable-sync=false, but takes around an hour with enable-sync=true.

This change keeps the enable-sync=true behaviour that fixes the "text file busy" issue we were seeing, but keeps the performance on par with enable-sync=false.

see https://github.com/rundeck/rundeck/issues/9544

sync is now called with filepath so that it only syncs that specific file of all outstanding changes, see the sync docs for more info:

https://man7.org/linux/man-pages/man1/sync.1.html

Without the filepath, the sync command synchronizes all cached data for the current user to the permanent memory.

 
see rundeck google group discussion:
https://groups.google.com/g/rundeck-discuss/c/Eqetx33hKgM

I have been testing with enable-sync=false and file-busy-err-retry=true in hopes that this would prevent the text file busy issue while also disabling the sync command.  This did not work.

I also tried setting
   file-copy-destination-dir="/var/lib/rundeck/rundeckTmpfs"

and mounting a tmpfs filesystem.  

I would have thought setting this along with enable-sync=false could workaround our problem, as the tmpfs location would be in memory and the sync shouldn't be needed/have no effect.  This also didn't fix the issue.  We still see intermittent text file busy failures.


fdevans:
Merging to a local branch so we can run CI/CD and will issue a new PR for final review by the Engineering team.  Thanks so much for the contribution.

PR #9552, opened 2025-02-05
RCLOUD-1578: Shrink execution uuid length

JoseOrtiz:

Enhancement. This will improve insert for execution uuid index

Shrink the database definition for uuid present on Execution

**Additional context**

There was a deadlock for the LogFileStorageRequest entity that was tied to Execution being updated


sjrd218 (changes requested):
There is no required pattern for job uuid. It might or might not be an actual uuid. Customers were allowed to put whatever they wanted in this column, so it cannot be assumed that it will conform to any kind of uuid pattern.

PR #9554, opened 2025-02-06
Fixes Maven Publish

ChuckCrawford:

**Additional context**


ChuckCrawford:
> question: Shouldn't milestone be 5.9.0 since this is a fix for the prev release?

I didn't do that because the 5.9.0 fix went into the `release/5.9.x` branch separately. This is the same fix but for 5.10 and beyond.

PR #9557, opened 2025-02-08
RUN-3135: cache some API response in GUI

gschueler:

* Enables caching API response for plugin details in the GUI.
* Improves browser behavior on some pages

**Additional context**

### DIFF
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/library/stores/RootStore.ts
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/library/stores/RootStore.ts
@@ -1,20 +1,21 @@
-import { ExecutionOutputStore, ExecutionOutput } from "./ExecutionOutput";
+import { RundeckClient } from "@rundeck/client";
+import { reactive, UnwrapNestedRefs } from "vue";
+import { api } from "../services/api";
+import { ExecutionOutputStore } from "./ExecutionOutput";
 
 import { JobPageStore } from "./JobPageStore";
-import { WorkflowStore } from "./Workflow";
 import { NavBar } from "./NavBar";
-import { UtilityBar } from "./UtilityBar";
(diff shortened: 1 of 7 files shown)


rundeckapp/grails-spa/packages/ui-trellis/src/library/stores/RootStore.ts
```diff
@@ -52,4 +54,26 @@ export class RootStore {
     this.nodeSourceFile = reactive(new NodeSourceFile(this, client));
     this.jobPageStore = reactive(new JobPageStore());
   }
+
+  /**
+   * Get API response for a path, caching the result
+   * @param path API url path
+   * @returns Promise<any>
+   */
+  api(path: string) {
```
smartinellibenedetti, line 63: question: what do you think about setting a timeout to invalidate the cache so that folks don't need to worry about doing it manually?

Also, a tiny nit: would it be possible to have a different name than API so that it's clear that when using this one, the response will be cached? I have a small concern that with the current name, it might end up being used somewhere that it wasn't supposed to because of the assumption that it's the same behavior as the main API client.

gschueler: I'd suggest we wait until that use case is necessary to add timeout behavior, it didn't seem necessary for these changes.

we can change the name, I guess I was trying to make it as simple as possible. do you have a suggestion? perhaps `getApiCached` would be better

smartinellibenedetti: Makes sense, `getApiCached` sounds good, or perhaps `cachedApi`?

smartinellibenedetti (approved):
Marking the PR approved as I already confirmed the changes

PR #9559, opened 2025-02-11
RCLOUD-1588: Ability to sideload plugins into Docker image.

ChuckCrawford:
This is a small enhancement to allow for "sideloading" external plugins  when using Rundeck in complex containerized environments.

In certain containerized runtimes like Kubernetes, it can be desirable to mount any dynamic portions of each container's filesystem on a separate volume. In these scenarios, [the volume](https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/) is typically empty at the time the container is created, making it difficult to get external plugins into the container.

This solution introduces a `plugins-to-sideload/` directory to Rundeck home. External plugins can be placed here [either by extending the Rundeck image or mounting an external directory]. Plugins in this directory are then copied into Rundeck's external plugin directory `libext` as the container starts up.

* Modifying the Rundeck plugin system to load plugins from multiple directories.
* Changing how Rundeck handles bundled plugins.

Both of these alternatives seemed to require more effort than the chosen solution for such a container-specific use-case.

**Additional context**

### DIFF
--- a/docker/official/Dockerfile
+++ b/docker/official/Dockerfile
@@ -6,9 +6,9 @@ COPY --chown=rundeck:root .build .
 RUN java -jar rundeck.war --installonly
 
 # Directories and permissions
-RUN mkdir -p libext etc var/logs \
+RUN mkdir -p libext etc plugins-to-sideload var/logs \
         # Adjust permissions for OpenShift
-        && chmod -R 0775 libext server user-assets etc var
+        && chmod -R 0775 libext plugins-to-sideload server user-assets etc var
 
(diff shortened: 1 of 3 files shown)


gschueler:
@ChuckCrawford seems fine. suggestion: Maybe use a shorter dir name like "extra-plugins" or something?

jbrookspd:
Naming things is hard of course, length is one thing but maybe something more descriptive like "container-plugins"? I'll go ahead and approve because we can always change the name later.

jbrookspd (approved):

PR #9574, opened 2025-02-19
RUN-3130 Fix Error handler failing when is used on a Job Reference with a workflow step

Jesus-Osuna-M:
**Bug Fix**
When using an error handler with the "continue on failure" option, if the job didn't exist or was deleted, and if the referenced job was of type workflow step, a null pointer exception was thrown.

**Solution**
To resolve this issue, the code was modified to validate that the variable is not null before attempting to access any data from the result. This change prevents the null pointer exception from occurring and addresses the reported bug.

Additionally, two functional tests were added. These tests verify that error handlers function correctly for both scenarios: when the referenced job is a workflow step and when it is a node step.

To test it manually use the job provided on the functional tests, on both cases the referenced job doesn't exist and since it has an error handler it should succeed showing the message added on the error handler

### DIFF
--- a/functional-test/src/test/groovy/org/rundeck/util/container/BaseContainer.groovy
+++ b/functional-test/src/test/groovy/org/rundeck/util/container/BaseContainer.groovy
@@ -16,6 +16,7 @@ import spock.lang.Specification
 import java.text.SimpleDateFormat
 import java.time.Duration
 import java.util.function.Consumer
+import java.util.function.Function
 import java.util.jar.JarEntry
 import java.util.jar.JarOutputStream
 import java.util.jar.Manifest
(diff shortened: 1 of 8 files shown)


ronaveva (changes requested):
Please add tests to the modified method

functional-test/src/test/groovy/org/rundeck/util/container/BaseContainer.groovy
```diff
@@ -570,4 +571,39 @@ abstract class BaseContainer extends Specification implements ClientProvider, Wa
         waitForRundeckAppToBeResponsive()
     }
 
+    List<String> getAllLogs(String execId, Function<Map, String> paramGen, int max=20) {
```
ronaveva, line 574: this method also exist in "ExecutionOutputSpec" class and "ProBaseContainer" class. It would be good to mark those as deprecated and just use this new one instead

ronaveva (approved):
LGTM

PR #9583, opened 2025-02-27
RCLOUD-1560: Adds resilience to new takeover method.

ChuckCrawford:
This is a follow up to add additional resilience to the takeover mechanism implemented in https://github.com/rundeck/rundeck/pull/9522.

This change surrounds the "claim job" and "schedule job" portions of the new takeover method with retries.

In the current implementation of the `reclaimAndScheduleJobByJob` function, the individual operations must be retried individually because they are committed individually. For example, if the "claim job" portion of the takeover succeeds, retrying the entire takeover would miss this job in the "find jobs to takeover" stage.

An alternate approach to this problem would be to implement a takeover with an inverted order of operations (schedule job -> claim job) along with a change that stops a newly scheduled job from unscheduling itself if it triggers before claim job commits.

**Additional context**

* I have tested that this works properly locally by:
  * Manually locking a Scheduled Execution record to force the "claim job" retries to kick in.
  * Manually deleting a Misfire Tracker record causing the "schedule job" retries to kick in.

### DIFF
--- a/rundeckapp/src/integration-test/groovy/rundeck/services/ScheduledExecutionServiceIntegrationSpec.groovy
+++ b/rundeckapp/src/integration-test/groovy/rundeck/services/ScheduledExecutionServiceIntegrationSpec.groovy
@@ -48,6 +48,43 @@ class ScheduledExecutionServiceIntegrationSpec extends Specification {
         ] + overrides
     }
 
+    def "reclaimAndScheduleJobByJob successfully moves a scheduled job to the new current server"() {
+        given:
+        def project = 'testProject'
+        def seOneId = UUID.randomUUID().toString()
+
+        def se = ScheduledExecution.withNewTransaction {
+            def workflow = new Workflow(commands: []).save(flush: true, failOnError: true)
+            return new ScheduledExecution(
+                    jobName: 'callisto-one',
+                    groupPath: 'group/reclaimAndScheduleJobByJob',
+                    uuid: seOneId,
+                    serverNodeUUID: TEST_UUID1,
+                    project: project,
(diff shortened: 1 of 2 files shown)


ChuckCrawford:
Follow up of https://github.com/rundeck/rundeck/pull/9522

rundeckapp/src/integration-test/groovy/rundeck/services/ScheduledExecutionServiceIntegrationSpec.groovy
```diff
@@ -48,6 +48,43 @@ class ScheduledExecutionServiceIntegrationSpec extends Specification {
...
+        }
+        service.jobSchedulesService = Mock(JobSchedulesService) {
+            1 * getSchedulesJobToClaim(TEST_UUID2, null, true, null, null) >> [se]
+        }
+
+        when:
+        def results = service.reclaimAndScheduleJobByJob()
+
+        then:
+        results[seOneId]["success"] == true
```
gschueler, line 85: suggestion: verify the new serverNodeUUID value for the job

gschueler (approved):

rundeckapp/src/integration-test/groovy/rundeck/services/ScheduledExecutionServiceIntegrationSpec.groovy
ChuckCrawford: https://github.com/rundeck/rundeck/pull/9583/commits/90c9da300546cbbb7971a2149552c8dfdd059f39

PR #9590, opened 2025-03-05
RUN-2240 Make public key download from GUI disabled as default

carlosrfranco:

Bugfix to prevent a security finding where an user can upload malicious content to the key storage what could be spread to other systems

From what was previously reviewed on this issue, our mitigations for this case is:
- As stated in their report, the customer should set up an ICAP-enabled proxy to filter and scan file upload requests before they hit the rundeck server; this alternative would be significantly simpler than adding scanning features to rundeck, as most popular proxy servers support ICAP.
- Restrict access to file upload features using access control.
- "Download" option can be disabled by configuration: https://docs.rundeck.com/docs/administration/configuration/gui-customization.html#rundeck-gui-keystorage-downloadenabled and this PR makes this configuration disabled by default
- Add a new feature flag `apiPublicKeysDownload` to disable Key Storage content retrieval at API level (it is disabled by default)

**Additional context**

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/config/RundeckConfigBase.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/config/RundeckConfigBase.java
@@ -459,6 +459,7 @@ public class RundeckConfigBase {
         Enabled nodeExecutorSecureInput = new Enabled();
         Enabled alphaUi = new Enabled();
         Enabled enhancedJobTakeoverQuery = new Enabled();
+        Enabled apiPublicKeysDownload = new Enabled();
 
 
         @Data
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/StorageController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/StorageController.groovy
@@ -199,6 +199,11 @@ class StorageController extends ControllerBase{
             return renderError("unauthorized")
         }
         if((askedForContent || anyContent) && !maskContent) {
+            //check if download is enabled by feature flag
+            if(!forceDownload && !configurationService.getBoolean("feature.apiPublicKeysDownload.enabled", false)){
+                response.status=403
+                return renderError("download is not enabled")
+            }
             response.contentType=resContentType
             if(forceDownload){
                 def filename= resource.path.name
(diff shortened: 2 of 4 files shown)


core/src/main/java/com/dtolabs/rundeck/core/config/RundeckConfigBase.java
```diff
@@ -459,6 +459,7 @@ public static class RundeckFeatureConfig {
         Enabled nodeExecutorSecureInput = new Enabled();
         Enabled alphaUi = new Enabled();
         Enabled enhancedJobTakeoverQuery = new Enabled();
+        Enabled apiPublicKeysDownload = new Enabled();
```
gschueler, line 462: new feature flags should also be added to the docker remco files

rundeckapp/grails-app/controllers/rundeck/controllers/StorageController.groovy
```diff
@@ -199,6 +199,11 @@ class StorageController extends ControllerBase{
             return renderError("unauthorized")
         }
         if((askedForContent || anyContent) && !maskContent) {
+            //check if download is enabled by feature flag for api public keys retrieval
+            if(!forceDownload && request.method == "GET" && !configurationService.getBoolean("feature.apiPublicKeysDownload.enabled", false)){
```
gschueler, line 203: I think could just do this:
```
boolean downloadEnabled = configurationService.getBoolean("feature.apiPublicKeysDownload.enabled", false)
if(downloadEnabled && (askedForContent || anyContent) && !maskContent) {
```

fyi, `forceDownload` just means force browser to download a file by using content-disposition header.

gschueler (changes requested):

rundeckapp/grails-app/controllers/rundeck/controllers/StorageController.groovy
carlosrfranco: Based [on this test](https://app.circleci.com/pipelines/github/rundeck/rundeck/17326/workflows/21a55519-8147-4bdb-9d1c-1d60b9fe04b2/jobs/200038/tests), it is also called for POST request. Should this feature flag block for POST too?

gschueler: @carlosrfranco yes, any request method

core/src/main/java/com/dtolabs/rundeck/core/config/RundeckConfigBase.java
gschueler: scratch that, remco uses a generic way to set feature flags!

gschueler (commented):
* The key storage Vue UI maybe is not receiving the "downloadenabled" setting from storage.gsp.  I can see it is false in the HTML, but the GUI still shows the download Link.
* The setting used by the GUI should be && with the new feature flag

gschueler (approved):

PR #9593, opened 2025-03-10
Add project key deletion request to function test's BaseContainer

juliannagreen1:
### Is this a bugfix, or an enhancement? Please describe.

This is an enhancement to better allow testing of key storage in selenium tests.

### Describe the solution you've implemented

I've added a data-testid attribute to the key directory button for better discoverability in test suites and added the key deletion request to the selenium BaseContainer class.

### Describe alternatives you've considered

I could have defined the key deletion request directly in a test but making it available for wider use seems wise, especially since there are similar functions in this class already.

### Additional context

N/A


smartinellibenedetti (approved):
A well-documented util is always a delight to see 😌

PR #9647, opened 2025-05-02
RUN-3422: fix: blank string value for "Options" property type causes Exception

gschueler:

If an Option type PluginProperty is configured with a blank string value, it causes an exception.

**Additional context**


jayas006 (approved):
LGTM 🎉

PR #9674, opened 2025-05-27
RUN-3202: Include Last Login on API Responses

jayas006:

Enhancement

In `UserController.groovy`, applied a condition so the `lastLogin` field is only shown when the API version is 53 or higher.

**Additional context**

### DIFF
--- a/rundeckapp/src/main/groovy/com/dtolabs/rundeck/app/api/ApiVersions.groovy
+++ b/rundeckapp/src/main/groovy/com/dtolabs/rundeck/app/api/ApiVersions.groovy
@@ -50,9 +50,9 @@ class ApiVersions {
      * Current API version Configuration
      */
     // References the current API version
-    public final static int API_CURRENT_VERSION = V52
+    public final static int API_CURRENT_VERSION = V53
     // Hardcoded inline string constant for the current version used in API doc generation
-    public final static String API_CURRENT_VERSION_STR = "52"
+    public final static String API_CURRENT_VERSION_STR = "53"
 
     /**
      * Version span configurations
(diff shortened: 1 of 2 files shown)


rundeckapp/src/main/groovy/com/dtolabs/rundeck/app/api/ApiVersions.groovy
```diff
@@ -40,6 +40,7 @@ class ApiVersions {
     public static final int V50 = 50
     public static final int V51 = 51
     public static final int V52 = 52
+    public static final int V53 = 53
```
carlosrfranco, line 44: Since you added a new api version, the final string API_CURRENT_VERSION and API_CURRENT_VERSION_STR should also be updated to 53. I've seen that you reverted this to fix some tests failing but probably the issue with the tests was because you also need to update the current version used to build the requests from tests here: https://github.com/rundeck/rundeck/blob/4a969f7a388dcfe9a2e23252a058b2c9e4a03ec1/functional-test/src/test/groovy/org/rundeck/util/container/RdClient.groovy#L30

carlosrfranco (changes requested):

rundeckapp/src/main/groovy/com/dtolabs/rundeck/app/api/ApiVersions.groovy
jayas006: > Since you added a new api version, the final string API_CURRENT_VERSION and API_CURRENT_VERSION_STR should also be updated to 53. I've seen that you reverted this to fix some tests failing but probably the issue with the tests was because you also need to update the current version used to build the requests from tests here:
> 
> https://github.com/rundeck/rundeck/blob/4a969f7a388dcfe9a2e23252a058b2c9e4a03ec1/functional-test/src/test/groovy/org/rundeck/util/container/RdClient.groovy#L30

Thanks Carlos for pointing it out.

PR #9692, opened 2025-06-25
RUN-3335 Restart api call should be disabled if user have AppAdmin

Jesus-Osuna-M:
### What does this PR do?

This PR addresses [RUN-3335](https://jira.atlassian.com/browse/RUN-3335): **Restart API call should be disabled if user has AppAdmin**.

- Disables the "Restart" API call for users who have the AppAdmin role.
- Adds validation in the API endpoint to check user roles before allowing the restart action.
- Updates related tests to reflect the change in access permissions.

### Why is this needed?

Previously, users with the AppAdmin role were able to perform restart actions via the API, which is not intended behavior. This change ensures that only authorized roles have access to the restart functionality, improving security and compliance with permission requirements.

### How was this tested?

- Unit tests updated and passing.
- Manual verification: confirmed that users with AppAdmin cannot trigger the restart API, while other authorized roles still can.

### Additional Notes

- Please review the changes to the role validation logic.
- No breaking changes expected for existing users without the AppAdmin role.

[RUN-3335]: https://pagerduty.atlassian.net/browse/RUN-3335?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rundeckapp/grails-app/controllers/rundeck/interceptors/PostApiTokenInterceptor.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/interceptors/PostApiTokenInterceptor.groovy
@@ -16,6 +16,7 @@ class PostApiTokenInterceptor {
             session.user=null
             request.subject=null
             session.subject=null
+            session.invalidate()
         }
         return true
     }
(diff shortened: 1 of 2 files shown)


ronaveva (approved):
LGTM

PR #9714, opened 2025-07-17
RUN-3433 Avoid create multiple execution report if notification fails

carlosrfranco:

When a job finishes, it creates a execution report and, in case this process fails, it will retry a couple of times. Although, in the same process it will also trigger the notification process after execution report is saved and the transaction commited. If the notification fails, it will retry all the process again and save multiple execution reports in the database

Separate the notification process from the execution report save method so that, in case only notification fails, it will retry to trigger only notification event.

**Additional context**

### DIFF
--- a/rundeckapp/src/integration-test/groovy/rundeck/quartzjobs/ExecutionJobIntegrationSpec.groovy
+++ b/rundeckapp/src/integration-test/groovy/rundeck/quartzjobs/ExecutionJobIntegrationSpec.groovy
@@ -19,6 +19,7 @@ import rundeck.services.FrameworkService
 import rundeck.services.JobSchedulerService
 import rundeck.services.JobSchedulesService
 import rundeck.services.MissingScheduledExecutionException
+import rundeck.services.events.ExecutionCompleteEvent
 import rundeck.services.execution.ThresholdValue
 import spock.lang.Specification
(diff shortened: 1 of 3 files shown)


rundeckapp/src/integration-test/groovy/rundeck/quartzjobs/ExecutionJobIntegrationSpec.groovy
```diff
@@ -238,6 +240,59 @@ class ExecutionJobIntegrationSpec extends Specification {
...
+        }, _, _
+        ) >> {
+            fail3times.call()
+            return new ExecutionCompleteEvent()
+        }
+
+        def fail5times = throwXTimes(5)
+        4 * mockes.triggerJobCompleteNotifications(_, _) >> {
+            fail5times.call()
+        }
```
gschueler, line 271: this seems to simulate: saveExecutionState fails 3 times, then succeeds, then triggerJobCompleteNotifications fails 4 times.  

I think there should be a test around the original scenario of the bug:

* if saveExecutionState does not fail, but triggerJobCompleteNotifications does, then saveExecutionState should only run 1 time.

gschueler (changes requested):
1 test case should be added

gschueler (approved):

PR #9742, opened 2025-07-30
RUN-3564: Additional Fixes for CVE-2025-48976

fdevans:
**Summary**
Identified and fixed the commons-fileupload vulnerability issue comprehensively across the Rundeck workspace. Here's what was done:

**Root Cause**
The vulnerability was coming from transitive dependencies through the Grails framework, specifically through org.grails:grails-web-fileupload:6.1.2 which was pulling in the vulnerable commons-fileupload:1.5.

**Changes Made**
build.gradle - Added constraint to force commons-fileupload:1.6.0 for all transitive dependencies build.gradle - Added the same constraint (this was specifically mentioned in your Snyk finding) build.gradle - Added constraint since it uses grails-web-boot build.gradle - Added constraint since it uses grails-web-boot build.gradle (root) - Added global force resolution strategy to ensure all subprojects use the secure version What This Fixes
Forces all transitive dependencies of commons-fileupload to use version 1.6.0 instead of the vulnerable 1.5 Addresses CVE-2025-48976 across all Grails modules that might pull in the vulnerable dependency Provides both local constraints and global force strategy for comprehensive coverage The dependency tree now shows that the vulnerable version 1.5 is being consistently upgraded to 1.6.0 across all dependency paths, which should resolve the CVE.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -141,7 +141,7 @@ allprojects {
     configurations.all {
         resolutionStrategy {
             force "commons-io:commons-io:${commonsIoVersion}"
-            force "commons-fileupload:${commonsFileuploadVersion}"
+            force "commons-fileupload:commons-fileupload:${commonsFileuploadVersion}"
         }
     }
 }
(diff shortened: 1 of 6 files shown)


Jesus-Osuna-M (approved):

ronaveva (approved):
LGTM

Issue #9655, opened 2025-05-06
Crtitical issue with node filters


xavier8854:
**Describe the bug**

One of my users reported this (awkwardly translated from the french ticked he filed to me)

We noticed that node filters do not appear in interface. This problem is present in PagerDuty Automation as well in Rundeck community. I already filed a ticket to the PägerDuty support, but perhaps someone here has some ideas.

Although a job has a node filter defined in a job reference (tab "override node filter") when you edit the job, it does not appear anymore in the "node filter" field. Critical impact is that if you save the job without setting again the node filter, Rundeck considers the former as empty, and removes the previously set node filter

That means that you can inadvertently remove node filters. Most, you cannot see the previously defined node filter, you have to download the job definition (in XML, eg.), where you can clearly see the node filter in the job ref tag. (see attached job definition)

More, you can see that Rundeck detects the presence of the node filter in the job ref, as although the node filter appears as empty, the tab "override node filters" is opened while it is normally closed when no filter is set. (see attached screenshot)

**My Rundeck detail**

* Rundeck version: 5.11.1
* install type: rpm
* OS Name/version: RHEL9
* DB Type/version: mysql
**To Reproduce**
1. Edit a job
2. Click on Override Node Filters
3. See that previosuly defined node filters has disappeared 

**Expected behavior**
Keep defined node filters

**Screenshots**

![Image](https://github.com/user-attachments/assets/0c1f3b44-2c35-41c2-8340-cc856a17e4a3)

**Desktop (please complete the following information):**
 - OS: Any
 - Browser any
 - Version any

**Additional context**
A job definition with the problem

[PP_MESIRH_A000_TM_ImportRenoirhRudaAvecRepriseAnnuleRemplace.xml.txt](https://github.com/user-attachments/files/20053701/PP_MESIRH_A000_TM_ImportRenoirhRudaAvecRepriseAnnuleRemplace.xml.txt)

MegaDrive68k:
Confirmed. Thanks for your feedback, Xavier!

xeon1400:
Are there any Updates to this topic ?

vblikstad:
I also have the same issue.

PR #9749, opened 2025-08-06
RUN-3462+RUN-3686: Fix node filter input not updating value correctly

smartinellibenedetti:
Node filter input wasn't correctly updating its value, causing issues in the following pages: 
- job edit (RUN-3686) - upon editing a job that has a job ref step, the node filter set in the step wouldn't be populated;
- running a job (RUN-3462) - upon trying to override the node filters for a job that allows it. Wouldn't update correctly either with a new filter nor if user selected a tag within a node;
- command page - Same as above, wouldn't update correctly either with a new filter nor if user selected a tag within a node;

Changed the syntax from v-model to value + added an immediate watcher on the value. There's a second PR on rundeckpro.

**Additional context**


jayas006 (approved):

fdevans:
fixes #9655

smartinellibenedetti:
Update: this PR is being excluded from 5.15's release notes as the fix was included in 5.14.1

PR #9750, opened 2025-08-07
RUN-2921: Add index for retry_execution_id in the execution table

carlosrfranco:

Customer is experiencing deadlock issues when using oracle database. It can be affected by cleaning execution job that affect the performance in the database causing the deadlocks

It adds a new index in the execution database for retry_execution_id that improves the performance when rundeck triggers the clean execution job and, as a side effect, this reduces deadlock cases in the database.
In a local test using postgresql, without the index, the cleaning process took 8 seconds to delete +/- 550  execution. After adding the index, it was reduced to 2 seconds.

**Additional context**


ronaveva (approved):
LGTM

PR #9761, opened 2025-08-13
RUN-3707: Add Title format validation to Checks

fdevans:
PR Titles need to start with “RUN-####:” but many are having to be fixed.  
Adding check to GH workflows.


smartinellibenedetti:
Btw, do you have a preference for cases when 1 PR = 2 tickets?

smartinellibenedetti (approved):

PR #9775, opened 2025-08-20
RUN-2430: Back previous behavior for node filtering combining two filters by clicking on it

carlosrfranco:

When we have a lot of nodes, and we want to filter by adding more tags automatically by clicking on the node tags, this is not possible on Rundeck 5.1.0. The new tags added automatically by clicking on them, are overwriting the existing one, and not adding them sequentially to get a filter more granular. This was working fine on 4.17.3.

Back the previous behavior to add filters sequentially when clicking on them.

**Additional context**


smartinellibenedetti (approved):
:shipit:

PR #9787, opened 2025-08-27
[RUN-3754] exclude providers based on a plugin metadata

ltamaster:

Enhancement. Adds a mechanism for plugin authors to mark a provider as **internal-use-only**, so it is excluded from the provider lists shown in the GUI/API (e.g. the dropdowns used to pick a default node executor, file copier, etc.), while it remains fully resolvable by name for direct/programmatic use.

This is useful for providers that exist to be used internally by another component (for example, an orchestrator or wrapper plugin that delegates to a different provider under the hood) and were never meant to be an option an end user picks directly — but today show up in every provider list because there was no way to hide them.

- Added `ExecutionEnvironmentConstants.INTERNAL_USE_ONLY`, a new plugin metadata key.
- Plugin authors can set it via `@PluginMetadata(key = ExecutionEnvironmentConstants.INTERNAL_USE_ONLY, value = "true")` on a provider class, or via any other mechanism that populates `Description.metadata`.
- `RundeckPluginRegistry.listPluginDescriptors(...)` now checks this metadata and excludes the provider from its results when set to `true`. This is the enumeration path used to build GUI/API provider listings; direct provider lookup by name (`loadPluginByName`, used when a provider is actually invoked) is unaffected, so marking a provider internal-use-only only hides it from listings — it does not disable it.
- The check is applied to both provider sources handled by this method: Spring/Groovy bean plugins and jar/zip plugins resolved via `service.listDescriptions()`.
- Added/updated unit tests in `RundeckPluginRegistrySpec` covering both exclusion paths.

Filtering by a hardcoded list of provider names in each place a list is rendered — rejected because it doesn't generalize and would need to be kept in sync across every consumer of the plugin registry.

**Additional context**

This mechanism is used by a downstream (commercial) plugin to hide providers that are only meant to be used internally by that plugin's own orchestration and were causing misconfiguration issues when selectable as a project default.


ronaveva (approved):
LGTM

PR #9805, opened 2025-09-04
RUN-3435: separated query from the Run Later Job for the Job Takeover process

ltamaster:
Separated the query from the Run Later Job for the Job Takeover process to improve the performance of the process

**Additional context**

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/schedule/SchedulesManager.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/schedule/SchedulesManager.java
@@ -100,6 +100,14 @@ public interface SchedulesManager {
     List getSchedulesJobToClaim(String toServerUUID, String fromServerUUID, boolean selectAll, String projectFilter, List<String> jobids);
 
 
+    /**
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/dtolabs/rundeck/core/schedule/SchedulesManager.java
```diff
@@ -99,6 +99,17 @@ public interface SchedulesManager {
...
+
+    /**
+     * Gets a list of jobs that has a schedule for execution later than now
+     * @param toServerUUID
+     * @param fromServerUUID
+     * @param selectAll
+     * @param projectFilter
+     * @return List
+     */
+    List getSchedulesExecutionLater(String toServerUUID, String fromServerUUID, boolean selectAll, String projectFilter);
```
ronaveva: i dont see a reason for this method to have that many params, it looks like the only one needed is the first one "toServerUUID"

ronaveva (approved):
LGTM, there is just a small observation that needs to be solved before merging it

PR #9812, opened 2025-09-12
RUN-3717: improve refexecs

ronaveva:
This pull request introduces a new `rundeck-data-models` module to the project, transitioning from using an external dependency to an internal project module. It adds a set of Java interfaces and supporting classes that define core data models for authentication tokens, executions, and jobs in Rundeck. Additionally, the build configuration is updated to support this new module.

The most important changes are:

**Build and Dependency Management:**
- Switched the dependency in `core/build.gradle` from the external `org.rundeck:rundeck-data-models:1.0.6` to the new local `:rundeck-data-models` project module, ensuring the project uses the internal version going forward.
- Added a new `rundeck-data-models/build.gradle` file, configuring it as a Java/Groovy library with Lombok support and JUnit platform for testing.

**Referenced executions query change:**
With the change to GormReferencedExecutionDataProvider.groovy class we dont need to iterate over the uuids list result to get the referenced executions of an specific job, we can now just use the database id directly as a list, this will improve the job/show loading time and it will also take into account the referenced executions when doing the pagination.

### DIFF
--- a/rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormExecReportDataProvider.groovy
+++ b/rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormExecReportDataProvider.groovy
@@ -267,8 +267,10 @@ class GormExecReportDataProvider implements ExecReportDataProvider, DBExecReport
                         if(execsIds && execsIds.size() > 0){
                             and{
                                 List execsIdPartitioned = Lists.partition(execsIds, 10000)
-                                for(def execIdPartition : execsIdPartitioned){
-                                    'in'('executionId', execIdPartition)
+                                or {
+                                    for (def execIdPartition : execsIdPartitioned) {
+                                        'in'('executionId', execIdPartition)
+                                    }
                                 }
                                 List execProjectsPartitioned = Lists.partition(query.execProjects, 1000)
                                 or{


rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormExecReportDataProvider.groovy
```diff
@@ -97,6 +97,16 @@ class GormExecReportDataProvider implements ExecReportDataProvider, DBExecReport
         }
     }
 
+    @Override
+    int countExecutionReportsWithTransaction(RdExecQuery query, boolean isJobs, String jobId, List<Long> execsId) {
+        return ExecReport.withTransaction {
+            ExecReport.createCriteria().count {
+                applyExecutionCriteria(query, delegate, isJobs, jobId, [])
+                applyExecutionCriteria(query, delegate, isJobs, jobId, execsId)
```
gschueler: why do this with an empty list and then with the execution Ids?

ronaveva: we dont, i'll fix it

rundeckapp/src/main/groovy/org/rundeck/app/data/providers/GormExecReportDataProvider.groovy
```diff
@@ -255,9 +264,12 @@ class GormExecReportDataProvider implements ExecReportDataProvider, DBExecReport
...
                 if (query.execProjects && seId) {
                     or {
-                        if(execUuids && execUuids.size() > 0){
+                        if(execsIds && execsIds.size() > 0){
                             and{
-                                'in'('executionUuid', execUuids)
+                                List execsIdPartitioned = Lists.partition(execsIds, 10000)
+                                for(def execIdPartition : execsIdPartitioned){
+                                    'in'('executionId', execIdPartition)
+                                }
```
gschueler, line 274: Do you need to wrap the 'in' sections in an `or{ }`? otherwise won't it will basically be like

```
and{
 'in'('executionId',[...])
 'in'('executionId',[...])
 'in'('executionId',[...])
}
```
which would not work

PR #9822, opened 2025-09-24
RUN-632: Multiline Job Options (WIP)

gschueler:

Adds support for Multiline Job Options as a new choice in the "Option Type" dropdown.

(Beta)

**Additional context**


rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/options/tests/OptionItem.spec.ts
```diff
@@ -43,9 +43,9 @@ describe("OptionItem", () => {
       editable: true,
     });
 
-    let findAll = wrapper.findAll(".btn");
+    const findAll = wrapper.findAll(".btn");
     expect(findAll.length).toBe(6);
-    expect(wrapper.findAll(".btn-group").length).toBe(2);
+    expect(wrapper.findAll("btn-group").length).toBe(2);
```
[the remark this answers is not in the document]
gschueler: matcher is finding the component not the css class

smartinellibenedetti (commented):
Few things I've noticed:

- upon disabling the multilineJobOptions: 
   - both nextUi and currentUi will still render the field as multiline in the job in the job/show page;
  
   - but upon editing the multiline option:
      - nextUi will show the multiline icon, but upon edit, the option type is empty, while still rendering the defaultValue as a multiline field;

Image

   - currentUi will show the multilineicon, but upon edit, the option type is text, and the string won't have any line breaks;
   
Image

I did a test to see if the enforced setting would persist upon re-enabling the multilineJobOption flag, it didn't, but the list of values persisted and I couldn't get the option to work anymore despite changing the value to text, saving, and then changing it back to multiline again. Also duplicating the field as multiline also duplicated the list property. The only solution was to delete the option and recreate it completely.

- also a small nit: the multilineJobOptions isn't showing up in the system config gui - need to manually add the flag;

I don't mind to approve now and leave the fixes for a separate PR as it's clear in the docs that it's a beta feature, but might be worth to add a note there that if people decide to disable it for any reason, it might break some jobs. Please let me know what you prefer!

rundeckapp/src/main/groovy/org/rundeck/app/config/FeatureFlagConfigurable.groovy
```diff
@@ -0,0 +1,37 @@
...
+@CompileStatic
+class FeatureFlagConfigurable implements SystemConfigurable {
+    List<SysConfigProp> systemConfigProps = [
+        featureConfig(
+            Features.MULTILINE_JOB_OPTIONS,
+            "Multiline Job Options (Beta)",
+            "(Beta Feature) Enable support for multiline job options in job definitions and the GUI.",
+            'app_admin'
+        ),
+        //TODO: include additional feature flags here
```
smartinellibenedetti, line 18: question: is the idea to concentrate all feature flags that we might want to remove at some point here?

smartinellibenedetti (approved):
Fixes confirmed :shipit:

rundeckapp/src/main/groovy/org/rundeck/app/config/FeatureFlagConfigurable.groovy
gschueler: I meant for this new class to eventually hold code definitions for rundeck feature flags, that are currently only described in rundeckpro configmeta

PR #9826, opened 2025-10-02
Improve log format escaping memory usage

gschueler:
avoid regular expression matching

**Additional context**


core/src/main/java/com/dtolabs/rundeck/core/logging/internal/RundeckLogFormat.java
```diff
@@ -117,25 +117,44 @@ public String outputEvent(LogEvent entry) {
...
         //mesg
-        sb.append(backslashEscape(dMesg, DELIM));
+        appendEscaped(sb, dMesg, DELIM);
         //end
         sb.append(DELIM);
 
         return sb.toString();
     }
 
     static String backslashEscape(String dMesg, String chars) {
```
ltamaster: It appears that `backslashEscape` is no longer used. Can we clean that up?

gschueler: I left it because the tests call that method, so it was simpler to do it that way.  We can deprecate it, or follow up with removing it and changing the tests

gschueler: moved it to the test

PR #9873, opened 2025-10-27
RUN-2210: Update Key Value Data with new Match Substrings checkbox

edbaltra:
## Release Notes

This fix addresses an issue where the Key Value Data log filter required regex patterns to match the entire log line due to its use of `Matcher.matches()`. Users found that patterns working in external tools failed in Rundeck because they didn't consume the full line. To resolve this, we've added a new matchSubstrings configuration property that allows users to toggle between full-line matching (using `matches()`) and substring matching (using `find()`). This provides the flexibility to use partial patterns like `^.*\.[A-Z]([0-9]+)\.` without requiring them to match the entire line, while maintaining backward compatibility by defaulting to the original full-line matching behavior.

## PR Details
JIRA TICKET: https://pagerduty.atlassian.net/browse/RUN-2210
This pull request adds a new feature to the `KeyValueDataLogFilterPlugin` that allows users to choose between full-line and substring regular expression matching. It introduces a new plugin property, updates the matching logic, and adds comprehensive tests to ensure correct behavior across both modes.

### Feature: Substring vs. Full-Line Regex Matching

* Added a new `matchSubstrings` plugin property to `KeyValueDataLogFilterPlugin`, allowing users to toggle between Java's `find()` (substring match) and `matches()` (full-line match) methods when applying the regular expression.
* Updated the matching logic in `handleEvent` to use `find()` if `matchSubstrings` is enabled, otherwise defaulting to `matches()`. This enables more flexible pattern matching on log lines.

### Testing and Validation

* Added new parameterized tests in `KeyValueDataLogFilterPluginSpec.groovy` to verify that the `matchSubstrings` toggle correctly controls partial vs. full-line matching, and that data extraction remains stable across both modes.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -17,7 +17,7 @@
 # Rundeck Build Properties
 #####################################################################
 group=org.rundeck
-currentVersion = 5.17.0
+currentVersion = 5.18.0
 grailsVersion=6.1.2
 grailsGradlePluginVersion=6.1.2
 springBootVersion=2.7.18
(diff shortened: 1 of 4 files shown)


jayas006 (approved):
LGTM 🎉

PR #9874, opened 2025-10-27
RUN-3858: Improve cleanup default values

Jesus-Osuna-M:
This PR implements configurable default values for execution cleanup settings in new Rundeck projects. The feature allows administrators to set system-wide defaults for execution history cleanup parameters (retention days, minimum executions to keep, maximum deletion size, and cleanup schedule) via configuration, which will be automatically applied when execution cleanup is enabled.

Key changes:

1. Added new feature flags and configuration properties for default execution cleanup settings
2. Modified project creation and editing logic to use configurable defaults instead of hardcoded constants
3. Updated UI to conditionally display default values in labels based on whether the new feature is enabled

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/config/RundeckConfigBase.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/config/RundeckConfigBase.java
@@ -65,6 +65,7 @@ public class RundeckConfigBase {
     RundeckJobsConfig jobs;
     JobsImport jobsImport;
     Startup startup;
+    RundeckExecutionCleanupConfig executionCleanupConfig;
 
     RundeckAwsgateway awsgateway;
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/dtolabs/rundeck/core/config/RundeckConfigBase.java
```diff
@@ -461,6 +460,11 @@ public static class RundeckFeatureConfig {
         Enabled enhancedJobTakeoverQuery = new Enabled();
         Enabled publicKeysDownload = new Enabled();
         Enabled guiHideRoiInstructions = new Enabled();
+        Enabled defaultExecutionCleanup = new Enabled();
```
ltamaster, line 465: I think we can move these to a separate class like 
````
@Data
public static class RundeckExecutionCleanupConfig {
        Integer defaultExecutionDaysToKeep;
        Integer defaultExecutionMinimumToKeep;
        Integer defaultExecutionMaximumDeletionSize;
        String defaultExecutionCleanupScheduleCron;
}
````

You can set this as a field in RundeckConfigBase 

````
RundeckExecutionCleanupConfig executionCleanupConfig
````

finally the config value should be

````
rundeck.executionCleanupConfig.defaultExecutionDaysToKeep=xxx
rundeck.executionCleanupConfig.defaultExecutionMinimumToKeep=xxx
rundeck.executionCleanupConfig.defaultExecutionMaximumDeletionSize=xxx
rundeck.executionCleanupConfig.defaultExecutionCleanupScheduleCron=xxx
````

PR #9877, opened 2025-10-29
RUN-3907: Fix error to list webhooks created by an user that was removed

carlosrfranco:
### Overview

This PR fixes a NullPointerException that occurs when listing webhooks created by users who have been removed from the system. The issue arose because when a user is deleted, all associated auth tokens (including webhook tokens) were removed, causing NPEs when the webhook listing tries to access token properties.

Bugfix for the case when webhooks are created by users that were removed and causes the webhooks page stop working correctly

This issue happens because when a user is removed, the authTokens related to it are also removed. It causes a NPE when listing webhooks that contains authTokens created by removed users.

Also, this PR add a method to retrieve auth tokens filtering by user id and type to the TokenDataProvider interface. It will allow to remove only user type tokens to be removed when an user is deleted and preserves webhooks type tokens so that webhooks doesnt stop to work if an user is removed.

**Additional context**

### DIFF
--- a/rundeck-data-models/src/main/java/org/rundeck/app/data/providers/v1/authtoken/TokenDataProvider.java
+++ b/rundeck-data-models/src/main/java/org/rundeck/app/data/providers/v1/authtoken/TokenDataProvider.java
@@ -97,7 +97,7 @@ public interface TokenDataProvider extends DataProvider {
      * @param type of the AuthenticationToken, format AuthTokenType
      * @return list of AuthenticationTokens
      */
-    List<AuthenticationToken> findAllByUserAndType(String userId, AuthTokenType type);
+    default List<AuthenticationToken> findAllByUserAndType(String userId, AuthTokenType type){ return null; }
 
     /**
      * Retrieves a List of AuthenticationTokens for the creator that are expired.


rundeck-data-models/src/main/java/org/rundeck/app/data/providers/v1/authtoken/TokenDataProvider.java
```diff
@@ -90,6 +90,15 @@ public interface TokenDataProvider extends DataProvider {
...
     List<AuthenticationToken> findAllByUser(String userId);
 
+    /**
+     * Retrieves a List of AuthenticationToken for the user and type.
+     *
+     * @param userId of the AuthenticationTokens, format String
+     * @param type of the AuthenticationToken, format AuthTokenType
+     * @return list of AuthenticationTokens
+     */
+    List<AuthenticationToken> findAllByUserAndType(String userId, AuthTokenType type);
```
ronaveva: to avoid the need to implement this new method on the pro side (for multi tenancy), we can just use default

ronaveva: example
https://github.com/rundeck/rundeck/pull/9812/files#diff-472fe1c4b3ad5fad71966484ca71a4b67c23890619a926e516f07b97e6b04b0eR54

PR #9896, opened 2025-11-14
RUN-3729: Allow KeyValueDataLogFilterPlugin to capture multiple values 

Jesus-Osuna-M:
## Release Notes

Enhanced the Key Value Data Log Filter Plugin to support capturing multiple key-value pairs from a single log line through a new optional allowMultipleMatches property. This enables parsing of complex log formats containing multiple matches (e.g., "user=john role=admin session=abc123") while maintaining full backward compatibility with the default single-match behavior.

## Ticket Details

  Enhancement - This adds new functionality to the Key Value Data Log Filter Plugin to support capturing multiple matches from a single log line.

  - Added a new allowMultipleMatches boolean property (default: false) to the KeyValueDataLogFilterPlugin
  - Updated the matchSubstrings property description to clarify it only captures the first match unless allowMultipleMatches is enabled
  - Refactored the handleEvent method to support both single match (original behavior) and multiple match modes
  - Extracted common match processing logic into a new processMatch method to avoid code duplication
  - When allowMultipleMatches=true and matchSubstrings=true, the plugin now iterates through all matches in a line using matcher.find() in a while loop
  - Maintains full backward compatibility - existing behavior is unchanged when the new property is not enabled

  1. Always capture multiple matches: Would break existing behavior and expectations
  2. Modify existing matchSubstrings behavior: Would be a breaking change for current users
  3. Use arrays for duplicate keys: Would complicate the data structure and output format

  **Additional context**

  The original implementation of matchSubstrings using matcher.find() only captured the first match per line, which was limiting for use cases where multiple key-value pairs exist on the same log line.

  **Use cases this enhancement enables:**
  - Parsing log lines with multiple key=value pairs: "user=john role=admin session=abc123"
  - Extracting multiple IDs from formatted logs: "Processing ID:123, ID:456, ID:789"
  - Capturing repeated patterns in structured logs

  **Testing:**
  - Added comprehensive test coverage for the new functionality
  - All existing tests pass to ensure backward compatibility
  - Tests cover various scenarios including single/multiple captures, invalid key patterns, and edge cases

  The enhancement maintains the principle of least surprise by keeping the default behavior unchanged while providing opt-in functionality for users who need multiple match support.

### DIFF
--- a/rundeckapp/src/main/groovy/com/dtolabs/rundeck/server/plugins/logging/KeyValueDataLogFilterPlugin.groovy
+++ b/rundeckapp/src/main/groovy/com/dtolabs/rundeck/server/plugins/logging/KeyValueDataLogFilterPlugin.groovy
@@ -99,7 +99,9 @@ when applying the regular expression. This allows substring matches instead of r
 
     @PluginProperty(
             title = "Allow Multiple Matches",
-            description = '''When "Match Substrings" is enabled, this option allows capturing multiple matches from the same line.''',
+            description = '''When "Match Substrings" is enabled, this option allows capturing multiple matches from the same line.
+
+This option has no effect when "Match Substrings" is disabled.''',
             defaultValue = "false"
     )
     Boolean allowMultipleMatches


edbaltra (approved):
LGTM 👍

PR #9899, opened 2025-11-18
[RUN-3956] NextUI: Vue migration for User and System Configuration menus

gschueler:
## Tell us about your PR

Vue migration for some menus

- convert User menu and System Configuration menus to Vue, using a common MainbarMenu.vue component
- add unit tests
- remove legacy GSP implementations
- (fix: jest tests report wrong line number for failures)
- adjust storybook config, remove/update some old stories in the library section

- System configuration menu items list is still constructed inside GSP and passed as json to the Vue component
- Future change should move all items in the list to the MenuItem interface, and provide the list via an API

**Additional context**


smartinellibenedetti (approved):
Leaving it approved but asides the comments from copilot, there are some things that would be great to adjust if there's time:

- both new components have hardcoded strings that didn't get localized (there's a skill in rundeckpro for addressing this if you'd like to test: i18n-vue-template);
- the unit tests have a bit of code-smell (I had hoped that by having the rules for unit tests in rundeckpro would affect the agents working in the submodule, but looks like it's not guaranteed. There's a skill for writing the unit tests but this case might work best to just ask agent to check the changes against the verification-checklist in the write-jest-tests skill folder);
- I didn't check if the reason was the changes in the stories pattern or this is also happening on main but storybook is currently broken, it loads but won't render the stories due to build errors. Would suggest to undo the pattern change in the `.storybook/main.ts` and I'll get back to it later to sort it out;

gschueler:
@smartinellibenedetti i fixed the i18n and tests per your suggestions, and reverted the storybook tweak

PR #9905, opened 2025-11-21
[RUN-4040] Use Execution API in Activity Page instead EventAjax

ltamaster:
## Tell us about your PR

Replace the use of eventAjax (Dynamic queries for BaseReport table) with the use of the Execution API for the Activity Page. This provides better performance, cleaner architecture, and improved scalability for installations with large execution tables.

For the execution API, it needs some changes to cover the Activity Page functionality:
- Improve the criteria builder with better filters and optimized sorting
- Add optimized HQL count for cross-project job references (includeJobRef)
- Add simple HQL count without JOINs (configurable via `countPerformance.enabled`)
- Add execution count caching

---

### Components Modified

| Layer | Files | Changes |
|-------|-------|---------|
| **Frontend** | `activityList.vue`, `executions.ts` | Use Execution API instead of eventAjax |
| **API** | `ExecutionController.groovy` | New filters, `olderFilter` support, partial match descriptions |
| **Backend** | `ExecutionService.groovy` | Criteria builder improvements, simple HQL count, count caching |
| **Query** | `ExecutionQuery.groovy` | New filters, HQL count methods, simple count, LIKE escaping |
| **Domain** | `Execution.groovy` | Removed hasOne relationship for performance |
| **Tests** | `ExecutionServiceExecutionApiIntegrationSpec.groovy` | Comprehensive filter and count tests |

---

## Feature 1: Improved Criteria Builder

Enhancements to the execution query criteria builder for better performance and functionality.

**Goals:**
- **Remove extra queries**: Eliminate unnecessary database queries triggered by `executionSummary` (workflow steps) and `LogFileStorageRequest` (hasOne relationship)
- **Better index usage**: Optimize sorting to leverage database indexes (using `id` instead of `dateStarted` for secondary sort)
- **Legacy status compatibility**: Support both legacy (`'false'`, `'true'`) and new (`'failed'`, `'succeeded'`) status values in filters

### 1.1 Optimized Sort Fields

Changed the default sorting from `dateCompleted DESC, dateStarted DESC` to `dateCompleted DESC, id DESC`:

```groovy
order('dateCompleted', 'desc')
order('id', 'desc')  // Primary key - always indexed, faster than dateStarted
```

**Why `id` instead of `dateStarted`?**
- **Indexed**: Primary key is always indexed
- **Sequential**: IDs are assigned in creation order
- **Unique**: Guarantees deterministic ordering for pagination
- **Fast**: No additional index lookup required

### 1.2 New Query Filters

| Filter | Type | Description |
|--------|------|-------------|
| `olderFilter` | String | Relative date (e.g., "60d", "24h") - executions older than |
| `recentFilter` | String | Relative date for recent executions |
| `nodeFilter` | String | Filter by node name (partial match with ILIKE) |
| `optionFilter` | String | Filter by option values (partial match with ILIKE) |
| `adhocStringFilter` | String | Filter adhoc commands by title (partial match) |
| `excludeRunning` | Boolean | Exclude running executions (`dateCompleted IS NOT NULL`) |
| `executionSummary` | Boolean | Include/exclude summary field in response |

### 1.3 LIKE Pattern Escaping (Security)

User input in `nodeFilter`, `optionFilter`, and `adhocStringFilter` is sanitized to prevent SQL wildcard injection.

### 1.4 Disabled executionSummary by Default

The `executionSummary` field is now **disabled by default** in the JSON response, avoiding extra queries to workflow tables.

### 1.5 Domain Model Optimization

Removed `hasOne` relationship between `Execution` and `LogFileStorageRequest` to prevent unwanted JOINs during list queries.

### 1.6 Legacy Status Value Support

The status filter now correctly matches both legacy and new status values:

| Status Filter | Matches |
|--------------|---------|
| `failed` | `'failed'` OR `'false'` (legacy) |
| `succeeded` | `'succeeded'` OR `'true'` (legacy) |

---

## Feature 2: Cross-Project Job Reference Count Optimization

Optimized HQL-based count queries for `includeJobRef` queries that count executions across multiple projects.

### Optimized UNION Query for includeJobRef

For queries with `includeJobRef` (cross-project job references), uses a dual-query approach:

```sql
-- COUNT QUERY 1: Referenced executions (via workflow job reference steps)
SELECT COUNT(e.id)
FROM Execution e
INNER JOIN ReferencedExecution re WITH re.execution.id = e.id
WHERE e.scheduledExecution IS NOT NULL
AND re.jobUuid = :jobUuid
AND e.project IN (:projects)

-- COUNT QUERY 2: Direct executions (without ReferencedExecution)
SELECT COUNT(e.id)
FROM Execution e
WHERE e.project = :project
AND e.scheduledExecution IS NOT NULL
AND e.jobUuid = :jobUuid
AND NOT EXISTS (SELECT 1 FROM ReferencedExecution re WHERE re.execution.id = e.id)
```

**Performance improvement**: 50-100x faster than EXISTS-based approach using INNER JOIN.

### When is this used?

The HQL count optimization is used when `shouldUseUnionQuery()` returns true:
- `includeJobRef` is enabled
- A single job UUID filter is specified
- Both `projFilter` and `execProjects` are set

---

## Feature 3: Simple HQL Count (Configurable)

Optimized HQL count query that avoids expensive JOINs with `scheduled_execution` table.

### The Problem

The default criteria-based count generates a JOIN query:

```sql
-- Default criteria count (slower under heavy load)
SELECT count(*)
FROM execution this_
INNER JOIN scheduled_execution schedulede1_
    ON this_.scheduled_execution_id = schedulede1_.id
WHERE this_.scheduled_execution_id IS NOT NULL
  AND schedulede1_.uuid = 'job-uuid'
  AND schedulede1_.project = 'MyProject'
  AND this_.project = 'MyProject'
  AND date_completed BETWEEN '2025-01-16' AND '2026-01-21';
```

### The Solution

Use a simple HQL query that filters directly on `execution.job_uuid`:

```sql
-- Simple HQL count (faster, no JOIN)
SELECT count(execution0_.id)
FROM execution execution0_
WHERE execution0_.project = 'MyProject'
  AND execution0_.date_completed BETWEEN '2025-01-16' AND '2026-01-21'
  AND execution0_.job_uuid = 'job-uuid';
```

### Configuration

```properties
# Enable simple HQL count optimization (default: false)
rundeck.api.executionQueryConfig.countPerformance.enabled=true
```

| Setting | Behavior |
|---------|----------|
| `countPerformance.enabled=true` | Uses simple HQL count (no JOINs) |
| `countPerformance.enabled=false` | Uses criteria-based count (default) |

### Required Index

This optimization requires a new composite index:

```sql
CREATE INDEX EXEC_IDX_JOB_PROJ_DATE
ON execution (job_uuid, project, date_completed, status, cancelled);
```

**Note**: This index is created in a separate PR ([#9964](https://github.com/rundeck/rundeck/pull/9964)).

### When is Simple Count Used?

The `canUseSimpleCount()` method returns `true` when:
- `countPerformance.enabled=true`
- No ILIKE filters (`nodeFilter`, `optionFilter`, `adhocStringFilter`)
- Uses job UUID filters (not numeric IDs)

### Performance Comparison

| Query Type | Time (11M rows) | Notes |
|------------|-----------------|-------|
| Criteria with JOIN | 7-50 seconds | Under heavy load |
| Simple HQL (no JOIN) | &lt;1 second | With index |

---

## Feature 4: Execution Count Cache

Caches the results of expensive `COUNT(*)` queries to avoid repeated database scans.

### Configuration

```properties
# Enable/disable count cache (default: false)
rundeck.api.executionQueryConfig.countCache.enabled=true

# Cache TTL in seconds (default: 30)
rundeck.api.executionQueryConfig.countCache.ttl=30
```

| Setting | Behavior |
|---------|----------|
| `countCache.enabled=true` | Caches COUNT query results |
| `countCache.enabled=false` | Disables caching, runs COUNT every time (default) |

### Features

- **Thread-Safe Cache Initialization**: Uses double-checked locking with volatile field
- **Stale Cache Detection**: Automatically detects and refreshes stale cached counts
- **Cache Key Structure**: Type-safe immutable cache keys with static invalidation methods

---

## Performance Impact

| Metric | Before | After |
|--------|--------|-------|
| Activity page load | 7-50 seconds | &lt;2 seconds |
| COUNT query time | 7-50 seconds | &lt;100ms (cache hit) or &lt;1s (simple HQL) |
| MySQL connections | Exhausted under load | Stable |

---

## Testing

- **Unit Tests**: `ExecutionServiceTests.groovy`, `ExecutionQuerySpec.groovy` - Cache key generation, cacheability, LIKE escaping, canUseSimpleCount
- **Integration Tests**: `ExecutionServiceExecutionApiIntegrationSpec.groovy` - Cache behavior, filter logic, count queries
- **Functional Tests**: `JobActivityHistorySpec.groovy` - Activity page functionality

---

## Limitations

1. **Multiple Job IDs**: Queries with multiple job IDs are not cached
2. **Absolute Dates**: Queries with absolute date filters are not cached
3. **Start Date Filters**: `startafterFilter`, `startbeforeFilter` not cached
4. **Stale Counts**: Counts may be up to TTL seconds stale (when cache is enabled)
5. **ILIKE Filters**: Simple HQL count cannot be used when ILIKE filters are present

---

## Release Notes

Performance optimization for Activity Page: Replaced eventAjax with Execution API. Includes four major improvements:
1. **Criteria Builder**: Optimized sorting (`id` instead of `dateStarted`), new filters, LIKE escaping for security, legacy status value support (`'false'`/`'true'`)
2. **Cross-Project Job Reference Count**: HQL-based count queries for `includeJobRef` avoiding expensive EXISTS subqueries
3. **Simple HQL Count**: Configurable HQL count without JOINs for job-filtered queries (requires index, configurable via `countPerformance.enabled`)
4. **Count Cache**: Caches COUNT results to reduce database load (configurable via `countCache.enabled`)

Reduces database load and improves response times for large execution tables.


rundeckapp/src/main/groovy/com/dtolabs/rundeck/app/support/ExecutionQuery.groovy
```diff
@@ -461,4 +487,66 @@ class ExecutionQuery extends ScheduledExecutionQuery implements Validateable{
...
+                        eq("jobUuid", jobUuid)
+                    }
+
+                    count += (count1 as Long) + (count2 as Long)
+
+                }
+            }
+        }
+        return count
+    }
```
gschueler: odd logic, it tries to convert id's to Long, then loops and only uses strings.. Is the purpose to exclude the Long values? the logic could be improved

ltamaster: I will return to it. I am still checking if I can get a good response time for that case; the count with GROM and Criteria takes a long time.

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
```diff
@@ -45,23 +45,6 @@ class JobActivityHistorySpec extends SeleniumBase {
         }
     }
 
-    def "review job activity history from Job List Page Saved Filters"() {
```
ronaveva, line 48: why is this test being deleted?

rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
```diff
@@ -190,8 +190,8 @@ class MenuController extends ControllerBase implements ApplicationContextAware{
         model = executionService.finishQueueQuery(query,params,model)
 
         //include id of last completed execution for the project
-        def eid=executionService.lastExecutionId(query.projFilter)
-        model.lastExecId=eid
+        //def eid=executionService.lastExecutionId(query.projFilter)
```
ronaveva: if this is not needed, it should be removed instead of commented

rundeckapp/grails-app/domain/rundeck/Execution.groovy
```diff
@@ -186,6 +186,13 @@ class Execution extends ExecutionContext implements EmbeddedJsonData, ExecutionD
         return logFileStorageRequest?.id
     }
 
+    Boolean getJobDeleted() {
```
ronaveva: since this returns a boolean a more appropriate name would be "isJobDeleted"

rundeckapp/grails-app/services/rundeck/services/ApiService.groovy
```diff
@@ -764,7 +764,8 @@ class ApiService implements WebUtilService{
                 def href=execdata.href
                 def status=execdata.status
                 def summary=execdata.summary
-                def Execution e = Execution.get(execdata.execution.id)
+                //Execution e = Execution.get(execdata.execution.id)
```
ronaveva:
```suggestion
```

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -4095,12 +4097,22 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
         }
         and {
           order('dateCompleted', 'desc')
-          order('dateStarted', 'desc')
```
ronaveva, line 4314: wouldn't this impact the order of the non completed executions?

rundeckapp/grails-spa/packages/ui-trellis/src/library/services/executions.ts
```diff
@@ -9,14 +9,16 @@ export interface PagedResult<T> {
 export interface PagingParams {
   max: number;
   offset: number;
+  total?: number;
+  count?: number;
```
ronaveva, line 13: whats the difference between total and count?

rundeckapp/src/main/groovy/com/dtolabs/rundeck/app/support/ExecutionQuery.groovy
```diff
@@ -291,7 +302,7 @@ class ExecutionQuery extends ScheduledExecutionQuery implements Validateable{
 
             txtfilters.each { key, val ->
               if (query["${key}Filter"]) {
-                ilike(val, '%' + query["${key}Filter"] + '%')
+                like(val, '%' + query["${key}Filter"] + '%')
```
ronaveva: this would affect the current behavior when searching, are we sure that we want this case sensitive? if thats so we would need to include it on the docs

rundeckapp/src/main/groovy/com/dtolabs/rundeck/app/support/ExecutionQuery.groovy
```diff
@@ -405,22 +421,14 @@ class ExecutionQuery extends ScheduledExecutionQuery implements Validateable{
...
+          eq('status',  'failed')
           eq('cancelled', false)
-          or{
-            eq('status',  'failed')
-            eq('status',  'false')
-          }
-        }else if(state == ExecutionService.EXECUTION_SUCCEEDED){
           isNotNull('dateCompleted')
+        }else if(state == ExecutionService.EXECUTION_SUCCEEDED){
+          eq('status',  'succeeded')
```
ronaveva: the status "true" is no longer considered, is this an old behavior?

rundeckapp/src/main/groovy/com/dtolabs/rundeck/app/support/ExecutionQuery.groovy
```diff
@@ -461,4 +495,53 @@ class ExecutionQuery extends ScheduledExecutionQuery implements Validateable{
       return criteriaClos
     }
 
+    /**
+     * Execute count query using dual-query approach for includeJobRef scenarios
+     * Executes two separate COUNT queries and sums them (database-agnostic)
+     */
+    def executeJobReferenceCount() {
```
ronaveva: if query.execProjects and query. projFilter have the same project, this could lead to a miscount by having the same exec more than one time? 🤔  i dont think so, i think it would only happen if a job references it self, but its worth giving it another thought

ronaveva (dismissed):
Hello @ltamaster theres a couple small obs to take a look at

rundeckapp/grails-spa/packages/ui-trellis/src/library/services/executions.ts
ltamaster: This is an example
```
"paging": {
        "count": 20,
        "total": 10668394,
        "offset": 0,
        "max": 20
    }
```
`count` returns the page size, and `total` is the total number of executions of the query

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/jobs/JobActivityHistorySpec.groovy
ltamaster: because the `Saved Filters` are not used in the job list page (Activity Tab). The feature saved the filter without any real field inside.

rundeckapp/src/main/groovy/com/dtolabs/rundeck/app/support/ExecutionQuery.groovy
ltamaster: I will return to the previous version

rundeckapp/src/main/groovy/com/dtolabs/rundeck/app/support/ExecutionQuery.groovy
ltamaster: Yes, the status "true" was to be used for succeeded executions. We can add a migration code in the PR where we will add the index, but I am not sure if it's worth it because executions with that state could be very old

rundeckapp/src/main/groovy/com/dtolabs/rundeck/app/support/ExecutionQuery.groovy
ltamaster: I will add a small change to avoid the problem, but it is OK for a job be returned for the referenced table and for the direct execution table.

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
ltamaster: We have a separate endpoint for running executions, so that is not affecting it. 
In other cases, we always set a complete: 
1. If you kill the job, it is set to `abort` status, and a `dateCompleted` is set
2. When the server goes down, and it exists running execution: in that case, there is a startup process that sets the execution to `incomplete` status and sets a `dateCompleted` date. 

So, this wouldn't affect how the executions are listed

rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
```diff
@@ -186,12 +186,7 @@ class MenuController extends ControllerBase implements ApplicationContextAware{
         def model = metricService?.withTimer(MenuController.name, actionName+'.queryQueue') {
             executionService.queryQueue(query)
         } ?: executionService.queryQueue(query)
-        //        System.err.println("nowrunning: "+model.nowrunning);
         model = executionService.finishQueueQuery(query,params,model)
-
-        //include id of last completed execution for the project
-        def eid=executionService.lastExecutionId(query.projFilter)
-        model.lastExecId=eid
```
smartinellibenedetti, line 194: heads up: I couldn't find where exactly this attribute `lastExecId` is being rendered in the ui, but I saw references to it in `main/menu/jobs.gsp` and `view/menu/jobs.gsp`. Probably they can also be removed in a later PR

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -4107,15 +4148,289 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
...
+        }
+        if (query.doendbeforeFilter && query.endbeforeFilter) {
+            parts << "endBefore:${query.endbeforeFilter.time}"
+        }
+        if (query.dostartafterFilter && query.startafterFilter) {
+            parts << "startAfter:${query.startafterFilter.time}"
+        }
+        if (query.dostartbeforeFilter && query.startbeforeFilter) {
+            parts << "startBefore:${query.startbeforeFilter.time}"
+        }
```
smartinellibenedetti, line 4660: question: isn't there a chance that by including the date filters in the cache key, the cache won't be reused correctly? I saw that when the user sets the period in the search modal, the timestamp will be fixed, but I think recentFilter actually uses a relative date, so endafterFilter would always be different

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -4107,15 +4148,289 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
...
+            // Standard criteria count (may include JOINs)
+            log.debug("Using criteria count for execution query")
+            total = Execution.createCriteria().count(criteriaClos.curry(true))
+        }
+
+        // Store in cache if enabled
+        if (isExecutionCountCacheEnabled() && cacheKey) {
+            getExecutionCountCache().put(cacheKey, total as Long)
+        }
+    }
```
smartinellibenedetti, line 4515: question: I saw that the executions endpoint is being called with excludeRunning = true from the activityPage, but it seems like neither countDirectAndReferencedExecutions nor countExecutionsSimple is including a query to check that dateCompleted is not null. Is this intentional?

smartinellibenedetti: Other parameters that seem to be missing:

  countExecutionsSimple: excludeRunning, nodeFilter, optionFilter, adhocStringFilter
  countDirectAndReferencedExecutions: all filters but jobUuid, project, scheduledExecution

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -4107,15 +4148,289 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
...
+            cacheIsStale = true
+        } else if (resultSize == 0 && offset > 0 && total > offset) {
+            // Empty result on a non-first page but cached total says there should be data
+            log.debug("Stale cache detected: empty page at offset ${offset} but total is ${total}")
+            cacheIsStale = true
+        } else if (resultSize > 0 && resultSize < max && max > 0 && offset + resultSize < total) {
+            // Got fewer results than max (partial page) but cached total says there's more
+            log.debug("Stale cache detected: partial page with ${resultSize} results but total is ${total}")
+            cacheIsStale = true
+        }
```
smartinellibenedetti, line 4536: not a blocker but: would it be a lot of work to mark the cache as invalid when an execution completes or gets deleted? I suspect that the scenarios covered here won't really happen unless all executions get removed OR there are enough results to create another page

smartinellibenedetti (approved):
There's just this comment that I think it would be good to take a look/address if there's time before merging: https://github.com/rundeck/rundeck/pull/9905#discussion_r2761342033

But leaving the PR approved in case that comment isn't applicable and you'd like to merge as is.

rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
ltamaster: you are right, I will rollback this

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
ltamaster: I clean that code

rundeckapp/src/main/groovy/com/dtolabs/rundeck/app/support/ExecutionQuery.groovy
```diff
@@ -211,7 +225,7 @@ class ExecutionQuery extends ScheduledExecutionQuery implements Validateable{
                                       }
                                   }
                               })
-                              eq("uuid", theid)
+
                           }
```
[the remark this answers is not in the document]
ltamaster: @copilot the referenced executions is not the common case

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
ltamaster: I will sort by ID, which has a similar use, and it performs better

ltamaster:
@copilot resolve the merge conflicts in this pull request

ltamaster:
@copilot resolve the merge conflicts in this pull request

PR #9909, opened 2025-11-25
RUN-3768: Add new parameters to executions/metrics endpoint, pre-aggregating metrics about executions

smartinellibenedetti:
## Overview
This PR implements a high-performance metrics API endpoint that uses pre-aggregated daily execution statistics instead of querying the execution table directly. This dramatically improves response times for job metrics queries, especially when fetching metrics for multiple jobs simultaneously.

## The Bigger Problem

The Job Metrics UI plugin was unusable in production environments with millions of executions:

**Scale Context:**
- Test environment: 3+ million execution records
- UI plugin making 100+ API calls per page load
- Each call querying the `execution` table directly
- Queries timing out or causing high CPU usage
- Performance degrading linearly with execution count

**Root Cause:** Directly querying the `execution` table for metrics aggregation is fundamentally incompatible with this scale. Even with database indexes, at 3+ million rows, queries scan large portions of the table and cannot provide acceptable performance.

## What This PR Solves

This backend PR addresses the performance problem by implementing pre-aggregated metrics storage and new API endpoints:

**Problem:** The existing `/api/40/project/{project}/executions/metrics` endpoint had significant performance limitations:
- **Slow queries**: Direct queries against the execution table for each job (timing out at scale)
- **Database load**: Multiple queries caused database contention
- **Scalability issues**: Performance degraded linearly with number of jobs
- **No batch support**: Required individual API calls for each job (N+1 problem)

## Solution

**Core Concept:** Write metrics incrementally as executions complete instead of computing them on-demand from the execution table.

This PR introduces:
1. **Pre-aggregated metrics storage**: Daily metrics stored in `SCHEDULED_EXECUTION_STATS` table's JSON field as executions complete
2. **Stats-based metrics API**: New `useStats=true` parameter uses pre-aggregated data (single row read vs table scan)
3. **Batch mode**: New `groupByJob=true` parameter returns metrics for all jobs in a project in a single request
4. **API Versioning**: New parameters require API version 57 or higher (enforced via version check)
5. **Backward compatibility**: Original execution table queries remain available when `useStats=false` or when using older API versions

### Data Structure
Daily metrics are stored in `SCHEDULED_EXECUTION_STATS.contentMap.dailyMetrics`:
```groovy
[
  "2024-01-15": [
    total: 10,
    succeeded: 8,
    failed: 1,
    aborted: 1,
    timedout: 0,
    duration: 45000,  // milliseconds
    hourly: [0, 0, 0, 1, 2, 3, ...]  // 24-element array
  ],
  "2024-01-16": [...]
]
```

### API Response Format
**Single job** (`useStats=true`, no `groupByJob`):
```json
{
  "total": 100,
  "succeeded": 95,
  "failed": 3,
  "aborted": 2,
  "timedout": 0,
  "successRate": 95.0,
  "duration": {"average": 45000},
  "daily_breakdown": {
    "2024-01-15": {"total": 10, "succeeded": 9, ...},
    "2024-01-16": {"total": 12, "succeeded": 11, ...}
  },
  "hourly_heatmap": [0, 0, 1, 2, 5, 8, ...]
}
```

**Batch mode** (`useStats=true`, `groupByJob=true`):
```json
{
  "jobs": {
    "job-uuid-1": { /* metrics object */ },
    "job-uuid-2": { /* metrics object */ }
  }
}
```

### Backward Compatibility
- Original execution table queries remain default (`useStats=false` or omitted)
- Existing API consumers continue to work without changes (API versions < 57)
- Stats-based mode is opt-in via feature flag and API version 57+
- **API Version Enforcement**: Parameters `useStats` and `groupByJob` are only available in API version 57 or higher
  - Requests with these parameters using older API versions will return an error
  - This ensures proper versioning and prevents accidental use of new features with incompatible clients

### Feature Flag
- **Configuration**: `rundeck.executionDailyMetrics.enabled=true`
- **UI flag**: `guiUseExecutionDailyMetrics` (exposed via feature API)
- **Backend flag**: Controls whether daily metrics are aggregated during execution completion

### Migration Notes
- Daily metrics are populated automatically when feature flag is enabled
- Historical data: Metrics are only available from the point when feature flag is enabled
- No data migration required - stats accumulate over time
- Pruning ensures table size remains manageable (90-day window)

### DIFF
--- a/rundeckapp/src/test/groovy/rundeck/controllers/ExecutionController2Spec.groovy
+++ b/rundeckapp/src/test/groovy/rundeck/controllers/ExecutionController2Spec.groovy
@@ -1043,7 +1043,9 @@ class ExecutionController2Spec extends Specification implements ControllerUnitTe
         when: "getMetricsFromStats is called with begin and end dates"
             def begin = "2025-11-22T00:00:00Z"
             def end = "2025-11-24T23:59:59Z"
-            def result = controller.getMetricsFromStats(jobUuid, begin, end)
+            def beginDate = ReportsController.parseDate(begin)
+            def endDate = ReportsController.parseDate(end)
+            def result = controller.getMetricsFromStats(jobUuid, beginDate, endDate)
 
         then: "only metrics within the date range are returned"
             result != null
@@ -1082,7 +1084,9 @@ class ExecutionController2Spec extends Specification implements ControllerUnitTe
         when: "getMetricsFromStats is called with a date range that has no metrics"
             def begin = "2025-11-22T00:00:00Z"
             def end = "2025-11-24T23:59:59Z"
-            def result = controller.getMetricsFromStats(jobUuid, begin, end)
+            def beginDate = ReportsController.parseDate(begin)
+            def endDate = ReportsController.parseDate(end)
+            def result = controller.getMetricsFromStats(jobUuid, beginDate, endDate)
 
         then: "returns empty metrics object with all zeros"
             result != null
@@ -1122,7 +1126,9 @@ class ExecutionController2Spec extends Specification implements ControllerUnitTe
         when: "getMetricsFromStats is called with begin and end matching boundary dates"
             def begin = "2025-11-22T00:00:00Z"
             def end = "2025-11-24T23:59:59Z"
-            def result = controller.getMetricsFromStats(jobUuid, begin, end)
+            def beginDate = ReportsController.parseDate(begin)
+            def endDate = ReportsController.parseDate(end)
+            def result = controller.getMetricsFromStats(jobUuid, beginDate, endDate)
 
         then: "boundary dates are included"
             result != null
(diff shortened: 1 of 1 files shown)


functional-test/src/test/groovy/org/rundeck/util/container/RdClient.groovy
```diff
@@ -33,7 +33,7 @@ class RdClient {
     /**
      *  The current (latest) released version of the Rundeck API
      */
-    public static final int API_CURRENT_VERSION = 56
+    public static final int API_CURRENT_VERSION = 57
```
carlosrfranco, line 36: Do we have a doc PR for the API version updating?

smartinellibenedetti: Thanks for the catch: https://github.com/rundeck/docs/pull/1747/files

PR #9737, opened 2025-07-24
UI/backend: job stats cleanup

gschueler:
**Is this a bugfix, or an enhancement? Please describe.**
cleanup job stats section shown in job/execution show pages:
* move stats calculation out of gsp code
* convert ui to vue

**Describe the solution you've implemented**

**Describe alternatives you've considered**

**Additional context**

PR #9924, opened 2025-12-08
Run-3744: NextUI: job status display fix

gschueler:
## Tell us about your PR

Follow up from  #9737 

Fix: (Next UI mode only) Job success rate stat was not shown if success rate was 0%

**Additional context**

### DIFF
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/app/pages/job/show/JobStatsBasic.vue
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/app/pages/job/show/JobStatsBasic.vue
@@ -87,7 +87,7 @@ export default defineComponent({
           {{ formatNumber(execCount) }}
         </span>
         <span class="text-table-header">
-          {{ $tc("execution", execCount) }}
+          {{ $t("execution", execCount) }}
         </span>
       </div>
     </ui-socket>


rundeckapp/src/main/groovy/org/rundeck/app/data/job/metadata/JobStatsMetadataComponent.groovy
```diff
@@ -42,7 +42,7 @@ class JobStatsMetadataComponent implements JobMetadataComponent {
             [
                 averageDuration: stats?.averageDuration ?: 0,
                 executionCount : stats?.execCount ?: 0,
-                successRate    : stats?.successRate ?: -1,
+                successRate    : stats?.successRate,
```
smartinellibenedetti, line 45:
```suggestion
                successRate    : stats?.successRate ?: 0,
```

wouldn't it be better to have it set to 0 if there are no successes? (saw that the tests are assuming that the success rate will be -1 as well)

gschueler: the data provider returns -1 if there is no data, which hides the stat info in the UI in that case.  The previous error was that the stats provider returned 0 correctly (e.g. 0/10 runs were successful), but the `?: -1` was converting that to -1.  In both cases we just want to pass along what the stats provider returns.

rundeckapp/grails-spa/packages/ui-trellis/src/app/pages/job/show/JobStatsBasic.vue
smartinellibenedetti:
```suggestion
          {{ $t("execution", execCount) }}
```

Sorry just saw this one now, tc got deprecated in the version 10 of vue-i18n

rundeckapp/src/main/groovy/org/rundeck/app/data/job/metadata/JobStatsMetadataComponent.groovy
smartinellibenedetti: Ah gotcha, thanks for the explanation

PR #9934, opened 2025-12-12
RUN-4020: Invalid regex in acl project name can cause 500 error when uploading

gschueler:
## Tell us about your PR

Bug fix: project name in acl context should be validated as a regex

- additional fix: source name for invalid ACL created in the GUI shown as "null", should be "editor".

An alternative is to allow invalid regex by catching pattern syntax exception, and not using the regex matcher for those acls, only using exact string matches.  However, a valid project name is already constrained to a basic charset, so any characters that make the name an invalid regex would also make it an invalid project name for exact string matching. If a user attempts to create a regex (e.g. assuming the string `*` will match all projects), it is better to cause a validation error than to silently match nothing.

**Additional context**

Project name in the context part is matched as a regular expression as well as via equality match when filtering ACLs for a project.


edbaltra (approved):
LGTM

Issue #8262, opened 2023-04-17
BaseDir for file storage provider ignored in Docker configuration


leonboot:
When trying to configure a file storage provider through the `RUNDECK_STORAGE_PROVIDER_#_*` environment variables (for the Rundeck Docker image), the value for `config.baseDir` is ignored. This will result in an exception being thrown when starting Rundeck.

**My Rundeck detail**

* Rundeck version: 4.11.0
* install type: docker
* OS Name/version: Debian 11
* DB Type/version: MariaDB 10.11

**To Reproduce**
1. Create an .env file with the following config and provide any other mandatory configuration, but make sure also contains this:
   ```
   RUNDECK_STORAGE_PROVIDER_1_TYPE=file
   RUNDECK_STORAGE_PROVIDER_1_PATH=/
   RUNDECK_STORAGE_PROVIDER_1_CONFIG_BASEDIR=/home/rundeck/var/storage
   ```
3. Start the Docker image
4. An error should appear in the logs about the missing baseDir. When entering the container, note that the baseDir line is absent in /home/rundeck/server/config/rundeck-config.properties.

**Expected behavior**

The file storage provider to work properly.

**Desktop (please complete the following information):**
 - OS: N/A
 - Browser N/A
 - Version N/A

**Additional context**
The variable is not accounted for in [the Remco template](https://github.com/rundeck/rundeck/blob/1d7f6d595858021f5cef7443c0f4a7aaca903f8d/docker/official/remco/templates/rundeck-config-storage.properties#L6-L13). I've tried adding the following: before the `{% endif %}`:
```
{% if type == 'file' %}
rundeck.storage.provider.{{index}}.config.baseDir={% set basedir = printf("%s/config/basedir", provider) %}{{ getv(basedir, "/home/rundeck/var/storage")}}
{% endif %}
```
But the line would still not appear. Removing the surrounding `if/endif` made the line appear in the generated rundeck-config.properties file.

PR #9947, opened 2025-12-24
Fixing lack of support for file storage provider config basedir.

zepag:
## Tell us about your PR

It is a bug fix for the remco template for storage properties, that allows configuring the base configdir.
See #8262 

I changed the name of the type variable to allow testing twice (otherwise it is being reset to the reference inside of the if block, therefore disallowing an additional test).
I then test it again for file only, in order to add the config basedir property.
This then functions as expected naturally and described in #8262 's premise.

Apart from tweaking templates and working around, no real clean alternative was found. We'd prefer for the baseline template to be correct :) .

**Additional context**

## Release Notes

FIX: container can now use file storage provider with config basedir, using a `RUNDECK_STORAGE_PROVIDER_*_CONFIG_BASEDIR` variable

### DIFF
--- a/docker/official/remco/templates/rundeck-config-storage.properties
+++ b/docker/official/remco/templates/rundeck-config-storage.properties
@@ -5,10 +5,13 @@
 
 {%- macro storage_provider(provider) %}
     {%- set index = provider | base %}
-    {%- set type = getv(printf("%s/type", provider), "db") %}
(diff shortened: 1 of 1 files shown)


zepag:
Is there some additional coming from me as submissioner? It seems to require a milestone, though that does more sound like a job for the project's team I guess.

PR #9948, opened 2025-12-31
RUN-3855: Update ubuntu image back to current and update selenium version to 4.36

smartinellibenedetti:
## Tell us about your PR

This PR updates the Selenium testing infrastructure and improves test reliability by upgrading from Selenium 4.34.0 to 4.36.0 and updating the Ubuntu CI image from a specific version (2024.11.1) to current. 

With Selenium's update, several tests broke or became flake due to stale element references and race conditions. 

**Additional context**

### DIFF
--- a/functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/appadmin/AccessControlSpec.groovy
+++ b/functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/appadmin/AccessControlSpec.groovy
@@ -63,7 +63,7 @@ class AccessControlSpec extends SeleniumBase {
             aclPage.waitForPoliciesCountToBeAtLeast 1
         then:
             aclPage.policiesTitleList.size() == 1
-            aclPage.policiesTitleList.any { it.getText() == validPolicyName }
+            aclPage.policiesTitleList.get 0 getText() equals validPolicyName
     }
 
     def "upload form warns of duplicate name"() {


functional-test/src/test/groovy/org/rundeck/util/container/SeleniumBase.groovy
```diff
@@ -37,11 +37,12 @@ class SeleniumBase extends BaseContainer implements WebDriver, SeleniumContext {
...
     WebDriver getDriver() {
         if (null == _driver) {
-            def prefs = ["download.default_directory": downloadFolder]
+            def prefs = ["download.default_directory": downloadFolder, "profile.password_manager_leak_detection": false]
             LoggingPreferences logPrefs = new LoggingPreferences()
             logPrefs.enable(LogType.BROWSER, Level.ALL)
 
-            ChromeOptions options = new ChromeOptions()
+            ChromeOptions options = new ChromeOptions();
+            options.setBrowserVersion("140"); // Pinned to Chrome 140 major version (uses latest stable 140.x)
```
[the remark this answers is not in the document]
smartinellibenedetti: Inaccurate, the latest CDP that 4.36 works with is 140, which is stable https://github.com/SeleniumHQ/selenium/blob/trunk/java/CHANGELOG#L50

https://googlechromelabs.github.io/chrome-for-testing/

functional-test/src/test/groovy/org/rundeck/util/gui/pages/project/ProjectEditPage.groovy
```diff
@@ -210,16 +211,36 @@ class ProjectEditPage extends BasePage {
...
+        if (str == null) {
+            return "''"
+        }
+        // Escape backslashes first, then single quotes, then handle newlines
+        return "'" + str.replace('\\', '\\\\')
+                       .replace("'", "\\'")
+                       .replace('\n', '\\n')
+                       .replace('\r', '\\r')
+                       .replace('\t', '\\t') + "'"
+    }
```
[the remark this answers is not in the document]
smartinellibenedetti: Not needed for this case.

.circleci/config.yml
```diff
@@ -91,7 +90,7 @@ executors:
   # Machine executor
   machine-builder:
     machine:
-      image: ubuntu-2204:2024.11.1
+      image: ubuntu-2204:current
```
ronaveva: are we sure we want to use the "current" version?

functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/appadmin/AccessControlSpec.groovy
```diff
@@ -55,16 +54,16 @@ class AccessControlSpec extends SeleniumBase {
...
             aclPage.uploadFileField.sendKeys createTempYamlFile(validPolicyData)
             aclPage.uploadNameField.clear()
             aclPage.uploadNameField.sendKeys validPolicyName
             aclPage.uploadSubmitButton.click()
+            aclPage.waitForPoliciesCountToBeAtLeast 1
         then:
-            aclPage.validatePage()
-            aclPage.policiesTitleList.size() == 1
-            aclPage.policiesTitleList.get 0 getText() equals validPolicyName
+            aclPage.policiesTitleList.size() >= 1
```
ronaveva: this should be an exact value, otherwise it would not be accurate

ronaveva (changes requested):
looks good overall, just a minor observation to take care of

.circleci/config.yml
smartinellibenedetti: Good question, can raise it on the standup to get everyone's input.

The current tag was the original value we had on circleci's pipelines when we had that situation where a lot of selenium tests broke out of nowhere in Oct 2 (around that time a new machine image got released). At the time, the consensus was that we were going to unblock the release by going to the previous image and then return to the current image afterwards. 

Circleci's official policy is that they'd update their images around every 3 months, there's a thread in their support discussing [this](https://discuss.circleci.com/t/no-updates-to-ubuntu-2204-machine-images-since-nov-2024/53297), but even now there is just one version after the [2024.11.1](https://circleci.com/developer/machine/image/ubuntu-2204).  

So possibly it won't be a problem either way. Any suggestions if we should stay on the hardcoded version or not?

PR #9951, opened 2026-01-07
RUN-3948: Change Data Model to store workflow data as JSON

carlosrfranco:
## Pull request overview

This PR migrates workflow storage for `ScheduledExecution` and `Execution` from the legacy GORM `Workflow`/`WorkflowStep` model to a JSON-backed model (`workflow_json`), updating converters, services, controllers, views, and tests to use `getWorkflowData()`/`setWorkflowData()`.

**Changes:**
- Add `workflow_json` columns (Liquibase) and introduce JSON serialization/deserialization helpers on `ScheduledExecution` and `Execution`.
- Introduce JSON-backed workflow/step implementations (`WorkflowDataImpl`, `WorkflowStepDataImpl`) and update converters/services/controllers/views to consume `WorkflowData`.
- Update/extend unit + integration + functional tests for the new workflow representation.

### DIFF
--- a/rundeckapp/src/integration-test/groovy/rundeck/WorkflowJsonStorageIntegrationSpec.groovy
+++ b/rundeckapp/src/integration-test/groovy/rundeck/WorkflowJsonStorageIntegrationSpec.groovy
@@ -64,7 +64,7 @@ class WorkflowJsonStorageIntegrationSpec extends Specification {
         then: "Job should have workflow data"
         retrieved != null
         retrieved.workflowJson != null
-        retrieved.workflow == null // Old format should be null
+        retrieved.workflow != null // Old format should not be null for backwards compatibility
 
         and: "Workflow data should be retrievable"
         def retrievedWorkflow = retrieved.getWorkflowData()
@@ -110,7 +110,7 @@ class WorkflowJsonStorageIntegrationSpec extends Specification {
         then: "Execution should have workflow data"
         retrieved != null
         retrieved.workflowJson != null
-        retrieved.workflow == null // Old format should be null
+        retrieved.workflow != null // Old format should not be null for backwards compatibility
 
         and: "Workflow data should be retrievable"
         def retrievedWorkflow = retrieved.getWorkflowData()
@@ -325,7 +325,7 @@ class WorkflowJsonStorageIntegrationSpec extends Specification {
         then: "Should now use JSON format"
         def migrated = ScheduledExecution.get(savedId)
         migrated.workflowJson != null
-        migrated.workflow == null
+        migrated.workflow != null // Old format should still be present for backwards compatibility
 
         and: "Data should be preserved"
         def migratedWorkflow = migrated.getWorkflowData()


core/src/main/java/com/dtolabs/rundeck/core/execution/ExecutionContext.java
```diff
@@ -229,4 +229,11 @@ public interface ExecutionContext {
      * @return An {@link ExecutionReference} to the execution, or null if doesn't apply.
      */
     public ExecutionReference getExecution();
+
+    /**
+     * Retrieves the workflow JSON representation for the execution.
+     *
+     * @return the workflow JSON string
+     */
+    public String getWorkflowJson();
```
ronaveva: It would be better to have this to return an instance of "WorkflowData" instead of the json string

rundeck-data-models/src/main/java/org/rundeck/app/data/model/v1/job/workflow/WorkflowData.java
```diff
@@ -40,6 +41,8 @@ public interface WorkflowData {
      */
     List<WorkflowStepData> getSteps();
 
+    List<WorkflowStepData> getCommands();
```
ronaveva, line 44: why is this being added and not just use "getSteps"?

rundeckapp/grails-app/controllers/rundeck/controllers/ScheduledExecutionController.groovy
```diff
@@ -2483,7 +2485,8 @@ Authorization required: `delete` on project resource type `job`, and `delete` on
         }
         def props=[:]
         props.putAll(execution.properties)
-        props.workflow=new Workflow(execution.workflow)
+        def execWorkflowData = execution.getWorkflowData()
+        props.workflow=execWorkflowData ? new Workflow(execWorkflowData) : new Workflow()
```
ronaveva: wouldn't be better to just use "workflowData" instead of creating a new instance of "Workflow"?

rundeckapp/grails-app/controllers/rundeck/controllers/WorkflowController.groovy
```diff
@@ -1194,10 +1195,13 @@ class WorkflowController extends ControllerBase {
                 ScheduledExecution sched = ScheduledExecution.getByIdOrUUID(params.scheduledExecutionId)
                 if (!sched) {
                     session.editWF[wfid] = new Workflow()
```
ronaveva: could we use an instande of workflowData instade?

rundeckapp/grails-app/controllers/rundeck/controllers/WorkflowController.groovy
```diff
@@ -1272,7 +1279,7 @@ class WorkflowController extends ControllerBase {
     static Map _validatePluginStep(FrameworkService frameworkService, WorkflowStep exec) {
         if (exec instanceof PluginStep) {
             PluginStep item = exec as PluginStep
-            def description = getPluginStepDescription(frameworkService, item.nodeStep, item.type)
+            def description = WorkflowController.getPluginStepDescription(frameworkService, item.nodeStep ?: false, item.type)
```
ronaveva: i dont see the need to check for "nodeStep"

```suggestion
            def description = getPluginStepDescription(frameworkService, item.nodeStep, item.type)
```

rundeckapp/src/main/groovy/org/rundeck/app/data/workflow/WorkflowDataImpl.groovy
```diff
@@ -0,0 +1,262 @@
...
+import rundeck.data.validation.validators.plugin.RundeckPluginValidator
+import rundeck.data.validation.validators.workflowstep.WorkflowStepValidatorFactory
+
+/**
+ * Implementation of WorkflowData for deserialized workflow data.
+ * This class represents workflow data that has been deserialized from JSON storage.
+ * It provides a runtime representation without GORM persistence overhead.
+ */
+@JsonIgnoreProperties(["errors"])
+class WorkflowDataImpl implements WorkflowData, Validateable  {
```
ronaveva, line 27: We already have a WorkflowData implementation called RdWorkflow, why not just use that one instead?

rundeckapp/grails-app/domain/rundeck/JobExec.groovy
```diff
@@ -108,7 +108,7 @@ public class JobExec extends WorkflowStep implements IWorkflowJobItem{
         if(!useName && uuid){
             return uuid
         }
-        return (null==jobGroup?'':jobGroup+"/")+jobName;
+        return (!jobGroup?'':jobGroup+"/")+jobName;
```
ronaveva, line 111: what is this change for?

rundeckapp/grails-app/domain/rundeck/Workflow.groovy
```diff
@@ -36,7 +36,7 @@ public class Workflow implements WorkflowData {
 
     Integer threadcount=1
     Boolean keepgoing=false
-    List<WorkflowStep> commands
+    List<WorkflowStepData> commands
```
ronaveva: i think we should not change the class here, we should keep using the domain clases since they should only be used to show the current jobs and its executions

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -1700,7 +1700,14 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
         if(execMap instanceof Execution) {
             executionReference = execMap.asReference()
         }
-        
+
+        // Get workflowData from execMap
+        def wfData = null
+        if(execMap instanceof Execution) {
```
ronaveva, line 1707: this validation seems irrelevant

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -1720,6 +1727,9 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
             executionListener(listener)
             workflowExecutionListener(wlistener)
             execution(executionReference)
+            if(wfData) {
```
ronaveva: could the workflow ever be null/empty?

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -2633,9 +2635,12 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
             }
         }
 
-        Workflow workflow = new Workflow(se.workflow)
-        //create duplicate workflow
-        props.workflow = workflow
+        // Clone workflow from job definition using new JSON storage approach
+        def workflowData = se.getWorkflowData()
+        if (workflowData) {
```
ronaveva: same as before, is it possible for the workflow to be null?

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -2633,9 +2635,12 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
...
+            Workflow workflow = new Workflow(workflowData)
```
ronaveva: why are we using Workflow domain class here? shouldn't we be using WorkflowDataImpl? or RdWorkflow

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -2657,10 +2662,9 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
             execution.setExtraMetadataMap(extraMeta)
         }
 
-        if (workflow && !workflow.save(flush:true)) {
-            execution.workflow.errors.allErrors.each { log.error(it.toString()) }
-            log.error("unable to save execution workflow")
-            throw new ExecutionServiceException("unable to create execution workflow")
+        // Set workflow data if we have a workflow
+        if (props.workflow) {
```
ronaveva: same as before, is it possible for the workflow to be null?

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -4689,8 +4700,7 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
         missed.user = scheduledExecution.user
         missed.userRoleList = scheduledExecution.userRoleList
         missed.serverNodeUUID = scheduledServer
-        missed.workflow = new Workflow()
-        missed.workflow.save()
+        missed.setWorkflowData(new Workflow())
```
ronaveva: what happens if we just leave the workflow data empy?
if needed, would be better to use a non domain class instead

rundeckapp/grails-app/services/rundeck/services/NotificationService.groovy
```diff
@@ -998,6 +998,9 @@ public class NotificationService implements ApplicationContextAware{
                         storageTree(storageService.storageTreeWithContext(evt.authContext))
                         framework(frameworkService.rundeckFramework)
                         frameworkProject(job.project)
+                        if(job.getWorkflowData()) {
```
ronaveva, line 1002: the workflowData to use should come from the execution and not the scheduledExecution

rundeckapp/grails-app/domain/rundeck/Execution.groovy
```diff
@@ -572,5 +577,123 @@ class Execution extends ExecutionContext implements EmbeddedJsonData, ExecutionD
...
     }
 
+    /**
+     * Get the workflow data, checking both old and new storage formats.
+     * Provides backwards compatibility by checking workflow field first.
+     * @return WorkflowData instance (either Workflow domain or deserialized from JSON)
+     */
+    WorkflowData getWorkflowData() {
+        // Backwards compatibility: check old workflow field first
+        if (workflow != null) {
```
ronaveva: it would be safer to check for the workflowJson first, this way we would check for the newest first and not rely on the process to actually clean the old workflow field

rundeckapp/grails-app/services/rundeck/services/WorkflowService.groovy
```diff
@@ -42,6 +42,8 @@ import com.google.common.cache.Cache
 import com.google.common.cache.CacheBuilder
 import grails.converters.JSON
 import grails.util.Environment
+import org.rundeck.app.data.model.v1.job.workflow.WorkflowData
```
ronaveva: looks like these imports are not being used

rundeckapp/src/main/groovy/org/rundeck/app/data/job/converters/ExecutionToRdExecutionConverter.groovy
```diff
@@ -2,6 +2,7 @@ package org.rundeck.app.data.job.converters
 
 import com.fasterxml.jackson.databind.ObjectMapper
 import rundeck.Execution
+import rundeck.Workflow
```
ronaveva: is this import being used?

rundeckapp/src/main/groovy/org/rundeck/app/data/workflow/WorkflowStepDataImpl.groovy
```diff
@@ -0,0 +1,157 @@
...
+
+import java.util.Collection
+
+/**
+ * Implementation of WorkflowStepData for deserialized workflow step data.
+ * This class represents workflow step data that has been deserialized from JSON storage.
+ * It provides a runtime representation without GORM persistence overhead.
+ */
+@JsonIgnoreProperties(["errors"])
+class WorkflowStepDataImpl implements WorkflowStepData, PluginProviderConfiguration, Validateable  {
```
ronaveva, line 24: we already have a RdWorkflowStep class, why not use that one?

rundeck-data-models/src/main/java/org/rundeck/app/data/model/v1/job/workflow/WorkflowData.java
carlosrfranco: It was added just to avoid to change the workflow.command calls in a lot of places in the project and to keep compatibility

rundeckapp/grails-app/controllers/rundeck/controllers/ScheduledExecutionController.groovy
carlosrfranco: Yes, it should be fixed to use WorkflowData

rundeckapp/grails-app/domain/rundeck/JobExec.groovy
carlosrfranco: It was made to fix a case where JobExec is created with jobGroup as an empty string. It makes the jobIdentifier to be something like "/jobName" and it doesn't match with the real job name

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
carlosrfranco: The if clause can be removed

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
carlosrfranco: The if clause can be removed

rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
carlosrfranco: Yes it should. I missed to change this place

rundeckapp/src/main/groovy/org/rundeck/app/data/workflow/WorkflowDataImpl.groovy
carlosrfranco: Honestly, I just realized that after started working on these changes. I will evaluate the size of this refactoring to use RdWorkflow

rundeckapp/grails-app/controllers/rundeck/controllers/ScheduledExecutionController.groovy
```diff
@@ -2718,7 +2722,8 @@ Authorization required: `delete` on project resource type `job`, and `delete` on
             return [success:false,failed:true,error:'disabled',message:msg]
         }
 
-        params.workflow=new Workflow(scheduledExecution.workflow)
+        def jobWorkflowData = scheduledExecution.getWorkflowData()
+        params.workflow=jobWorkflowData ? new Workflow(jobWorkflowData) : new Workflow()
```
ronaveva: change to use the non domain class

rundeck-data-models/src/main/java/org/rundeck/app/data/model/v1/execution/ExecutionData.java
```diff
@@ -58,6 +58,7 @@ public interface ExecutionData {
     WorkflowData getWorkflow();
     OrchestratorData getOrchestrator();
```
ronaveva, line 59: this seems to be irrelevant now, we should only keep "getWorkflowData" method

rundeckapp/src/main/groovy/org/rundeck/app/data/workflow/WorkflowDataImpl.groovy
```diff
@@ -0,0 +1,308 @@
...
```
ronaveva, line 27: we should remove this class and use RdWorkflow class only

rundeckapp/src/main/groovy/org/rundeck/app/data/workflow/WorkflowStepDataImpl.groovy
ronaveva: we should remove this class and use RdWorkflowStep class only

rundeckapp/src/main/groovy/org/rundeck/app/data/workflow/WorkflowDataImpl.groovy
carlosrfranco: We can make it in a separated ticket as it is not a small change

rundeck-data-models/src/main/java/org/rundeck/app/data/model/v1/execution/ExecutionData.java
carlosrfranco: I'm going to keep this for now because there are some places that use this to access workflow to set it to json format

PR #9970, opened 2026-01-26
RUN-4102: Add groupBy annotation for plugins and update endpoints to return it

smartinellibenedetti:
## Tell us about your PR

In order for the new modal to show the step plugins grouped it is required to add a new plugin annotation to make it simple to categorize them. Besides updating the plugins endpoint to return the metadata when api version is 57. 

**Additional context**

## Release Notes

Adds support for plugin grouping metadata (groupBy) and exposes it (plus resolved groupIconUrl) via plugin listing responses for API v57+.

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/PluginApiService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/PluginApiService.groovy
@@ -332,7 +332,7 @@ class PluginApiService {
                                 absolute: true)
 
                     if(uiDesc.providerMetadata) {
-                        pluginDesc.providerMetadata.put(uiDesc.providerMetadata)
+                        pluginDesc.providerMetadata.putAll(uiDesc.providerMetadata)
                     }
                 }
 
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/PluginController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/PluginController.groovy
@@ -202,6 +202,7 @@ Since: v33
     def listPlugins() {
         String service = params.service
         boolean includeGroupMetadata = request.api_version >= ApiVersions.V57
+        boolean includeV40Fields = request.api_version >= ApiVersions.V40
 
         def providers = []
         pluginApiService.listPlugins(includeGroupMetadata).each { svc ->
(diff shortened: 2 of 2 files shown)


rundeckapp/grails-app/services/rundeck/services/PluginApiService.groovy
```diff
@@ -323,7 +331,55 @@ class PluginApiService {
                                 absolute: true)
 
                     if (uiDesc.providerMetadata) {
-                        pluginDesc.providerMetadata = uiDesc.providerMetadata
+                        pluginDesc.providerMetadata = new LinkedHashMap<>(uiDesc.providerMetadata)
+                    } else {
```
ronaveva: might be better to set the providerMetadata as empty when instantiating the pluginDesc map, this way we would not need that many else statements

rundeckapp/src/main/groovy/com/dtolabs/rundeck/app/api/plugins/ApiPluginListProvider.groovy
```diff
@@ -19,7 +19,7 @@ class ApiPluginListProvider {
     @Schema(description = 'URL to icon file for the plugin if present. Since: v40')
     String iconUrl
     @ApiVersion(40)
-    @Schema(description = 'Map of metadata about the plugin if present. Since: v40')
+    @Schema(description = 'Map of metadata about the plugin if present. Includes groupBy and groupIconUrl in v57+. Since: v40')
```
ronaveva:
```suggestion
    @Schema(description = 'Map of metadata about the plugin if present. Includes groupBy and groupIconUrl from v57+. Since: v40')
```

ronaveva (changes requested):
Minor observations

rundeckapp/grails-app/services/rundeck/services/PluginApiService.groovy
```diff
@@ -323,7 +331,55 @@ class PluginApiService {
...
+
+                    String groupIconUrl = null
+
+                    if (groupIconCache.containsKey(groupBy)) {
+                        groupIconUrl = groupIconCache[groupBy]
+                    } else {
+                        def pluginGroupType = provider.pluginGroupType
+                        if (pluginGroupType) {
+                            try {
+                                def method = pluginGroupType.getMethod('getGroupIconUrl', String.class)
```
ronaveva: it might be better to modify the PluginGroup.java interface to include the "getGroupIconUrl" with a default implementation that returns null, this way we would not need the use of reflection

PR #10004, opened 2026-03-05
[RUN-4196] Fix intermittent API token 403 with parallel requests

### LINKED ISSUE #9643: Parallel httpx.AsyncClient Requests to Rundeck 5.11.1 Intermittently Fail with 403 ‘Not Authorized’ Errors
**Describe the bug**
When using the Python `httpx.AsyncClient` to send multiple `/api/44/job/{jobId}` requests to a Rundeck 5.11.1 instance in parallel, some requests fail with a 403 error:

```
{
  "error": "admin is not authorized for: /api/44/job/<jobId>"
}
```

However, if each request is made sequentially with a brand new `AsyncClient` (fresh TLS handshake per request), all requests succeed.

**My Rundeck detail**

- Rundeck version: **5.11.1**
- Install type: **Photon OS**
- OS Name/version:
- DB Type/version: **PostgreSQL**

**To Reproduce**

1. Configure a single `httpx.AsyncClient`:
   ```python
   async with AsyncClient(verify=False) as client:
       results = await asyncio.gather(*[
           client.get(f"{BASE_URL}/api/44/job/{job_id}")
           for job_id in job_ids
       ])
   ```
2. Observe that some responses return HTTP 403 with the above error message.
3. Change to instantiate and close a new client for each request:

   ```python
   async def fetch_job(id):
       async with AsyncClient(verify=False) as client:
           return await client.get(f"{BASE_URL}/api/44/job/{id}")

   results = await asyncio.gather(*(fetch_job(id) for id in job_ids))
   ```

4. All requests succeed without authorization errors.

**Expected behavior**
Parallel requests using a single `AsyncClient` instance should not cause intermittent authorization failures and should return HTTP 200 for all job queries.

**Additional context**

- Creating a new client per request works around the issue but incurs significant overhead due to repeated TLS handshakes.
- Requests are authenticated using a Rundeck API token.
- Likely related to HTTP connection pooling or HTTP/2 multiplexing being rejected by the Rundeck server or load balancer.

Lukas20039:
## Tell us about your PR

Bugfix.

Fixes intermittent `403 Unauthorized` API responses when using a single pooled client (`httpx.AsyncClient`) to issue multiple `/api/44/job/{jobId}` requests in parallel with token auth.

Fixes #9643

`SetUserInterceptor` previously only performed token authentication when `session.user` was empty:

- old condition: `request.api_version && !session.user && authtoken`

With parallel requests, `session.user` can already be populated by another in-flight request, so token auth was skipped for valid token requests and the request could be marked invalid.

This change makes token auth always run for API requests that include an auth token:

- new condition: `request.api_version && authtoken`

This keeps authentication request-scoped and prevents cross-request session-state interference.

A regression test was added in `SetUserInterceptorSpec` to verify that API token requests still authenticate correctly even when `session.user` is already populated.

- Keeping the session gate (`!session.user`) and synchronizing session access, which is more complex and still fragile under concurrency.
- Adding endpoint-specific exceptions, which is incomplete and harder to maintain.

Always authenticating token-bearing API requests is the minimal and robust fix.

**Additional context**
Reported reproduction: parallel `httpx.AsyncClient` requests against Rundeck `5.11.1` intermittently failed with:

`admin is not authorized for: /api/44/job/<jobId>`

while sequential requests with fresh clients succeeded.

## Release Notes
Fixed intermittent API token authorization failures for parallel API requests by ensuring token authentication is always evaluated for token-bearing API calls, even if `session.user` is already set.


jayas006 (approved):
LGTM

PR #10044, opened 2026-04-14
[RUN-4320] Fix PostgreSQL SQLGrammarException caused by Grails 7 backtick column mappings

Jesus-Osuna-M:
## Summary

Fixes runtime `SQLGrammarException` errors when running Rundeck with a PostgreSQL backend. Queries against `scheduled_execution` failed because Hibernate generated double-quoted uppercase identifiers (e.g. `"MINUTE"`) that did not match the actual lowercase column names stored in PostgreSQL.

## Root Cause

The Grails 7 upgrade introduced two changes that collided and broke PostgreSQL:

### Change 1: New naming strategy (Grails 7 / Spring Boot 3.x)

```yaml
# BEFORE (Grails 5/6 — Spring Boot 2.x)
physical-strategy: org.springframework.boot.orm.jpa.hibernate.SpringPhysicalNamingStrategy

# AFTER (Grails 7 — Spring Boot 3.x)
physical-strategy: org.hibernate.boot.model.naming.CamelCaseToUnderscoresNamingStrategy
```

`CamelCaseToUnderscoresNamingStrategy` respects backtick-quoted identifiers as force-quoted, converting them to double-quoted identifiers in PostgreSQL. The old `SpringPhysicalNamingStrategy` ignored backticks.

### Change 2: Backtick mappings added for H2 (commit `315112c580`)

The commit "Grails 7: Critical fixes for BC encryption, UI plugins, and SQL compatibility" added:

```groovy
minute  column: "`MINUTE`"
hour    column: "`HOUR`"
month   column: "`MONTH`"
seconds column: "`SECONDS`"
year    column: "`YEAR`"
```

These were added for supposed H2 compatibility with reserved SQL keywords. However, the correct fix for H2 reserved keywords is the `NON_KEYWORDS` JDBC parameter — already present in `application.groovy` since the H2 2.x upgrade in 2022, but missing from `DataSource.groovy`.

### How the two changes collided

| State | Hibernate generates | PostgreSQL sees | Result |
|---|---|---|---|
| **Before Grails 7** (no backticks, old strategy) | `this_.minute` | Finds column `minute` | ✅ Works |
| **After Grails 7** (backticks + new strategy) | `this_."MINUTE"` | Looks for exact column `MINUTE` | ❌ Fails |

PostgreSQL treats double-quoted identifiers as case-sensitive, so `"MINUTE"` ≠ `minute`.

**Error observed:**
```
ERROR: column schedulede0_.MINUTE does not exist  Position: 219
org.hibernate.exception.SQLGrammarException: could not extract ResultSet
```

### 1. Remove backtick column mappings from `ScheduledExecution.groovy`

Without them, Hibernate generates unquoted lowercase identifiers (`minute`, `hour`, etc.) that match the existing column names on all supported databases.

### 2. Add `NON_KEYWORDS` to `DataSource.groovy` test URL

Ensures consistency with `application.groovy`, which already had this parameter. With `NON_KEYWORDS=MONTH,HOUR,MINUTE,YEAR,SECONDS` set in both config files, H2 treats these words as regular identifiers and Liquibase creates columns as unquoted uppercase (`MINUTE`), matching what Hibernate generates.

### 3. Remove Liquibase rename changesets

The H2-only rename changesets (`4.6.0-1` through `4.6.0-4`) and the PostgreSQL-specific rename changesets (`4.6.0-1-pg` through `4.6.0-5-pg`) are removed. With `NON_KEYWORDS` consistently set, the schema is created correctly from the start without needing column renames.

## Final behavior

| Database | Column created as | Hibernate generates | Result |
|---|---|---|---|
| **PostgreSQL** | `minute` (lowercase) | `minute` | ✅ |
| **MySQL** | case-insensitive | `minute` | ✅ |
| **H2 (test)** | `MINUTE` (uppercase, via `NON_KEYWORDS`) | `minute` → normalized to `MINUTE` | ✅ |

## Changes

- `DataSource.groovy`: added `NON_KEYWORDS=MONTH,HOUR,MINUTE,YEAR,SECONDS` to H2 test URL for consistency with `application.groovy`
- `ScheduledExecution.groovy` (domain): removed backtick column mappings for `minute`, `hour`, `month`, `seconds`, `year`
- `ScheduledExecution.groovy` (migrations): removed H2-only rename changesets (`4.6.0-1` through `4.6.0-4`) and PostgreSQL rename changesets (`4.6.0-1-pg` through `4.6.0-5-pg`)

## Test plan

- [x] Start Rundeck with a fresh PostgreSQL 17 database — no `SQLGrammarException` on startup
- [x] Verify jobs can be listed and scheduled
- [x] Verify `rescheduleJobs` completes without errors in startup logs
- [x] Confirm H2 test suite passes


ronaveva (commented):
LGTM

ronaveva (approved):

PR #10052, opened 2026-04-16
[RUN-4287] Fix conditional fields to accept data and global variables

carlosrfranco:
## Jira Ticket

[RUN-4287](https://rundeckpro.atlassian.net/browse/RUN-4287)

## Summary

Fix variable replacement in conditional step plugin entries so that global variables (${export.*}) and data values (${data.*}) are properly replaced with their values during job execution.

### Core Implementation
- **ConditionalEvaluator**: Updated to use SharedDataContext instead of DataContext for scope-aware variable expansion
- **Strategy Profiles**: Implemented hybrid evaluation approach with context accumulation:
  - Added `updatedSharedContext` field to accumulate context from all completed steps
  - Merge each completed step's context into accumulated context
  - Evaluate conditionals for ALL steps after each step completes
  - Use accumulated context for evaluations
- **ConditionalExecutionService**: Updated node step evaluation to use SharedDataContext

### OSS Core Changes (rundeck submodule)
- **WorkflowStrategyProfile**: Added `updateConditionalStatesAfterStep()` method
- **StepOperation**: Calls profile method after each step completes
- **EngineWorkflowExecutor**: Passes profile to StepOperation constructor
- **ExecutionService**: Removed conditions parameter from createExecutionItemForWorkflow

### Testing
- **Unit Tests**: Added 44 tests in ConditionalEvaluatorSpec covering:
  - ${export.*} global variables
  - ${data.*} from log filters
  - ${1:data.*} step-specific references
  - Nested patterns (verified NOT expanded)
  - Backward compatibility for ${option.*}, ${job.*}, ${node.*}
- **Functional Tests**: Added 2 end-to-end tests in ConditionalWorkflowProSpec:
  - Export variable test with key-value-data log filter
  - Data variable test with pattern-based log filter

## Technical Approach

The fix uses a hybrid evaluation approach:
1. Conditionals require state set BEFORE rules engine checks conditions
2. After each step completes, evaluate conditionals for future steps
3. Accumulate SharedDataContext across steps for complete data access
4. Update workflow state with evaluation results to trigger rules engine

This enables conditionals to access ${export.*} and ${data.*} from previous steps while working within the rules engine architecture.

[RUN-4287]: https://pagerduty.atlassian.net/browse/RUN-4287?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/execution/workflow/EngineWorkflowExecutor.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/execution/workflow/EngineWorkflowExecutor.java
@@ -577,7 +577,8 @@ public class EngineWorkflowExecutor extends BaseWorkflowExecutor {
                             stepRunTriggerState,
                             stepSkipTriggerState,
                             stepStartTriggerConditions,
-                            stepSkipTriggerConditions
+                            stepSkipTriggerConditions,
+                            profile
                     )
             );
         }
--- a/core/src/main/java/com/dtolabs/rundeck/core/execution/workflow/WorkflowStrategyProfile.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/execution/workflow/WorkflowStrategyProfile.java
@@ -1,6 +1,7 @@
 package com.dtolabs.rundeck.core.execution.workflow;
 
 import com.dtolabs.rundeck.core.rules.Condition;
+import com.dtolabs.rundeck.core.rules.MutableStateObj;
 import com.dtolabs.rundeck.core.rules.StateObj;
 
 import java.util.Set;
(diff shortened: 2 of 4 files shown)


smartinellibenedetti:
### Code review

Found 1 issue:

1. `updateConditionalStatesAfterStep()` is a no-op — the fix mechanism is inert at runtime. The new hook is defined as a default no-op in `WorkflowStrategyProfile` and called in `StepOperation.apply()`, but neither `ConditionalSequentialStrategyProfile` nor `ConditionalParallelStrategyProfile` override it. Conditionals are still evaluated once at initialization via `loadConditionals()` in the constructor. Without an implementation that accumulates and re-evaluates context, `${data.*}` and `${export.*}` references in conditionals will never resolve against completed-step data — which is the stated goal of this PR.

https://github.com/rundeck/rundeck/blob/fd19dd5aa643e41adaadb49dccd11d907cda2822/core/src/main/java/com/dtolabs/rundeck/core/execution/workflow/WorkflowStrategyProfile.java#L53-L69

https://github.com/rundeck/rundeck/blob/fd19dd5aa643e41adaadb49dccd11d907cda2822/core/src/main/java/com/dtolabs/rundeck/core/execution/workflow/engine/StepOperation.java#L140-L150

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

smartinellibenedetti (approved):

PR #10067, opened 2026-04-21
[RUN-4320] Fix H2 reserved keyword column name errors for PostgreSQL compatibility

Jesus-Osuna-M:
## Problem

When removing backtick column mappings (e.g. `minute column: "\`MINUTE\`"`) from the `ScheduledExecution` domain class — which were breaking PostgreSQL since backtick quoting is not valid PostgreSQL syntax — H2 integration tests started failing with:

```
SQLGrammarException: Column "SCHEDULEDE0_.MINUTE" not found
```

**Root cause:** Liquibase's internal reserved keyword list includes words like `month`, `hour`, `year`, `minute`. When these are used as column names in H2 without the `NON_KEYWORDS` JDBC parameter, Liquibase auto-quotes them as lowercase case-sensitive identifiers (e.g. `"month"`). Hibernate, however, uses `CamelCaseToUnderscoresNamingStrategy` and generates unquoted uppercase identifiers (e.g. `MONTH`), causing a column name mismatch.

The existing H2-only Liquibase changesets (`4.6.0-1` to `4.6.0-4`) that rename these columns used `columnExists` as a precondition, which is **case-insensitive** in H2. This meant the rename was attempted even on clean installs where the column was already created as uppercase — causing `ALTER TABLE ... RENAME COLUMN "month" TO MONTH` to fail with *Column "month" not found*.

## Solution

- **Remove backtick column mappings** from `ScheduledExecution` static mapping (fixes PostgreSQL)
- **Update H2 Liquibase changesets** (`4.6.0-1` to `4.6.0-4`): replace the `columnExists` precondition with a `grailsPrecondition` that runs a case-sensitive SQL query against `INFORMATION_SCHEMA.COLUMNS`. The rename only executes when the column genuinely exists as a lowercase quoted identifier (legacy installs). On clean installs where `NON_KEYWORDS` is set and the column is already uppercase, the precondition skips the rename safely.

## Affected databases

| Database | Impact |
|---|---|
| PostgreSQL | Fixed — backtick syntax no longer used |
| H2 (legacy install) | Fixed — rename changeset runs correctly when lowercase column exists |
| H2 (clean install) | Fixed — rename changeset is safely skipped when column is already uppercase |

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -317,7 +317,7 @@ jobs:
     <<: *job-defaults
     executor:
       name: docker-builder
-      resource_class: medium
+      resource_class: large
     steps:
       - setup_docker
       - checkout
--- a/rundeckapp/grails-app/domain/rundeck/ScheduledExecution.groovy
+++ b/rundeckapp/grails-app/domain/rundeck/ScheduledExecution.groovy
@@ -190,13 +190,6 @@ class ScheduledExecution extends ExecutionContext implements JobData, EmbeddedJs
         pluginConfig(type: 'text')
         workflowJson(type: 'text')
         lastModifiedBy(type: 'string')
-        // Escape SQL reserved words with backticks for H2 compatibility
-        minute column: "`MINUTE`"
-        hour column: "`HOUR`"
-        month column: "`MONTH`"
-        seconds column: "`SECONDS`"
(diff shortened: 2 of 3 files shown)


ronaveva (approved):
LGTM

PR #8305, opened 2023-05-05
Vue3 compat build

sjrd218:
Convert Vue2 to Vue3 using @vue/compat package.

Update deprecated Vue2 api to Vue 3 apis.

PR #10131, opened 2026-05-06
[RUN-4609] ui-trellis production minification

yoshi-taka:
## Tell us about your PR

Enhancement.

The ui-trellis library webpack config currently disables both chunk splitting and minimization.

While preserving the existing non-split output structure may still be important for downstream consumers, production builds do not necessarily need to disable webpack's default minimizer as well.

This PR re-enables minimization for production builds only.

- Keep splitChunks = false unchanged
- Preserve the current non-production unminified behavior for easier debugging
- Allow production builds to use webpack's default minimizer again

This intentionally does not change chunk splitting behavior or library output structure.

Describe alternatives you've considered

One alternative was restoring webpack optimization defaults entirely, including chunk splitting.

However, previous bundling-related fixes suggest that some downstream consumers may still rely on the existing per-entry output behavior. Because of that, this PR intentionally limits the scope to production minimization only.

Another option considered was always enabling minimization regardless of environment, but preserving unminified non-production builds seemed safer and more convenient for debugging.

Some of the earlier changes also appear to have been introduced during the Vue 3 compatibility / migration period, where conservative bundling behavior may have been safer. However, the current ui-trellis package is already on modern Vue 3 dependencies, so re-enabling production minimization now appears lower risk than it may have been at the time.

Additional context

Relevant historical changes include:

- ccdc0afd74 — RCLOUD-888: Vue3 compat build (#8305)
- e839d3a7fe — Remove minimizer
- 0f1212ad1f — RUN-3968: Fix webpack bundling configuration (#9916)

This PR keeps the compatibility-sensitive parts of the current configuration intact while narrowing the scope of disabled optimizations.

## Release Notes

Re-enable webpack minimization for production ui-trellis library builds while preserving existing non-split bundle behavior.


smartinellibenedetti:
Thanks for digging into this. I agree that minification shouldn't just be off, that part's right. Closing for now though: `ui-trellis` has other repos depending on it (e.g. rundeckpro), so this needs to account for that before it goes in. Also, we're looking at moving to Vite, so investing more time tuning the webpack config isn't the priority right now with everything else that needs sorting first. We'll revisit this once that's clearer. Appreciate the effort on this!

PR #10145, opened 2026-05-12
[RUN-4253] Fix schedule owner update regression

jayas006:
## Tell us about your PR

Bugfix for [RUN-4253 ](https://pagerduty.atlassian.net/browse/RUN-4253)- Schedule owner update regression 

Removes conditional checks on user field assignment in ScheduledExecutionService to restore the previous behavior where the schedule owner (user/roles) updates  
  to whoever modifies the job. This prevents stuck schedule owners that cause job execution failures when the original creator is deleted or loses permissions.

 Initially considered removing 'user' from EXCLUDED_AUDIT_FIELDS entirely, but kept it there for security - this prevents malicious job imports from overwriting  
  schedule owners while still allowing normal UI edits to update the owner.  

**Additional context**


ronaveva (approved):
LGTM

PR #10146, opened 2026-05-13
[RUN-3932] Make UUID field read-only in job Other tab

Jesus-Osuna-M:
## Release Notes
UUID is no longer be editable on Jobs Other tab

## Summary
- Removes the editable `<input>` for UUID in the job editor's **Other** tab
- The UUID field now always renders as read-only static text
- **New job (no UUID yet):** shows *Auto-generated on save* placeholder
- **Existing job:** shows the current UUID as read-only text

## Root cause
The field used two separate `v-if` / `v-if` blocks (not `v-if` / `v-else`). When creating a new job the input rendered because `modelData.uuid` was falsy. Typing a single character made it truthy, instantly destroying the `<input>` and rendering the static `<p>` — making the field appear frozen after one keystroke.

## Jira
[RUN-3932](https://pagerduty.atlassian.net/browse/RUN-3932)

## Test plan
- [x] Create a new job → Other tab → UUID field is visible but not editable, shows *Auto-generated on save*
- [x] Save the job → reopen it → UUID field shows the auto-generated UUID as read-only text
- [x] Confirm no input/editable field is present in either scenario

[RUN-3932]: https://pagerduty.atlassian.net/browse/RUN-3932?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ronaveva (approved):
LGTM

PR #10154, opened 2026-05-19
[RUN-4291] Fix cron expression selector not updating in Execution History Clean

Jesus-Osuna-M:
## Tell us about your PR

This is a bugfix. In the Project Settings under **Execution History Clean**, the predefined cron expression selector list fails to update the "Schedule clean history job (Cron expression)" text field when the user clicks an option. The field retains its initial value regardless of the selection, forcing users to type the cron expression manually.

Fixes [RUN-4291](https://pagerduty.atlassian.net/browse/RUN-4291)

The root cause is a JavaScript library mismatch. `editProject.gsp` declares `<meta name="skipPrototypeJs" content="true"/>`, which prevents Prototype.js from loading. However, `_editProjectForm.gsp` was using Prototype.js-specific syntax throughout the cron expression section:

- `$('cronTextField')` — Prototype's `$()` finds elements by ID. Without Prototype, this falls through to jQuery's `$()`, which interprets `'cronTextField'` as a **tag-name selector**, finds nothing, and silently no-ops.
- `$F(elem)` — Prototype's form field value helper; throws `TypeError` when Prototype is absent.
- `new Ajax.Updater(...)` — Prototype's AJAX helper; not available without Prototype.
- `.size()` — Prototype's Array extension; not available in plain JS.

All Prototype.js calls have been replaced with vanilla JS / jQuery equivalents:

| Before (Prototype.js) | After |
|---|---|
| `$('cronTextField').value = this.value` | `document.getElementById('cronTextField').value = this.value` |
| `$('crontooltip')` | `document.getElementById('crontooltip')` |
| `$F(elem)` | `elem.value` |
| `new Ajax.Updater(...)` | `jQuery.ajax(...)` |
| `sub.split(' ').size()` | `sub.split(' ').length` |

Removing the `skipPrototypeJs` meta tag from `editProject.gsp` was considered, but re-introducing Prototype.js on a page that already uses jQuery exclusively would risk regressions. The correct fix is to modernize the isolated cron script block.

## Release Notes

Fix cron expression selector in Project Settings > Execution History Clean. Selecting a predefined cron option now correctly updates the schedule input field.

[RUN-4291]: https://pagerduty.atlassian.net/browse/RUN-4291?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ronaveva (approved):
LGTM

fdevans:
This had unit test fails related to `checkCrontab`.  Let's check those out before finalizing reviews

PR #10155, opened 2026-05-19
[RUN-2811] Fix KeyStorageSelector state not resetting on reopen

Jesus-Osuna-M:
## Tell us about your PR

Bugfix. When creating a PagerDuty V3 webhook in Rundeck, selecting a key from the Key Storage popup (Advanced tab → Shared Secret Key → "Select") closes the modal but the field remains empty.

Fixes [RUN-2811](https://pagerduty.atlassian.net/browse/RUN-2811)

`KeyStorageSelector.vue` is a Vue 3 component that emits `update:modelValue`. Its `openSelector()` method did not call `clean()` before opening, meaning that if a user opened the selector, made a selection, then cancelled and reopened, the internal `selectedKey` state still held the previous (unsaved) value rather than the current `modelValue`.

Added a `clean()` call inside `openSelector()` so the component always initialises `selectedKey` from the current `modelValue` before the modal is shown.

Note: the main binding fix (Vue 2 `@input`/`:value` → Vue 3 `@update:model-value`/`:model-value`) is in the companion PR in `rundeckpro/rundeckpro`.

Watching `modelValue` to keep `selectedKey` in sync — rejected as unnecessarily complex for the problem at hand.

## Release Notes

Fix PagerDuty V3 webhook: selecting a key from the Key Storage selector now correctly populates the Shared Secret Key field.

[RUN-2811]: https://pagerduty.atlassian.net/browse/RUN-2811?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/KeyStorageSelector.vue
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/KeyStorageSelector.vue
@@ -120,6 +120,7 @@ export default defineComponent({
       this.modalOpen = false;
     },
     openSelector() {
+      this.clean();
       this.modalOpen = true;
     },
     openEditor(uploadSetting: UploadSetting) {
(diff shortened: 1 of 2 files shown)


ronaveva (approved):
LGTM

PR #10157, opened 2026-05-20
[RUN-3050] Fix node filter attributes incorrectly wrapped in double quotes

Jesus-Osuna-M:
## Tell us about your PR

Bugfix. When a user saves a node filter containing multiple attributes (e.g. `osFamily: unix name:localhost`) or attributes with already-quoted values (e.g. `osArch: "aarch64" osName: "Mac OS X"`), Rundeck incorrectly wraps the entire remainder of the filter string in double quotes, treating it as a single attribute value instead of preserving the original multi-attribute structure.

Fixes [RUN-3050](https://pagerduty.atlassian.net/browse/RUN-3050)

The root cause is in `NodeFilterLink.getFilter()`. The method searches for the first `: ` separator to locate the attribute value, then checks whether the rest of the string contains a space. If it does, it wraps the entire remainder in quotes — without checking whether those spaces are attribute separators or part of a single value.

This causes two failure modes:

1. **Multi-attribute filter quoted as one:** `osFamily: unix name:localhost` → `osFamily: "unix name:localhost"`
2. **Already-quoted values double-quoted:** `osArch: "aarch64" osName: "Mac OS X"` → `osArch: ""aarch64" osName: "Mac OS X""`

**Fix:** Before quoting, two guards are now checked:

- **`alreadyQuoted`** — if the value portion starts with `"`, it is already quoted; return as-is to prevent double-quoting.
- **`hasMultipleAttrs`** — if the value portion matches `/\s+!?\w[\w.-]*:/`, it contains another attribute key (e.g. ` name:`, ` !tags:`, ` osName:`); this is a multi-attribute filter and must not be quoted.

Only when neither guard triggers and the value contains spaces is it wrapped in quotes — which is the correct behaviour for a single attribute with a space-containing value (e.g. `osName: Mac OS X` → `osName: "Mac OS X"`).

The regex handles all Rundeck filter attribute name variants: standard names, custom names with hyphens/dots (`custom-attr`), namespace attributes (`ui:status:text`), and exclude prefixes (`!tags`).

Removing quoting entirely was considered but rejected — single-attribute values with spaces (e.g. `osName: Mac OS X`) genuinely need quotes to be parsed correctly by the Rundeck filter engine.

## Release Notes

Fix node filter: saving a filter with multiple attributes (e.g. `osFamily: unix name:localhost`) no longer incorrectly wraps them in double quotes. Pre-quoted values are also preserved correctly when re-loading a saved filter.

[RUN-3050]: https://pagerduty.atlassian.net/browse/RUN-3050?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/resources/NodeFilterLink.vue
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/resources/NodeFilterLink.vue
@@ -102,18 +102,28 @@ export default defineComponent({
 
     getFilter() {
       if (this.nodeFilter) {
-        let nodeFilterCpy = this.nodeFilter.trim();
-        const idxKey = nodeFilterCpy.indexOf(": ") + 2;
-        if (nodeFilterCpy.slice(idxKey).includes(" ")) {
-          nodeFilterCpy += '"';
-          nodeFilterCpy =
-            nodeFilterCpy.slice(0, idxKey) + '"' + nodeFilterCpy.slice(idxKey);
-
+        const nodeFilterCpy = this.nodeFilter.trim();
+        const sepIdx = nodeFilterCpy.indexOf(": ");
+        if (sepIdx < 0) {
           return nodeFilterCpy;
         }
-        return this.nodeFilter;
+        const valueStart = sepIdx + 2;
+        const valueAndRest = nodeFilterCpy.slice(valueStart);
+        // Already quoted — leave as-is to avoid double-quoting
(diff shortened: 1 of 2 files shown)


smartinellibenedetti (approved):
LGTM, would suggest just addressing this comment to be safe (https://github.com/rundeck/rundeck/pull/10157/changes#r3270627208)

PR #10159, opened 2026-05-21
i18n: sync ja_JP.js with messages_ja.properties for new UI keys

tt-pd:
## Tell us about your PR

Enhancement. Keys actively referenced by `$t()` in the new Vue UI (`ja_JP.js`)
were missing or had untranslated (English) values, causing Japanese users
to see English text in those UI elements.

Synced `ja_JP.js` with `messages_ja.properties` (the legacy Grails i18n
file, which contains a more complete Japanese translation set).

**What changed in `ja_JP.js`:**

- 4 keys overwritten with Japanese translations that existed in
  `messages_ja.properties` but were English in `ja_JP.js`:
  `Delete`, `execution`, `button.action.Cancel`, `button.action.Save`
- 92 keys added: present in `messages_ja.properties` with Japanese
  translation, actively used by `$t()` in the new UI, but missing
  from `ja_JP.js`
- Removed orphan key `Duplicate` (not present in `en_US.js`)
- Fixed `page.home.search.project.title` placeholder `{0}` → `{n}`
  to match `en_US` plural syntax
- Removed `app.firstRun.md` whose value in `messages_ja.properties`
  was stale OSS-era content incompatible with the current `en_US` value
- Sanitized values imported from `.properties` format:
  HTML entities decoded (`&hellip;` → `…` etc.),
  property separator escapes stripped (`\:` → `:`)

**Keys were excluded when any of the following applied:**

- No Japanese translation existed in `messages_ja.properties`
  (en_US fallback is sufficient; adding English-only strings adds noise)
- Adding the key would create a conflict with an existing nested key tree
  in `ja_JP.js` (a string key cannot coexist with an object key at the
  same path, e.g. existing `"filter.save.button"` blocks adding `"filter"`)
- The Japanese value in `messages_ja.properties` was found to have drifted
  from the current `en_US` meaning upon comparison (see below)

**Validation performed:**

1. **Key coverage**: Compared flat key sets of `ja_JP.js`, `en_US.js`,
   `fr_FR.js`, and `es_419.js`. After this change, `ja_JP` covers ~37%
   of `en_US` keys, above the ~28% average of `de_DE`/`zh_CN`/`pt_BR`.
   No orphan keys (keys in `ja_JP` absent from `en_US`) remain.

2. **Placeholder consistency**: Verified all `{0}`, `{1}`, `{n}` etc.
   placeholders in changed keys match `en_US`. The only plural-form
   divergence (`execution`, `Execution`) is intentional — Japanese has
   no plural forms.

3. **Nested key structure**: Confirmed no leaf/object type conflicts
   across all 266 nested keys in `ja_JP.js` against `en_US.js`.

4. **Semantic drift check**: All 400 keys present in both `ja_JP.js` and
   `en_US.js` were compared side-by-side against `fr_FR.js` and `es_419.js`.
   Two keys were identified where `messages_ja.properties` had drifted from
   the current `en_US` meaning and were excluded from this PR:
   - `form.option.secureExposed.true.label`: props has `"セキュア"` but
     `en_US` now reads `"Plain Text with Password Input"`
   - `job.filter.apply.button.title`: props has `"フィルタ"` but
     `en_US` now reads `"Search"`

5. **HTML entity / escape handling**: Confirmed values imported from
   `.properties` (which uses Unicode escapes and HTML entities) are
   correctly decoded before insertion into the JS file.

Syncing all keys from `messages_ja.properties` regardless of new-UI usage.
Rejected: would add ~1500 untested keys including stale content, with no
way to verify correctness without full UI review.

**Additional context**

Known pre-existing drift between `ja_JP.js` and `en_US.js` (out of scope):

| Key | en_US | ja_JP |
|-----|-------|-------|
| `scheduledExecution.property.orchestrator.description` | English only | Japanese prefix + English remainder (partial translation, never completed) |
| `scheduledExecution.property.defaultTab.description` | Full descriptive sentence | Truncated to label text only |

These originate from `messages_ja.properties` and predate this PR.

## Release Notes

Improved Japanese (`ja_JP`) translations in the Vue UI: added 92 missing
keys and corrected 4 untranslated values, sourced from the existing
`messages_ja.properties` translation file.


smartinellibenedetti (approved):
Thank you for your contribution @tt-pd ! I'll recreate your PR to pass local pipelines, in the meantime I'd like to ask you if you could check the copilot comments about two possible typos. If the comments are not applicable, please let me know.

tt-pd:
@smartinellibenedetti Thank you for the review. Incorporated the Copilot suggestion. The original escaped Japanese text already had typos.

smartinellibenedetti (approved):

PR #10160, opened 2026-05-22
i18n: add missing key gui.menu.FindPlugins to messages.properties and…

tt-pd:

Bugfix. The i18n key `gui.menu.FindPlugins` was missing from both `messages.properties` and `messages_ja.properties`, causing the menu item title to display the raw key name instead of a translated label when `pluginRead && repoEnabled` is true.

Added the missing key to both files:
- `messages.properties`: `gui.menu.FindPlugins=Find Plugins`
- `messages_ja.properties`: `gui.menu.FindPlugins=プラグインを検索`

Adjacent keys `gui.menu.InstalledPlugins` and `gui.menu.UploadPlugin` were already defined; this key was simply omitted.

Adding a `default=` attribute to the `g.message()` call in `_sysConfigNavMenuJson.gsp` as a fallback, but adding the key to the message bundle is the correct approach.

**Additional context**

Verified by code inspection. Display not confirmed due to lack of an environment with `repoEnabled=true`.

Note: `gui.menu.InstalledPlugins` and `gui.menu.UploadPlugin` remain untranslated in `messages_ja.properties` — out of scope for this fix.

**Release Notes**

Fix: menu item "Find Plugins" displaying raw i18n key name (`gui.menu.FindPlugins`) when plugin repository is enabled.

### DIFF
--- a/rundeckapp/grails-app/i18n/messages.properties
+++ b/rundeckapp/grails-app/i18n/messages.properties
@@ -699,6 +699,7 @@ gui.menu.SystemInfo=System Report
 gui.menu.Security=Security
 gui.menu.SystemConfig=System Configuration
 gui.menu.UserProfiles=User Profiles
+gui.menu.FindPlugins=Find Plugins
 gui.menu.InstalledPlugins=Installed Plugins
 gui.menu.UploadPlugin=Upload Plugin
 gui.menu.ListPlugins=List Plugins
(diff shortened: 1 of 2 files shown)


smartinellibenedetti (approved):
Thank you for your contribution, will recreate the PR locally to finalize the process to merge it.

PR #10161, opened 2026-05-22
fix(i18n): register dynamic messages into fallback locale (en_US) in …

tt-pd:

Bugfix. When `updateLocaleMessages` is called to dynamically register messages (e.g. from plugins via `addUiMessages`), messages are only added to the active locale (e.g. `ja_JP`). If a key has no translation in that locale, vue-i18n attempts to fall back to `en_US`, but since the key was never registered there either, the raw key name is displayed in the UI instead of the English label.

In `updateLocaleMessages`, after merging into the target locale, also merge the same messages into `en_US` (skipped when the target locale is already `en_US`). This ensures dynamically added keys are always resolvable via the fallback chain.

Requiring all callers of `updateLocaleMessages` / `addUiMessages` to explicitly register into `en_US` as well — but this would require changes at every call site (`ui/main.ts`, `storage/main.ts`, `uiSocketObserver.ts`, `navbar/main.ts`, etc.) and is error-prone. Fixing it once in `updateLocaleMessages` is safer.

**Additional context**

`addUiMessages` is the DI-injected function used by plugin components to register their i18n messages at runtime. The bug affects any non-`en_US` locale where a plugin provides a partial translation — keys missing from the locale file fall through to `en_US`, but since `en_US` also lacks them, the key name is rendered as-is.

Note: the three call sites that bypass `commonAddUiMessages` and call `updateLocaleMessages` directly (`ui/main.ts`, `storage/main.ts`, `uiSocketObserver.ts`) also benefit from this fix without any changes. Consolidating those to use `commonAddUiMessages` is a separate cleanup.

**Release Notes**

Fix: dynamically registered i18n messages (e.g. from plugins) now correctly fall back to English when a locale-specific translation is missing, preventing raw key names from appearing in the UI.

### DIFF
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/app/utilities/i18n.ts
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/app/utilities/i18n.ts
@@ -50,6 +50,13 @@ const updateLocaleMessages = async (i18n: I18n, locale: string, lang: string, me
   //include previously loaded messages
   const merged = mergeDeep((i18n.global.getLocaleMessage(locale) as Record<string, any>) || {}, messages);
   i18n.global.setLocaleMessage(locale, merged);
+  // Also register into the fallback locale (en_US) so that dynamically added
+  // messages (e.g. from plugins) are resolvable when a locale-specific
+  // translation is missing.
+  if (locale !== "en_US") {
+    const mergedFallback = mergeDeep((i18n.global.getLocaleMessage("en_US") as Record<string, any>) || {}, messages);
+    i18n.global.setLocaleMessage("en_US", mergedFallback);
+  }
   return nextTick();
 };
 /**


smartinellibenedetti:
Thanks for the contribution and the detailed write-up, the bug is real and we appreciate you tracking it down.

After reviewing this, we've decided to take a different approach before merging. The current fix copies the entire `messages` payload into `en_US`, which works today because plugins initially ship with English implementation. The problem is that `messages` has no locale structure, so as soon as a plugin adds translations for other locales, those translated strings would also get written into `en_US`, and English users would see the interface in the wrong locale.

The fix we want to land is deeper:
- Switch `updateLocaleMessages` to use `i18n.global.mergeLocaleMessage` (the idiomatic vue-i18n v10 API) instead of the manual `mergeDeep` + `setLocaleMessage` pattern.
- Widen `commonAddUiMessages` to accept a locale-keyed messages object (`{ en_US: {...}, ja_JP: {...} }`) so the function can register both the active locale and `en_US` explicitly, with correct content for each.
- Migrate all call sites — which turns out to span ~50 locations across the `rundeck` and `rundeckpro` repositories.

That's a larger refactor than what this PR signed up for, so we'll carry it forward internally. We're closing this PR but wanted to explain the reasoning rather than just declining. Thanks again for taking the time, the refactor will be done with the ticket RUN-4558 if you'd like to track it's progress. I can also let you know once changes are ready if you'd like to test them locally.

PR #10162, opened 2026-05-22
[RUN-2633] Fix Hostname capitalization in tool tip on node filter input

ddarby-hike:
## Tell us about your PR

This is a Bug Fix, originally reported by Syntax as Jira RUN-2633.  The tooltip for the node filter was referencing the Hostname name title which was capitalized, and also internationalized which it should not have been.  The fix was to not use the variable and hardcode the node filter example to be "hostname: ".

Correct "Hostname" case to be "hostname" (which is what the node filter requires).

None.

**Additional context**

None.

## Release Notes

Minor fix of capitalization on the word "hostname".

### DIFF
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/resources/NodeFilterInput.vue
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/resources/NodeFilterInput.vue
@@ -150,7 +150,7 @@
 
         <strong>{{ $t("use.regular.expressions") }}</strong>
         <p>
-          <code>{{ $t("node.metadata.hostname") }}: dev(\d+).test.com</code>.
+          <code>hostname: dev(\d+).test.com</code>.
         </p>
 
         <strong>{{ $t("regex.syntax.checking") }}:</strong>


rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/resources/NodeFilterInput.vue
```diff
@@ -150,7 +150,7 @@
 
         <strong>{{ $t("use.regular.expressions") }}</strong>
         <p>
-          <code>{{ $t("node.metadata.hostname") }}: dev(\d+).test.com</code>.
+          <code>hostname: dev(\d+).test.com</code>.
```
smartinellibenedetti, line 153: Hey, thank you for your contribution, could you please update the case here: https://github.com/rundeck/rundeck/blob/main/rundeckapp/grails-spa/packages/ui-trellis/src/app/utilities/locales/en_US.js#L637

It's better to update the value there instead of hardcoding it in the UI, I can revert the change for you in this file if you'd like.

ddarby-hike:
Hi Sarah,

  At a quick glance, your recommendation would seem correct to fix the
Hostname -> hostname.

  However, there are actually 2 Bugs here.  The first simply being the
capitalization correction.  But there is a 2nd bug.  The message being
displayed is an example REGEX expression in the help message for node
filters.  This REGEX should NOT be localized as Rundeck only accepts "
*hostname:* abc" as valid patterns regardless of the language a user is
using.  Therefore, the help message should not be using the
"node.metadata.hostname"
variable (which is localized) and should be hardcoded.  That variable is
also used elsewhere in Rundeck where it is expected to be Capitalized to
match the usage of the other "headers" (node.metadata.username,
node.metadata.tag).

  Lert me know if that makes sense or you still have questions.

Thanks,
-David


David Darby

Principal Professional Services Consultant

***@***.***

rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/resources/NodeFilterInput.vue
ddarby-hike: There are actually 2 bugs fixed here.  First is the case (Hostname -> hostname).  The 2nd fix is this is an example tooltip for using a REGEX pattern to filter nodes by "hostname: <abc>:".  The Variable "node.metadata.hostname" is localized based on the language and the localized value should NOT be used as the example regex.  Rundeck only understands US "hostname:" as the regex, thus hardcoding the tooltip to "hostname: " rather than using the localized variable is correct.

smartinellibenedetti (approved):

PR #10168, opened 2026-05-25
[RUN-3774] Improve audit log userInfo.username showing null or anonymous for job runs

Jesus-Osuna-M:
## Tell us about your PR

Bugfix. The `userInfo.username` field in audit logs for job run events shows `null` (scheduled jobs) or `__grails.anonymous.user__` (API/webhook triggers) instead of the real username.

Fixes [RUN-3774](https://pagerduty.atlassian.net/browse/RUN-3774)

When `createExecution()` published the audit event for `ActionTypes.RUN`, no username was explicitly set on the event builder. The dispatch logic in `AuditEventsService.dispatchEvent()` fell back to `SecurityContextHolder.context.authentication` to determine the username. This produces incorrect results in two scenarios:

- **Scheduled jobs (Quartz thread):** No web session exists, so `SecurityContextHolder` has no authentication → `username = null`
- **API / webhook calls:** The request may carry an `AnonymousAuthenticationToken` → `username = __grails.anonymous.user__`

The execution record (`newExec.user`) already holds the correct username at this point:
- Scheduled jobs: the user who last saved the job (set by `ExecutionJob.initRun()` from `scheduledExecution.user`)
- Interactive runs: the authenticated user
- API calls: the API token owner
- Run-as override: the `runAsUser` value

**Fix:** add `.setUsername(newExec.user)` to the audit event builder in `createExecution()` so the audit log always reflects the real execution user rather than the transient security context.

Normalising the fallback values (`null` → `"scheduler"`, `__grails.anonymous.user__` → `"anonymous"`) in `extractUsername()` was considered. Rejected because it still shows a synthetic value instead of the actual user — and existing SumoLogic/audit consumers may have filters on those strings that would silently break.

**Additional context**

Validated locally against three execution types:

| Exec | Type | `userInfo.username` before | After |
|---|---|---|---|
| Quartz scheduled | `scheduled` | `null` | `admin` |
| Webhook trigger | `user` | `__grails.anonymous.user__` | `admin` |
| API call | `user` | `__grails.anonymous.user__` | `admin` |

Evidence from `rundeck.audit.events.log` on local instance:
```
# Scheduled — sessionID='null', userAgent='null' confirms no web session
ActionType='run', UserInfo={username='admin', sessionID='null', userAgent='null'}, ...

# Webhook / API — now shows token owner instead of anonymous
ActionType='run', UserInfo={username='admin', sessionID='node0q4v...', userAgent='curl/8.7.1'}, ...
```

## Release Notes

Fix audit log `userInfo.username` for job run events. Scheduled jobs no longer show `null` and API/webhook-triggered jobs no longer show `__grails.anonymous.user__` — both now correctly reflect the user associated with the execution.

[RUN-3774]: https://pagerduty.atlassian.net/browse/RUN-3774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -2895,12 +2895,15 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
         Execution newExec = int_createExecution(se, authContext, runAsUser, input, securedOpts, secureExposedOpts)
 
         // Publish audit event for new job run.
+        // Explicitly set the username from the execution record so that the audit log
+        // always reflects the real user, even when there is no web session in the
+        // security context (e.g. Quartz-scheduled jobs, API/webhook calls).
         if(auditEventsService) {
             auditEventsService.eventBuilder()
                 .setResourceType(ResourceTypes.JOB)
-                .setResourceName("${jobReference.project}:${jobReference.jobAndGroup}")
```
ronaveva, line 2901: This change doesn't seem needed. It might affect customers that could be relying  on this 🤔

Jesus-Osuna-M: Hey @ronaveva there is no breaking change here. The setResourceName("${se.project}:${se.uuid}:...") line was already present in the original code before this PR. Since it appeared after the other setResourceName(...) call in the builder chain, it was already always overwriting it (dead code). My PR only added .setUsername(newExec.user) and removed the unreachable first call as cleanup. The resourceName format itself has not changed.

ronaveva (approved):
LGTM

PR #10183, opened 2026-06-02
[RUN-4502] Fix ZipResourceLoader asset path resolution for Grails 7 multi-segment paths

ltamaster:
## Tell us about your PR

Bug fix. `ZipResourceLoader.resolveAssetPath()` fails to find manifest entries for paths with multiple segments (e.g. `js/menu/aclmanager.js`), causing a 404 for the `rundeckpro-ui-acl-manager` plugin script in Grails 7 builds.

Jira: https://pagerduty.atlassian.net/browse/RUN-4502

Grails 7 asset-pipeline strips the top-level type directory (`js/`, `css/`) when writing `manifest.properties` keys. The old code only stripped the **last** segment via `lastIndexOf('/')`, producing `aclmanager.js` for the path `js/menu/aclmanager.js` — which has no manifest entry.

Fix: progressively strip **leading** segments until a manifest key is found:

| Step | Value |
|---|---|
| Requested | `js/menu/aclmanager.js` |
| Try | `menu/aclmanager.js` → found ✓ |
| Returns | `menu/aclmanager-{hash}.js` |

Could require all plugin.yaml paths to match manifest keys exactly (i.e. already omit the type directory). Rejected: would require changes to all plugins and existing plugin.yaml files.

**Additional context**

The previous single-segment stripping (`js/common.js` → `common.js`) still works — the progressive loop handles that case on the first iteration.

## Release Notes

Fixes 404 for plugin assets with nested paths (e.g. `js/menu/aclmanager.js`) in Grails 7 deployments where asset-pipeline fingerprints files and writes a `manifest.properties`.

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/plugins/ZipResourceLoader.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/plugins/ZipResourceLoader.java
@@ -35,7 +35,7 @@ public class ZipResourceLoader implements PluginResourceLoader {
     List<String> resourcesList;
     String resourcesBasedir;
     List<String> basedirListing;
-    Properties assetManifest; // Grails 7: asset-pipeline manifest for hashed filenames
+    volatile Properties assetManifest; // Grails 7: asset-pipeline manifest for hashed filenames
 
     public ZipResourceLoader(
             final File cacheDir,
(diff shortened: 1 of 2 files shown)


ronaveva (approved):
LGTM

PR #10186, opened 2026-06-03
[RUN-2580] Fix: API REST metrics trigger SQL errors

ncofreortiz-hub:
## Release Notes
Fixed an issue where retrieving execution metrics through the REST API generated repeated SQL conversion errors in the logs when using the H2 database, ensuring clean logs and reliable metrics responses.

### 📋 Description / Rationale
This PR resolves the persistent `Data conversion error converting "INTERVAL DAY TO SECOND to TIME"` exception that clutters the logs in H2 database environments (specifically when fetching metrics or initializing log storage checks).

The issue was traced back to `ExecutionService.isSqlCompatible()`, where a dry-run criteria projection was forcing a date-subtraction operation `(date_completed - date_started)` into `StandardBasicTypes.TIME`. In H2, this operation natively yields an `INTERVAL` type, causing a low-level JDBC driver crash before evaluating any dynamic database scenario filters.

### 🛠️ Changes Made
- Added `ExecutionService.isH2Datasource()`, which detects H2 at runtime via `DataSource#getConnection().metaData.databaseProductName` (same JDBC-metadata pattern already used elsewhere in the codebase, e.g. `ReportService.isOracleDatasource()`), rather than parsing the configured Hibernate dialect/config strings.
- **For H2:** Substituted the generic date-subtraction with a native `DATEDIFF('SECOND', date_started, date_completed)` projected as `StandardBasicTypes.LONG`, both in the `isSqlCompatible()` probe and in a dedicated H2 criteria branch in `getCriteriaScenarios()` so the real metrics query (not just the probe) uses the correct SQL.
- **For other DBs:** Kept the original fallback logic intact to ensure backwards compatibility.
- `isH2Datasource()` closes the JDBC connection it opens and swallows/logs (at debug) any failure, defaulting to `false` rather than propagating.
- Added unit tests covering `isH2Datasource()` (H2 and non-H2 cases) and `isSqlCompatible()`'s exception-safe fallback.

### ✅ Verification / Testing Done
- Rebuilt the Docker image and spawned a fresh container using the local H2 configuration.
- Validated that the endpoint `GET /api/34/project/.../executions/metrics` returns a correct `200 OK` response payload.
- Audited the real-time `docker logs` and verified that the `SqlExceptionHelper` spam has been completely eliminated from the output.
- Added Spock unit tests for the new/changed methods.

### 🔗 Ticket Reference
- **Ticket ID:** RUN-2580
- **Target Release:** 6.1.0 (Ready for Review)


carlosrfranco (changes requested):
IMO, the copilot made some important comments and perhaps they would need to be addressed/reviewed.

ltamaster:
## Code Review Notes

### 1. Use the existing pattern for database detection

The codebase already has an established pattern for runtime database detection via JDBC metadata — see `ReportService.isOracleDatasource()` (line 544):

```groovy
private boolean isOracleDatasource() {
    def dataSource = applicationContext.getBean('dataSource', DataSource)
    def databaseProductName = dataSource?.getConnection()?.metaData?.databaseProductName
    return (databaseProductName == 'Oracle') ? true : false
}
```

The current approach in this PR (parsing `sessionFactory.properties.get('hibernate.dialect')` + `grailsApplication.config.getProperty('dataSource.url')`) is fragile: the dialect may not be present in the properties if it was auto-detected by Grails/GORM, and parsing config strings instead of querying the connection directly is prone to fail in non-standard configurations.

The consistent approach would be `dataSource.getConnection().metaData.databaseProductName == 'H2'`.

### 2. Incomplete fix — H2 will still generate errors in `getCriteriaScenarios()`

`isSqlCompatible()` is only a probe. When it returns `true` for H2 (thanks to this fix), the code proceeds to `queryExecutionMetricsByCriteria()`, which internally calls `getCriteriaScenarios()` — which tries **`metricCriteriaA`** first:

```groovy
sqlProjection '(date_completed - date_started) as durationSum', 'durationSum', StandardBasicTypes.TIME
```

That SQL is exactly the one that fails in H2 with `INTERVAL DAY TO SECOND to TIME`. The log spam is removed from the probe but may reappear in the actual data query.

The correct pattern was already established in `RUN-4485` for Oracle: **add an H2-specific criteria directly in `getCriteriaScenarios()`** using `DATEDIFF('SECOND', date_started, date_completed)` as `LONG`, rather than modifying `isSqlCompatible()`. This way the multi-DB fallback system handles H2 the same way it handles Oracle and PostgreSQL.

### 3. Missing unit tests

The `isSqlCompatible()` method has no test coverage. Since this PR introduces conditional behavior (H2 branch vs. non-H2), it would be important to add at least:

- A test verifying that with an H2 dialect/URL, `DATEDIFF` is used as `LONG`
- A test verifying that with any other dialect, the original behavior is preserved
- A test verifying that an exception returns `false` without propagating the error

carlosrfranco (commented):
Tested and it looks good but would be good to check the latest comments (medium) from copilot. These are some small improvements that would be good to have @ncofreortiz-hub

carlosrfranco (approved):

PR #10190, opened 2026-06-04
[RUN-4513] Add default time filter to activity/executions listing

ltamaster:
## Tell us about your PR

Enhancement. Adds a configurable default time filter to the activity/executions listing page, gated by a feature flag. Without this, the page loads all executions with no time scope on first load, which is slow and overwhelming in large environments.

Jira: https://pagerduty.atlassian.net/browse/RUN-4513

Backend-driven injection in `ReportsController`. When the feature flag `ACTIVITY_DEFAULT_TIME_FILTER` is enabled, the controller reads `gui.activity.defaultTimeFilter` (default: `1m`) and injects `recentFilter` into the query object — but only when no `*Filter` params are already present in the request. This flows to the Vue component via the existing `window._rundeck.data.query` mechanism that `activityList.vue` already reads on mount. No changes to the Vue component are needed.

Changes:
- `Features.java` — new `ACTIVITY_DEFAULT_TIME_FILTER` enum constant
- `RundeckConfigBase.java` — `gui.activity.defaultTimeFilter` typed field for Spring Binder
- `ReportsController.groovy` — inject `recentFilter` when feature flag enabled and no filters active
- `MenuController.groovy` — same injection for the jobs page activity tab
- `jobs.gsp` / `jobs.next.gsp` — pass `defaultRecentFilter` from controller to inline JS
- `reports/index.gsp` — pass `defaultRecentFilter` to `window._rundeck.data.query`
- `SysConfigProp.java` / `SystemConfig.java` — add `values` field for Select-type config props
- `rundeck-config.properties` — remco template block for Docker deployments
- `ReportsControllerSpec.groovy` — 6 unit tests covering all acceptance criteria

Frontend-driven alternative (read config via API on mount): rejected because it adds an extra round-trip before the first execution load, and requires adding conditional logic to `activityList.vue`'s mount lifecycle.

**Additional context**

The feature is off by default (`ACTIVITY_DEFAULT_TIME_FILTER.enabled=false`). Accepted values for `gui.activity.defaultTimeFilter`: `1h`, `1d`, `1w`, `1m`. Invalid values fall back to `1m`.

The `rundeckpro` side adds the System Config GUI props (feature flag toggle + Select for the time period) via `GuiSystemConfigurable` and `FeatureFlagSystemConfigurable`.

## Release Notes

Adds a feature flag `rundeck.feature.activityDefaultTimeFilter.enabled` that, when enabled, scopes the activity/executions page to a configurable recent time window on first load. The window defaults to the last month and is configurable via `gui.activity.defaultTimeFilter` (accepted values: `1h`, `1d`, `1w`, `1m`).

### DIFF
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
@@ -244,13 +244,7 @@ class MenuController extends ControllerBase implements ApplicationContextAware{
                 return redirect(jobListLinkHandler.generateRedirectMap([project:params.project]))
             }
         }
-        def defaultRecentFilter = null
-        if (!params.find { it.key.endsWith('Filter') }) {
-            if (featureService?.featurePresent(Features.ACTIVITY_DEFAULT_TIME_FILTER)) {
-                def configured = configurationService?.getString('gui.activity.defaultTimeFilter', '1m') ?: '1m'
-                defaultRecentFilter = (configured in ['1h', '1d', '1w', '1m']) ? configured : '1m'
-            }
-        }
+        def defaultRecentFilter = executionService.getActivityDefaultTimeFilter(params)
 
         if (request.getCookies().find { it.name == 'nextUi' }?.value == 'true' || params.nextUi == 'true') {
             params.nextUi = true
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
@@ -244,9 +244,17 @@ class MenuController extends ControllerBase implements ApplicationContextAware{
                 return redirect(jobListLinkHandler.generateRedirectMap([project:params.project]))
             }
         }
+        def defaultRecentFilter = null
+        if (!params.find { it.key.endsWith('Filter') }) {
+            if (featureService?.featurePresent(Features.ACTIVITY_DEFAULT_TIME_FILTER)) {
+                def configured = configurationService?.getString('gui.activity.defaultTimeFilter', '1m') ?: '1m'
+                defaultRecentFilter = (configured in ['1h', '1d', '1w', '1m']) ? configured : '1m'
+            }
+        }
+
         if (request.getCookies().find { it.name == 'nextUi' }?.value == 'true' || params.nextUi == 'true') {
             params.nextUi = true
-            return render(view: 'jobs.next', model: [:])
+            return render(view: 'jobs.next', model: [defaultRecentFilter: defaultRecentFilter])
         }
 
         if(configurationService.getBoolean('gui.paginatejobs.enabled',false)) {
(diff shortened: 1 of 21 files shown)


ltamaster:
Fixed in commit a2ae7f3: changed both lookups in `ReportsController` to use `configurationService.getString('gui.activity.defaultTimeFilter', '1m')`, matching `MenuController` and the unit test mocks. Also added `ConfigurationService configurationService` injection. The `rundeck.gui.*` key prefix was indeed wrong — the System Config GUI stores under `gui.activity.defaultTimeFilter`.

rundeckapp/grails-app/controllers/rundeck/controllers/ReportsController.groovy
```diff
@@ -130,7 +141,12 @@ class ReportsController extends ControllerBase{
             params.recentFilter=null
         }
         if(null!=query && !params.find{ it.key.endsWith('Filter')}){
-            //no default filter
+            if (featureService.featurePresent(Features.ACTIVITY_DEFAULT_TIME_FILTER)) {
+                def configured = configurationService.getString('gui.activity.defaultTimeFilter', '1m') ?: '1m'
+                def effectiveFilter = (configured in ['1h', '1d', '1w', '1m']) ? configured : '1m'
+                params.recentFilter = effectiveFilter
```
smartinellibenedetti, line 140: question: is there any benefit of setting params.recentFilter here? I'd expect that setting query would be enough.

rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
```diff
@@ -244,9 +244,17 @@ class MenuController extends ControllerBase implements ApplicationContextAware{
                 return redirect(jobListLinkHandler.generateRedirectMap([project:params.project]))
             }
         }
+        def defaultRecentFilter = null
+        if (!params.find { it.key.endsWith('Filter') }) {
+            if (featureService?.featurePresent(Features.ACTIVITY_DEFAULT_TIME_FILTER)) {
+                def configured = configurationService?.getString('gui.activity.defaultTimeFilter', '1m') ?: '1m'
+                defaultRecentFilter = (configured in ['1h', '1d', '1w', '1m']) ? configured : '1m'
+            }
+        }
```
smartinellibenedetti: tiny nit: could this logic be a common method somewhere? looks like this block got repeated in multiple places with very minor differences

smartinellibenedetti (approved):
Code looks good, the only suggestion is to document the enhancement in case customers would like to setup a different time filter by default, and also to give a heads-up that when the feature is on, this will impact the search params in the activity page (so they might need to adjust it depending of what they are looking for).

rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
ltamaster: Refactor done

PR #10198, opened 2026-06-08
[RUN-4521] Fix node sources randomly getting deleted — Vue 3.5 null vnode crashes and JSON body parse data loss

ltamaster:
## Tell us about your PR

Bug fix for [RUN-4521](https://pagerduty.atlassian.net/browse/RUN-4521): Node sources randomly getting deleted — two independent root causes identified and fixed.

### Root Cause 1: Vue 3.5 null vnode crashes corrupt component state

Four related Vue 3.5 rendering bugs were identified and fixed in the node sources configuration components:

**`ProjectPluginConfig.vue`**
- Removed `editFocus` from `:key` of the `additional_config` plugin-config component, which was causing forced destroy/recreate on every `editFocus` change — leaving null entries in Vue 3.5 block tracking `dynamicChildren`
- Changed two adjacent `v-if` spans (Edit/Save buttons) to `v-if`/`v-else-if` since they are mutually exclusive — prevents simultaneous null vnodes in the same block
- Wrapped `addPlugin` state changes in `$nextTick` to separate modal unmount from v-for array push, avoiding `Cannot set properties of null (setting '__vnode')` during concurrent DOM reconciliation

**`pluginConfig.vue`**
- Changed `<span v-for> + <plugin-prop-view v-if>` to `<template v-for> + <span v-if>` pattern — in Vue 3.5, a `v-if` without `v-else` inside a non-block `<span>` produces literal `null` in the children array, which causes `unmountChildren` to crash

**`pluginPropVal.vue`**
- Fixed two separate `v-if` blocks for Boolean `true`/`false` display to use `v-if`/`v-else-if` — eliminates null entries in the Boolean template's Fragment children array

**`pluginPropView.vue`**
- Replaced `v-if="allowCopy"` with `v-show="allowCopy"` on all copy icon `<i>` elements — since `allowCopy` defaults to `false`, `v-if` was producing null in children arrays for every `PluginPropView` instance in show mode
- Added guard `prop.desc ? $t(prop.desc) : ''` to prevent `SyntaxError: 17` in vue-i18n when `prop.desc` is `undefined`
- Added `<template v-else></template>` for Boolean template `v-if` and Options `v-else-if` cases

---

### Root Cause 2: `JsonUtil.parseRequestBody` silently drops request body — causes destructive save

**`JsonUtil.groovy`** — `parseRequestBody` had a Servlet spec violation that caused intermittent silent data loss:

The old code called `getInputStream()` first (Attempt 1). When `available() == 0` (common on proxied/ALB network connections due to TCP buffering timing), it fell through without reading anything. Attempt 2 then called `getReader()`, which throws `IllegalStateException` because `getInputStream()` was already called. That exception was caught silently, returning `null`.

With `body == null`, `saveProjectPlugins` proceeded with `plugins = []`, then `updateFrameworkProjectConfig` deleted **all** `resources.source.*` project properties and wrote nothing back. This is the random deletion in production.

**Fix:**
- `JsonUtil.parseRequestBody`: call `getReader()` first (never `getInputStream()` before `getReader()`). Fall back to `getInputStream().readAllBytes()` only if `getReader()` throws. Added `log.error` at each failure point for diagnosability.
- `FrameworkController.saveProjectPlugins`: added null body guard — returns HTTP 400 with a clear error message instead of silently proceeding with an empty plugin list.

**`ResourceModelRunnerFilter.vue`** (rundeckpro)
- Added `desc: ""` to `propsData` entries to prevent `$t(undefined)` → `SyntaxError: 17` in vue-i18n, which left `subTree = null` and triggered the null vnode crash cascade.

---

- Upgrading Vue to a patch version that handles null children in `unmountChildren` more gracefully — not viable as it would require full regression testing
- Using `v-memo` on v-for items to prevent unnecessary re-renders — partially effective but doesn't address the null vnode root cause
- Using `@RequestBody` annotation on `saveProjectPlugins` to let Spring/Micronaut handle body parsing — valid long-term fix, but requires broader controller refactoring

**Additional context**

The root cause of null vnodes in Vue 3.5 is that `v-if` without `v-else` inside non-block elements (compiled with `createElementVNode` rather than `createElementBlock`) generates `null` as the false branch in the children array. When `unmountChildren` iterates these arrays, it calls `unmount(null)` which crashes with `TypeError: Cannot destructure property 'type' of 'vnode' as it is null`.

The `JsonUtil` bug is intermittent because `available()` on a `ByteArrayInputStream` (unit tests) always returns > 0, but on Tomcat's real network stream with an AWS ALB, it can return 0 before all TCP segments arrive.

## Release Notes

Fixed intermittent node sources deletion in the project nodes configuration page. Two independent bugs were addressed: Vue 3.5 null vnode crashes that could corrupt component state during Add/Delete operations, and a request body parsing bug in `JsonUtil.parseRequestBody` that caused `saveProjectPlugins` to silently proceed with an empty plugin list — deleting all node sources — when the HTTP input stream was not immediately available.

[RUN-4521]: https://pagerduty.atlassian.net/browse/RUN-4521?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/tests/pluginPropView.spec.ts
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/tests/pluginPropView.spec.ts
@@ -5,21 +5,39 @@ import * as Clipboard from "../../../utilities/Clipboard";
 jest.mock("../../../utilities/Clipboard");
 const mockedClipboard = Clipboard as jest.Mocked<typeof Clipboard>;
 
+interface MountOptions {
+  props?: Record<string, any>;
+  stubs?: Record<string, boolean | object>;
+}
+
 let originalMutationObserver: typeof MutationObserver;
 
-const createWrapper = async (options: { props?: Record<string, any> } = {}) => {
+const createWrapper = async (options: MountOptions = {}) => {
   const wrapper = mount(PluginPropView, {
     props: {
       prop: { type: "String", title: "My Prop", desc: "Prop description" },
       value: "hello",
+      allowCopy: false,
       ...options.props,
+    },
+    global: {
+      stubs: {
+        "ace-editor": true,
+        expandable: {
+          template: '<div><slot name="label"></slot><slot></slot></div>',
+        },
+        ...options.stubs,
+      },
     },
   });
   await wrapper.vm.$nextTick();
   return wrapper;
 };
 
-describe("pluginPropView", () => {
+const findByTestId = (wrapper: ReturnType<typeof mount>, testId: string) =>
+  wrapper.find(`[data-testid="${testId}"]`);
+
+describe("PluginPropView", () => {
   beforeEach(() => {
     originalMutationObserver = global.MutationObserver;
     global.MutationObserver = jest
@@ -46,12 +64,10 @@ describe("pluginPropView", () => {
         },
       });
 
-      expect(
-        wrapper.find('[data-testid="boolean-prop-title"]').exists(),
-      ).toBe(true);
-      expect(
-        wrapper.find('[data-testid="boolean-prop-title"]').text(),
-      ).toContain("Enable Feature");
+      expect(findByTestId(wrapper, "boolean-prop-title").exists()).toBe(true);
+      expect(findByTestId(wrapper, "boolean-prop-title").text()).toContain(
+        "Enable Feature",
+      );
     });
 
     it("shows 'yes' text when value is true", async () => {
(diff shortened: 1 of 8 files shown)


rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/tests/pluginPropView.spec.ts
```diff
@@ -460,6 +460,42 @@ describe("pluginPropView", () => {
...
+        createWrapper({
+          props: {
+            prop: { type: "Boolean", title: "Enable", desc: undefined },
+            value: "true",
+          },
+        }),
+      ).resolves.toBeDefined();
+    });
+  });
+
```
smartinellibenedetti, line 503: Not a blocker, but it would be great if you could run the skill write-jest-tests (or ask agent to check the changes against the verification-checklist.md in the skill) to adjust these tests as they don't match the format we expect .

ltamaster: done

ronaveva (approved):
LGTM

Issue #10208, opened 2026-06-12
Menu loads slow due to excessive querying


ReneKroon:
**Describe the bug**
Menu loading is unnecessarily slow due to excessive querying.

**My Rundeck detail**

This issue is there for ~9 years already. In general if the menu lists contains 2000 VMs with 12 possible tasks per VM the menu starts to become slow to load.

**To Reproduce**
Open the menu for a large environment.

**Expected behavior**
No excessive querying, just the menu.

**Screenshots**

**Desktop (please complete the following information):**
For the last 9 years.

**Additional context**

fdevans:
@ReneKroon - We need more details about versions and steps to reproduce this.  Thanks for the PR but it was causing other issues within the product.  If you can reopen with more specifics we can take a better look.

PR #10209, opened 2026-06-12
[RUN-4530] Improve the recursive pattern of menu building.

ReneKroon:
This rest on the following assumption:

JobSchedulesService.isScheduled(uuid) is ScheduledExecution.findByUuid(jobUUID)?.scheduled
 it just reloads the row to read the same column. The objects in schedlist are fully hydrated ScheduledExecutions from
  listWorkflows, so it.scheduled returns the same value without the round-trip.

The behaviour should therefore not change, but the query count should be reduced.

Process fllowed
1. Followed the guidelines, however some part of the test suite requires a token. I ran focused tests for this change.
2. See https://github.com/rundeck/rundeck/issues/10208

bugfix to speed up menu loading

There is no functional change but database queries are avoided -> equivalent solution to what was already there.

More optimizations were considered, but this is the core change. Let's keep it simple and just do this.

**Additional context**

## Release Notes

Job menu loading could be slow for large environments. This is now faster as one query per job is avoided.

### DIFF
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
@@ -527,7 +527,7 @@ class MenuController extends ControllerBase implements ApplicationContextAware{
 
         def finishq=scheduledExecutionService.finishquery(query,params,qres)
 
-        def allScheduled = schedlist.findAll { jobSchedulesService.isScheduled(it.uuid) }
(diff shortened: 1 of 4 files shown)


jayas006:
The PR was rejected because it replaced `JobSchedulerService` with direct scheduler/domain calls, and that breaks project-level schedule handling.

Key reason:

-   Direct checks only look at whether the job itself is scheduled.
-   They do **not** consider schedules defined at the **project level**.
-   That means some valid scheduled jobs would stop appearing / being handled correctly.

PR #10212, opened 2026-06-12
[RUN-4214] Audit failed login attempts when using rundeck.jaaslogin=true

Jesus-Osuna-M:
## Tell us about your PR

This is a bug fix. When Rundeck is configured with `rundeck.jaaslogin=true`, failed login attempts are not captured in the audit log, leaving no security trail for failed authentication events. Successful logins are audited correctly.

Related Jira ticket: [RUN-4214](https://pagerduty.atlassian.net/browse/RUN-4214)

`AuditEventsService` only listened to `AuthenticationFailureBadCredentialsEvent`. However, when JAAS is enabled, Spring Security wraps the JAAS `LoginException` via `DefaultLoginExceptionResolver` into `AuthenticationServiceException`, which is published as `AuthenticationFailureServiceExceptionEvent` — a different event type.

Added `@EventListener` handlers for `AuthenticationFailureServiceExceptionEvent` in:

- **`AuditEventsService`**: new `handleAuthenticationServiceExceptionFailureEvent()` method that mirrors the existing `BadCredentials` handler — increments `userLogin.failure` metric counter and dispatches a `LOGIN_FAILED` audit event with username, roles, and action type.
- **`UserActionService`**: new `handleAuthenticationFailureServiceExceptionEvent()` method that logs a warning with the username of the failed JAAS login attempt.

Both changes are surgical and isolated — the existing `AuthenticationFailureBadCredentialsEvent` handlers are unchanged, so non-JAAS login auditing is unaffected.

- **Listen to `AbstractAuthenticationFailureEvent`** (parent class of all failure events): would cover all providers in one handler but risks catching unintended events from OAuth2 or X509 flows.
- **Override `RundeckJaasAuthenticationProvider`** to re-throw `BadCredentialsException`: changes provider behavior and could affect callers; more invasive than necessary.

**Additional context**

Root cause: `DefaultLoginExceptionResolver` maps `LoginException` → `AuthenticationServiceException`, so the published Spring Security event type differs from non-JAAS failures. Successful JAAS logins are audited correctly because `ProviderManager` always publishes `AuthenticationSuccessEvent` regardless of provider type.

## Release Notes

Fixed: Failed login attempts are now audited when Rundeck is configured with `rundeck.jaaslogin=true`. Previously, JAAS authentication failures were silently dropped from the audit log.

[RUN-4214]: https://pagerduty.atlassian.net/browse/RUN-4214?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/events/UserActionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/events/UserActionService.groovy
@@ -57,11 +57,15 @@ class UserActionService implements LogoutHandler{
         }
     }
 
-    private String extractUsername(Authentication authentication) {
-        if (!authentication){
-            log.error("Null authentication on event")
-            return null
-        }
-        return authentication.name ?: authentication.principal.name ?: null
+    /**
+     * Extracts the username from an authentication object without side effects.
+     * Returns null if authentication is null or no username can be determined;
+     * callers are responsible for logging when null is returned.
+     */
+    private static String extractUsername(Authentication authentication) {
+        if (!authentication) return null
+        if (authentication.name) return authentication.name
+        if (authentication.principal instanceof String) return authentication.principal as String
+        return authentication.principal?.name ?: null
     }
 }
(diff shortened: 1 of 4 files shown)


ronaveva (approved):
LGTM

PR #10219, opened 2026-06-16
[RUN-3911] Fix compacted execution output API returning mixed-type entries array

ncofreortiz-hub:
## Release Notes

Fixed the compacted execution output API (`GET /api/*/execution/{id}/output?compacted=true&format=json`), which previously returned a mix of full objects and bare strings in its `entries` array—causing errors in clients that expected every entry to be an object with a `log` field. As of API v59, all compacted log entries are returned as consistent objects (omitting only unchanged fields), while API v58 and earlier keep their existing behavior for backward compatibility.

## Summary

- **Bug**: `GET /api/*/execution/{id}/output?compacted=true&format=json` returned a mixed-type `entries` array — first entry as a full object, subsequent log-only entries as bare strings (e.g. `"   \"blockdevices\": ["`).
- **Impact**: Most JSON clients (Python, JS, Java) iterate entries and call `.log` on each element; bare string entries lack `.log`, causing runtime errors. Log content with `"` characters (e.g. output from `lsblk --json`) also produced confusing `\"` escape sequences.
- **Fix**: Introduced **API v59**. In v59+, compacted entries are always objects with only changed fields omitted — never bare strings. API v58 and below retain existing behavior (backward compatible).

## Changes

| File | Change |
|------|--------|
| `ApiVersions.groovy` | Added `V59 = 59`, bumped `API_CURRENT_VERSION` to V59 |
| `ExecutionController.groovy` | `renderOutputFormatJson` accepts `apiVersion` param; bare-string collapse gated to `apiVersion < V59`; call site passes `request.api_version`; updated inline API docs |
| `ExecutionControllerSpec.groovy` | New test: V59 compacted entries are always `Map` instances, never bare strings |

## Before / After (API v59+)

**Before (`compacted=true`, v58):**
```json
"entries": [
  {"time": "16:54:06", "log": "{", "level": "NORMAL", ...},
  "   \"blockdevices\": [",
  "      {"
]
```

**After (`compacted=true`, v59+):**
```json
"entries": [
  {"time": "16:54:06", "log": "{", "level": "NORMAL", ...},
  {"log": "   \"blockdevices\": ["},
  {"log": "      {"}
]
```

## Test plan

- [x] Existing `api execution output compacted json` test still passes (v21 behavior unchanged)
- [x] New `api execution output compacted json v59 entries always objects` test passes
- [x] Manual: run job with `lsblk --json` step, call output API with `?compacted=true`, verify entries are all objects in v59

### DIFF
--- a/functional-test/src/test/groovy/org/rundeck/util/container/RdClient.groovy
+++ b/functional-test/src/test/groovy/org/rundeck/util/container/RdClient.groovy
@@ -33,7 +33,7 @@ class RdClient {
     /**
      *  The current (latest) released version of the Rundeck API
      */
-    public static final int API_CURRENT_VERSION = 58
+    public static final int API_CURRENT_VERSION = 59
 
     final ObjectMapper mapper = new ObjectMapper()
     String baseUrl
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/ExecutionController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/ExecutionController.groovy
@@ -1065,7 +1065,9 @@ the result data.
 In this mode, Log Entries are compacted by only including the changed values from the
 previous Log Entry in the list.  The first Log Entry in the results will always have complete information.  Subsequent entries may include only changed values.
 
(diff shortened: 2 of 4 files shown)


ronaveva (approved):
LGTM

PR #10220, opened 2026-06-16
[RUN-3651] Preserve accumulated data vars when secureOption storagePath uses node variable

ncofreortiz-hub:
## Summary

Fixes RUN-3651. This is a re-submission of #10203 as a branch on the main repo (per reviewer request) with a test added.

**Bug:** When a secure option's `storagePath` contains a node context variable (e.g. `keys/project/foo/${node.name}`), `data.*` variables captured by log filter plugins (e.g. `key-value-data`) in earlier workflow steps are lost and unavailable in subsequent steps. With a hardcoded storage path the bug does not occur.

**Root cause:** In `ExecutionServiceImpl.executeNodeStep()`, when `secureOptionReplaced=true`, the code constructed a brand-new `WFSharedContext` containing only the global view:

```java
// Before — discards all non-global views (data.* from log filters, etc.)
.sharedDataContext(WFSharedContext.with(ContextView.global(), context.getDataContextObject()))
```

`WFSharedContext.with(ContextView, DataContext)` creates an empty context and merges only that single view, losing every accumulated view from prior steps.

**Fix:** Copy the existing `sharedDataContext` first, then merge the updated global data on top:

```java
WFSharedContext updatedSharedContext = new WFSharedContext(context.getSharedDataContext());
updatedSharedContext.merge(ContextView.global(), context.getDataContextObject());
nodeContext = new ExecutionContextImpl.Builder(context)
        .singleNodeContext(node, true)
        .sharedDataContext(updatedSharedContext)
        .build();
```

This preserves all accumulated views while still propagating the resolved secure option value.

## Changes

- `core/src/main/java/com/dtolabs/rundeck/core/execution/ExecutionServiceImpl.java` — the fix
- `core/src/test/groovy/com/dtolabs/rundeck/core/execution/ExecutionServiceImplSpec.groovy` — new Spock test covering:
  - `secureOptionReplaced=true`: `data.*` vars from prior steps survive AND resolved option value is present
  - `secureOptionReplaced=false`: normal path continues to work

## Test plan

- [ ] New unit tests pass: `./gradlew :core:test --tests "com.dtolabs.rundeck.core.execution.ExecutionServiceImplSpec"`
- [ ] Reproducer: job with `storagePath: keys/project/foo/${node.name}` + `key-value-data` log filter writing `data.variable` — subsequent step echoing `${data.variable}` now prints correctly

## Additional context

Reported by customer LSEG Business Services Limited. Reproduces on Rundeck 5.13.0.

## Release Notes

Fixed an edge case where `data.*` variables set by log filter plugins (e.g. `key-value-data`) were lost in subsequent workflow steps when a secure option's storage path contained a node context variable such as `${node.name}`.

### DIFF
--- a/core/src/test/groovy/com/dtolabs/rundeck/core/execution/ExecutionServiceImplSpec.groovy
+++ b/core/src/test/groovy/com/dtolabs/rundeck/core/execution/ExecutionServiceImplSpec.groovy
@@ -1,0 +1,202 @@
+/*
+ * Copyright 2026 Rundeck, Inc. (http://rundeck.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.dtolabs.rundeck.core.execution
+
+import com.dtolabs.rundeck.core.common.IExecutionProviders
+import com.dtolabs.rundeck.core.common.NodeEntryImpl
+import com.dtolabs.rundeck.core.common.PluginControlService
+import com.dtolabs.rundeck.core.data.BaseDataContext
+import com.dtolabs.rundeck.core.dispatcher.ContextView
+import com.dtolabs.rundeck.core.execution.workflow.StepExecutionContext
+import com.dtolabs.rundeck.core.execution.workflow.WFSharedContext
+import com.dtolabs.rundeck.core.execution.workflow.steps.node.NodeStepExecutionItem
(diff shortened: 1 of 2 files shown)


core/src/test/groovy/com/dtolabs/rundeck/core/execution/ExecutionServiceImplSpec.groovy
```diff
@@ -0,0 +1,202 @@
...
+            capturedNodeContext = ctx
+            def result = Mock(NodeStepResult)
+            result.isSuccess() >> true
+            result
+        }
+
+        and: "providers returning the mock interpreter"
+        def providers = Mock(IExecutionProviders)
+        providers.getNodeStepExecutorForItem(_, _, _) >> interpreter
+
```
Jesus-Osuna-M, line 110: Both test cases share ~30 lines of identical mock setup (framework, nodes, execution listener, etc.). Consider extracting them to a Spock `setup()` method or a private helper to reduce duplication. Non-blocking.

Jesus-Osuna-M (commented):
Overall the fix is correct and well-tested. One minor suggestion for maintainability: both test cases share ~30 lines of identical mock setup. Consider extracting them to a Spock `setup()` or `given:` fixture method to reduce duplication. Non-blocking.

Jesus-Osuna-M (approved):
Fix is correct and well-tested. The copy constructor approach properly preserves accumulated data vars by chaining the original context as base while applying the new merge on top. LGTM!

PR #10225, opened 2026-06-17
[RUN-4539] Fix StorageTreeFactory stopping at first gap in provider index sequence

Jesus-Osuna-M:
## Release Notes

Fixed an issue where key storage providers and converters configured with non-contiguous index numbers were silently skipped, making any keys stored under the skipped providers inaccessible. Rundeck now loads every configured storage provider regardless of gaps in the index sequence, so administrators no longer need to keep provider indexes perfectly consecutive when configuring key storage outside the UI (for example, via `rundeck-config.properties` or automated/API-driven setups).

## About this change

**Jira Ticket:** [RUN-4539](https://pagerduty.atlassian.net/browse/RUN-4539)

**Companion rundeckpro PR:** [rundeckpro/rundeckpro#4770](https://github.com/rundeckpro/rundeckpro/pull/4770)

---

### Root Cause

\`StorageTreeFactory.buildTree()\` iterated storage and converter provider indices with a sequential \`while\` loop that stopped at the **first missing index**:

\`\`\`java
int storeIndex = 1;
while (config.containsKey(prefix + "." + storeIndex + ".type")) {
    configureStoragePlugin(builder, storeIndex, config);
    storeIndex++;   // stops here if storeIndex+1 is absent
}
\`\`\`

Any providers configured at indices **beyond the first gap** were silently never loaded — making all keys stored under those providers inaccessible.

**Example:** \`rundeck-config.properties\` with:
\`\`\`properties
rundeck.storage.provider.1.type=db
rundeck.storage.provider.1.path=keys
rundeck.storage.provider.4.type=vault
rundeck.storage.provider.4.path=keys/vault
\`\`\`
Provider at index 4 is never loaded because index 2 is absent. The loop stops when it checks for \`provider.2.type\` and finds nothing.

The same issue affected converter plugins (\`storage.converter.N\`).

This is a pre-existing bug surfaced during work on RUN-4539, where the companion rundeckpro PR introduces per-provider index reservation that can produce non-contiguous index sequences.

---

### Fix

Replace both \`while\` loops with **\`extractConfiguredIndices()\`**:

1. Scans the full config map for all keys matching \`prefix.N.type\`
2. Collects all distinct integer indices \`N\` (ignoring non-numeric and nested keys)
3. Returns them sorted — iterates the **complete set regardless of gaps**

\`extractConfiguredIndices()\` is package-visible static for direct unit testing.

---

### Testing

**New unit tests in \`StorageTreeFactorySpec\`:**

| Test | Covers |
|------|--------|
| Gap-skipping | Indices 1, 3, 5 all returned even when 2 and 4 are absent |
| Empty config | Returns empty list |
| Prefix isolation | Keys for a different prefix are ignored |
| Non-integer / nested key rejection | \`provider.abc.type\` and \`provider.1.config.type\` are skipped |
| Duplicate deduplication | Multiple keys for same index (type, path, config.*) counted once |

- [x] All new specs pass
- [x] Existing \`StorageTreeFactorySpec\` test passes

**Kind of Change**

- [x] Bug Fix

**Development Checklist**

- [x] Unit Tests added (5 new Spock specs)
- [x] Documentation Not needed

[RUN-4539]: https://pagerduty.atlassian.net/browse/RUN-4539?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/storage/StorageTreeFactory.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/storage/StorageTreeFactory.java
@@ -29,8 +29,11 @@ import org.rundeck.storage.conf.TreeBuilder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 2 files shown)


jayas006 (approved):
LGTM

PR #10231, opened 2026-06-19
[RUN-4445] Fix job reference step label not being saved on edit

nachodolce:
## Is this a bugfix, or an enhancement?

Bugfix. Fixes #RUN-4445 — when editing a Job Reference step, the label (description) field was not being saved.

## Describe the solution you've implemented

`saveEditStep` was reading `editModel.description` for job reference steps, but `JobRefForm` overwrites `editModel` via its `update:modelValue` emit before `saveEditStep` runs, clearing the user's input. The label input is already bound to `editExtra.description`, so the fix reads from there instead (one-line change). Two unit tests added to cover the regression.

## Alternatives considered

Moving the v-model binding to a separate data property was considered, but editExtra.description was already the correct source of truth — only the read side needed fixing.

## Release Notes

Fixed an issue in the job editor where changing the label of a Job Reference step and saving would not take effect—the label reverted to its previous value. Job Reference step labels now save correctly.

### DIFF
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/workflow/WorkflowSteps.vue
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/workflow/WorkflowSteps.vue
@@ -548,7 +548,7 @@ export default defineComponent({
 
         if (result.valid) {
           if (stepForValidation.jobref) {
-            saveData.description = this.editModel.description;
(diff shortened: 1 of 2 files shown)


fdevans:
Thank you @nachodolce. After discussing with Eng this isn't the right fix and breaks other things.  Please see internal ticket for more comments and path forward.

PR #10232, opened 2026-06-19
[RUN-4440] Restore job reference name autocomplete lost in Vue migration

nachodolce:
Bugfix. Fixes RUN-4440 — job reference name autocomplete stopped working after upgrade to 5.20.

## Describe the solution

The jQuery-based _initJobPickerAutocomplete function was not ported when the workflow editor was migrated to Vue 3, leaving the job name field as a plain static input with no suggestions. Replaced it with PrimeVue's AutoComplete component calling /menu/jobsSearchJson on keypress (300ms debounce). On selection, name, group and UUID are all populated automatically — matching the original behavior.

## Alternatives considered

Re-implementing with vanilla JS was considered but rejected in favor of the existing PrimeVue AutoComplete already used elsewhere in the component.

## Release Notes

Restored job name autocomplete in the Job Reference step form, which was lost during the Vue 3 UI migration in 5.20. Customers with many jobs can now search by name instead of browsing manually.

### DIFF
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/workflow/JobRefFormFields.vue
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/workflow/JobRefFormFields.vue
@@ -91,16 +91,19 @@
       >
         <label class="col-sm-2 control-label"></label>
         <div class="col-sm-5">
-          <input
+          <PtAutoComplete
             :id="`jobNameField${rkey}`"
-            v-model="modelValue.name"
-            type="text"
+            v-model="jobNameInputText"
+            :suggestions="(jobNameSuggestions as any)"
+            option-label="name"
             name="jobName"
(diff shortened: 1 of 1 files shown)


smartinellibenedetti (changes requested):
Thank you for your contribution, could you please update the component from AutoComplete to `PtAutoComplete` that lives in ui-trellis folder?

nachodolce:
Hi, done Thanks

nachodolce:
Hi @smartinellibenedetti, thanks for the feedback! The component has already been updated to `PtAutoComplete` in the latest commit (`eeb43ac`). The direct `AutoComplete` import has been replaced with the internal wrapper. Please take another look when you get a chance!

fdevans:
@nachodolce Are you able to address the CoPilot comments here?

fdevans:
This is going to take a more significant refactor that the Eng team will take on.

PR #10235, opened 2026-06-23
[RUN-4445] Fix label not being saved

smartinellibenedetti:
## Release Notes

Fixed an issue where the label (description) on a Job Reference step was not saved—both when adding a new job reference step and when editing an existing one—causing the label to disappear or revert after saving. Job Reference step labels are now preserved correctly.

## Tell us about your PR

Bugfix. When adding or editing a Job Reference step, the label (description) the user typed was lost on save.

In `WorkflowSteps.vue`, `saveEditStep` explicitly overwrote `saveData.description` with `editModel.description` for job reference steps. Because `JobRefForm` emits `update:modelValue` (without a `description` field) before save runs, `editModel.description` had already been cleared—so the overwrite wiped the user's label. The fix removes that overwrite so the label captured in the step editor flows through to `saveData` unchanged. Also standardized the step-item test selector from `data-test` to `data-testid` and added a stable `data-testid` on the step description element.

N/A

**Additional context**

Adds Jest regression tests covering both scenarios: preserving the label on an existing job reference step, and rendering the typed label after adding a new one.


jayas006 (approved):
LGTM

PR #10238, opened 2026-06-23
[RUN-4559] Fix Stored XSS in autoLink taglib via raw job name

Jesus-Osuna-M:
## Tell us about your PR

Security bugfix for a Stored XSS vulnerability in the `autoLink` taglib (`UtilityTagLib.groovy`). An authenticated user with `create_job` permission can craft a job name containing an HTML/JS payload (e.g. ``). When a user with `delete_execution` permission deletes an execution of that job, the flash message renders the `{{Job UUID}}` token via `autoLink`, which calls `g.link()` with the raw, un-encoded job name as the anchor body — executing arbitrary JavaScript in the victim's browser context.

Reported via HackerOne (#3815645 / PS-1691). Jira: https://pagerduty.atlassian.net/browse/RUN-4559

Added `text?.encodeAsHTML()` before passing the job name to `g.link()` inside the `autoLink` taglib:

```groovy
// Before (vulnerable)
return g.link(lparams, text)

// After (fixed)
return g.link(lparams, text?.encodeAsHTML())
```

This single-line change covers all surfaces where `autoLink` resolves `{{Job UUID}}` tokens:
- Execution delete flash message (`ExecutionController`)
- Bulk job action result messages (`ScheduledExecutionController` / `menu/jobs.gsp`)
- Cluster heartbeat job/member views

A Spock regression test was added to `UtilityTagLibSpec` covering three XSS payload variants (``, `<script>`, `<svg onload>`).

- Sanitizing at the job name validation level (only blocks `/`, too narrow).
- Encoding in the GSP templates (already attempted via `<g:enc>` wrapper but it runs before `autoLink` substitution, leaving the replacement unprotected).
- Encoding in the controller flash message (doesn't cover all `autoLink` surfaces).

The taglib fix is the correct single point of control.

**Additional context**

Reproduction steps (from HackerOne report):
1. Create a job with name ``
2. Run the job once to create an execution
3. Navigate to `/execution/show/<id>` and delete the execution
4. Observe XSS alert on redirect to reports page (session cookie exposed)

## Release Notes

Fixed a Stored XSS vulnerability in the `autoLink` taglib where job names containing HTML/JS payloads were rendered unencoded in flash messages. Job names are now HTML-encoded before being placed inside anchor tags.

### DIFF
--- a/rundeckapp/grails-app/taglib/rundeck/UtilityTagLib.groovy
+++ b/rundeckapp/grails-app/taglib/rundeck/UtilityTagLib.groovy
@@ -661,7 +661,7 @@ class UtilityTagLib{
                     def lparams= [:]+opts.linkParams
                     lparams.id=it[2]
                     def text = opts.textValue?opts.textValue(it[2]):it[1]
-                    return g.link(lparams,text)
+                    return g.link(lparams, text?.encodeAsHTML())
                 }else if(opts.hrefParams){
                     def lparams= [:]+opts.hrefParams
                     return g.createLink(lparams)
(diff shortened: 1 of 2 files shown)


carlosrfranco (approved):
LGTM

PR #10245, opened 2026-06-24
[RUN-2580] Fix H2 dialect detection and debug logging in isSqlCompatible

ncofreortiz-hub:
## Tell us about your PR

Bugfix. Addresses two issues found in code review of the H2 SQL compatibility probe in `ExecutionService.isSqlCompatible()`.

1. `sessionFactory.configurationProperties` is not a standard Hibernate 5.6 API — calling it throws `MissingPropertyException` at runtime, causing the method to silently fall back to in-memory metrics even when H2 is in use.
2. Silent `catch (Exception)` block made it impossible to diagnose why metrics were consistently falling back to the in-memory path.

Fixes https://pagerduty.atlassian.net/browse/RUN-2580

- Replaced `sessionFactory.configurationProperties?.getProperty('hibernate.dialect')` with `sessionFactory?.properties?.get('hibernate.dialect')` (standard Hibernate API).
- Added `dataSource.url` fallback: H2 is also detected when `dataSource.url` starts with `jdbc:h2:`, so detection works even when `hibernate.dialect` is not explicitly configured.
- Added `log.debug(...)` in the catch block to surface unexpected failures without polluting logs in normal operation.

Could have only fixed the API call without the `dataSource.url` fallback, but that would still miss unconfigured-dialect H2 setups (common in test environments).

**Additional context**
These changes were identified via Copilot code review on the original PR that introduced H2 dialect branching.

## Release Notes

Fix H2 dialect detection in execution metrics SQL compatibility probe to prevent silent fallback to in-memory metrics calculation.

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
@@ -4707,8 +4707,9 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
         boolean isCompatible = false
         try {
             def sessionFactory = applicationContext.getBean("sessionFactory")
-            def dialectName = sessionFactory.configurationProperties?.getProperty("hibernate.dialect") ?: ""
-            boolean isH2 = dialectName.contains("H2Dialect")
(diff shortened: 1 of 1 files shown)


ncofreortiz-hub:
Cerrado por error — los cambios van en #10186.

PR #10246, opened 2026-06-24
[RUN-4558] Fix i18n fallback for dynamically registered plugin messages

smartinellibenedetti:
## Tell us about your PR

Bug fix. When plugins registered i18n messages via `addUiMessages`, only the active locale was registered into the vue-i18n instance. When vue-i18n fell back to `en_US` (e.g. for a Japanese user with a missing translation key), the fallback lookup failed because `en_US` was never registered — causing the raw key string to render instead of the English text.

Jira: RUN-4558

- Changed `updateLocaleMessages` to use `i18n.global.mergeLocaleMessage` instead of the previous `setLocaleMessage`/`mergeDeep` approach, ensuring messages are merged rather than overwriting existing locale state.
- Widened `commonAddUiMessages` to accept either `UiMessage[]` (legacy flat array) or the new `LocalizedMessages = Record<string, Record<string, LocaleMessageValue>>` type.
- When given a `LocalizedMessages` object, `commonAddUiMessages` now registers both the active locale AND `en_US` (when locale is not already `en_US`), so vue-i18n's built-in fallback chain always resolves to English text instead of raw keys.
- Exported the new `LocalizedMessages` and `LocaleMessageValue` types from `library/index.ts` so rundeckpro plugins can import and use them directly.
- Updated all inline `addUiMessages` provider implementations in the rundeck submodule (`ui/main.ts`, `uiSocketObserver.ts`, `storage/main.ts`, `project-nodes-config/main.js`, `job/editor/main.js`) to delegate to `commonAddUiMessages`.
- Migrated existing call sites (`readme-motd/main.ts`, `RulesetStrategyEditor.vue`) to pass the full `LocalizedMessages` object instead of a locale-sliced flat array.

Alternative: register all locale keys at build time (static imports). Rejected — plugins are dynamically loaded and their locales are not known at build time.

Alternative: use `setLocaleMessage` with a pre-merged object. Rejected — this overwrites any previously registered messages for the locale rather than merging.

**Additional context**

- Backward compatible: the legacy `UiMessage[]` code path is preserved unchanged.
- rundeckpro plugin migration is in a companion PR against the rundeckpro repo.

## Release Notes

Fixed an issue where plugin-provided i18n messages would show raw translation keys instead of English text for users whose locale was not `en_US` and a specific key was missing in their locale's message catalogue.

### DIFF
--- a/rundeckapp/grails-spa/packages/ui-trellis/jest.config.js
+++ b/rundeckapp/grails-spa/packages/ui-trellis/jest.config.js
@@ -22,6 +22,7 @@ module.exports = {
     "^@/(.*)$": "<rootDir>/src/$1",
     "^primevue/(.*)": "<rootDir>/node_modules/primevue/$1",
     "^@primevue/(.*)": "<rootDir>/node_modules/@primevue/$1",
+    "^vue-i18n$": "<rootDir>/node_modules/vue-i18n/dist/vue-i18n.cjs",
   },
   modulePathIgnorePatterns: ["<rootDir>/public"],
   testMatch: ["**/*.test.ts", "**/*.spec.ts"],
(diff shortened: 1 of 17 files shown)


smartinellibenedetti:
@copilot+claude-sonnet-4.6 address the comments of this PR

Jesus-Osuna-M (approved):

PR #10295, opened 2026-06-30
[RUN-4581] Fix ACL policy API returning JSON instead of YAML

fdevans:
## Release Notes
Fixed an issue where the ACL policy API returned JSON with the policy wrapped in a `contents` field even when YAML was requested via the `Accept: application/yaml` header. The endpoint now correctly returns the raw YAML policy document again, restoring compatibility for ACL import, diff, and backup tooling that relies on YAML responses.

### Jira Ticket
[RUN-4581](https://pagerduty.atlassian.net/browse/RUN-4581) — ACL API YAML responses broken.

### Summary
After the Grails 7 / Micronaut 4 migration, the ACL policy API endpoints stopped honoring the `Accept: application/yaml` header and instead returned JSON with the policy wrapped in a `contents` field, regardless of the requested format. This broke clients that depend on raw YAML responses (the customer's RBA import/diff utility being one).

### Root cause
The migration reimplemented `ApiService.extractResponseFormat` to parse the `Accept` header explicitly (because `request.format` is no longer populated from it in Grails 7). That reimplementation handled `application/json`, `application/xml`, and `text/plain` but omitted `application/yaml` / `text/yaml`. YAML requests therefore fell through to the default format and were rendered as JSON with the `contents` wrapper.

Both affected handlers route their format decision through this single method:
- System ACL: `FrameworkController.apiSystemAclsGetResource`
- Project ACL: `ProjectController.apiProjectAclsGetResource` (and the project PUT response path)

### Changes
- `ApiService.extractResponseFormat`: add explicit `application/yaml` / `text/yaml` handling to the Accept-header negotiation so YAML requests resolve to the `yaml` format and are streamed directly. The branch only triggers when the endpoint's allowed-format list already includes `yaml`, keeping the blast radius minimal.
- `ApiServiceSpec`: add unit coverage that exercises the real Accept-header resolution with `request.format` unset (the Grails 7 condition). Prior controller specs mocked `extractResponseFormat`, which is why this regressed undetected.

### Testing
- `./gradlew :rundeckapp:test --tests "rundeck.services.ApiServiceSpec"` → BUILD SUCCESSFUL, 42 tests, 0 failures, 0 skipped.
- The new test is a true regression guard: without the fix it returns `all`/`json` instead of `yaml` and fails.

[RUN-4581]: https://pagerduty.atlassian.net/browse/RUN-4581?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ronaveva (approved):
LGTM

PR #10298, opened 2026-06-30
[RUN-0000] Align okhttp-urlconnection with okhttpVersion

fdevans:
## Summary
- `functional-test` pinned `okhttp-urlconnection` to a separate `okhttpUrlConnectionVersion` (5.4.0) while `okhttp` itself remained at `okhttpVersion` (4.12.0).
- Mixing okhttp 4.x and 5.x on the same test classpath causes conflicting transitive resolution (Gradle silently upgrades okhttp to 5.x via the urlconnection artifact).
- This consolidates `okhttp-urlconnection` onto `${okhttpVersion}` and removes the now-redundant `okhttpUrlConnectionVersion` property, matching the pattern used elsewhere in the build.

## Test plan
- [ ] `functional-test` resolves a single, consistent okhttp version (4.12.0) with no version conflict.
- [ ] Functional test build/compile succeeds.


ronaveva (approved):
LGTM

PR #10315, opened 2026-07-08
[RUN-4260] Fix SQL Error 8117 log spam on MSSQL by caching isSqlCompatible result

nachodolce:
## Release Notes

Fixed excessive `SQL Error: 8117` messages filling the logs on Microsoft SQL Server deployments. An incompatible probe query previously ran on every execution-metrics lookup; that check now runs once and its result is reused, eliminating the repeated errors and continuous log growth (which had no impact on job execution).

## Summary

Fixes constant `SQL Error: 8117` log spam on Microsoft SQL Server instances.

**Root cause:** `ExecutionService.isSqlCompatible()` ran a `datetime2` arithmetic probe query on every call to `queryExecutionMetrics()`. On MSSQL, this operation is invalid — the JDBC driver logs the error before Hibernate catches the `JDBCException`. Since the probe result was never cached, this fired on every metrics query, causing continuous log growth with no functional impact.

**Fix:** Cache the `boolean` result in `sqlCompatibleCache` after the first probe. Subsequent calls short-circuit immediately without touching the database.

Also fixes a pre-existing Java 21 compatibility issue in `ExecutionJob.groovy`: `Thread.stop()` was removed in Java 21 and caused a `@CompileStatic` compilation failure. Replaced with `Thread.interrupt()`.

## Changes

- `ExecutionService.groovy`: added `private Boolean sqlCompatibleCache = null` field; `isSqlCompatible()` returns cached value on subsequent calls
- `ExecutionService2Spec.groovy`: new Spock spec verifying the cache short-circuits for both `true` and `false` values
- `ExecutionJob.groovy`: `Thread.stop()` → `Thread.interrupt()` (Java 21 compat)

## Test plan

- [ ] Run `./gradlew :rundeckapp:test --tests "rundeck.ExecutionService2Spec.isSqlCompatible*"` — passes
- [ ] Verify on MSSQL instance: `SQL Error: 8117` appears at most once per app startup in logs

### DIFF
--- a/rundeckapp/grails-app/jobs/rundeck/quartzjobs/ExecutionJob.groovy
+++ b/rundeckapp/grails-app/jobs/rundeck/quartzjobs/ExecutionJob.groovy
@@ -502,7 +502,7 @@ class ExecutionJob implements InterruptableJob {
                     killcount++;
                 } else {
                     //reached pre-set kill limit, so shut down
-                    thread.stop()
+                    ((Thread) thread).interrupt()
                 }
             }
         }
--- a/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
@@ -374,6 +374,7 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
     }
 
     private boolean applicationIsShutdown
+    private Boolean sqlCompatibleCache = null
 
     public boolean isApplicationShutdown() {
         return applicationIsShutdown
@@ -4707,6 +4708,9 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
     }
 
     private boolean isSqlCompatible() {
+        if (sqlCompatibleCache != null) {
+            return sqlCompatibleCache
+        }
         boolean isCompatible = false
         try{
             Execution.createCriteria().list(max:1) {
@@ -4720,6 +4724,7 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
             isCompatible = false
         }
 
+        sqlCompatibleCache = isCompatible
         return isCompatible
     }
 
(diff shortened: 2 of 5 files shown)


Jesus-Osuna-M:
## Review
Thanks for the MSSQL log-spam fix — caching `isSqlCompatible()` is the right idea for RUN-4260. A few things should come out of this PR before merge.
### Blocker: `Thread.stop()` → `((Thread) thread).interrupt()` is not a replacement
This hunk is unrelated to RUN-4260 and changes kill semantics.
`thread` is already a `WorkflowExecutionServiceThread` (`ServiceThreadBase` → `Thread`). The kill loop calls `thread.abort()` up to `killLimit` (100) times, ~1s apart. `abort()` already does:
```java
aborted = true;
interrupt();
```
So by the time we hit the `else`, the thread has already been interrupted ~100 times. Calling `interrupt()` again is a no-op.
`Thread.stop()` was a last-resort force-kill when cooperative abort failed. `interrupt()` is cooperative. If the workflow ignores interrupts (native SSH, `Process.waitFor()`, swallowed `InterruptedException`), `while (thread.isAlive())` never exits and a Quartz worker can hang forever.
The cast is also unnecessary.
Please **drop this change from this PR**. Java 21 removal of `Thread.stop()` is real (`@CompileStatic` on `executeCommand`), but it needs its own change that:
1. Also covers the same `thread.stop()` in `ExecutionUtilService.runRefJobWithTimer` (this PR leaves that one in place).
2. After `killLimit`, **leaves the wait loop**, marks the execution failed, and logs that the workflow thread is still alive — instead of pretending another `interrupt()` will kill it.
### Blocker: two tickets in one PR
Commit `507a24e` is RUN-4445 (`WorkflowSteps.vue` job-ref label). Title and release notes only mention RUN-4260. Please split RUN-4445 into its own PR.
### Tests
The new `isSqlCompatible` spec only assigns `service.@sqlCompatibleCache` and reads it back. It does not show that the SQL probe is skipped on a cache hit, or that the first miss populates the cache.
The kill-path change has no tests.
### Nits (non-blocking if the cache lands alone)
- `sqlCompatibleCache` on a Grails singleton should be `volatile Boolean`.
- Current `main` already branched `isSqlCompatible()` for H2 vs non-H2. This branch looks based on the older probe — rebase and apply the cache on top of that, don’t replace it.
**Verdict: request changes.** For RUN-4260, ship only the cache (+ a test that covers miss → probe → hit) after rebase. Move RUN-4445 and the Java 21 kill change to separate PRs.

PR #10319, opened 2026-07-08
[RUN-4536] Parse blankIfUnexpandable from script plugin YAML config

Jesus-Osuna-M:
## Release Notes

Fixed an issue where shell variable references such as `${VAR}` and `${VAR:-default}` in scripts run by script-based step plugins could be stripped out before the shell had a chance to evaluate them, causing those variables to resolve to empty values. Script plugins can now preserve these expressions so they expand correctly at runtime.

## Summary

- Fixes the script plugin property loader to parse `blankIfUnexpandable` from `plugin.yaml` config sections.
- Without this fix, script plugins could not opt out of having unresolved `${...}` patterns blanked during data-context expansion, breaking shell variable references like `${VAR}` and `${VAR:-default}`.

## Problem

`AbstractDescribableScriptPlugin.createDescription()` reads property attributes from plugin.yaml (`type`, `name`, `title`, `description`, `required`, `default`, `values`, `labels`, `scope`, `renderingOptions`) but never read `blankIfUnexpandable`. The `PropertyBuilder` defaults this to `true`, so **all** script plugin string properties had unresolvable `${...}` patterns replaced with empty string.

This caused the [nixy-step-plugins issue #20](https://github.com/rundeck-plugins/nixy-step-plugins/issues/20) where `${VAR}` and `${VAR:-default}` in user scripts were blanked before bash could expand them.

## Fix

Added parsing of `blankIfUnexpandable` (both Boolean and String forms) from the plugin.yaml config map, calling `pbuild.blankIfUnexpandable(value)` before building the property. This allows script plugins to declare:

```yaml
config:
  - type: String
    name: script
    blankIfUnexpandable: false
```

## Test plan

- [x] Verified locally: rebuilt Rundeck core + nixy plugin with `blankIfUnexpandable: false` → `${VAR}` and `${VAR:-default}` now expand correctly in the *nixy / local-script step
- [x] Unit test for `AbstractDescribableScriptPlugin` parsing `blankIfUnexpandable`

## Related

- Plugin fix: https://github.com/rundeck-plugins/nixy-step-plugins/pull/37
- Upstream issue: https://github.com/rundeck-plugins/nixy-step-plugins/issues/20
- Jira: https://pagerduty.atlassian.net/browse/RUN-4536

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/plugins/AbstractDescribableScriptPlugin.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/plugins/AbstractDescribableScriptPlugin.java
@@ -59,6 +59,7 @@ import java.util.*;
  *     config.X.default = default string of the property
  *     config.X.values = comma-separated values list for Select or FreeSelect properties
  *     config.X.scope = scope of the property, from {@link PropertyScope}
+ *     config.X.blankIfUnexpandable = true/false, if unresolvable ${...} references should be blanked (default: true)
  * </pre>
  *
  * @author Greg Schueler <a href="mailto:greg@dtosolutions.com">greg@dtosolutions.com</a>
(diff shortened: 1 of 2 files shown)


ronaveva (approved):
LGTM

PR #10331, opened 2026-07-10
[RUN-4615] Bump attribute-match-node-enhancer to 1.0.3

fdevans:
## Release Notes
The bundled Attribute Match Node Enhancer plugin now supports attribute value substitution, letting you build new node attributes and tags from a node's existing attributes using `${attribute}` syntax (for example, `image-${ec2.imageId}` or `endpoint=${host}:${port}`). This makes it possible to derive richer, dynamic metadata for nodes without external scripting. The update also adds a new "is present" match operator (`~~`), so enhancement rules can target nodes based simply on whether an attribute exists, regardless of its value, complementing the existing "not present" (`!!`) operator.

## PR Details
Bumps the bundled `attribute-match-node-enhancer` plugin from `1.0.1` to `1.0.3` by updating `attributeMatchNodeEnhancerVersion` in `gradle.properties`.

### What's new in the plugin (1.0.1 -> 1.0.3)
- **Attribute value substitution** — a new opt-in "Enable Attribute Substitution" setting. When enabled, the attributes and tags you add can reference existing node attribute values with `${attribute}` syntax:
  - Attributes: `newattr=some-${oldattr}/${otherattr}`
  - Tags: `tag1,image-${ec2.imageId}`
  
  Unknown references resolve to an empty string. The setting is off by default, so existing node enhancer configurations behave exactly as before.
- **New `~~` "is present" match operator** — matches when an attribute exists (no value required), complementing the existing `!!` "not present" operator. This makes it easy to add attributes/tags to any node that simply has a given attribute set.
- **Reliability fix** — attribute values containing `$` or `\` (such as Windows paths or passwords) are now substituted literally instead of being misinterpreted, preventing errors and mangled values.

### Change in this repo
- Updated `attributeMatchNodeEnhancerVersion` in `gradle.properties` (`1.0.1` -> `1.0.3`).

### References
- Tracked by RUN-4615
- Plugin changes: https://github.com/rundeck-plugins/attribute-match-node-enhancer/pull/24


ltamaster (approved):
LGTM

PR #10399, opened 2026-07-23
[RUN-4680] Fix CVE-2026-41907 nested uuid dependency

fdevans:
## Tell us about your PR

Security fix for CVE-2026-41907 / SNYK-JS-UUID-16133035. Direct `uuid` was already at 11.x, but scanners still reported vulnerable nested `uuid@3.4.0` via `@rundeck/client` → `@azure/ms-rest-js@2.0.7`.

- Pin `uuid` to `^11.1.1` (patched version)
- Add npm overrides for `uuid: $uuid` so nested copies cannot remain on 3.x/8.x/9.x
- Override `@azure/ms-rest-js` to `2.7.0`, which uses `uuid.v4()` and is compatible with uuid 11 (2.0.7 used the removed `require('uuid/v4')` deep import)

- Overriding only `uuid` to 11 without upgrading `ms-rest-js` — breaks `@azure/ms-rest-js@2.0.7` because it requires `uuid/v4`
- Waiting for a new `@rundeck/client` release — longer path; override clears the finding now

**Additional context**
- Jira: [RUN-4680](https://pagerduty.atlassian.net/browse/RUN-4680)
- Path previously flagged: `@rundeck/ui-trellis` → `@rundeck/client@0.2.5` → `@azure/ms-rest-js@2.0.7` → `uuid@3.4.0`
- Verified with `npm list uuid` that all paths resolve to `uuid@11.1.1`

## Release Notes

[RUN-4680]: https://pagerduty.atlassian.net/browse/RUN-4680?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rundeckapp/grails-spa/packages/ui-trellis/package-lock.json
+++ b/rundeckapp/grails-spa/packages/ui-trellis/package-lock.json
@@ -113,7 +113,7 @@
         "ts-jest": "^29.1.2",
         "ts-node": "^10.9.1",
         "typescript": "^5.4.5",
-        "uuid": "^11.0.0",
+        "uuid": "^11.1.1",
         "vue-eslint-parser": "^9.4.0",
         "vue-loader": "^17.4.2",
         "walk": "^2.3.15",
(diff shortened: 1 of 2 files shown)


jtobard (approved):
Deep review — ran this locally against the PR branch (checked out in the actual `rundeck` submodule, restored to `main` afterward) rather than just reading the diff:

- `npm ci` installs cleanly.
- `npm ls uuid` — every path (`@rundeck/client`, `@storybook/addon-actions`, `@types/uuid`, `sockjs` via `webpack-dev-server`, `jest-junit`) dedupes to a single `uuid@11.1.1`. No stray old versions remain.
- `npm ls @azure/ms-rest-js` — resolves to `2.7.0 overridden`, single instance.
- Verified the actual installed `@azure/ms-rest-js@2.7.0` source uses `require('uuid')` + `.v4()` (not the removed `uuid/v4` deep import that `2.0.7` used) — confirms the stated reason for bumping ms-rest-js is correct and sufficient.
- `npm run test:unit` — full jest suite: **117 suites / 1330 tests, all passed.**

The `overrides` + `"uuid": "$uuid"` self-reference pattern matches existing precedent already used in this file for `axios`, `tough-cookie`, etc. (e.g. "Update axios to 1.16.1 to fix security vulnerability"), so this is consistent with how CVE fixes are normally done here.

No blockers, no breaking-change surface actually exercised. LGTM.

PR #10410, opened 2026-07-27
[RUN-4699] Allow Java 21 and 25 in preinst.sh version check

ltamaster:
## Tell us about your PR

Bugfix. [RUN-4699](https://pagerduty.atlassian.net/browse/RUN-4699): `packaging/lib/rpm/scripts/preinst.sh` hard-gates on the runtime `java -version` output and aborts the RPM install unless the major version is exactly `11` or `17`:

```sh
if [[ "$JAVA_MAJOR_VERSION" != "11" && "$JAVA_MAJOR_VERSION" != "17" ]]; then
  echo "Java 11 or 17 is required"
  exit 1
fi
```

Rundeck 6.0 documents support for Java 17 through 25, but this script still rejects 21 and 25. On RHEL/Rocky Linux 10 (which no longer ships any `java-17-*` package at all), this independently blocks installation even once the RPM's `Requires:` metadata is fixed to accept newer JDKs (see [rundeck/packaging#77](https://github.com/rundeck/packaging/pull/77), which alone is not sufficient because of this gate).

Extended the version check to also accept `21` and `25`, matching the documented system requirements:

```sh
if [[ "$JAVA_MAJOR_VERSION" != "11" && "$JAVA_MAJOR_VERSION" != "17" && "$JAVA_MAJOR_VERSION" != "21" && "$JAVA_MAJOR_VERSION" != "25" ]]; then
  echo "Java 11, 17, 21 or 25 is required"
  exit 1
fi
```

Considered checking `>= 17` instead of an exact allow-list, but kept the existing exact-match style (just extended) to stay consistent with the current code and avoid silently accepting untested future JDK versions.

**Additional context**

Verified this fix together with [rundeck/packaging#77](https://github.com/rundeck/packaging/pull/77) end-to-end: built a real RPM (using the actual published 6.0.0 WAR + this patched `preinst.sh`) and confirmed a full working install and boot on Rocky Linux 10 with only `java-25-openjdk` installed, and again with only `java-21-openjdk`, plus a Rocky Linux 9 / Java 17 regression check (still installs cleanly).

Note: the DEB package has no equivalent script (confirmed via `dpkg-deb -e` on the real published `.deb` — only `postinst`/`postrm`/`prerm` exist, none of which check `java -version`), so this fix is RPM-specific.

## Release Notes

Fixed RPM installation failing on hosts running Java 21 or 25 because the installer incorrectly rejected those JVM versions. The RPM pre-install check now accepts Java 17, 21, and 25, matching Rundeck's documented system requirements and allowing installation on newer Linux distributions such as RHEL/Rocky Linux 10 that no longer ship Java 17.

[RUN-4699]: https://pagerduty.atlassian.net/browse/RUN-4699?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


fdevans (approved):

fdevans:
Shouldn't we take out the 11?

fdevans (approved):

PR #10138, opened 2026-05-11
[RUN-4197] Convert adhoc page to Vue SPA

smartinellibenedetti:
## Release Notes

The Commands (adhoc) page has been rebuilt as a modern Vue single-page application, available when the new NextUI option is enabled. This delivers faster, more responsive command execution and a consistent look and feel with the rest of the redesigned Rundeck interface.

## Tell us about your PR

**Is this a bugfix, or an enhancement? Please describe.**

Enhancement — converts the legacy Adhoc/Command page to a fully Vue-driven SPA (`adhocNext`), replacing the old GSP-rendered page.

**Describe the solution you've implemented**

The adhoc page has been rewritten as a Vue SPA (`adhocNext` route). The new implementation covers:

- `AdhocCommandForm` — command entry, target node filter, runner/executor selection
- `NodeFilterSection` — node filter input + results (reuses existing `NodeFilterInput`/`NodeFilterResults` components)
- `ExecutionOutput` — live execution log output + node dispatch status
- `ActivitySection` — recent command history
- `adhocService` — API layer for all backend calls
- Controller changes (`ScheduledExecutionController`, `FrameworkController`, `ExecutionController`) to support the new endpoints needed by the SPA
- URL mappings and `NextUiInterceptor` wired for the new route

## Feedback requested

Two areas where a second set of eyes would be particularly useful:

1. **Runner manual selection widget** — please give the runner selector an extra look; want to make sure the UX and state management are solid there.
2. **Node dispatch settings** — confirm that all node dispatch options (thread count, keep-going, rank attribute/order) are working correctly end-to-end.

## Known follow-up

The `ExecutionOutput` header section does not yet match the original page 1:1 — a follow-up ticket will address that alignment.

PR #10421, opened 2026-07-31
[RUN-4716] Hide adhoc inline endpoints from public OpenAPI spec

fdevans:
## Jira Ticket
[RUN-4716](https://pagerduty.atlassian.net/browse/RUN-4716)

## Summary

`runAdhocInline` and `runAdhocInlineApi` (added in #10138 / RUN-4197 for the adhoc Vue SPA) ended up surfaced in the public 6.1.0 `rundeck-api.yml` OpenAPI spec, but were never documented in `docs/api/index.md` and weren't called out in the 6.1.0 release notes.

Investigation found that the ajax endpoint these replace (`runAdhocInline`, pre-#10138) was never part of the public API surface: it was invoked only via a plain `createLink(controller:'scheduledExecution', action:'runAdhocInline', ...)` from the old GSP page (not under `/api/{version}/...`), carried no OpenAPI annotations, and authenticates via CSRF token + session (`withForm{}.invalidToken{}`) rather than an API key.

#10138 added a new `/api/{version}/project/{project}/run/command/inline` URL mapping and full `@Operation`/`@ApiResponse` annotations to that same session/CSRF-authenticated action — which doesn't fit the public API auth model used elsewhere (e.g. `apiRunCommandv14` behind the already-documented `POST /project/[PROJECT]/run/command`, which uses `apiService.requireApi`). `runAdhocInlineApi` does use proper API-key auth, but its own description states it's "intended for use by the new Vue.js UI" — a UI-support endpoint, not general customer integration surface.

## Changes

- Add `@Hidden` to both `runAdhocInline` and `runAdhocInlineApi` in `ScheduledExecutionController.groovy` so they're excluded from generated OpenAPI spec / API Explorer, while remaining fully functional for the adhoc Vue SPA.

## Testing

- Endpoints are unchanged functionally — only the OpenAPI annotation processing changes. Verified via existing adhoc Selenium/functional coverage (no behavior change expected).

[RUN-4716]: https://pagerduty.atlassian.net/browse/RUN-4716?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/ScheduledExecutionController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/ScheduledExecutionController.groovy
@@ -2700,7 +2700,10 @@ Authorization required: `delete` on project resource type `job`, and `delete` on
     /**
      * execute the job defined via input parameters, but do not store it.
      * REST API endpoint: /api/{api_version}/project/{project}/run/command/inline
+     * Internal-only: backs the adhoc Vue SPA's session/CSRF-authenticated ajax call, not a
+     * documented public API surface. See RUN-4716.
      */
+    @Hidden
     @Post(uri='/project/{project}/run/command/inline', produces = MediaType.APPLICATION_JSON)
     @Operation(
         method='POST',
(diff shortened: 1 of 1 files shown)


ltamaster (approved):
LGTM

PR #10422, opened 2026-07-31
[RUN-4714] Fix: project PluginGroup config silently discarded when page-level Save is clicked mid-edit

jgarces-pd:
## Tell us about your PR

Bugfix. On a project's Configuration > Plugins page, a `PluginGroup` provider (e.g. the ServiceNow plugin) only writes its edited field values into the form data submitted by the page's Save button once that plugin's own inline Save link is clicked (`ProjectPluginGroups.vue`'s `savePlugin()`). If a user instead clicks the page-level Configuration Save while a plugin entry is still open for editing, the in-progress edit is silently dropped — no validation runs, no error is shown, the data is just gone.

Reported via Salesforce case / RUN-4714:
1. Open a project's "Edit Configuration" page
2. Select a plugin (e.g. ServiceNow) under Plugins
3. Enter values for its fields
4. Click the page-level Save button for the Configuration (not the plugin's own inline Save)
5. All data entered for the plugin is lost, with no warning

Per review feedback from Luis Toledo (RUN-4714) — this now matches the existing UX pattern used by the resource model source page (`ProjectPluginConfig.vue`), where the page's main Save button is hidden entirely while a plugin config form is open, rather than being clickable and then rejected.

`ProjectPluginGroups.vue` emits a `project-plugin-group-editing` event on the shared Rundeck event bus (`window._rundeck.eventBus` / `getRundeckContext().eventBus`, the same singleton used elsewhere, e.g. `logViewer.vue`) whenever a Plugin Group entry is opened or closed for editing (`editFocus` toggling to/from `-1`), plus a safety-net `emit(false)` on unmount so the button can't get stuck hidden. `editProject.gsp` listens for that event and toggles the visibility of the page-level Save button (`#editProjectSaveButton`) accordingly, so it simply can't be clicked while a plugin edit is in progress.

Also added unit tests (`ProjectPluginGroups.spec.ts`) covering: no emit on mount, `editing=true` when a plugin is opened, `editing=false` after that plugin's own Save is clicked, `editing=false` on Cancel, and `editing=false` as a safety net on unmount.

An earlier version of this PR instead let the page-level Save be clicked and then blocked the submit with an inline error message if a plugin edit was still open. Review feedback pointed out this was inconsistent with how the resource model source page handles the same situation (hiding the main Save button instead of letting it be clicked), so this was reworked to match that pattern for UI consistency across the two pages.

A server-side backstop (validating each submitted `PluginGroup` entry in `FrameworkController.save()`, the way sibling "project.plugin" scoped configs already are a few lines below) would add defense in depth, but is left out of this PR to keep scope to the actual reported bug.

**Additional context**

Verified locally: `npm ci`, `npx jest src/app/pages/project-config src/library/components/plugins` (166 tests passing, no regressions), `npx tsc -p tsconfig.app.json --noEmit` (clean).

## Release Notes

Fixed: entered values for a project's Plugin Group configuration (e.g. ServiceNow) could be silently lost if the page-level Save was clicked before the plugin's own inline Save. The Save button is now hidden while a plugin's configuration form is open, matching the resource model source page's behavior.

### DIFF
--- a/rundeckapp/grails-app/views/framework/editProject.gsp
+++ b/rundeckapp/grails-app/views/framework/editProject.gsp
@@ -104,6 +104,12 @@
     });
     function init(){
         jQuery('input[type=text]').on('keydown', noenter);
+        // Mirror the resource model source page: hide the main Save button
+        // while a Plugin Group config form is open for editing, so it can't
+        // be clicked mid-edit and silently discard the unsaved values.
+        window._rundeck.eventBus.on('project-plugin-group-editing', function(editing){
+            jQuery('#editProjectSaveButton').toggle(!editing);
+        });
     }
     var _storageBrowseSelected=confirm.setNeedsConfirm;
     jQuery(init);
@@ -173,7 +179,7 @@
         </div>
         <div class="card-footer">
           <g:submitButton name="cancel" value="${g.message(code:'button.action.Cancel',default:'Cancel')}" class="btn btn-default reset_page_confirm"/>
-          <g:submitButton name="save" value="${g.message(code:'button.action.Save',default:'Save')}" class="btn btn-cta reset_page_confirm"/>
+          <g:submitButton id="editProjectSaveButton" name="save" value="${g.message(code:'button.action.Save',default:'Save')}" class="btn btn-cta reset_page_confirm"/>
 
             <div class="project-config-plugins-vue">
                 <page-confirm :event-bus="EventBus"
--- a/rundeckapp/grails-app/views/framework/editProject.gsp
+++ b/rundeckapp/grails-app/views/framework/editProject.gsp
@@ -107,8 +107,10 @@
         // Mirror the resource model source page: hide the main Save button
         // while a Plugin Group config form is open for editing, so it can't
         // be clicked mid-edit and silently discard the unsaved values.
+        // Selected by name (not #save id) because Selenium page objects rely
+        // on the id Grails auto-assigns from name="save".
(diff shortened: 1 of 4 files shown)


ltamaster:
Hey @jgarces-pd — while testing this fix locally (adding a plugin group entry and clicking its own inline Save) I found a related-but-distinct bug in the same `savePlugin()` flow and pushed a fix for it on this branch (d0a547b).

**Repro:** open "+ Plugin Config", pick a provider whose fields are optional/defaulted (e.g. Ansible), and click that entry's own Save **without touching any field**. The entry silently disappears from the list — no error, no warning.

**Root cause:** `savePlugin()` treated an empty `plugin.entry.config` (`{}`) as a signal to discard the entry via `removePlugin()`, before validation ever ran. Elsewhere in this codebase (`EditStepCard.vue`, `InlinePluginConfigForm.vue`, `EditPluginModal.vue`) an empty config is treated as a valid "create" state, not something to delete — this was the only spot that dropped it silently.

**Fix:** removed that early-return guard so the flow falls through to the existing `pluginService.validatePluginConfig()` call. A plugin with required fields now correctly surfaces a validation error instead of vanishing; a plugin with only optional/defaulted fields saves as-is.

Verified locally: existing `ProjectPluginGroups.spec.ts` suite still passes, `tsc --noEmit` clean, and manually reproduced/fixed against a running instance via browser automation.

This is separate from the RUN-4714 page-level-Save scenario this PR already fixes, but same file/flow so folding it in here rather than opening a second PR — let me know if you'd rather split it out.

ltamaster (approved):

PR #10426, opened 2026-08-03
[RUN-4723] Fix NPE in NotificationService.triggerPlugin when execution context is null

### LINKED ISSUE #10415: Slack (and other plugin-type) onfailure/onsuccess notifications fail with NPE in NotificationService.triggerPlugin after upgrade to 6.0.1
**Rundeck version:** 6.0.1.20260715-1 (upgraded from 5.19.0.20260202-1)
**Java:** OpenJDK 25.0.4 (also reproduced on OpenJDK 17)
**Install:** Debian package, MariaDB backend, DB-based project/key storage

## Summary

After upgrading from Rundeck 5.19 to 6.0.1, all job notifications using a generic notification plugin (confirmed with `SlackNotification`; likely affects any non-`email`/`url` plugin type such as PagerDuty, MS Teams, etc.) fail with a `NullPointerException`. Built-in `email` and `url` (webhook) notification types are unaffected.

## Steps to reproduce

1. Configure a job with an `onfailure` (or `onsuccess`) notification using a plugin-based notification type, e.g. `SlackNotification`.
2. Cause the job to fail (or succeed, for `onsuccess`).
3. Observe `rundeck.log`.

## Actual behavior

```
ERROR services.NotificationService [pool-3-thread-17] - Error sending notification: Notification{eventTrigger='onfailure', type='SlackNotification', ...}: class java.lang.NullPointerException: Cannot get property 'storageTree' on null object
java.lang.NullPointerException: Cannot get property 'storageTree' on null object
	at org.codehaus.groovy.runtime.NullObject.getProperty(NullObject.java:92)
	at org.codehaus.groovy.runtime.InvokerHelper.getProperty(InvokerHelper.java:167)
	at org.codehaus.groovy.runtime.callsite.NullCallSite.getProperty(NullCallSite.java:48)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callGetProperty(AbstractCallSite.java:329)
	at rundeck.services.NotificationService.triggerPlugin(NotificationService.groovy:800)
	...
	at rundeck.services.NotificationService.triggerJobNotification(NotificationService.groovy:296)
	...
	at rundeck.services.NotificationService$__tt__asyncTriggerJobNotification_closure22.doCall(NotificationService.groovy:187)
```

The notification is silently dropped — no Slack message is sent, no exception surfaces to the user, only this ERROR line in the log.

## Expected behavior

The plugin-based notification should be delivered, same as under Rundeck 5.19 with an identical job/notification configuration (no config changes were made — this reproduces immediately after the platform upgrade).

## Root cause (from source, tag `v6.0.1`, `rundeckapp/grails-app/services/rundeck/services/NotificationService.groovy`)

`triggerJobNotification` correctly computes a `context` local variable and passes it into `triggerPlugin` as an explicit parameter:

```groovy
}else if (n.type) {
    def execMap = null
    Map context = null
    (context, execMap) = generateNotificationContext(content.execution, content, source)
    didsend = triggerPlugin(trigger, execMap, n.type, source.project, n.configuration, content, context)
}
```

But inside `triggerPlugin` (~line 800), the `context` parameter is never used for this purpose — instead the code reads from an unrelated, never-populated field on the `content` map:

```groovy
private boolean triggerPlugin(String trigger, Map data, String type, String project, Map config, Map content, Map context){
    ...
    Map<Class, Object> servicesMap = [:]
    servicesMap.put(KeyStorageTree, content.context.storageTree)   // <-- content.context is always null here
    ...
}
```

`content.context` is never set by any caller; the correctly-computed `context` parameter is what should be used (e.g. `context.storageTree` or `context?.sharedDataContext?...` depending on intended type). This looks like a leftover/incomplete refactor from the Grails 7 / Spring Boot 3 migration in 6.0.0.

`email` and `url` notification types are handled in separate branches earlier in `triggerJobNotification` and don't call `triggerPlugin`, so they are unaffected.

## Impact

Any job relying on a plugin-based notification (Slack, and presumably PagerDuty/Teams/other notification plugins) for failure/success alerting silently stops delivering after upgrading to 6.0.1. In our instance this affects 100+ job notification configs across multiple projects; ~54 of those jobs have *no* email/url fallback configured, so they currently have no working failure alerting at all.

## Suggested fix

In `triggerPlugin`, use the `context` parameter instead of `content.context`, e.g.:
```groovy
servicesMap.put(KeyStorageTree, context?.storageTree)
```
(or whatever the correct accessor is for the `KeyStorageTree` on the `ExecutionContext`/context object built by `generateNotificationContext`).

fdevans:
## Jira Ticket
[RUN-4723](https://pagerduty.atlassian.net/browse/RUN-4723)

## Related Issue
Fixes #10415

## Summary
Plugin-type job notifications (Slack, PagerDuty, MS Teams, and any non-`email`/`url` notification plugin) were silently failing to send on `onfailure`/`onsuccess` triggers whenever the notification's execution context (`content.context`) was null. `NotificationService.triggerPlugin()` read `content.context.storageTree` with no null-guard; the resulting `NullPointerException` was swallowed by `triggerJobNotification`'s per-notification `catch(Throwable)` block, so the notification silently dropped with only an `ERROR` log line and no visible failure to the user.

Built-in `email` and `url` (webhook) notification types are handled in separate branches that don't call `triggerPlugin`, so they were unaffected.

## Root Cause
`content.context` can be null when the caller derives it via an unguarded `execRun?.thread?.context` chain (e.g. `ExecutionService.groovy`) rather than a guaranteed-non-null `ExecutionContextImpl.builder()...build()`. The same file already defends against a null `content.context` a few lines up (line 673), so this was simply an inconsistency rather than a new bug pattern.

## Fix
```diff
-        servicesMap.put(KeyStorageTree, content.context.storageTree)
+        servicesMap.put(KeyStorageTree, content.context?.storageTree)
```

## Testing
- Added a regression test (`"plugin notification still sends when execution context is null"` in `NotificationServiceSpec.groovy`) that fails on the old code (`TooFewInvocationsError` — the plugin's `postNotification` never gets invoked because the NPE aborts the call first) and passes with the fix.
- Ran the full `NotificationServiceSpec` suite locally — all 41 tests pass.

## Test plan
- [x] `./gradlew :rundeckapp:test --tests "rundeck.services.NotificationServiceSpec"` passes
- [x] New test fails against the pre-fix code, passes against the fix (verified both ways locally)

[RUN-4723]: https://pagerduty.atlassian.net/browse/RUN-4723?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/NotificationService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/NotificationService.groovy
@@ -797,7 +797,7 @@ public class NotificationService implements ApplicationContextAware, EventBusAwa
             }
         }
         Map<Class, Object> servicesMap = [:]
-        servicesMap.put(KeyStorageTree, content.context.storageTree)
+        servicesMap.put(KeyStorageTree, content.context?.storageTree)
 
         def services = new RundeckSpiBaseServicesProvider(
                 services: servicesMap
--- a/rundeckapp/src/test/groovy/rundeck/services/NotificationServiceSpec.groovy
+++ b/rundeckapp/src/test/groovy/rundeck/services/NotificationServiceSpec.groovy
@@ -1180,6 +1180,48 @@ class NotificationServiceSpec extends Specification implements ServiceUnitTest<N
 
     }
 
+    def "plugin notification still sends when execution context is null"() {
+        given: 'a plugin-type notification dispatched without an execution context'
+        def (job, execution) = createTestJob()
+
+        def content = [
+                execution: execution,
+                context  : null
+        ]
+
+        job.notifications = [
+                new Notification(
+                        eventTrigger: 'onstart',
+                        type: 'SimpleNotificationPlugin'
+                )
+        ]
+        job.save()
+        service.frameworkService = Mock(FrameworkService) {
+            _ * getRundeckFramework() >> Mock(Framework) {
+                _ * getWorkflowStrategyService()
+            }
+            _ * getPluginControlService(_) >> Mock(PluginControlService)
+        }
+
+        service.grailsLinkGenerator = Mock(LinkGenerator) {
+            _ * link(*_) >> 'alink'
+        }
+        service.pluginService = Mock(PluginService)
+        service.executionService = Mock(ExecutionService) {
+            getEffectiveSuccessNodeList(_) >> []
+        }
+
+        def pluginInst = Mock(NotificationPlugin)
+
+        when: 'trigger notification'
+        def result = service.triggerJobNotification('start', job, content)
+
+        then: 'the plugin still runs and the notification is sent'
+        1 * service.pluginService.configurePlugin('SimpleNotificationPlugin', _, _, _, _) >> new ConfiguredPlugin(pluginInst, [:])
+        1 * pluginInst.postNotification(_, _, _) >> true
+        result
+    }
+
     def "export variables replace values in webhook url"() {
         given:
         def (job, execution) = createTestJob()


ltamaster (approved):
LGTM

PR #10432, opened 2026-08-07
[RUN-2580] Fix execution metrics SQL time-parse errors on MySQL/MariaDB

ncofreortiz-hub:
## Release Notes

Fixed SQL time-parse errors logged on every call to the execution metrics API when using MySQL, MariaDB, or Oracle databases. Metrics data was still returned correctly via a fallback path, but server logs were spammed and response times were slower than necessary. Duration aggregation now uses database-appropriate queries for these backends.

## Summary

- [PR #10186](https://github.com/rundeck/rundeck/pull/10186) fixed the `/executions/metrics` SQL time-parse error (`cannot be parse as time. time must have "99:99:99" format`) for H2 only. Production customers mostly run MySQL/MariaDB, which hits the same underlying bug: aggregating execution durations by casting a raw `date_completed - date_started` diff to SQL `TIME` doesn't work — MySQL's naive datetime subtraction isn't a valid `TIME` value at all, and even where it superficially parses, `TIME` overflows past `838:59:59` once durations are summed across many executions.
- Add `isMySqlDatasource()` and a dedicated MySQL/MariaDB branch (mirroring the existing H2 branch) using `TIMESTAMPDIFF(SECOND, ...)` projected as `LONG`, applied in both `getCriteriaScenarios()` and the `isSqlCompatible()` compatibility gate.
- The `isSqlCompatible()` gate was the more subtle half of the bug: it silently caught the SQL exception and fell back to in-memory metrics calculation, so the API kept returning correct data while still spamming `service.log` with the SQL error on every call — matching the exact symptom in the original report.
- Follow-up commit also fixed the same bug class on Oracle (TIMESTAMP interval arithmetic), adding `isOracleDatasource()` and an explicit Oracle branch.

## Test plan

- [x] `ExecutionServiceSpec`: added unit tests for `isMySqlDatasource()` (MySQL, MariaDB, non-MySQL) and for `getCriteriaScenarios()` selecting the MySQL branch. Full suite: 295 tests, 0 failures.
- [x] Verified end-to-end against a real `mysql:8.4`-backed Rundeck instance (built from this branch, isolated docker-compose): before the fix, `/executions/metrics` logged `cannot be parse as time` on every call; after, zero SQL errors across repeated calls, and response latency dropped (SQL aggregation path instead of in-memory fallback).
- [x] Verified end-to-end against Oracle (`gvenzl/oracle-free`): zero SQL errors across repeated `/executions/metrics` calls, correct duration data.
- [x] PostgreSQL verified live — existing `extract(EPOCH ...)` path unaffected, no changes needed.

### DIFF
--- a/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
+++ b/rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
@@ -4757,14 +4757,35 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
         }
     }
 
+    /**
+     * Detects MySQL/MariaDB datasources, which require dedicated duration-sum SQL:
+     * a plain {@code date_completed - date_started} projected as TIME overflows
+     * MySQL's TIME range (max 838:59:59) once the aggregated duration across the
+     * queried executions exceeds it.
+     */
+    boolean isMySqlDatasource() {
+        try {
+            def dataSource = applicationContext.getBean('dataSource', DataSource)
+            return dataSource.getConnection().withCloseable { conn ->
+                conn.metaData.databaseProductName in ['MySQL', 'MariaDB']
+            }
+        } catch (Exception ex) {
+            log.debug("Unable to determine datasource type, assuming non-MySQL", ex)
+            return false
+        }
+    }
+
     private boolean isSqlCompatible() {
         boolean isCompatible = false
         try {
             boolean isH2 = isH2Datasource()
+            boolean isMySql = isMySqlDatasource()
             Execution.createCriteria().list(max:1) {
                 projections {
                     if (isH2) {
                         sqlProjection 'DATEDIFF(\'SECOND\', date_started, date_completed) as durationSum', 'durationSum', StandardBasicTypes.LONG
+                    } else if (isMySql) {
+                        sqlProjection 'TIMESTAMPDIFF(SECOND, date_started, date_completed) as durationSum', 'durationSum', StandardBasicTypes.LONG
                     } else {
                         sqlProjection '(date_completed - date_started) as durationSum', 'durationSum', StandardBasicTypes.TIME
                     }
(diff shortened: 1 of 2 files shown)


rundeckapp/grails-app/services/rundeck/services/ExecutionService.groovy
```diff
@@ -4757,14 +4757,35 @@ class ExecutionService implements ApplicationContextAware, StepExecutor, NodeSte
...
     private boolean isSqlCompatible() {
         boolean isCompatible = false
         try {
             boolean isH2 = isH2Datasource()
+            boolean isMySql = isMySqlDatasource()
             Execution.createCriteria().list(max:1) {
                 projections {
                     if (isH2) {
                         sqlProjection 'DATEDIFF(\'SECOND\', date_started, date_completed) as durationSum', 'durationSum', StandardBasicTypes.LONG
+                    } else if (isMySql) {
```
carlosrfranco, line 4808: I think this issue should be handled for any database system. Did you try to find out any solution that could be more generic so that we don't need to have a code for each supported database? Or does it happens only for H2 and MySQL? Did you tested in the other database systems that we support like Oracle and PostgreSQL

ncofreortiz-hub: Good questions, answered all three by actually testing rather than guessing:

**Does it happen only for H2 and MySQL?** Yes. The bug is specifically that `date_completed - date_started` cast to SQL `TIME` doesn't survive a round trip through the JDBC driver:
- H2: produces a fractional-seconds numeric like `7.486000` that H2's own TIME parser rejects (`cannot be parse as time`) — the original reported bug.
- MySQL/MariaDB: same failure, different mechanism — naive datetime subtraction yields a non-TIME pseudo-numeric, plus TIME overflows past `838:59:59` once durations are summed.
- PostgreSQL (`metricCriteriaB`, untouched by this PR): already projects via `extract(EPOCH FROM ...)` as `LONG`, never touches `TIME`. Just spun up a real `postgres:16`-backed instance to confirm — ran 8 executions, hit `/executions/metrics` repeatedly: zero `cannot be parse as time` errors, and response times (~30-90ms) confirm it's using the SQL aggregation path, not the in-memory fallback.
- Oracle (`metricCriteriaC`, untouched): projects via `(date - date) * 86400` as `LONG`, same reasoning applies — never casts to `TIME`. I didn't spin up a live Oracle instance (no test image handy / licensing), but it's worth noting Oracle isn't in this repo's own documented support matrix (`.claude/rules/database-migrations.md` lists MySQL/MariaDB, PostgreSQL, H2 as the supported databases) — the Oracle branch predates this fix and is out of scope here.

**Is there a more generic fix so we don't need per-DB code?** Looked into this. The root cause of *why* this needs per-dialect handling at all is that `sqlProjection(...)` here is raw native SQL injected as a literal string — it bypasses Hibernate's dialect abstraction entirely, so nothing translates it automatically. A truly portable version would mean moving this aggregation off the legacy Hibernate `Criteria`/`sqlProjection` API and onto HQL, where Hibernate 5's built-in `timestampdiff(second, ...)` function *is* dialect-portable (each dialect registers its own translation). That's a real option, but it means rewriting how `ExecutionQuery.createCriteria()` applies its dynamic filters (currently built directly against the Hibernate `Criteria` delegate), which is a materially bigger, riskier change than this hotfix. I'd rather do that as a deliberate follow-up than fold it into a bug fix PR — happy to file a ticket for it if that seems worthwhile.

ncofreortiz-hub: Update — I need to correct my earlier reply in this thread. I said Oracle "isn't in this repo's documented support matrix," citing `.claude/rules/database-migrations.md`. That's wrong: that file is just this repo's own AI-assistant dev convention doc, not Rundeck's actual product documentation. The real docs (`administration/configuration/database/oracle.md`) confirm Oracle **is** an officially supported backend (driver goes in `$RDECK_BASE/server/lib`, with `dataSource.dialect = org.rundeck.hibernate.RundeckOracleDialect`).

So I went and tested it live — spun up `gvenzl/oracle-free` + the driver via the documented mechanism, built an image from this branch, and ran real executions against it. That surfaced two more real bugs, both the same root cause: `date_started`/`date_completed` are TIMESTAMP columns, and Oracle's `TIMESTAMP - TIMESTAMP` produces an `INTERVAL DAY TO SECOND`, not the `NUMBER` the existing Oracle code assumed:

1. `isSqlCompatible()`'s fallback branch cast the diff to SQL `TIME` — same bug class as the original H2/MySQL issue, just never exercised until now (Postgres's interval type tolerates it, so it never showed up there). Failed with `ORA-17004: getTime not implemented for T4CIntervaldsAccessor`.
2. The pre-existing Oracle criteria (`metricCriteriaC`) computed `(date_completed - date_started) * 86400` assuming classic `DATE - DATE` arithmetic — doesn't hold for TIMESTAMP columns, so it hit `ORA-17004: getLong not implemented` too.
3. `getCriteriaScenarios()` had no explicit Oracle branch, so every call tried the Postgres-shaped query first (`extract(EPOCH from ...)`), invalid syntax on Oracle (`ORA-00907`), before falling through to the (also broken) Oracle criteria.

Fixed all three: added `isOracleDatasource()` + an explicit Oracle branch in both gate methods (same pattern as H2/MySQL now), casting both timestamps to `DATE` before subtracting to force classic date arithmetic instead of interval arithmetic. Verified live: 0 SQL errors across repeated `/executions/metrics` calls, correct duration data, on both MySQL and Oracle now.

Postgres was also verified live (separately, clean, no changes needed — its existing `extract(EPOCH ...)` path was never affected by this bug class).

Pushed as a follow-up commit on this same branch.

carlosrfranco (approved):
LGTM. Maybe it's worth to consider collapsing isH2Datasource/isMySqlDatasource/isOracleDatasource into one connection-probing helper to cut duplication.

PR #10437, opened 2026-08-10
[RUN-4671] Fix arbitrary file read via writeable file resource-model source

carlosrfranco:
## Tell us about your PR

This is a **security bugfix** for RUN-4671 / PS-1693.

Fixes a critical security vulnerability where users with project-configure permission could read arbitrary files on the Rundeck server by registering a file resource-model source with an unconstrained path (e.g., `/etc/passwd`, `realm.properties`).

Jira: https://pagerduty.atlassian.net/browse/RUN-4671

Implemented a two-layer defense-in-depth approach:

1. **Path Validation (Layer 1)**
   - Added `validateWriteableSource()` method to `WriteableModelSource` interface with default no-op implementation
   - Implemented path validation in `FileResourceModelSource` that:
     - Canonicalizes file paths to prevent traversal attacks (../, symlinks)
     - Enforces containment within project directory by default for all users
     - Supports optional configuration property `rundeck.resourceModelSource.file.allowedBasePaths` for administrators to specify additional allowed directories
   - Validation occurs when opening the edit node source page (`editProjectNodeSourceFile`)

2. **Content Validation (Layer 2 - Defense-in-Depth)**
   - Files must parse as valid node source format using `FileResourceModelSource.parseFile()`
   - Invalid files return empty content instead of exposing file contents
   - Prevents reading arbitrary non-node-source files

**Security Properties:**
- Secure by default: most restrictive configuration without additional setup
- Path traversal attacks blocked via canonical path resolution
- Error messages don't leak sensitive path information
- Validation failures logged for audit
- Configuration-based (not ACL/role-based) for consistency

Alternative approaches considered and documented in the spec:
- **ACL-based restrictions**: Rejected in favor of configuration-based approach for simpler administration
- **Validation at save time**: Rejected because it would require bridging core module with Grails ConfigurationService
- **Sensitive path blacklist**: Deferred as optional future enhancement; current whitelist approach is more secure

**Additional context**

**Test Coverage:**
- 8 unit tests in `TestFileResourceModelSource` (all passing)
- Path validation logic (within project, outside project, traversal, allowed paths, canonicalization, error messages)
- 4 functional tests in `FrameworkController2Spec` (all passing)

**Configuration Example:**
```properties
# In rundeck-config.properties (optional)
rundeck.resourceModelSource.file.allowedBasePaths=/opt/shared/nodes,/var/custom/resources
```

**Files Modified:**
- `core/src/main/java/com/dtolabs/rundeck/core/resources/WriteableModelSource.java`
- `core/src/main/java/com/dtolabs/rundeck/core/resources/FileResourceModelSource.java`
- `core/src/test/java/com/dtolabs/rundeck/core/resources/TestFileResourceModelSource.java`
- `rundeckapp/grails-app/controllers/rundeck/controllers/FrameworkController.groovy`
- `rundeckapp/src/test/groovy/rundeck/controllers/FrameworkController2Spec.groovy`

## Release Notes

**Security Fix: File Resource Model Source Path Validation (RUN-4671)**

Fixed a security vulnerability where users with project-configure permission could read arbitrary files on the Rundeck server filesystem.

**What changed:**
- File resource model sources are now restricted to the project directory by default
- Administrators can configure additional allowed paths via `rundeck.resourceModelSource.file.allowedBasePaths` in `rundeck-config.properties`
- Path traversal attacks (../, symlinks) are now blocked
- Invalid node source files return empty content instead of exposing file contents

**Impact:**
- Existing file sources within project directories continue to work without changes
- File sources pointing outside the project directory will show a validation error when editing
- To allow file sources outside the project directory, administrators must explicitly configure allowed paths

**Configuration:**
```properties
# Optional: Allow file resource model sources to read from additional directories
rundeck.resourceModelSource.file.allowedBasePaths=/opt/shared/nodes,/var/custom/resources
```


ronaveva (approved):
LGTM

It just needs doc

PR #10445, opened 2026-08-11
[RUN-4543] Fix stored CSS injection via unguarded style attribute in HTML sanitizer

ncofreortiz-hub:
## Tell us about your PR

Security bugfix. `SanitizedHTMLCodec` allows the `style` attribute on `td`, `th`, and several SVG elements (`svg`, `g`, `text`, `path`, `polygon`, `rect`) but never calls `allowStyling()` and configures no `CssSchema`, so arbitrary CSS passes through the sanitizer verbatim. A job option description (or other input rendered via `encodeAsSanitizedHTML`/`SanitizedHTMLCodec.encode`) containing HTML with a `style` attribute like `position:fixed;inset:0;...;background:url('//attacker/leak')` executes that CSS in the browser of anyone who views the page — including admins. This affects job run pages, execution log HTML, and email notifications, since they all route through the same sanitizer policy.

`MarkdownCodec`'s `HtmlRenderer` also had no `escapeHtml(true)`, so raw inline HTML in markdown-authored fields reached the sanitizer as live markup rather than escaped text, giving an extra delivery path for the same payload.

- `SanitizedHTMLCodec.groovy`: added `.allowStyling(CssSchema.DEFAULT)` to both `HtmlPolicyBuilder` chains that permit `style` (the SVG builder, and the `td`/`th` builder in the main policy). This makes the sanitizer parse and constrain CSS property values instead of passing the attribute through unvalidated — `position:fixed` and `url(...)` are stripped, generic styling (`color`, `font-weight`, `text-align`, `opacity`, etc.) still works.
- `MarkdownCodec.groovy`: added `.escapeHtml(true)` to the `HtmlRenderer` builder so raw inline HTML in markdown is rendered as literal escaped text rather than live markup before it reaches the sanitizer.
- All `encodeAsSanitizedHTML`/`SanitizedHTMLCodec.encode()` call sites (`ExecutionController`, `NotificationService`, job option views, etc.) share the single `POLICY` object in `SanitizedHTMLCodec`, so this one fix covers all render paths without touching each call site individually.

The report suggested `CssSchema.MAXIMUM_SAFE`, but that constant doesn't exist in the `owasp-java-html-sanitizer` version currently pinned in this repo (checked the sources jar for `20260313.1`) — only `CssSchema.DEFAULT` is exposed publicly, so that's what's used here.

Also considered removing `style` from `allowAttributes()` entirely instead of guarding it with a `CssSchema`. Went with the guard so existing legitimate styling (colored table cells, etc.) keeps working rather than breaking outright.

**Additional context**

Known trade-off: `fill`/`stroke`/`stroke-width` aren't modeled in this sanitizer's built-in CSS property definitions at all (only generic properties like `color`/`background-color`/`opacity` are), so `style="fill:...;stroke:..."` on SVG `text`/`g`/`path`/`polygon` will now be stripped. Didn't hand-roll a custom `CssSchema.Property` for these since the internal bit-flag grammar for property values isn't part of the public API — guessing it wrong in a CSS-injection fix risks reopening the hole. `rect` still has a bare `fill` attribute allowed directly and is unaffected.

Verified with standalone `groovyc`/`groovy` runs against the actual pinned dependency jars (full `./gradlew build` isn't runnable from this worktree checkout due to an unrelated `grgit`/JGit worktree limitation):
- Exploit payload `<td style="position:fixed;inset:0;...;background:url('//attacker/leak')">` → CSS collapses to safe `width`/`height` only.
- Same payload via SVG `rect` → CSS fully stripped.
- Same payload via Markdown → raw `<td>` tag is HTML-escaped, never becomes live markup.
- Legit styling (`color:red;font-weight:bold;text-align:center` on `td`) still passes through unchanged.

Source: PS-1683 (internal security report).

## Release Notes

Fixed a stored CSS injection vulnerability where the `style` attribute on table cells and SVG elements was not validated by the HTML sanitizer, allowing arbitrary CSS (including full-page overlays and outbound `url()` requests) to be stored via job option descriptions and rendered to other users.

### DIFF
--- a/functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/project/MotdSpec.groovy
+++ b/functional-test/src/test/groovy/org/rundeck/tests/functional/selenium/project/MotdSpec.groovy
@@ -44,17 +44,17 @@ class MotdSpec extends SeleniumBase {
             projectEditPage.save()
             projectEditPage.validateConfigFileSave()
             motdPage.clickMOTD()
-        then: "validate that motd is shown in the navbar and it has the right value"
-            motdPage.waitForMessageShownInProject("a simple message")
+        then: "validate that motd is shown in the navbar and it has the right value, with the script tag inert but not stripped"
+            motdPage.waitForMessageShownInProject(motdMessage)
         when:
             projectDashboard.go("/project/${projectName}/home")
         then: "validate motd shown on project dashboard"
-            motdPage.waitForMessageShownInProject("a simple message")
+            motdPage.waitForMessageShownInProject(motdMessage)
         when:
             homePage.setLoadPath("/menu/home")
             homePage.go()
         then:
-            motdPage.waitForMessageShownInHome("a simple message")
+            motdPage.waitForMessageShownInHome(motdMessage)
         cleanup:
             deleteProject(projectName)
 
@@ -86,17 +86,17 @@ class MotdSpec extends SeleniumBase {
             projectEditPage.clickNavLink(NavProjectSettings.USER_INTERFACE)
             projectEditPage.selectAllMotdPlaces()
             projectEditPage.save()
-        then: "validate that motd is shown in the navbar and it has the right value"
-            motdPage.waitForMessageShownInProject("a simple message")
+        then: "validate that motd is shown in the navbar and it has the right value, with the script tag inert but not stripped"
+            motdPage.waitForMessageShownInProject(motdMessage)
         when:
             projectDashboard.go("/project/${projectName}/home")
         then: "validate motd shown on project dashboard"
-            motdPage.waitForMessageShownInProject("a simple message")
+            motdPage.waitForMessageShownInProject(motdMessage)
         when:
             homePage.setLoadPath("/menu/home")
             homePage.go()
         then:
-            motdPage.waitForMessageShownInHome("a simple message")
+            motdPage.waitForMessageShownInHome(motdMessage)
         cleanup:
             deleteProject(projectName)
     }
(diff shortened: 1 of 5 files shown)


jayas006:
Hi @ncofreortiz-hub , the sanitizer/escapeHtml changes look right to me. One gap: we don’t have unit tests for the actual fix. SanitizedHTMLCodecTest still only covers links/<script>, and there’s no MarkdownCodec test at all. MotdSpec covers the MOTD text change, but it wouldn’t catch this regressing on job option descriptions or logs.
Could you add a few unit tests for the cases you already verified locally?
- malicious style on td / SVG (position:fixed, url(...)) gets stripped
- normal styling still works (color, font-weight, text-align)
- raw HTML in markdown is escaped, while normal markdown like **bold** still renders

That would lock the security behavior in so it can’t silently come back. Thanks!

ncofreortiz-hub:
Thanks for the review! Added unit tests covering exactly those three cases:

- `SanitizedHTMLCodecTest`: `testTdStyleMaliciousCssStripped` (`position:fixed`/`url()` on `td` gets stripped, leaving only safe properties), `testSvgRectStyleMaliciousCssStripped` (same on SVG `rect`, style attribute fully discarded), and `testTdStyleNormalCssStillWorks` (`color`/`font-weight`/`text-align` still pass through unchanged).
- New `MarkdownCodecTest`: `testRawHtmlInMarkdownIsEscaped` (raw `<td style=...>` in markdown source is rendered as escaped literal text) and `testNormalMarkdownStillRenders` (`**bold**` still renders as `<strong>`).

Pushed in the latest commit — let me know if you'd like additional cases covered.

jayas006 (approved):
LGTM 🎉

PR #10449, opened 2026-08-12
[RUN-4646] [Backend] Swapping and/or logic in a Conditional step

carlosrfranco:
## Jira Ticket

[RUN-4646](https://pagerduty.atlassian.net/browse/RUN-4646)

## Summary

Add a toggle to the Conditional workflow step that reverses its default condition-grouping logic, from AND-within-a-set/OR-across-sets to OR-within-a-set/AND-across-sets, and wire it end-to-end through the backend (data model, evaluator, job definition serialization).

- `ConditionalSet` gains `isInvertLogic()`; `ConditionalSetImpl`/`ConditionalStep` add an `invertLogic` boolean field (default `false`, unset jobs unaffected) threaded through `fromMap`/`toMap`.
- `ConditionalEvaluator` honors the flag: `true` evaluates conditions within a group as OR and groups as AND; `false`/unset preserves current behavior.
- `ExecutionUtilService` (the live job-execution flattening/combine path) and `WorkflowDataWorkflowExecutionItemFactory` (the validation-time path) both normalize nested condition sets to DNF before combining parent/child conditionals, so combined sets are always evaluated correctly regardless of either side's polarity.

## Related PR

- rundeckpro PR: (added once opened)

## Testing

- Unit tests added/updated in `ConditionalEvaluatorSpec`, `ConditionalSetSpec`, `ConditionalStepSpec`, `WorkflowDataWorkflowExecutionItemFactoryConditionalSpec`, `ExecutionUtilServiceConditionalSpec`.
- Manually verified via Rundeck MCP against a local dev instance: created and validated a job with `invertLogic: true`, confirmed round-trip through YAML export.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

[RUN-4646]: https://pagerduty.atlassian.net/browse/RUN-4646?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rundeck-data-models/src/main/java/org/rundeck/app/data/model/v1/job/workflow/ConditionalSet.java
+++ b/rundeck-data-models/src/main/java/org/rundeck/app/data/model/v1/job/workflow/ConditionalSet.java
@@ -20,10 +20,10 @@ import java.util.Map;
 
 /**
  * Represents a set of conditions organized in AND/OR groups.
- * The structure uses nested lists where:
+ * The structure uses nested lists where, by default (isInvertLogic() == false):
  * - Each inner list represents an AND group (all conditions must be true)
  * - The outer list represents OR groups (any group being true passes the ConditionalSet)
- * 
+ *
  * Example:
  * conditionGroups = [
  *   [Condition1, Condition2],  // AND group 1: Condition1 AND Condition2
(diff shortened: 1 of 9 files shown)


smartinellibenedetti (approved):
Leaving the PR approved, but asides from the comments from copilot that seem worth to address (most important ones https://github.com/rundeck/rundeck/pull/10449#discussion_r3767898478 and https://github.com/rundeck/rundeck/pull/10449#discussion_r3767898592 ), is there any chances that an exported job will have the wrong value for the flag as the invertLogic is showing up now both in the top and nested layers? 

In fact now I'm unsure, but wasn't the original plan to have the flag just at the top layer?

PR #10456, opened 2026-08-13
[RUN-4756] Show conditional substeps in Definition modal

jayas006:
Syntax customer feedback: Definition tab only showed top-level conditional steps. Render nested substeps with Sub-steps label and 2.1-style numbering in read-only workflow view, matching design feedback for nested conditionals.

## Tell us about your PR

Bugfix — Syntax reported that the Definition modal on job show only displays top-level conditional steps and does not show substeps inside the conditional.
When viewing a job’s Definition tab, workflows with conditional steps only showed the conditional step itself (e.g. description nested-2), not the commands/jobs nested inside it. Customers could not read the full workflow structure from Definition.

This change renders conditional substeps in the read-only workflow list (legacy GSP), including nested conditionals, with clear hierarchy numbering per design feedback.

- In _wflistContent.gsp, when edit is false and the step is a ConditionalStep with subSteps, render a new partial _wflistSubStepsContent.gsp.
- Recursive partial lists substeps under Sub-steps: with hierarchical labels (2.1, 2.1.1, etc.) based on the parent step number.
- Substep error handlers still use the existing on error: pattern.
- Scoped to read-only Definition (!edit) so the legacy job edit workflow UI is unchanged.
- Styling in process-flow.scss (nested list, left border, inline numbering) aligned with existing Definition / error-handler patterns.
- i18n: Workflow.conditional.subSteps.label=Sub-steps:
- Note: This does not add the workflow editor “If” UI from Figma; Definition still uses the legacy GSP view and shows step descriptions, not condition expressions.

**Additional context**

- Ticket: RUN-4756
- Customer: Syntax feedback (M3)
- UX: Design reviewed Option A (numbered substeps) vs simple nested list
- Test: Job with nested conditionals → Job show → Definition tab; verify substeps and numbering; confirm plain steps and error handlers unchanged
- ## Release Notes

### DIFF
--- a/rundeckapp/grails-app/i18n/messages.properties
+++ b/rundeckapp/grails-app/i18n/messages.properties
@@ -231,6 +231,7 @@ Workflow.stepErrorHandler.description=The error handler will execute if the step
 Workflow.stepErrorHandler.title=Error Handler
 Workflow.stepErrorHandler.keepgoingOnSuccess.label=Keep going on success.
 Workflow.stepErrorHandler.keepgoingOnSuccess.description=If the Workflow keepgoing is false, this allows the Workflow to continue when the Error Handler is successful
+Workflow.conditional.subSteps.label=Sub-steps\:
 Workflow.stepErrorHandler.label.on.error=on error\:
 Workflow.stepErrorHandler.label.keep.going.on.success=keep going on success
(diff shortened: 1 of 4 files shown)


ltamaster (approved):
LGTM

PR #10457, opened 2026-08-13
[PS-1689] Auto-enable Secure flag on JSESSIONID cookie for https deployments

ncofreortiz-hub:
## Tell us about your PR

Security bugfix. Ref: [PS-1689](https://pagerduty.atlassian.net/browse/PS-1689) — the `JSESSIONID` cookie is set with `HttpOnly` and `SameSite=Lax` but never `Secure`. On an HTTP-served instance (default `server.port: 4440`, default `http://` `serverURL`, HSTS off by default), the session cookie can be transmitted in cleartext and captured by an on-path attacker.

- `JettyServletContainerCustomizer` now auto-enables the `Secure` flag on the session cookie when `grails.serverURL` uses `https`, unless the operator has already set `server.servlet.session.cookie.secure` explicitly (that setting is respected either way). This fixes TLS deployments without breaking default HTTP-only deployments (forcing `Secure` unconditionally would stop the cookie being sent back to the browser over `http://`, breaking login).
- Added `JettyServletContainerCustomizerSpec` covering: https → secure enabled, http → unchanged, explicit `secure: false` respected, no `serverUrl` configured, and the `isHttpsUrl` scheme parsing helper.
- Added a comment in `application.yml` documenting the auto-detection behavior and how to override it.
- Documented the existing (previously unexplained) `rundeck.web.jetty.servlet.stsMaxAgeSeconds` / `stsIncludeSubdomains` properties in `resources.groovy`, including recommended values for TLS deployments, and clarified that HSTS is only ever added when the connector has a secure port (per `JettyServletHstsCustomizer#checkSSL`).

- Setting `server.servlet.session.cookie.secure: true` unconditionally in `application.yml` — rejected because it would break the still-common default HTTP-only deployment (browser won't return a `Secure` cookie over plain HTTP, breaking session/login).
- Changing the default HSTS `stsMaxAgeSeconds` from `-1` to a positive value for all TLS deployments — rejected as out of scope for this PR; forcing HSTS on by default is a larger behavior change (risk of lockout if TLS is later misconfigured/rolled back) that deserves its own discussion. Left as documented, opt-in guidance instead.
- Hardening-guide documentation update (remediation item 3 from the finding) — the hardening guide lives in the separate `rundeck-docs` repo, not in this codebase, so it's not included here.

**Additional context**
Verified locally: `./gradlew :rundeckapp:test --tests "rundeckapp.init.servlet.JettyServletContainerCustomizerSpec" --tests "rundeckapp.init.servlet.JettyServletHstsCustomizerSpec"` — 13 tests, 0 failures.

## Release Notes
Session cookies are now automatically marked `Secure` when Rundeck's configured server URL uses `https`, closing a gap where session cookies could be transmitted over an insecure connection. HTTP-only deployments are unaffected.

[PS-1689]: https://pagerduty.atlassian.net/browse/PS-1689?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rundeckapp/grails-app/conf/application.yml
+++ b/rundeckapp/grails-app/conf/application.yml
@@ -399,6 +399,9 @@ server:
             cookie:
                 http-only: true
                 same-site: "Lax"
+                # secure is intentionally left unset here: JettyServletContainerCustomizer auto-enables
+                # it when grails.serverURL uses https, so HTTP-only deployments keep working. Set this
+                # explicitly (true/false) to override that auto-detection.
             tracking-modes: 'cookie'
 rundeck:
     web:
--- a/rundeckapp/grails-app/conf/spring/resources.groovy
+++ b/rundeckapp/grails-app/conf/spring/resources.groovy
@@ -923,6 +923,12 @@ beans={
         serverUrl = grailsApplication.config.getProperty('grails.serverURL', String.class)
     }
 
+    // HSTS is only ever added when the connector has a secure port configured (see
+    // JettyServletHstsCustomizer#checkSSL), so these properties only take effect for TLS deployments.
+    // A negative stsMaxAgeSeconds (the default) disables the Strict-Transport-Security header.
+    // For TLS-terminated-at-Rundeck deployments, operators should set:
+    //   rundeck.web.jetty.servlet.stsMaxAgeSeconds: 31536000 (1 year)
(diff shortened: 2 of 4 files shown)


ronaveva (approved):
LGTM

PR #10477, opened 2026-08-18
[RUN-4553] Migrate webhook auth tokens off LEGACY plaintext storage to SECURED (PS-1686)

ncofreortiz-hub:
## Tell us about your PR

Security fix. PS-1686 — webhook auth tokens were forced to `AuthTokenMode.LEGACY` (`ApiService.groovy:114`), storing the raw token value in the DB with no hashing, unlike every other API token type which is SHA-256 hashed at rest (`AuthTokenMode.SECURED`). A DB dump or backup exposure yielded webhook tokens directly usable to trigger job executions.

Full design rationale: [`docs/superpowers/specs/2026-08-18-webhook-tokens-secured-design.md`](docs/superpowers/specs/2026-08-18-webhook-tokens-secured-design.md). Four changes, each with test coverage:

1. **Root cause fix** — `WebhookService.saveHook` was publishing the *persisted* token value (`at.token`) as the public webhook URL secret instead of the transient clear value (`at.clearToken`). This is why WEBHOOK was hard-forced to LEGACY in the first place: flipping the mode without this fix would have published a SHA-256 hash as the "secret" and broken every webhook.
2. **SECURED by default** — removed the `WEBHOOK -> LEGACY` special-case in `ApiService.generateAuthToken` and the duplicated branch in `GormTokenDataProvider.createWithId`. New webhook tokens are now hashed at rest like every other token type.
3. **Fixed a silently-breaking lookup** — `GormTokenDataProvider.findByTokenAndType` used a raw-only GORM finder that would never match a SECURED row's hash; it now delegates to the existing hybrid SECURED-hash-or-LEGACY-raw matcher (`tokenLookupWithType`), fixing `RundeckAuthTokenManagerService.importWebhookToken`/`.updateAuthRoles`.
4. **Startup migration for existing tokens** — `BootStrap.groovy` gained `migrateWebhookTokensToSecured()` (mirrors the pre-existing `ensureTypeOnAuthToken()` pattern), a one-time, idempotent, per-row fixup that re-hashes existing LEGACY/null-mode webhook rows to SECURED. It never touches the separate `webhook.auth_token` column (the public URL secret), so existing webhook URLs keep working with zero user action.

- Forcing re-issuance of all existing webhook tokens instead of re-encoding in place — rejected as a breaking change for every external system calling existing webhook URLs.
- A Liquibase SQL changeset (DB-specific SHA-256 functions) for the data migration — rejected in favor of a DB-agnostic Groovy fixup reusing the existing hashing utility, consistent with the precedent already set by `ensureTypeOnAuthToken()`.
- Feature-flag-gated rollout of the mode switch — rejected, no precedent for gating token-mode behavior elsewhere in the codebase.

**Additional context**
Due date 2026-09-17. Out of scope (tracked separately per the security report): moving `SECURED` off plain SHA-256 to a slow KDF.

Verification: 109/109 tests passing across the 5 affected Spock spec files (`WebhookServiceSpec`, `ApiServiceSpec`, `RundeckAuthTokenManagerServiceSpec`, `GormTokenDataProviderSpec`, `BootStrapTest`), plus a full `./gradlew build -x check` success. Each of the 4 changes went through independent spec-compliance and code-quality review, plus a final holistic review confirming the end-to-end token lifecycle (creation → publish → lookup) and the existing-token migration path are both correct and that no other WEBHOOK/LEGACY call site was missed.

## Release Notes

Webhook auth tokens are now hashed at rest (SHA-256), matching every other API token type. Existing webhook tokens are automatically migrated on next server startup; no action is required and existing webhook URLs continue to work unchanged.


ronaveva (changes requested):
I have a couple of changes to request

1. Use @Subscriber("rundeck.bootstrap") to trigger the migration process
2. Remove the md file
3. Check for the migration to work with all supported DBs
4. Would be good to add a test with an old unsecured token, check for the WH to work, migrate the token and check for the WH to continue working

ncofreortiz-hub:
Addressed all 4 review points, @ronaveva:

1. **@Subscriber('rundeck.bootstrap')** — moved the migration into its own service, `WebhookTokenSecureMigrationService`, triggered via `@Subscriber('rundeck.bootstrap')` instead of a direct call from `BootStrap.init`. The primary-server/cluster gate that used to be a local variable inside `BootStrap.groovy` is reconstructed using already-injectable beans (`frameworkService.isClusterModeEnabled()`/`getServerUUID()`, `configurationService.getString("primaryServerId")`) — no new plumbing needed. Covered by 5 tests including cluster-primary, cluster-non-primary, and cluster-with-no-primaryServerId-configured cases.
2. **Removed the stray .md file** from the PR.
3. **Verified DB compatibility** — the migration's SQL is plain `SELECT`/`UPDATE` with `IS NULL`/`=`/`AND`/`OR` predicates and JDBC `?` placeholders only, no vendor-specific functions/sequences/DDL. Checked this against every `dbms="oracle"`-specific Liquibase changeset already in this codebase (e.g. `ExecReportJcExecIdToExecutionId.groovy`'s `bigint`→`NUMBER(19,0)` cast, sequence-existence checks in `HibernateIndex.groovy`) — all of those are DDL/type-system differences that don't apply to this simple DML. Documented the reasoning directly in the new service's class-level Groovydoc. (Note: this repo has no automated cross-DB test matrix — MySQL/Postgres/Oracle/MSSQL/H2 compatibility for raw SQL fixups is verified by convention here, same as the pre-existing sibling `ensureTypeOnAuthToken()`.)
4. **Added an end-to-end test** (`WebhookTokenMigrationIntegrationSpec`) that creates a webhook in the old pre-fix shape (LEGACY auth_token row + matching plaintext `webhook.auth_token`), confirms both lookup paths resolve with the raw token, runs the real migration, and confirms both still resolve with the identical raw token afterward — plus that the auth_token row is now actually hashed and no duplicate/stale row is left behind.

113/113 tests pass across all 7 affected spec files, full `./gradlew build -x check` succeeds.

ronaveva (changes requested):
1. Use Gorm instead of native sql
2. Would this change be rollback able? 
3. I think it does not apply, but it would be good to check RbaTokenDataProvider on the PRO side and see if more changes are required there

ncofreortiz-hub:
Addressed the 3 new points, @ronaveva:

1. **Use GORM instead of native SQL** — replaced the raw `groovy.sql.Sql` SELECT/UPDATE with `AuthToken.createCriteria()` + a per-row `AuthToken.withNewTransaction { ... }` (same pattern already used in `NotificationService` for per-item failure isolation). Now that the migration lives in its own `@Subscriber`-triggered service (not raw-`BootStrap.init`), there's no reason to avoid GORM — the app context/Hibernate session is fully up by the time `rundeck.bootstrap` fires. This also fully resolves point 3 from the first review round (DB compatibility): Hibernate's dialect layer handles cross-database differences for every query here, so there's no hand-written SQL left to reason about at all.

2. **Rollback-able?** Two separate things here:
   - The *data* migration itself is one-way by design — a SHA-256 hash can't be reversed back to plaintext, and that's the whole point of the fix (nothing to "undo" without recreating the original security gap).
   - Rolling back the *application code* to a pre-PS-1686 version is safe for the main authentication path: `GormTokenDataProvider.tokenLookupWithType` (used by `SetUserInterceptor` to authenticate every incoming webhook POST) already had hybrid SECURED-hash-or-LEGACY-raw matching *before* this PR started, so it keeps authenticating already-migrated tokens correctly even on old code. The one narrow gap: rolling back past commit `2a2cea6898` (the `findByTokenAndType` fix) would break `RundeckAuthTokenManagerService.importWebhookToken`/`.updateAuthRoles` — i.e. re-importing a webhook or editing its roles — for tokens already migrated to SECURED, since the pre-fix version of that one finder was raw-only. That's an admin-triggered path, not the request-authentication path, so existing webhooks keep firing correctly either way.

3. **RbaTokenDataProvider (PRO side)** — can't inspect or modify PRO-only code from this OSS repo/session. What I *can* confirm from here: the public `TokenDataProvider` interface (`rundeck-data-models/.../TokenDataProvider.java`) was not touched by any commit in this PR — method signatures are unchanged, so `RbaTokenDataProvider` (or any other implementation) won't break at compile time. If it has its own analogous WEBHOOK-forced-LEGACY special-casing mirroring what `GormTokenDataProvider`/`ApiService` used to have here, that would need an equivalent fix tracked separately in the PRO repo — flagging for the team to check, since I have no visibility into that codebase.

113/113 tests pass across all 7 affected spec files, full `./gradlew build -x check` succeeds.

ronaveva (dismissed):
Would be good to mark the new service or the method with 

@Transactional(propagation = Propagation.REQUIRES_NEW)

ncofreortiz-hub:
Addressed, @ronaveva — marked \`migrateWebhookTokensToSecured()\` with \`@Transactional(propagation = Propagation.REQUIRES_NEW)\`, so it always runs in its own independent transaction regardless of whatever context (if any) the \`rundeck.bootstrap\` event fires from.

Put it on the method itself rather than extracting a new private helper, since Spring's \`@Transactional\` AOP proxy only intercepts calls that come from *outside* the bean — a self-invoked private method wouldn't actually get wrapped, it'd silently no-op. The existing per-row \`AuthToken.withNewTransaction\` stays as-is; it's solving a different problem (isolating one bad row's failure from the rest of the batch), and GORM's \`withNewTransaction\` isn't subject to that self-invocation limitation since it isn't a Spring-proxied call.

113/113 tests pass across all 7 affected spec files, full \`./gradlew build -x check\` succeeds.

Jesus-Osuna-M (approved):
Code looks good to approve: existing webhooks are migrated on startup, public URLs are unchanged, and rolling the binary back still authenticates via the existing hybrid lookup.

One process fix before merge: please update the PR title to our required format so the Jira ticket is associated and **PR Hygiene Check** passes:

`[RUN-4553] Migrate webhook auth tokens off LEGACY plaintext storage to SECURED`

The check requires a leading `[RUN-####]` (see `.github/workflows/check-pr.yml`). `[PS-1686]` in the title is not enough for that association.

PR #10490, opened 2026-08-24
[RUN-4693] Harden job option input validation (enforced fail-closed, undeclared-option rejection, OptsUtil quoting)

ronaveva:
## Tell us about your PR

Security-hardening follow-up for RUN-4693 (reopened), addressing review feedback from Garrett Rappaport on the merged opt-in option-input allowlist (#10429). Closes several gaps by which a `run`-only user could get unvalidated option values into the execution environment (option DataContext / `RD_OPTION_*` env vars, which also feed the `ssh-variable-export-pattern` sink of RUN-4579).

- **Enforced fail-closed** — an `enforced` option whose remote `valuesUrl`/plugin value list fails to load left `optionValues` null, so the enforced check never fired and the default allowlist skipped it (`!opt.enforced`) → zero validation. Introduces `hasOwnValueConstraint(opt) = opt.regex || (opt.enforced && opt.optionValues)`; the default allowlist now applies unless the option is genuinely self-constrained.
- **enforced + regex** now validates the regex (the per-option regex branches previously required `!opt.enforced`, silently ignoring the regex on enforced options).
- **typeFile exemption** — file-ref option values are system-generated UUIDs already validated by `fileUploadService`; they are now exempt from the default allowlist so a strict pattern doesn't reject them.
- **OptsUtil control-char quoting** — `escapeChars` quoted only `{space, " ' CR LF}` while `QuotedStringTokenizer.burst()` splits on any char `<= 32`; a TAB survived `join()` and was re-split by `burst()`, letting extra `-opt value` tokens be smuggled through the webhook payload → argString → runJob round-trip. `escapeChars` now also quotes values containing any char `<= 32`. Non-ASCII text (e.g. Japanese, full-width space U+3000) is above 32 and unaffected. Adds a `burst(join(x)) == x` round-trip test.
- **Reject undeclared options** — options not declared on the job bypass all server-side validation (validateOptionValues only iterates declared options) yet reach the option DataContext / `RD_OPTION_*` env vars. The execution is now **created and then failed at start** (in `executeAsyncBegin`, before any workflow step runs) with a clear message in the execution log output. Gated by `rundeck.execution.rejectUndeclaredOptions` (default `true`; System-Configuration-editable / `rundeck-config`-overridable).

**Behavior changes (default on unless noted):**
- enforced + regex: a value that is in the enforced allowed-values list but violates the regex now fails validation.
- undeclared options: an execution that provides options not defined on the job now fails at start (opt-out via `rundeck.execution.rejectUndeclaredOptions=false`). This affects re-running a job whose option set has since changed, and API/webhook callers that pass extra options.
- OptsUtil: values containing ASCII control chars (e.g. TAB) are now quoted (kept as a single token) instead of being re-split.
- Gated by an allowlist pattern being configured (no effect otherwise): enforced fail-closed, typeFile exemption.

Hardening the sink itself (escaping in `ReplaceTokenReader`/`ScriptVarExpander`, or gating `expandTokens`) is the only change that would close the finding by default, but it is the same class of change that caused the RUN-4175 → #10091 → revert regressions and has no context-free-correct form for a script body. Deferred as a separate, carefully-designed (likely flagged) change.

**Additional context**

- Config keys resolve through `ConfigurationService` and are registered as `SysConfigProp` (System Configuration UI): `rundeck.option.input.validation.default.pattern` (from #10429) and `rundeck.execution.rejectUndeclaredOptions`.
- Scope of undeclared-option rejection: top-level executions (UI/API/webhook/scheduled). The job-reference path (`createJobReferenceContext`) is author-controlled and left as a follow-up.
- Pending (not in this PR): sink hardening; documentation/release-note caveats (both config keys, `(?s).*` for newline matching, `\p{L}` guidance for non-ASCII/Japanese option values).

## Release Notes

Security hardening for job option input validation: enforced options whose remote value list fails to load are now validated by the configurable input allowlist; per-option regex is now applied to enforced options; `typeFile` options are exempt from the allowlist; option values containing control characters are quoted consistently to prevent argument smuggling; and executions that provide options not defined on the job now fail at start with a clear log message (configurable via `rundeck.execution.rejectUndeclaredOptions`, default enabled).

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/config/RundeckConfigBase.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/config/RundeckConfigBase.java
@@ -159,6 +159,13 @@ public class RundeckConfigBase {
         RetryConfig finalize;
         RetryConfig status;
         ExecutionLogs logs;
+        /**
+         * RUN-4693: when true (default), an execution that provides options not declared on the job
+         * is created and then failed at start. Set false to restore the legacy passthrough. Bound
+         * from {@code rundeck.execution.rejectUndeclaredOptions} so it is resolvable via
+         * ConfigurationService and editable in the System Configuration UI.
+         */
+        Boolean rejectUndeclaredOptions;
 
         @Data
         public static class RetryConfig {
--- a/core/src/main/java/com/dtolabs/rundeck/core/utils/OptsUtil.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/utils/OptsUtil.java
@@ -88,7 +88,12 @@ public class OptsUtil {
     }
 
     public static void escapeChars(StringBuilder out, String str) {
-        if (StringUtils.containsNone(str, CSV_SEARCH_CHARS) && !"".equals(str)) {
+        // A value is only safe to emit unquoted if it contains none of the CSV search chars AND no
+        // control/whitespace char with code <= 32. QuotedStringTokenizer.burst() splits on any char
+        // <= 32 (StrMatcher.trimMatcher()), so a value containing e.g. a TAB that is not quoted here
+        // would be re-split on burst() — breaking the join()/burst() round-trip and allowing extra
+        // tokens (arg smuggling). Quoting such values keeps the round-trip intact. See RUN-4693.
+        if (StringUtils.containsNone(str, CSV_SEARCH_CHARS) && !"".equals(str) && !containsControlChar(str)) {
             if (str != null) {
                 out.append(str);
             }
@@ -103,6 +108,25 @@ public class OptsUtil {
             out.append(c);
         }
         out.append(DBL_QUOTE);
+    }
+
+    /**
+     * @param str input (may be null)
+     * @return true if the string contains any character with code point &lt;= 32 (space and control
+     * characters such as TAB, VT, FF), which {@link QuotedStringTokenizer} treats as a delimiter.
+     * Only ASCII control/whitespace is matched; all non-ASCII text (e.g. Japanese, full-width space
+     * U+3000) is above 32 and is not affected.
+     */
+    private static boolean containsControlChar(final String str) {
+        if (str == null) {
+            return false;
+        }
+        for (int i = 0; i < str.length(); i++) {
+            if (str.charAt(i) <= 32) {
+                return true;
+            }
+        }
+        return false;
     }
 
 }
--- a/core/src/test/java/com/dtolabs/rundeck/core/utils/TestOptsUtil.java
+++ b/core/src/test/java/com/dtolabs/rundeck/core/utils/TestOptsUtil.java
@@ -201,5 +201,40 @@ public class TestOptsUtil extends TestCase {
         );
     }
 
+    /**
+     * RUN-4693: a value containing a TAB (or any char <= 32 not already covered) must be quoted by
+     * join() so burst() does not re-split it into extra tokens.
+     */
+    public void testJoinQuotesTab() {
+        assertEquals("-arg \"a\tb\"",
+                OptsUtil.join(new String[]{"-arg", "a\tb"})
+        );
+    }
+
+    /**
+     * RUN-4693: burst(join(x)) must round-trip. Values containing whitespace/control chars <= 32
+     * must survive as single tokens; non-ASCII text (Japanese, full-width space U+3000) is above 32
+     * and must be unaffected.
+     */
+    public void testJoinBurstRoundTrip() {
+        String[][] cases = new String[][]{
+                {"-arg", "plain"},
+                {"-arg", "a b"},
+                {"-arg", "a\tb"},                 // TAB
+                {"-arg", "a\r\nb"},               // CR/LF
+                {"-x", "foo\t-inject\tvalue"},    // smuggling attempt: must stay ONE token
+                {"-arg", "あ　日"},   // Japanese あ　日 incl. full-width space U+3000
+        };
+        for (String[] input : cases) {
+            String joined = OptsUtil.join(input);
+            String[] burst = OptsUtil.burst(joined);
+            assertEquals(
+                    "round-trip failed for " + Arrays.asList(input) + " joined=[" + joined + "]",
+                    Arrays.asList(input),
+                    Arrays.asList(burst)
+            );
+        }
+    }
+
 }
 
--- a/functional-test/src/test/groovy/org/rundeck/tests/functional/api/job/JobExecutionSpec.groovy
+++ b/functional-test/src/test/groovy/org/rundeck/tests/functional/api/job/JobExecutionSpec.groovy
@@ -868,6 +868,11 @@ class JobExecutionSpec extends BaseContainer {
                     "    <uuid>165ef9b9-61dc-470c-91aa-3f6dc248249d</uuid>\n" +
                     "  </job>\n" +
                     "  <job>\n" +
+                    "    <context>\n" +
+                    "      <options>\n" +
+                    "        <option name='opt2' />\n" +
+                    "      </options>\n" +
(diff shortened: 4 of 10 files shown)


carlosrfranco:
## Review against RUN-4693 reopening comment

I checked this PR against the 4 items Garrett Rappaport listed in the [reopening comment](https://pagerduty.atlassian.net/browse/RUN-4693?focusedCommentId=1068190), and live-tested both branches (main vs. this PR) to confirm behavior, not just read the code.

### 1. Enforced fail-closed + enforced-regex — ✅ Fixed, verified live
`hasOwnValueConstraint(opt) = opt.regex || (opt.enforced && opt.optionValues)` replaces the `!opt.enforced` guards, as requested.

- Enforced option with a `valuesUrl` that fails to resolve, value `x; echo INJECTED` → **main accepts it** (validation skipped entirely), **this branch rejects it** (`api.error.job.options-invalid`, falls through to the default allowlist).
- Enforced option with static list `["a b","cd"]` + regex `[a-zA-Z0-9]+`, value `a b` (in the enforced list, violates regex) → **main accepts it**, **this branch rejects it** with the regex error.

### 2. Reject undeclared options — ⚠️ Partially fixed, verified live
Different mechanism than requested: enforced in `executeAsyncBegin` (execution created, then failed at start) rather than at input marshalling (`parseJobOptionInput`/`generateJobArgline`).

- Job with zero declared options, run with `-ghost pwned` → **main runs it to completion** (option silently accepted, `hello` printed), **this branch fails at start** with `Execution rejected: option(s) not defined on this job were provided: [ghost]`.
- Gaps the PR itself documents: `generateJobArgline` still passes undeclared options through unchanged (covered by a new characterization test, not a fix), and the **job-reference path (`createJobReferenceContext`) is explicitly left as a follow-up** — a job step that triggers another job can still smuggle undeclared options into the referenced job's option context.

### 3. OptsUtil TAB/control-char quoting gap — ✅ Fixed
`escapeChars` now quotes any value containing a char ≤ 32, matching `QuotedStringTokenizer.burst()`'s split behavior. The new `burst(join(x)) == x` round-trip test covers TAB, CR/LF, and an explicit smuggling case, and confirms non-ASCII (Japanese, U+3000) is unaffected. (Verified via code + unit tests — didn't rig a webhook to exercise this end-to-end.)

### 4. Documentation of both config keys + caveats — ❌ Not done
Explicitly called out in the PR description as "Pending (not in this PR)."

### The bigger issue: item n. 1 in the comment ("the sink is unchanged")
Garrett's first and most fundamental point — that `ScriptNodeStepPlugin`/`ReplaceTokenReader` still substitutes `@option.x@` verbatim into inline scripts — is **not addressed**, by explicit design. I confirmed this live: with no allowlist configured (the default), running the original HackerOne PoC (`"; id; echo INJECTED"`) against an inline-script job produced the **same result on this branch as on main** — `id` executed, injected output printed. The PR's own new test (`OptionInputValidationSpec`) documents this same fact ("default is vulnerable").

So the branch remains exploitable via the exact original PoC as long as no allowlist is configured — which is the out-of-the-box state.

### Bottom line
This PR correctly fixes 3 concrete validation-bypass bugs found during review of #10429 — all 3 reproduced on main and fixed here in live testing. Measured against "what I'd like to see to get this closed out," it's incomplete:
- Undeclared-option rejection only covers top-level executions, not job-references (explicitly deferred).
- No documentation of the two config keys or caveats (explicitly deferred).
- The root cause — command injection in the default (opt-out) configuration — remains fully exploitable, deferred as a separate, more invasive change.

Recommend not closing RUN-4693 on the strength of this PR alone — good incremental hardening, but the ticket's core finding (pre-auth RCE via webhook → inline script, no config required) is still live on this branch.

ronaveva:
> ## Review against RUN-4693 reopening comment
> I checked this PR against the 4 items Garrett Rappaport listed in the [reopening comment](https://pagerduty.atlassian.net/browse/RUN-4693?focusedCommentId=1068190), and live-tested both branches (main vs. this PR) to confirm behavior, not just read the code.
> 
> ### 1. Enforced fail-closed + enforced-regex — ✅ Fixed, verified live
> `hasOwnValueConstraint(opt) = opt.regex || (opt.enforced && opt.optionValues)` replaces the `!opt.enforced` guards, as requested.
> 
> * Enforced option with a `valuesUrl` that fails to resolve, value `x; echo INJECTED` → **main accepts it** (validation skipped entirely), **this branch rejects it** (`api.error.job.options-invalid`, falls through to the default allowlist).
> * Enforced option with static list `["a b","cd"]` + regex `[a-zA-Z0-9]+`, value `a b` (in the enforced list, violates regex) → **main accepts it**, **this branch rejects it** with the regex error.
> 
> ### 2. Reject undeclared options — ⚠️ Partially fixed, verified live
> Different mechanism than requested: enforced in `executeAsyncBegin` (execution created, then failed at start) rather than at input marshalling (`parseJobOptionInput`/`generateJobArgline`).
> 
> * Job with zero declared options, run with `-ghost pwned` → **main runs it to completion** (option silently accepted, `hello` printed), **this branch fails at start** with `Execution rejected: option(s) not defined on this job were provided: [ghost]`.
> * Gaps the PR itself documents: `generateJobArgline` still passes undeclared options through unchanged (covered by a new characterization test, not a fix), and the **job-reference path (`createJobReferenceContext`) is explicitly left as a follow-up** — a job step that triggers another job can still smuggle undeclared options into the referenced job's option context.
> 
> ### 3. OptsUtil TAB/control-char quoting gap — ✅ Fixed
> `escapeChars` now quotes any value containing a char ≤ 32, matching `QuotedStringTokenizer.burst()`'s split behavior. The new `burst(join(x)) == x` round-trip test covers TAB, CR/LF, and an explicit smuggling case, and confirms non-ASCII (Japanese, U+3000) is unaffected. (Verified via code + unit tests — didn't rig a webhook to exercise this end-to-end.)
> 
> ### 4. Documentation of both config keys + caveats — ❌ Not done
> Explicitly called out in the PR description as "Pending (not in this PR)."
> 
> ### The bigger issue: item n. 1 in the comment ("the sink is unchanged")
> Garrett's first and most fundamental point — that `ScriptNodeStepPlugin`/`ReplaceTokenReader` still substitutes `@option.x@` verbatim into inline scripts — is **not addressed**, by explicit design. I confirmed this live: with no allowlist configured (the default), running the original HackerOne PoC (`"; id; echo INJECTED"`) against an inline-script job produced the **same result on this branch as on main** — `id` executed, injected output printed. The PR's own new test (`OptionInputValidationSpec`) documents this same fact ("default is vulnerable").
> 
> So the branch remains exploitable via the exact original PoC as long as no allowlist is configured — which is the out-of-the-box state.
> 
> ### Bottom line
> This PR correctly fixes 3 concrete validation-bypass bugs found during review of #10429 — all 3 reproduced on main and fixed here in live testing. Measured against "what I'd like to see to get this closed out," it's incomplete:
> 
> * Undeclared-option rejection only covers top-level executions, not job-references (explicitly deferred).
> * No documentation of the two config keys or caveats (explicitly deferred).
> * The root cause — command injection in the default (opt-out) configuration — remains fully exploitable, deferred as a separate, more invasive change.
> 
> Recommend not closing RUN-4693 on the strength of this PR alone — good incremental hardening, but the ticket's core finding (pre-auth RCE via webhook → inline script, no config required) is still live on this branch.

I would prefer not going the full way to avoid this behaviors by default, why not? because going this way would break customer jobs. This is the reason why this is an opt-in solution.

I could do the sink solution but it would require a more "core" change, since we can't just change the behavior from true to false (expandTokenInScriptFile) on ScriptNodeStepPlugin. So i would prefer to do that on a separated PR.

carlosrfranco (approved):

PR #10492, opened 2026-08-25
[RUN-4131] Warn and revert conditional-logic steps when the feature is disabled

smartinellibenedetti:
## Summary
- Job editor's `WorkflowEditor.vue` no longer has any knowledge of "conditional logic" - that concept now lives entirely in the rundeckpro plugin, connected via a new `UiSocket` extension point (`section="workflow-editor" location="steps"`) wrapping the step list.
- `createStepsData` reverted to its original, feature-agnostic single-argument form.
- Removed the unused/orphaned early-access `Banner.vue` (never actually wired up anywhere) and its dead i18n keys.
- Fixed a pre-existing cascade bug in the shared `message.scss` PrimeVue `Message` wrapper: PrimeVue's own runtime-injected theme rules for `.p-message-*` (background, outline, text color) were winning cascade ties against this wrapper's overrides. Raised specificity (`&.p-message-warn` instead of `&-warn`, etc.) so the wrapper's colors always win regardless of PrimeVue's injection order - fixes a thick double-outline and unreadable white text in dark mode.

## Test plan
- [x] `npm run test:unit` in `ui-trellis` - all workflow/job editor suites pass
- [x] Verified live in browser: job with a real conditional-logic step (`type: "conditional"` + top-level `conditionGroups`/`subSteps`, matching `ConditionalStep.toMap()`) renders correctly with the flag disabled via rundeckpro's new widget


jayas006 (approved):
LGTM

PR #10494, opened 2026-08-25
[RUN-4694] Fix XXE in ResourceXMLParser allowing arbitrary file read and SSRF

carlosrfranco:
## Jira Ticket

[RUN-4694](https://pagerduty.atlassian.net/browse/RUN-4694)

## Summary

Harden `ResourceXMLParser` against XML External Entity (XXE) attacks so that a user holding only the delegable, non-admin project `configure` permission can no longer use a writeable node source to read arbitrary local files or perform SSRF via the `POST /api/{ver}/project/{proj}/source/{idx}/resources` endpoint.

- `SAXReader` construction now uses `SAXReader.createDefault()`, which sets the JAXP anti-XXE features (`load-external-dtd`, `external-general-entities`, `external-parameter-entities` all disabled)
- `EntityResolver.resolveEntity()` null-guards the publicId comparison so a `SYSTEM`-only entity (null publicId) no longer throws an NPE
- `EntityResolver.resolveEntity()` now fails closed, returning an empty `InputSource` instead of `null` for any entity that isn't the bundled project DTD, so the parser never falls back to resolving external URIs itself
- Legitimate parsing of `resourcexml` documents using the bundled project DTD continues to work unchanged

## Testing

Added `ResourceXMLParserSpec.groovy` (Spock), replacing the legacy JUnit `TestResourceXMLParser.java`, covering:
- Existing parser behavior (missing file, receiver callbacks, early-stop, attribute parsing)
- Legitimate bundled-DTD parsing still succeeds
- `PUBLIC file://` entity no longer leaks local file content
- `PUBLIC` external parameter-entity chain no longer triggers an outbound HTTP request (SSRF)
- `EntityResolver.resolveEntity()` no longer throws `NullPointerException` for a null publicId
- A malformed DOCTYPE declaration fails cleanly via `ResourceXMLParserException`

Each new security test was verified to fail against the pre-fix code and pass after the fix.

[RUN-4694]: https://pagerduty.atlassian.net/browse/RUN-4694?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ronaveva (approved):
LGTM

PR #10496, opened 2026-08-25
[RUN-4757] Add toggle to hide skipped steps in execution results

jayas006:
## Tell us about your PR

Enhancement. Fixes RUN-4757
Syntax customer feedback: Conditionals that don't run make nodes show as "incomplete" in execution results, which can be confusing when the job actually succeeded.

Added "Hide skipped steps" checkbox to the execution page (Nodes tab). When enabled:                                                                                               
  - Nodes with only skipped conditionals show "All Steps OK" instead of "X Step not run"                                                                                             
  - INCOMPLETE count and COMPLETE % update accordingly  

- Display-only approach (separate displaySummary observable) - more complex, deferred
- Persistent job setting - could be follow-up enhancement

**Additional context**

- Checkbox only visible on Nodes tab
- Toggle affects overall stats (intentional per customer ask)
- Default is OFF (original behavior preserved)

## Release Notes

Added "Hide skipped steps" toggle to execution results page, allowing users to hide incomplete status for nodes where conditional steps were skipped.

### DIFF
--- a/rundeckapp/grails-app/assets/javascripts/executionStateKO.js
+++ b/rundeckapp/grails-app/assets/javascripts/executionStateKO.js
@@ -830,7 +830,10 @@ function RDNode(name, steps,flow){
      */
     self.summarize=function(){
         var currentStep=null;
-        var hideSkipped = self.flow.hideSkippedSteps && self.flow.hideSkippedSteps();
+        // Only hide skipped steps for completed successful executions
+        // NOT_STARTED could mean "never reached due to failure" not just "conditional skipped"
+        var hideSkipped = self.flow.hideSkippedSteps && self.flow.hideSkippedSteps()
+                          && self.flow.completed() && self.flow.executionState() === 'SUCCEEDED';
 
         //step summary info
         var summarydata = {
@@ -942,8 +945,9 @@ function RDNode(name, steps,flow){
         self._originalSummaryState = state;
         self._originalSummary = originalSummary;
 
-        // Check if hideSkippedSteps is enabled and adjust state accordingly
-        var hideSkipped = self.flow.hideSkippedSteps && self.flow.hideSkippedSteps();
+        // Only hide skipped steps for completed successful executions
+        var hideSkipped = self.flow.hideSkippedSteps && self.flow.hideSkippedSteps()
+                          && self.flow.completed() && self.flow.executionState() === 'SUCCEEDED';
 
         if (hideSkipped && (state === 'PARTIAL_NOT_STARTED' || state === 'NOT_STARTED')) {
             // When hiding skipped steps, show these as succeeded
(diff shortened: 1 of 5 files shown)


Jesus-Osuna-M:
## Review
Thanks for the `SUCCEEDED` + `completed` gate — that addresses Copilot’s original “in-progress/failed job shows All Steps OK” issue. A few things still need to land before merge.
### Blocker: collapsed nodes still remap `PARTIAL_NOT_STARTED` → `SUCCEEDED`
Most nodes never load steps. `resummarizeAllNodes` / `updateSummary` treat server `PARTIAL_NOT_STARTED` or `NOT_STARTED` as “All Steps OK” whenever the overall execution succeeded.
That is not equivalent to “only conditionals were skipped”. `StateMapping.summarizeForNode` (same order as `summarize()`) sets `PARTIAL_NOT_STARTED` as soon as **any** step is `NOT_STARTED`, even if another step is `ABORTED` / `other`. `FAILED` is excluded (checked first); abort is not.
So a `keepgoing` job that ends `SUCCEEDED` with abort + a skipped conditional on a collapsed node is displayed as success and moves into `completedNodes()`, which inflates Complete % and drops Incomplete.
The expanded path is stricter (`ranSteps == summarydata.SUCCEEDED`). Please apply the same bar to the collapsed path: only rewrite the summary when every non-`NOT_STARTED` step is `SUCCEEDED` (or keep a distinct skip marker from the backend). Do not infer skip-only from `PARTIAL_NOT_STARTED` alone.
### Blocker: no tests
`executionStateKO.test.js` already uses TestHarness. Please add coverage for:
- all steps `NOT_STARTED` on a `SUCCEEDED` execution → treated as complete
- mix of `SUCCEEDED` + `NOT_STARTED` → complete
- `NOT_STARTED` + `ABORTED`/`other` → **not** rewritten to `SUCCEEDED`
- failed / running execution → toggle is a no-op
- toggle off restores the original summary (collapsed and expanded)
### Blocker: label vs behavior
The control says “Hide skipped steps” / “Hide steps that were skipped due to conditional evaluation”, but skipped steps still appear when a node is expanded. This only remaps the node summary and the Complete/Incomplete counts.
Either change the copy to match (e.g. “Treat skipped conditionals as complete”) or actually filter `NOT_STARTED` steps in the tree.
### Nits
- The checkbox is still clickable on failed/running jobs and does nothing. Disable or hide it unless `completed && executionState === 'SUCCEEDED'`.
- New i18n keys are only in `messages.properties`.
- `updateSummary` early-return compares the **displayed** `summaryState` to the server. After remap, display is `SUCCEEDED` while the server may still be `PARTIAL_NOT_STARTED`; a same-second `lastUpdated` can skip refreshing `_original*`. Compare against `_originalSummaryState`.
- Copilot’s invalid HTML comment looks stale — the checkbox is no longer nested in the output `<span>`.
**Verdict: request changes.**

PR #10507, opened 2026-08-27
[RUN-4812] Fix rerun failed tests selecting no tests in functional jobs

Jesus-Osuna-M:
## Summary

Follow-up to #10482, which enabled CircleCI **Rerun failed tests** on the functional and Selenium jobs. The button appeared and CircleCI selected the right tests, but the rerun ran **nothing** and reported the job as green.

`circleci tests glob` yields test source paths, which is what `explicitTestIncluder` was written for. When rerunning only the failed tests, `circleci tests run` yields **fully qualified class names** instead, because Gradle's JUnit XML carries a `classname` attribute and no `file` one. `explicitTestIncluder` passed those through untouched, so the include matched no compiled class, the task was skipped as `NO-SOURCE`, and a failing job silently turned green.

This maps both forms to the same compiled class path. `pro-functional-test` in rundeckpro delegates to this same function, so it is covered too.

## Evidence

Branch history had a deliberate failure in `BadVersionSpec` (since reverted) to exercise the button on `T Functional API`, which globs 63 spec files across 3 containers.

Before the fix, job 303954:

```
1 test(s) failed out of 322 total tests. Rerunning tests from 1 classname
Received: org.rundeck.tests.functional.api.basic.BadVersionSpec

Compiled files to test: [org.rundeck.tests.functional.api.basic.BadVersionSpec]
> Task :functional-test:apiTest NO-SOURCE
BUILD SUCCESSFUL in 27s
```

Job status `success`, and the CircleCI test API reports `items: []`. A false green.

After the fix, job 304345 reports exactly one test and fails as it should:

```
failure | org.rundeck.tests.functional.api.basic.BadVersionSpec | test-project-bad-version
```

Two of the three containers halted with nothing assigned, and the job took 5m instead of the full suite.

## Test plan

- [x] `T Functional API` rerun executes only the previously failed spec
- [x] The rerun reports the job as failed when the spec still fails
- [x] Normal runs still receive source paths from `circleci tests glob` and are unaffected
- [x] Mapping verified for source paths, class names, `.java` files, the submodule paths used by rundeckpro, and mixed lists

### DIFF
--- a/functional-test/build.gradle
+++ b/functional-test/build.gradle
@@ -79,19 +79,30 @@ def checkWarFile(destination, fileName) {
 
 /** 
  * Returns a function that includes only the test files specified in the `testFiles` parameter.
- * The `testFiles` parameter is a string of paths to the source code of the tests to be run.
+ * The `testFiles` parameter is a string of test source paths, or of fully qualified
+ * class names, one entry per line.
  */
 def explicitTestIncluder(String testFiles) {
     // Convert test file source code paths to compiled class paths.
     // Example: `functional-test/src/test/groovy/org/rundeck/tests/functional/api/basic/BasicSpec.groovy`
     // needs to become `org/rundeck/tests/functional/api/basic/BasicSpec.class`.
+    //
+    // `circleci tests glob` yields those source paths, but when rerunning only the
+    // failed tests `circleci tests run` yields fully qualified class names instead,
+    // because Gradle's JUnit XML carries a `classname` attribute and no `file` one.
+    // Both forms must map to the same compiled class path, otherwise the include
+    // matches nothing and the task is skipped as NO-SOURCE, turning a failing job
+    // green without running a single test.
     def compiledFilesToTest = testFiles
         .split(System.lineSeparator())
-        .collect { testFilePath -> 
-            testFilePath
-            .trim()
-            .replaceFirst(".*\\/src\\/test\\/(groovy\\/|java\\/)", "")
-            .replaceFirst("(.groovy|.java)\$", ".class")
+        .collect { it.trim() }
+        .findAll { !it.isEmpty() }
+        .collect { testFileOrClass ->
+            testFileOrClass.contains("/")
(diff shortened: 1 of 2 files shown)


Jesus-Osuna-M:
### Result of the first rerun: false green

Clicking **Rerun failed tests** on `T Functional API` (job 303954) turned the job green **without running a single test**. The CircleCI test API reports `items: []` for that job.

CircleCI itself did its part correctly:

```
1 test(s) failed out of 322 total tests. Rerunning tests from 1 classname
DEBU  if all tests are being run instead of only failed tests, ensure your JUnit XML has a file or classname attribute.
DEBU  Received: org.rundeck.tests.functional.api.basic.BadVersionSpec
```

Two of the three containers halted with nothing assigned; the third got the failed spec. The problem is the handoff to Gradle. On a rerun `circleci tests run` emits a **fully qualified class name**, not a source path, because our JUnit XML has `classname` but no `file`. So the container ran:

```
./gradlew :functional-test:apiTest -PtestFiles=org.rundeck.tests.functional.api.basic.BadVersionSpec
```

`explicitTestIncluder` only knew how to map source paths, so the value passed through untouched and never matched a compiled class path:

```
Compiled files to test: [org.rundeck.tests.functional.api.basic.BadVersionSpec]
> Task :functional-test:apiTest NO-SOURCE
Skipping task ':functional-test:apiTest' as it has no source files and no previous output files.
BUILD SUCCESSFUL in 27s
```

That is worse than not having the feature at all: a rerun silently converts a red job into a green one.

`f346f34a71` makes `explicitTestIncluder` accept both source paths and class names. Since `pro-functional-test` delegates to this same function, the fix covers rundeckpro too.

Next: let this pipeline go red on `BadVersionSpec` again, hit **Rerun failed tests**, and confirm the job now **fails** while running only that spec. Then the fix moves to its own PR against `main` and this branch gets deleted.

ltamaster (approved):

PR #10528, opened 2026-08-31
[RUN-4853] Update Node.js to 24.20.0

luismalamoc:
## Tell us about your PR

Maintenance: bumps Node.js `24.19.0` → `24.20.0`. Supersedes Renovate PR #10527, which only updated `.nvmrc` and missed `test/.nvmrc`.

- `.nvmrc`: `24.19.0` → `24.20.0`
- `test/.nvmrc`: `v24.19.0` → `v24.20.0`
- `CLAUDE.md`: document that agent-created PRs must use the repo PR template

npm engine pins stay at `11.19.0` (unchanged). A companion rundeckpro PR aligns its `.nvmrc` in lockstep.

Merging Renovate PR #10527 as-is: rejected because it leaves `test/.nvmrc` at `v24.19.0`.

An earlier revision of this PR also disabled Renovate's node/npm engine group in `renovate.json`; that was reverted per review — renovate.json changes will be handled separately once the team agrees on the process.

**Additional context**

Jira: [RUN-4853](https://pagerduty.atlassian.net/browse/RUN-4853)

## Release Notes

Updated bundled Node.js toolchain pin to 24.20.0.

[RUN-4853]: https://pagerduty.atlassian.net/browse/RUN-4853?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


renovate.json
```diff
@@ -121,10 +121,10 @@
       "enabled": true
     },
     {
-      "description": "Group Node.js version (.nvmrc) with the npm engine pin so they update together",
+      "description": "Node/npm engine bumps are handled manually — .nvmrc, test/.nvmrc and npm engines must move in lockstep with rundeckpro, so a partial automated bump is worse than none",
       "matchManagers": ["nvm", "npm"],
       "matchDepNames": ["node", "npm"],
-      "groupName": "node and npm engine versions"
+      "enabled": false
     },
```
smartinellibenedetti, line 128: question: why this is being disabled?

luismalamoc: Renovate's node PRs are incomplete here: they only bump .nvmrc and miss test/.nvmrc, and the pin must also move in lockstep with the rundeckpro repo, so each bump needs a ticket anyway.
Disabling the group avoids half-applied automated bumps; node/npm updates are now handled manually per ticket (like this one).

elquimeras (approved):

renovate.json
smartinellibenedetti: I don't understand this reasoning, as there's nothing preventing us from updating the renovate's PR with the lockfile. The whole point of renovate was to originally help us to not miss the fact that there are dependencies that need an update, disabling the grouping without a process in place will go against that. Plus this PR has this change, but the equivalent change hasn't been done in rundeckpro, so things will get even more out of sync.

Could you please undo this change for now? I think that changes on renovate.json should be handled separately (and I have a proposal to discuss today that also help to simplify this process, so I'd suggest we revisit it once we decide how we want to move forward).

CLAUDE.md
```diff
@@ -51,6 +51,9 @@ Only make changes that are directly requested or clearly necessary. Don't add fe
 ### PR Title
 Format: `[RUN-XXXX] Description` when linked to an issue (recommended), or a clear description.
 
+### PR Body
+Always use this repo's PR template (`.github/pull_request_template.md`) as the PR body — read it and fill in every section (change type, solution, alternatives, context, release notes). Tools like `create_pull_request` do NOT auto-populate it.
```
smartinellibenedetti, line 55: > Tools like `create_pull_request` do NOT auto-populate it.

I do suspect this will confuse the agent in the scenarios where people were using skills to open the PRs, but let's see how it goes.

smartinellibenedetti (changes requested):

luismalamoc:
Good points — reverted the `renovate.json` change in 97385daa06; this PR now only bumps the node pins (`.nvmrc` + `test/.nvmrc`). Happy to revisit the Renovate grouping separately with your proposal.

CLAUDE.md
luismalamoc: I've noticed that the agents used by team members aren't using the template properly, or they're completely ignoring it.

renovate.json
luismalamoc: Removed that change!

smartinellibenedetti (approved):

Issue #7627, opened 2022-04-04
Syncing LDAP User Information Is Not Working (back on 4.0.0)


allenjeong:
_Note: Bug Reports not following this template may be closed without feedback._

**Getting Help**
Please use this issue template for reporting identified or suspected **bugs** *only*. For support on configuration or troubleshooting please see [Getting Help](https://docs.rundeck.com/docs/manual/02-getting-help.html#enterprise-support) for the appropriate channels.

**Describe the bug**
Enabling the LDAP sync feature (Active Directory) for a docker container via  these environment variables.
RUNDECK_JAAS_LDAP_USERFIRSTNAMEATTRIBUTE: givenName
RUNDECK_JAAS_LDAP_USERIDATTRIBUTE: sAMAccountName
RUNDECK_JAAS_LDAP_USERLASTNAMEATTRIBUTE: sn

**My Rundeck detail**
* Rundeck version: 4.0.0 (d344f689482d)
* install type: docker
* OS Name/version: Ubuntu 20.04
* DB Type/version: postgresql

**To Reproduce**

1. Perform an LDAP search and confirm the attributes are correct.
2. Set the appropriate environment variables.
3. Verify settings appear in files generated from templates.
4. Log in to Rundeck.
5. View user profile information and observe they are blank.

**Expected behavior**
Expect to see email address
**Screenshots**

**Desktop (please complete the following information):**
 - OS: [windows]
 - Browser [MS edge]
 - Version [96.0.1054.34]

**Additional context**
This is the same issue with [RUN-315: Syncing LDAP User Information Is Not Working ](https://github.com/rundeck/rundeck/issues/7175)
which was solved on 3.4.10 and back on 4.0.0

alexander-variacode:
I tried this issue and It happens randomly. However, I found this warning at Rundeck logs: 

`WARN  services.UserService - registerLogin: for john, caught OptimisticLockingFailureException: org.springframework.orm.hibernate5.HibernateOptimisticLockingFailureException: Object of class [rundeck.User] with identifier [1]: optimistic locking failed; nested exception is org.hibernate.StaleObjectStateException: Row was updated or deleted by another transaction (or unsaved-value mapping was incorrect) : [rundeck.User#1]`

The walkaround I found was do logout and login again and the user data appears on Profile section:
<img width="1413" alt="image" src="https://user-images.githubusercontent.com/80333592/163451393-377fdb02-aad9-476f-8cfe-c14ce60caea6.png">

[...]

Issue #7175, opened 2021-07-21
RUN-315: Syncing LDAP User Information Is Not Working


paulcalabro:
_Note: Bug Reports not following this template may be closed without feedback._

**Getting Help**

**Describe the bug**

I am enabling the LDAP sync feature for a Docker container via these environment variables:

- RUNDECK_JAAS_LDAP_USERFIRSTNAMEATTRIBUTE: givenName
- RUNDECK_JAAS_LDAP_USERIDATTRIBUTE: sAMAccountName
- RUNDECK_JAAS_LDAP_USERLASTNAMEATTRIBUTE: sn

These appear to be inserted into the correct configuration files as expected, however, the syncing of LDAP information is not occurring:

```
sh-4.4$ grep sync /home/rundeck/server/config/rundeck-config.properties
...
# user sync
rundeck.security.syncLdapUser=true
...
```

```
sh-4.4$ grep -E Attribute  /home/rundeck/server/config/jaas-loginmodule.conf
...
        userLastNameAttribute="sn"
        userFirstNameAttribute="givenName"
        userEmailAttribute="mail"
...
```

**My Rundeck detail**

* Rundeck version: 3.4.1-20210715 
* install type: docker
* OS Name/version: Ubuntu 18.04.5 LTS
* DB Type/version: mariadb

**To Reproduce**
1. Perform an LDAP search and confirm the attributes are correct.
2. Set the appropriate environment variables.
3. Verify settings appear in files generated from templates.
4. Log in to Rundeck.
5. View user profile information and observe they are blank.

**Expected behavior**
I expected to see my first name, last name, and company email address.

**Screenshots**
![Screen Shot 2021-07-20 at 9 13 08 PM](https://user-images.githubusercontent.com/585059/126429771-c83fa1e9-e2cd-4dff-80ce-ec7d96d19a96.png)

**Desktop (please complete the following information):**
 - OS: macOS 11.4
 - Browser Firefox 89

**Additional context**
N/A

mlachal:
Hi, I just observed the same issue on our 3.4.1 RPM platform. The syncLdapUser functionnality works fine on 3.3.11 and it looks like the issue is there : https://github.com/rundeck/rundeck/commit/c228eb199c92dd178c8b2908d30f299e956f7404#diff-f2e0d41b63fe1c660b65a51f9d507689ac2c1465ae0431ea1ce64b9ff29e5c00R41

paulcalabro:
[...]

PR #10532, opened 2026-08-31
[RUN-4860] Explicitly request demographic/password attributes in LDAP findUser()

### LINKED ISSUE #10469: LDAP profile sync receives null attributes when using JettyCachingLdapLoginModule
**Describe the bug**
LDAP authentication and role lookup work correctly with
`JettyCachingLdapLoginModule`, but Rundeck does not synchronize first name,
last name, or email into the user profile.

The LDAP attributes are present and readable with `ldapsearch`. However, the
same search through Java JNDI returns `null` for every profile attribute when
no explicit returning-attribute list is set.

As a result, Rundeck logs that it is updating the profile, but all supplied
values are `null`.

**My Rundeck detail**
- Rundeck version: `6.1.0`
- Install type: RPM
- Database: MariaDB
- LDAP: Active Directory over LDAPS
- Authentication module:
  `com.dtolabs.rundeck.jetty.jaas.JettyCachingLdapLoginModule`

**Relevant configuration**
java
forceBindingLogin="true"
forceBindingLoginUseRootContextForRoles="true"

userBaseDn="OU=Users,DC=example,DC=com"
userRdnAttribute="sAMAccountName"
userIdAttribute="sAMAccountName"
userPasswordAttribute="unicodePwd"
userObjectClass="user"

userFirstNameAttribute="givenName"
userLastNameAttribute="sn"
userEmailAttribute="mail"

roleBaseDn="OU=Groups,DC=example,DC=com"
roleNameAttribute="cn"
roleMemberAttribute="member"
roleObjectClass="group"

cacheDurationMillis="300000"

+ properties:
rundeck.security.syncLdapUser=true

**Command line**
-Drundeck.jaaslogin=true
-Djava.security.auth.login.config=/etc/rundeck/jaas-ldap.conf
-Dloginmodule.name=activedirectory

**Expected behavior**
After LDAP login, Rundeck should populate the user profile with:

first name from givenName
last name from sn
email from mail

**Actual behavior**
The profile remains empty. With DEBUG logging enabled for
org.rundeck.security.RundeckJaasAuthenticationSuccessEventListener, Rundeck
logs:
DEBUG ... Updating user profile from ldap attributes for user: testuser
DEBUG ... First Name: null
DEBUG ... Last Name: null
DEBUG ... Email: null
No database-update error is logged; the LDAP principals already contain
null.

**LDAP verification**
The LDAP user is found by sAMAccountName; its DN is different from the login
name, as is normal for Active Directory.

ldapsearch -LLL \
  -H ldaps://ldap.example.com:636 \
  -D 'ldap-bind@example.com' -W \
  -b 'OU=Users,DC=example,DC=com' \
  '(&(objectClass=user)(sAMAccountName=testuser))' \
  givenName sn mail

Result:
dn: CN=Example User,OU=Users,DC=example,DC=com
sn: User
mail: example.user@example.com
sAMAccountName: testuser
givenName: Example

Authentication and group lookup work with the same bind account and base DN.

**Java JNDI reproduction**
The following standalone Java program uses the same JNDI settings and search
controls as JettyCachingLdapLoginModule.findUser():

```
import javax.naming.*;
import javax.naming.directory.*;
import java.util.*;
import java.io.Console;

public class LdapAttrs {
  public static void main(String[] args) throws Exception {
    Console console = System.console();
    char[] password = console.readPassword("LDAP password: ");

    Hashtable<String,Object> env = new Hashtable<>();
    env.put(Context.INITIAL_CONTEXT_FACTORY,
            "com.sun.jndi.ldap.LdapCtxFactory");
    env.put(Context.PROVIDER_URL, "ldaps://ldap.example.com:636");
    env.put(Context.SECURITY_AUTHENTICATION, "simple");
    env.put(Context.SECURITY_PRINCIPAL, "ldap-bind@example.com");
    env.put(Context.SECURITY_CREDENTIALS, new String(password));

    DirContext ctx = new InitialDirContext(env);
    SearchControls c = new SearchControls();
    c.setCountLimit(1);
    c.setDerefLinkFlag(true);
    c.setSearchScope(SearchControls.SUBTREE_SCOPE);

    NamingEnumeration<SearchResult> r = ctx.search(
      "OU=Users,DC=example,DC=com",
      "(&(objectClass={0})({1}={2}))",
      new Object[]{"user", "sAMAccountName", "testuser"}, c);

    Attributes a = r.next().getAttributes();
    System.out.println("givenName = " + a.get("givenName"));
    System.out.println("sn        = " + a.get("sn"));
    System.out.println("mail      = " + a.get("mail"));
    ctx.close();
  }
}
```

Output:
givenName = null
sn        = null
mail      = null

**Investigation**
The Rundeck class in the installed WAR was inspected. It correctly calls
setDemographicAttributes(searchResult.getAttributes()) and adds the
corresponding LDAP principals in commit().

However, findUser() creates SearchControls without calling
setReturningAttributes(...).

It appears that, with this Active Directory/Java JNDI combination, the
requested attributes must be explicit. Could findUser() request the
configured user attributes, for example:

ctls.setReturningAttributes(new String[]{
    _userIdAttribute,
    _userPasswordAttribute,
    _userFirstNameAttribute,
    _userLastNameAttribute,
    _userEmailAttribute
});

Related historical issues:
#7175
#7627

fdevans:
## Tell us about your PR

Bugfix. Fixes #10469. `JettyCachingLdapLoginModule.findUser()` searches for the user entry without calling `SearchControls.setReturningAttributes(...)`, relying on the JNDI/directory default of "return all attributes."

Against Active Directory over JNDI, this can return `null` for every attribute on the search result — even though the same attributes are readable via `ldapsearch` using the exact same bind account and base DN (the reporter included a standalone JNDI repro program demonstrating this independent of Rundeck). No error is logged; the profile-sync code just silently receives `null` for every field.

This means `setDemographicAttributes()` (first/last name, email) and the `userPassword` lookup in `getUserCredentials()` — both of which read attributes off of `findUser()`'s result — never actually get populated for affected AD deployments.

Added an explicit returning-attributes list to `findUser()`'s `SearchControls`, covering every attribute this class reads from that particular search result: `userIdAttribute`, `userPasswordAttribute`, `userFirstNameAttribute`, `userLastNameAttribute`, `userEmailAttribute`. This matches the existing pattern already used for role searches elsewhere in the same class (see `getNonPaginatedRoles()`, which already calls `setReturningAttributes`), so it's consistent with the rest of the file rather than a new pattern.

- Leaving the returning-attributes list unset and instead adding a config flag to opt into requesting attributes explicitly — rejected as unnecessary complexity; explicitly requesting only the attributes this code actually reads is strictly more correct with no behavior change for directories that already worked (they'll return the same attributes, just now explicitly asked for), so no flag is needed.

**Additional context**

- Fixes #10469
- Related historical issues referenced by the reporter: #7175, #7627

## Release Notes

Fixed LDAP authentication (`JettyCachingLdapLoginModule`) so that first name, last name, and email are correctly synced from directories (e.g. Active Directory) that require attributes to be explicitly requested in the search, rather than silently syncing `null` values.

### DIFF
--- a/rundeckapp/src/main/groovy/com/dtolabs/rundeck/jetty/jaas/JettyCachingLdapLoginModule.java
+++ b/rundeckapp/src/main/groovy/com/dtolabs/rundeck/jetty/jaas/JettyCachingLdapLoginModule.java
@@ -1018,6 +1018,18 @@ public class JettyCachingLdapLoginModule extends AbstractLoginModule {
         ctls.setCountLimit(1);
         ctls.setDerefLinkFlag(true);
         ctls.setSearchScope(SearchControls.SUBTREE_SCOPE);
+        // Explicitly request the attributes this class reads off of the search result
+        // (setDemographicAttributes() and the userPassword lookup in getUserCredentials()).
+        // Some directories (e.g. Active Directory over JNDI) return null for every attribute
+        // when no returning-attribute list is set, even though the same attributes are readable
+        // via ldapsearch, so they must be requested by name.
+        ctls.setReturningAttributes(new String[]{
+            _userIdAttribute,
+            _userPasswordAttribute,
+            _userFirstNameAttribute,
+            _userLastNameAttribute,
+            _userEmailAttribute
+        });
 
         String filter = OBJECT_CLASS_FILTER;
 
--- a/rundeckapp/src/test/groovy/com/dtolabs/rundeck/jetty/jaas/JettyCachingLdapLoginModuleTest.groovy
+++ b/rundeckapp/src/test/groovy/com/dtolabs/rundeck/jetty/jaas/JettyCachingLdapLoginModuleTest.groovy
@@ -218,6 +218,69 @@ class JettyCachingLdapLoginModuleTest extends Specification {
         username | _
         'auser'  | _
     }
+
+    def "findUser requests demographic and password attributes explicitly"() {
+        // Some directories (e.g. Active Directory over JNDI) return null for every attribute
+        // when no returning-attribute list is set on the search, even though the same
+        // attributes are readable via ldapsearch. findUser() must request them by name.
+        JettyCachingLdapLoginModule module = new JettyCachingLdapLoginModule()
+        module._debug = true
+        module._forceBindingLogin = true
+        module._contextFactory = "notnull"
+        module._providerUrl = "notnull"
+        module._forceBindingLoginUseRootContextForRoles = false
+        module._roleBaseDn = 'roleBaseDn'
+        module.rolePagination = false
+        module._roleUsernameMemberAttribute = 'roleUsernameMemberAttribute'
+        module.setCallbackHandler(Mock(CallbackHandler) {
+            1 * handle(_) >> { it[0][0].name = username; it[0][1].object = 'apassword' }
+        })  // Use setter instead of @field access (Groovy 4)
+        def found = [Mock(SearchResult) {
+            getNameInNamespace() >> "cn=$username,dc=test,dc=com"
+            getAttributes() >> Mock(Attributes) {
+                get(module._userFirstNameAttribute) >> new BasicAttribute(module._userFirstNameAttribute, "First")
+                get(module._userLastNameAttribute) >> new BasicAttribute(module._userLastNameAttribute, "Last")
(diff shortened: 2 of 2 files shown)


fdevans:
@copilot resolve the merge conflicts in this pull request

smartinellibenedetti (approved):

PR #10535, opened 2026-08-31
[RUN-4861] Support bare JSON array bodies in JsonUtil for MenuController#listExport

### LINKED ISSUE #10468: JSON errors when browsing projects
**Describe the bug**

When NextUI is disabled (legacy UI), navigating to a project's job list page triggers a POST /menu/listExport request that sends a JSON array as the request body. The Rundeck backend expects a JSON object, causing a deserialization error.

This does not occur when NextUI is enabled.

The legacy UI sends a bare JSON array as the POST body : 

`["7add463a-93c1-4936-9ac0-01a25321554f","3b66b5b7-c066-46f9-9c4d-b448cb7cb990","f5df62e9-7355-4ff2-b5b1-023734374f07"]`

The server logs : 

```
ERROR com.dtolabs.rundeck.util.JsonUtil - parseRequestBody: JSON parsing failed
[MismatchedInputException: Cannot deserialize value of type
`java.util.LinkedHashMap<java.lang.Object,java.lang.Object>` from Array value
(token `JsonToken.START_ARRAY`)] — body was readable but not a JSON object
```

**My Rundeck detail**

* Rundeck version: 6.1.0
* install type: war
* OS Name/version: RHEL 8.10
* DB Type/version: Postgres

**To Reproduce**
1. Disable NextUI for the user
2. Open any browser (tested on Firefox/Chrome)
3. Navigate to any project's job list
4. Observe the error in rundeck.log

**Expected behavior**

The request body should be a JSON object or the backend should accept an array for this endpoint.

 - OS: Windows 11
 - Browser : Chrome/Firefox
 - Version : 150

fdevans:
## Tell us about your PR

Bugfix. Fixes #10468. When NextUI is disabled (legacy UI), the job list page POSTs a bare JSON array of visible job ids as the request body to `POST /menu/listExport`, e.g. `["7add463a-...","3b66b5b7-..."]`, rather than a JSON object.

`JsonUtil.parseRequestBody` (added as part of the Grails 7 migration, replacing the old `request.JSON` binding) always deserializes into `Map.class`, so this array body fails to parse:

```
[MismatchedInputException: Cannot deserialize value of type
`java.util.LinkedHashMap<java.lang.Object,java.lang.Object>` from Array value
(token `JsonToken.START_ARRAY`)]
```

This gets logged as an `ERROR` on every legacy job list page view, and the parse failure is swallowed — `MenuController#listExport` then builds its `ScheduledExecutionQuery` with `idlist=null` instead of the intended comma-joined id list, because the failure is masked by safe navigation (`data?.join(...)`).

Added `JsonUtil.parseRequestBodyAsList(request)`, sharing the existing reader/fallback logic (extracted into a private `readRequestBodyContent` helper — this logic has its own bug-fix history, RUN-4521, around `getReader()`/`getInputStream()` ordering, so it's reused rather than duplicated) but deserializing into `List.class` instead of `Map.class`. Switched `MenuController#listExport` to call the new method instead of `parseRequestBody`.

The other 9 call sites of `JsonUtil.parseRequestBody` (in `FrameworkController`, `WorkflowController`, other `MenuController` actions, `PluginController`, `ScheduledExecutionController`, `WebUtil`) are untouched — they all genuinely expect JSON objects, so `listExport` was the only actual mismatch.

- Changing `parseRequestBody` itself to parse into `Object.class` and letting every caller handle both `Map` and `List` shapes — rejected as unnecessarily broad; only one of the ten call sites (`listExport`) ever receives an array body, so a dedicated method is more targeted and doesn't risk changing behavior for the other nine.

**Additional context**

- Fixes #10468
- Jira: [RUN-4861](https://pagerduty.atlassian.net/browse/RUN-4861)

## Release Notes

Fixed the legacy (non-NextUI) job list page so that job ids are correctly parsed from the `POST /menu/listExport` request body, instead of silently failing to parse and logging an error on every page view.

[RUN-4861]: https://pagerduty.atlassian.net/browse/RUN-4861?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
@@ -3734,7 +3734,9 @@ if executed in cluster mode.
         def results=[:]
         if(request.format=='json' ) {
             // Grails 7: Parse body using Jackson instead of request.JSON
-            def data = com.dtolabs.rundeck.util.JsonUtil.parseRequestBody(request)
+            // The request body here is a bare JSON array of job ids (not an object), sent by the
+            // legacy (non-NextUI) job list page — see RUN-10468.
+            def data = com.dtolabs.rundeck.util.JsonUtil.parseRequestBodyAsList(request)
             def nextScheduled = data?.join(",")?.replaceAll(/"/, '')
             def query = new ScheduledExecutionQuery()
             query.idlist = nextScheduled
--- a/rundeckapp/src/main/groovy/com/dtolabs/rundeck/jetty/jaas/JettyCachingLdapLoginModule.java
+++ b/rundeckapp/src/main/groovy/com/dtolabs/rundeck/jetty/jaas/JettyCachingLdapLoginModule.java
@@ -1018,18 +1018,6 @@ public class JettyCachingLdapLoginModule extends AbstractLoginModule {
         ctls.setCountLimit(1);
         ctls.setDerefLinkFlag(true);
         ctls.setSearchScope(SearchControls.SUBTREE_SCOPE);
-        // Explicitly request the attributes this class reads off of the search result
-        // (setDemographicAttributes() and the userPassword lookup in getUserCredentials()).
-        // Some directories (e.g. Active Directory over JNDI) return null for every attribute
-        // when no returning-attribute list is set, even though the same attributes are readable
-        // via ldapsearch, so they must be requested by name.
-        ctls.setReturningAttributes(new String[]{
-            _userIdAttribute,
-            _userPasswordAttribute,
-            _userFirstNameAttribute,
-            _userLastNameAttribute,
-            _userEmailAttribute
-        });
 
         String filter = OBJECT_CLASS_FILTER;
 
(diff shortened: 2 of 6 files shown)


ronaveva (changes requested):
It looks ok, but it has changes that are not related to the ticket. Please remove those and i will apporve the PR

fdevans:
@ronaveva Good catch — this branch was accidentally built on top of an unrelated LDAP commit (from PR #10532/RUN-4860) instead of cleanly off `main`. Reverted that commit; the PR now contains only the RUN-4861 `JsonUtil`/`MenuController` changes (verified: `JsonUtilSpec` 8/8 and `MenuControllerSpec` 157/158, both passing, no failures). The LDAP fix itself is untouched and still present on its own PR (#10532).

fdevans:
Closing and restarting clean per review feedback — the revert commit approach left messy history that would be a pain at merge time. Recreated as #10554, a single clean commit cut directly from `main` with no LDAP history at all.

PR #10537, opened 2026-09-01
[RUN-4790] Fix rd system acls create JsonParseException regression

carlosrfranco:
## Release Notes

Fixed a regression where creating system ACL policies via the API or `rd` CLI with a YAML request body and a JSON Accept header could fail with a parse error even though the ACL was created successfully. The server now returns JSON when requested via the Accept header instead of defaulting to the request body's content type.

## Jira Ticket
[RUN-4790](https://pagerduty.atlassian.net/browse/RUN-4790)

## Summary
`rd system acls create` uploads a YAML `aclpolicy` body and requests a JSON response
(`Accept: application/json`). `ApiService.extractResponseFormat()` checked the
request's own `Content-Type`-derived format (`request.format`) *before* checking the
`Accept` header, so the server answered with raw YAML instead of the expected
`{"contents": "..."}` JSON body. The CLI then threw
`JsonParseException: Unrecognized token 'by'` trying to parse that YAML as JSON —
even though the ACL was created successfully server-side.

- Reordered `extractResponseFormat` so the `Accept` header is resolved first, falling
  back to `request.format`/`defformat` only when `Accept` doesn't map to an allowed
  format.
- Added unit test coverage in `ApiServiceSpec.groovy` for the conflicting-format case
  (YAML body + JSON Accept) plus non-regression rows confirming the fallback still
  works when `Accept` is absent or unrecognized.
- Added a functional test (`SystemAclCreateSpec`) exercising the system ACL create
  endpoint with a YAML body and `Accept: application/json`, asserting a JSON response.

This is a server-only fix; no `rd-cli-tool`/`rd-api-client` changes are needed since the
client's existing JSON parsing is correct once the server response is fixed.

## Testing
- `./gradlew test --tests "rundeck.services.ApiServiceSpec"` — pass
- `./gradlew test --tests "rundeck.controllers.FrameworkControllerSpec" --tests "rundeck.controllers.ProjectControllerSpec" --tests "rundeck.controllers.ProjectController2Spec"` — pass (these mock `extractResponseFormat`, confirming no regression)
- New functional spec compiles; requires a running rundeck instance to execute end-to-end

[RUN-4790]: https://pagerduty.atlassian.net/browse/RUN-4790?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/functional-test/src/test/groovy/org/rundeck/tests/functional/api/acl/SystemAclCreateSpec.groovy
+++ b/functional-test/src/test/groovy/org/rundeck/tests/functional/api/acl/SystemAclCreateSpec.groovy
@@ -1,0 +1,53 @@
+package org.rundeck.tests.functional.api.acl
+
+import okhttp3.Headers
+import org.rundeck.util.annotations.APITest
+import org.rundeck.util.container.BaseContainer
+
+/**
+ * Regression coverage for RUN-4790: creating a system ACL with a YAML request body while
+ * asking for a JSON response (as the CLI's API client does) must return a JSON-wrapped body,
+ * not raw YAML.
+ */
+@APITest
+class SystemAclCreateSpec extends BaseContainer {
+
+    def "create system acl with yaml body and json accept returns json"() {
+        given:
+        def aclPath = "run4790-${UUID.randomUUID()}.aclpolicy"
+        def yamlBody = '''by:
+  group: Run4790Group
+description: Allow [read] for project_acl
+for:
+  project_acl:
+  - allow:
+    - read
+context:
+  application: rundeck
+'''
+
+        when:
+        int code
+        String contentType
+        Map json
+        try (def resp = client.doPost(
+                "/system/acl/${aclPath}",
+                yamlBody,
+                'application/yaml',
+                Headers.of('Accept', 'application/json')
+        )) {
+            code = resp.code()
+            contentType = resp.header('Content-Type')
+            json = client.jsonValue(resp.body(), Map)
+        }
+
+        then:
+        code == 201
+        contentType != null
+        contentType.contains('application/json')
+        json.contents.contains('Run4790Group')
+
+        cleanup:
+        deleteSystemAcl(aclPath)
+    }
+}
--- a/functional-test/src/test/groovy/org/rundeck/tests/functional/api/acl/SystemAclCreateSpec.groovy
+++ b/functional-test/src/test/groovy/org/rundeck/tests/functional/api/acl/SystemAclCreateSpec.groovy
@@ -27,7 +27,7 @@ context:
 '''
 
         when:
-        int code
+        Integer code
         String contentType
         Map json
         try (def resp = client.doPost(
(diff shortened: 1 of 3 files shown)


ncofreortiz-hub:
## Review notes on the Accept-header precedence fix

Thanks for the fix — the reorder correctly addresses the RUN-4790 regression, and the test coverage for the ACL create path (unit + functional) looks solid. A few things worth confirming before merge:

1. **Blast radius beyond the ACL endpoint.** `extractResponseFormat` is called from 20+ places (`FrameworkController`, `ProjectController`, `ExecutionController`, `UserController`, `ExecutionService`). Since the Accept-header check now runs *before* `request.format`, any route using `.$format` URL-suffix mapping (e.g. `UrlMappings.groovy:46` execution output, `:313` events, `:332` exportWait) could have its explicit `.json`/`.xml` suffix overridden by a broad Accept header (e.g. a browser's default `Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8`). Was this checked for the other call sites, or is the fix intentionally scoped to `ApiService` regardless of blast radius?

2. **Possible dead code.** For the callers that pass `defformat = request.format` (`FrameworkController.groovy:4580,4612,4730,4776,4810`, `ProjectController.groovy:1684,1719,1929,2045`), the relocated `requestFormat`-vs-`allowed` check (`ApiService.groovy` ~line 1010) seems to never change the outcome — the following `if (defformat) return defformat` already returns the same value. Worth double-checking whether that block is still pulling weight, or if it can be dropped to avoid confusing future readers.

3. **Test gap.** The new `extractResponseFormat` where-table rows in `ApiServiceSpec.groovy` all set `defformat == request.format` (mirroring the ACL caller). Could we add a row or two where `defformat` is a fixed literal different from `request.format` (as in `ProjectController`/`ExecutionService`), to cover that interaction explicitly?

None of these block the fix itself — just flagging for visibility given how central this method is.

carlosrfranco:
Thanks for the thorough look, @ncofreortiz-hub — went through all three points against the actual call graph:

1. **Blast radius / URL-suffix override:** I checked all 20 call sites of `extractResponseFormat` against `UrlMappings.groovy` — none of them have a `.$format` URL-suffix mapping; they're all plain `/api/$api_version/...` routes where `request.format` comes only from the request body's Content-Type. The three routes cited (execution output, events, exportWait) don't call `extractResponseFormat` at all — they use Grails' native `withFormat{}` blocks, an unrelated code path. So this specific scenario doesn't occur in this codebase; no change needed here.
2. **Possible dead code:** Confirmed the redundancy for callers that pass `defformat = request.format` — but that's pre-existing on `main` today (the reorder just moved where it sits), and it's *not* dead for callers like `apiProjectConfigGet`/`apiProjectConfigKeyPut` that pass a fixed literal `defformat`. Added a test below that proves it's still load-bearing for that shape, so leaving it as-is.
3. **Test gap:** Good catch — added `ApiServiceSpec` coverage for the fixed-`defformat`-distinct-from-`request.format` interaction (mirrors `apiProjectConfigGet`'s shape). This also empirically confirms point 2's check isn't removable for that caller pattern.

ncofreortiz-hub (approved):
LGTM

PR #10539, opened 2026-09-01
[RUN-4866] Don't persist STATIC_TEXT plugin properties into saved job config

### LINKED ISSUE #10420: WorkflowStrategy shows raw html in frontend and in yaml job description
**Describe the bug**
When creating jobs in rundeck, the WorkflowStrategy displays raw html in the frontend and the downloaded yaml job desciption contains the same html.

e.g.

`- defaultTab: nodes
  description: TEST
  executionEnabled: true
  group: TEST
  id: a8522449-2935-444c-a752-a2dbb7885efc
  loglevel: INFO
  name: TEST
  nodeFilterEditable: false
  plugins:
    ExecutionLifecycle: {}
  scheduleEnabled: true
  sequence:
    commands:
    - description: TEST
      exec: echo "TEST"
    keepgoing: false
    pluginConfig:
      WorkflowStrategy:
        node-first:
          info: |-
            <table>
                <tr><td>1.</td><td class="text-info">NodeA</td> <td>step 1</td></tr>
                <tr><td>2.</td><td class="text-info">"</td> <td>step 2</td></tr>
                <tr><td>3.</td><td class="text-info">"</td> <td>step 3</td></tr>
                <tr><td>4.</td><td class="text-muted">NodeB</td> <td>step 1</td></tr>
                <tr><td>5.</td><td class="text-muted">"</td> <td>step 2</td></tr>
                <tr><td>6.</td><td class="text-muted">"</td> <td>step 3</td></tr>
            </table>
    strategy: node-first
  uuid: a8522449-2935-444c-a752-a2dbb7885efc`

We have a lot of yaml job definitions in our source control that we load into rundeck everytime we do an upgrade.

If the job is not edited in rundeck after being imported, the exported job definition does not contain the WorkflowStrategy information at all.

If the job is edited in rundeck after import, then the exported job defintion does contain the WorkflowStrategy information and it will contain the raw html mentioned above. 

**My Rundeck detail**

* Rundeck version: 6.0.1 (also present in 5.20.1, not present in 4.14.2)
* install type: rpm
* OS Name/version: RHEL8
* DB Type/version: mariadb

**To Reproduce**
Steps to reproduce the behaviour:
1. Create a new job
2. Add a command step
3. Save the job
4. export yaml job description
5. See error

**Expected behavior**
I would expect the html to be displayed properly in the frontend.
I would expect the job definition to be the same on export whether it had been edited inside rundeck or not.  
Not sure if the html should be present in the yaml file, but I don't see any advantage to having it in there.

**Screenshots**
Image

 - OS: Windows 11
 - Browser firefox
 - Version 153.0.1

Also tested in chrome 150.0.7871.187

fdevans:
## Problem

Fixes #10420.

`WorkflowStrategy`'s `node-first` provider has an `info` property that's a `STATIC_TEXT` display-only help table (a hardcoded HTML `<table>`, rendered from `prop.staticTextDefaultValue`, never user-editable — see [`NodeFirstWorkflowStrategy.java`](https://github.com/rundeck/rundeck/blob/main/core/src/main/java/com/dtolabs/rundeck/core/execution/workflow/NodeFirstWorkflowStrategy.java#L43)).

Editing and saving a job through the UI (including creating a new job, or duplicating an existing one) persists that raw HTML into `sequence.pluginConfig.WorkflowStrategy.node-first.info` in the job's config. This:
- Shows up as literal `<table>...` markup in the read-only "show"/Definition view.
- Gets written into exported YAML, causing spurious diffs for anyone using Git SCM export/import — jobs imported from YAML and never edited in the UI don't have it; the moment they're opened and saved in the UI, it appears.

## Root cause

`pluginConfig.vue`'s `prepareInputs()`/`exportInputs()` treat every plugin property the same way: `prepareInputs()` seeds any property's `defaultValue` into the form's `inputValues` in create mode, and `exportInputs()` unconditionally writes every property back into the saved config. Neither has an exemption for `STATIC_TEXT` properties, which are purely descriptive and were never meant to be treated as real config data.

Separately, [`WorkflowStrategy.vue`](https://github.com/rundeck/rundeck/blob/main/rundeckapp/grails-spa/packages/ui-trellis/src/app/components/job/workflow/WorkflowStrategy.vue)'s `model` watcher flips `mode` to `"create"` whenever it sees `newVal.type !== oldVal.type` — but `model` starts as `{}` in `data()` and only gets its real `type` assigned in `mounted()`, which the deep watcher sees as a "type change" from `undefined` on *every single mount*. So opening the job editor at all (not just genuinely creating a new job) transiently exercises the create-mode defaulting path for this sub-form.

## Fix

Rather than touch the `mode`-detection quirk (which affects more than just this one field and carries more risk), this excludes `STATIC_TEXT` properties from both `prepareInputs()`'s default-seeding and `exportInputs()`'s export, via a new `isStaticTextProp()` helper mirroring the existing `isPropHidden()` pattern. This:
- Stops new pollution going forward, regardless of the `mode`-detection quirk.
- **Self-heals** already-affected jobs — saving them again now strips the bogus `info` key out of `config` instead of carrying it forward.
- Makes the raw-HTML display in the "show" view moot too, since there's no persisted value left to render.

## Testing

Added 3 regression tests to `pluginConfig.spec.ts`:
- STATIC_TEXT props are never seeded with a default value in create mode.
- STATIC_TEXT props are never included in the exported config payload.
- An already-persisted STATIC_TEXT value is stripped out of the exported config on the next save (self-heals legacy data).

`npx jest pluginConfig.spec.ts` — 33/33 passing. `npx eslint` on both changed files — 0 errors, no new warnings.

## Release Notes

Fixed the Workflow Strategy configuration so its descriptive help text is no longer saved into the job definition and exported YAML as raw HTML.  This will not fix existing jobs, but opening and resaving the job manually will fix the issue.

### DIFF
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/pluginConfig.vue
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/pluginConfig.vue
@@ -457,7 +457,12 @@ export default defineComponent({
         if (config[prop.name]) {
           this.inputValues[prop.name] = config[prop.name];
         }
-        if (modeCreate && !this.inputValues[prop.name] && prop.defaultValue) {
+        if (
+          modeCreate &&
+          !this.isStaticTextProp(prop) &&
+          !this.inputValues[prop.name] &&
+          prop.defaultValue
+        ) {
           this.inputValues[prop.name] = prop.defaultValue;
         }
         if (
@@ -506,6 +511,12 @@ export default defineComponent({
       const values: { [index: string]: any } = {};
       //convert true boolean to 'true'
       this.props.forEach((prop: any) => {
+        if (this.isStaticTextProp(prop)) {
+          // STATIC_TEXT properties are display-only help text (e.g. workflow strategy
+          // "info" panels) and must never be persisted as real config data.
+          // See https://github.com/rundeck/rundeck/issues/10420
+          return;
+        }
         if (prop.type === "Boolean") {
           if (
             this.inputValues[prop.name] === true ||
@@ -604,6 +615,14 @@ export default defineComponent({
     },
     isPropHidden(testProp: any): boolean {
       return testProp.options && testProp.options["displayType"] === "HIDDEN";
+    },
+    isStaticTextProp(testProp: any): boolean {
+      // STATIC_TEXT properties (e.g. the WorkflowStrategy "info" help table) are
+      // rendered from prop.staticTextDefaultValue and are never user-editable, so
+      // they must not be seeded with a default value or exported as config data.
+      return (
+        testProp.options && testProp.options["displayType"] === "STATIC_TEXT"
+      );
     },
     isPropInScope(testProp: any): boolean {
       // determine if property is visible in scope
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/tests/pluginConfig.spec.ts
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/tests/pluginConfig.spec.ts
@@ -397,6 +397,86 @@ describe("PluginConfig", () => {
     });
   });
 
+  describe("STATIC_TEXT props", () => {
+    // Regression tests for https://github.com/rundeck/rundeck/issues/10420:
+    // WorkflowStrategy's "info" property is a STATIC_TEXT/display-only help table
+    // (rendered from prop.staticTextDefaultValue, never user-editable). It was
+    // being seeded from prop.defaultValue in create mode and then persisted into
+    // the saved job config as literal HTML, polluting exported YAML and causing
+    // spurious SCM diffs.
+    it("does not seed a STATIC_TEXT prop with its default value in create mode", async () => {
+      const wrapper = await createWrapper({
+        props: {
+          mode: "create",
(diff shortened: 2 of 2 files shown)


carlosrfranco (changes requested):
It looks like the fix didn't work. Tested locally using docker image and the raw html is still there:

Image
Image

fdevans:
Thanks for testing this — appreciate the screenshots. I traced the fix end-to-end (Java `@RenderingOption` → `PluginApiService` → `ApiProviderProp` → `pluginConfig.vue`'s `isStaticTextProp()`) and it's structurally correct, matching the existing `isPropHidden()` pattern; `pluginConfig.spec.ts` passes 44/44 including the new regression tests. My leading guess is the Docker image you tested wasn't built fresh off this branch's front-end bundle (ui-trellis has a separate build step). Could you confirm: (1) was the image built directly from this PR's commit, or an existing/cached tag, and (2) in the browser's Network tab, does the `/plugin/detail/WorkflowStrategy/node-first` response show `"options": {"displayType": "STATIC_TEXT", ...}`? That'll tell us in one look whether this is a stale-build issue or a real gap I'm missing.

carlosrfranco (approved):
The fix actually worked but it should resave the job to apply it

PR #10546, opened 2026-09-02
[RUN-4883] Remove identity id generator from User domain to fix login constraint violation

### LINKED ISSUE #10493: Login of a new user raises constraint violation on Rundeck 6.x
**Describe the bug**
We run a Rundeck instance that has seen a lot of upgrades over the years, including changes to the schema of the underlying Postgres.  The last upgrades done were 5.20.1 -> 6.0.1 -> 6.1.0.
After upgrade to Rundeck  6.1.0 the login of a new user raises constraint violation in table `rduser`.

```
[2026-08-25T10:22:44,502] ERROR spi.SqlExceptionHelper [qtp398275400-25602] - ERROR: null value in column "id" of relation "rduser" violates not-null constraint
org.springframework.dao.DataIntegrityViolationException: Hibernate operation: could not execute statement; SQL [n/a]; ERROR: null value in column "id" of relation "rduser" violates not-null constraint
Caused by: org.postgresql.util.PSQLException: ERROR: null value in column "id" of relation "rduser" violates not-null constraint
```

We use LDAP auth by means of `/etc/rundeck/jaas-ldap.conf`. Due to this we skipped version 6.0.0 because of the LDAP auth issue.

Definition of the `rduser` table:
```
rundeck=> \d rduser
                                Table "public.rduser"
        Column         |            Type             | Collation | Nullable | Default 
-----------------------+-----------------------------+-----------+----------+---------
 id                    | bigint                      |           | not null | 
 version               | bigint                      |           | not null | 
 dashboard_pref        | character varying(255)      |           |          | 
 date_created          | timestamp without time zone |           | not null | 
 email                 | character varying(255)      |           |          | 
 filter_pref           | character varying(255)      |           |          | 
 first_name            | character varying(255)      |           |          | 
 last_name             | character varying(255)      |           |          | 
 last_updated          | timestamp without time zone |           | not null | 
 login                 | character varying(255)      |           | not null | 
 password              | character varying(255)      |           |          | 
 last_logged_host_name | character varying(255)      |           |          | 
 last_logout           | timestamp without time zone |           |          | 
 last_session_id       | character varying(255)      |           |          | 
 last_login            | timestamp without time zone |           |          | 
Indexes:
    "rduser_pkey" PRIMARY KEY, btree (id)
Referenced by:
    TABLE "auth_token" CONSTRAINT "fk8kjy551iwoqqmsu2f1a88pwbu" FOREIGN KEY (user_id) REFERENCES rduser(id)
    TABLE "auth_token" CONSTRAINT "fk_aiqc20kpjasth5bxogsragoif" FOREIGN KEY (user_id) REFERENCES rduser(id)
```

**My Rundeck detail**

* Rundeck version: 6.1.0
* install type: deb
* OS Name/version: Debian Bookworm
* DB Type/version: postgres
* Java: Temurin JDK 25

**To Reproduce**
A user that has never done this before tries to login. He will be shown the cat image with 500 There was an unexpected error

**Expected behavior**
Login should work and the user should be added to the `rduser` table.

I asked Claude to take a look and in his opinion commit 3bdd415e2f72cdddf821ad7bc3da404758450613 might play into this because it added to `User.groovy`:
```
 id generator: 'identity'
``` 
This makes Rundeck rely on the DB to provide the ID via autoincrement, which is not the case for our table.

According to Claude this commit was included in Rundeck 6.0.0++. When we ran 6.0.1 we had no new users trying to login, so we came across this after 6.1.0 update, but it might have been an issue on 6.0.1 already.

Should some kind of DB migration have taken place during any of the upgrades? During startup I see this
```
[2026-08-24T15:18:24,665] INFO  liquibase.database [main] - Set default schema name to public
[2026-08-24T15:18:25,347] INFO  liquibase.changelog [main] - Reading from public.databasechangelog
[2026-08-24T15:18:27,433] INFO  liquibase.lockservice [main] - Successfully acquired change log lock
[2026-08-24T15:18:27,435] INFO  liquibase.command [main] - Using deploymentId: 7577480046
[2026-08-24T15:18:27,438] INFO  liquibase.changelog [main] - Reading from public.databasechangelog
[2026-08-24T15:18:27,483] INFO  liquibase.util [main] - UPDATE SUMMARY
[2026-08-24T15:18:27,483] INFO  liquibase.util [main] - Run:                          0
[2026-08-24T15:18:27,483] INFO  liquibase.util [main] - Previously run:             117
[2026-08-24T15:18:27,483] INFO  liquibase.util [main] - Filtered out:                28
[2026-08-24T15:18:27,484] INFO  liquibase.util [main] - -------------------------------
[2026-08-24T15:18:27,484] INFO  liquibase.util [main] - Total change sets:          145
[2026-08-24T15:18:27,484] INFO  liquibase.util [main] - FILTERED CHANGE SETS SUMMARY
[2026-08-24T15:18:27,484] INFO  liquibase.util [main] - DBMS mismatch:               28
[2026-08-24T15:18:27,487] INFO  liquibase.util [main] - Update summary generated
[2026-08-24T15:18:27,490] INFO  liquibase.command [main] - Update command completed successfully.
[2026-08-24T15:18:27,490] INFO  liquibase.ui [main] - Liquibase: Update has been successful. Rows affected: 0
[2026-08-24T15:18:27,501] INFO  liquibase.lockservice [main] - Successfully released change log lock
[2026-08-24T15:18:27,503] INFO  liquibase.command [main] - Command execution complete
```

fdevans:
## Problem

Fixes #10493. (Reopened from #10545, which had the wrong ticket number in its branch/commit.)

After upgrading to 6.x, the login of a brand new user (one who has never logged in before, so no `rduser` row exists yet for them) fails with:

```
org.postgresql.util.PSQLException: ERROR: null value in column "id" of relation "rduser" violates not-null constraint
```

This affects long-lived Postgres installations that have been upgraded across many Rundeck versions over the years.

## Root cause

`User.groovy`'s mapping declares:

```groovy
static mapping = {
    table "rduser"
    id generator: 'identity'
}
```

This was added in [3bdd415e2f](https://github.com/rundeck/rundeck/commit/3bdd415e2f72cdddf821ad7bc3da404758450613), bundled into a broader Grails 7 / Spring Boot 3.5.8 dependency-conflict-resolution commit. The commit message doesn't explain why `identity` specifically was needed — the actual documented fix for the Hibernate 6 compatibility issue (`0d5fc1c74d`, the day before) was just adding the explicit `Long id` field, which alone fixed 115 failing tests.

`identity` requires the database column itself to auto-generate the value on insert (a real Postgres `IDENTITY`/serial column with an attached sequence). The Liquibase changeset that originally creates `rduser.id` ([RDUSER.groovy](https://github.com/rundeck/rundeck/blob/main/rundeckapp/grails-app/migrations/core/RDUSER.groovy)) only sets a generic `autoIncrement: "true"` column flag — whether that reliably produces a real DB-level identity/sequence depends on the Liquibase/DB version that originally ran it, which varies across years of upgrades. On installations where it didn't, Hibernate's identity-strategy `INSERT` omits the `id` column entirely (expecting the DB to fill it in), which fails with the not-null constraint violation shown above.

**This isn't necessary**: two other domain classes with the identical situation — an explicit `Long id` field added for Hibernate 6 compatibility, created via the exact same `autoIncrement: "true"` Liquibase pattern — declare **no** generator override at all: `ScheduledExecution` and `PluginMeta`. Both work fine in production across upgraded installations. `User` was the only domain class in the codebase with an explicit generator override.

## Fix

Remove `id generator: 'identity'` from `User.groovy`, matching the (already proven-correct) pattern used by every other domain class.

## Testing

Added `UserIntegrationSpec`, a real-database GORM persistence test (not `DataTest`-mocked), since the existing `User`/`GormUserDataProviderSpec` unit tests all mock persistence and never exercise Hibernate's actual id-generator resolution — which is exactly why this regression wasn't caught by CI. `./gradlew :rundeckapp:integrationTest --tests rundeck.UserIntegrationSpec` — 2/2 passing.

## Release Notes

Fixed a login failure (`null value in column "id" of relation "rduser"`) that could occur for new users on Rundeck installations that have been upgraded across multiple releases, particularly on PostgreSQL.

### DIFF
--- a/rundeckapp/grails-app/domain/rundeck/User.groovy
+++ b/rundeckapp/grails-app/domain/rundeck/User.groovy
@@ -31,7 +31,6 @@ class User implements RdUser{
     
     static mapping = {
         table "rduser"
-        id generator: 'identity'
     }
     String dashboardPref
     String filterPref
--- a/rundeckapp/src/integration-test/groovy/rundeck/UserIntegrationSpec.groovy
+++ b/rundeckapp/src/integration-test/groovy/rundeck/UserIntegrationSpec.groovy
@@ -23,18 +23,30 @@ import spock.lang.Specification
 /**
  * Integration tests for {@link User} persistence against a real database.
  *
- * Regression coverage for https://github.com/rundeck/rundeck/issues/10493:
- * the {@code rduser} table's {@code id} column is created via a plain
- * Liquibase {@code autoIncrement: "true"} column definition, the same
- * pattern used for {@code scheduled_execution.id} and other core tables.
- * Long-lived (repeatedly upgraded) Postgres installations may not have a
- * real DB-level IDENTITY/SERIAL sequence attached to that column, so
- * mapping {@code User}'s id with an explicit {@code generator: 'identity'}
- * strategy -- which requires the database itself to generate the value --
- * caused saving a brand new user to fail with a not-null constraint
- * violation on {@code id}. These tests exercise real GORM persistence
- * (not mocked via DataTest) so they actually go through Hibernate's id
- * generator resolution, unlike the existing unit specs for User.
+ * Background for https://github.com/rundeck/rundeck/issues/10493: the
+ * {@code rduser} table's {@code id} column is created via a plain Liquibase
+ * {@code autoIncrement: "true"} column definition, the same pattern used for
+ * {@code scheduled_execution.id} and other core tables. Long-lived
+ * (repeatedly upgraded) Postgres installations may not have a real DB-level
+ * IDENTITY/SERIAL sequence attached to that column, so mapping {@code User}'s
+ * id with an explicit {@code generator: 'identity'} strategy -- which
+ * requires the database itself to generate the value -- caused saving a
+ * brand new user to fail with a not-null constraint violation on
+ * {@code id}.
+ *
+ * Note on coverage: the test datasource here uses Hibernate
+ * {@code ddl-auto: create-drop} (see {@code application-test.yml}), which
+ * generates the schema fresh from the current mapping rather than from
+ * Liquibase -- so it does not reproduce the specific legacy-schema mismatch
+ * from #10493 (a schema that predates/lacks a real identity column). What
+ * these tests verify is that real (non-DataTest-mocked) GORM persistence of
+ * a new User works end-to-end and assigns distinct ids, exercising
+ * Hibernate's actual id generator resolution instead of DataTest's mocked
+ * persistence, which the existing unit specs for User use and which would
+ * not have caught this class of bug either way. The specific regression --
+ * reintroducing an explicit {@code identity} generator on User -- is guarded
+ * against directly by {@code UserTests#"does not map id with an explicit
+ * generator"}, which asserts on the mapping itself.
  */
 @Integration
 @Rollback
--- a/rundeckapp/src/integration-test/groovy/rundeck/UserIntegrationSpec.groovy
+++ b/rundeckapp/src/integration-test/groovy/rundeck/UserIntegrationSpec.groovy
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2026 Rundeck, Inc. (http://rundeck.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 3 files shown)


luismalamoc:
The PR body is thorough on the root cause, but it doesn't follow the repo's PR template sections — in particular the **Release Notes** section is missing. Since this is a user-facing fix (new-user login failing with a 500 on upgraded Postgres installations), a short release note would be valuable for community users scanning the changelog. Suggested draft:

> Fixed a regression introduced in 6.0.0 where the first login of a new user could fail with a `null value in column "id" of relation "rduser"` constraint violation on long-lived PostgreSQL installations. (#10493)

luismalamoc (approved):

PR #10547, opened 2026-09-02
[RUN-4884] Sync activity filters into the URL so browser back/forward preserves them

### LINKED ISSUE #10031: Rundeck activity filtering reset on opening an activity
**Describe the bug**
After adding filter to activity page, if you go to an activity and come back, the filter is cleared

**My Rundeck detail**

* Rundeck version: 5.18.0-20251216
* install type: [rpm]
* OS Name/version: fedora 8.10
* DB Type/version: mysql

**To Reproduce**
1. Go to any job activity page
2. Add any filter to the activities
3. Open any activity
4. Go back using browser back 
5. Activity filter is reset

**Expected behavior**
The activity filter set in the same session should persist when opening an activity. That was the same behaviour in rundeck v3.x.x

fdevans:
## Problem

Fixes #10031.

Steps to reproduce (from the issue):
1. Go to any job's Activity page
2. Add any filter to the activities
3. Open any activity
4. Go back using the browser's back button
5. The activity filter is reset

Reporter notes this used to work correctly in Rundeck 3.x.

## Root cause

Filters applied on the Activity page live only in [`activityList.vue`](https://github.com/rundeck/rundeck/blob/main/rundeckapp/grails-spa/packages/ui-trellis/src/app/components/activity/activityList.vue)'s component `query` state — nothing writes them into the browser's address bar (no `history.pushState`/`replaceState` anywhere in the component). Opening an execution is a full-page navigation to a different URL; clicking "back" returns to the Activity page's *original* URL from browser history, which still has the pre-filter (or no) query string, since it was never updated. The page reloads and re-bootstraps from that stale URL, silently dropping the filter.

I confirmed the other half of the round-trip already exists server-side: [`rundeckapp/grails-app/views/reports/index.gsp`](https://github.com/rundeck/rundeck/blob/main/rundeckapp/grails-app/views/reports/index.gsp) builds the page's initial `query` bootstrap data from `params.subMap([...])` — i.e. whatever filter query params are present on the incoming request — so once the URL actually reflects the applied filters, a hard reload via back navigation correctly restores them. This is presumably how it worked pre-Vue-rewrite (classic server-rendered GET-with-params list page).

Also found `execnodeFilter` missing from that `subMap` allowlist even though `activityList.vue`'s `query` object includes it — so the node filter specifically could never round-trip regardless of the client-side fix. Added it alongside the existing filter keys.

## Fix

This codebase already has an established pattern for exactly this "don't lose it on browser back" problem: [`JobListPage.vue`](https://github.com/rundeck/rundeck/blob/main/rundeckapp/grails-spa/packages/ui-trellis/src/app/pages/job/browse/JobListPage.vue) uses `history.replaceState`/`pushState` to keep the Jobs list's browse path in sync with the URL. Applied the same idea here: `activityList.vue`'s `query` watcher now also calls a new `syncQueryToUrl()` method, which updates the current URL via `history.replaceState` (not `pushState`, so it doesn't spam the back-button history stack on every filter keystroke).

## Testing

Added 2 tests to `activityList.spec.ts`:
- Applying a filter calls `history.replaceState` with the filter reflected in the URL query string.
- No error is thrown in environments without `history.replaceState`.

`npx jest activityList.spec.ts` — 12/12 passing (10 existing + 2 new). `npx eslint` on all 3 changed files — 0 errors, no new warnings.

## Release Notes

Fixed the Activity page so that applied filters are preserved when navigating back from viewing an execution, instead of being silently reset.

### DIFF
--- a/rundeckapp/grails-app/views/reports/index.gsp
+++ b/rundeckapp/grails-app/views/reports/index.gsp
@@ -50,7 +50,8 @@
             'startafterFilter',
             'endbeforeFilter',
             'endafterFilter',
-            'filterName'
+            'filterName',
+            'execnodeFilter'
     ]) + (defaultRecentFilter && !params.recentFilter ? [recentFilter: defaultRecentFilter] : [:])}"/>
     <g:embedJSON id="eventsparamsJSON" data="${eventsparams}"/>
     <g:embedJSON id="pageparamsJSON" data="${pageparams}"/>
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/app/components/activity/activityList.vue
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/app/components/activity/activityList.vue
@@ -1236,7 +1236,9 @@ export default defineComponent({
       ) {
         return;
       }
-      const url = _genUrl(this.activityPageHref, this.fullQueryParams());
+      // Reuse the same URL the "N executions" summary link already computes,
+      // rather than re-deriving it, so there's a single source of truth.
+      const url = this.activityHref;
       history.replaceState(history.state, "", url);
     },
   },
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/app/components/activity/activityList.vue
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/app/components/activity/activityList.vue
@@ -746,6 +746,7 @@ export default defineComponent({
     query: {
       handler(newValue, oldValue) {
         this.reload();
+        this.syncQueryToUrl();
       },
       deep: true,
     },
@@ -1222,6 +1223,21 @@ export default defineComponent({
         }
       }
       return params;
+    },
+    syncQueryToUrl() {
+      // Reflect the current filters in the address bar (without navigating) so
+      // that browser back/forward restores the filter state instead of losing
+      // it. See https://github.com/rundeck/rundeck/issues/10031 -- filters
+      // previously only lived in component state, so opening an execution and
+      // going back returned to the URL as it was on the original page load.
+      if (
+        typeof history.replaceState !== "function" ||
+        !this.activityPageHref
+      ) {
+        return;
+      }
+      const url = _genUrl(this.activityPageHref, this.fullQueryParams());
+      history.replaceState(history.state, "", url);
     },
   },
 });
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/app/components/activity/activityList.vue
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/app/components/activity/activityList.vue
@@ -1236,6 +1236,24 @@ export default defineComponent({
       ) {
         return;
       }
+      // This component is also embedded on other pages that share the same
+      // bootstrap data shape (e.g. the ad hoc Command page sets
+      // activityPageHref to the standalone Activity page's URL even though
+      // the browser is actually on /command/run). Only touch the address bar
+      // when it already matches the Activity page itself, otherwise this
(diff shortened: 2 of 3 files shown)


luismalamoc:
Reviewed alongside Copilot's feedback — two changes requested before approval:

1. **Reuse `this.activityHref` in `syncQueryToUrl()`** — it is the exact same expression as the existing computed (`_genUrl(this.activityPageHref, this.fullQueryParams())`). Same point as Copilot's comment; keeps a single source of truth for URL generation.

2. **On the "replaceState unsupported" test**: rather than fixing that test as Copilot suggests, drop the `typeof history.replaceState !== "function"` guard and the test entirely. This bundle targets Vue 3 / TS `esnext` — every supported browser has shipped `history.replaceState` since IE10, and IE is not supported at all. The guard defends against an environment that cannot run the app, and the test exists only to cover the guard (and as written it doesn't exercise it anyway — the `delete` falls back to `History.prototype.replaceState`, which is still a function).

Minor, non-blocking: `syncQueryToUrl()` rebuilds the URL from `query` keys only, so non-filter params present on the initial load (e.g. `max`) are dropped from the address bar on the first filter change. Fine to leave as-is, just noting the behavior change.

luismalamoc (approved):
LGTM. Root cause well diagnosed, fix reuses the established `history.replaceState` pattern from JobListPage.vue, and `syncQueryToUrl()` now reuses the `activityHref` computed as a single source of truth. Review feedback (Copilot + mine) addressed in 28c703b, with the unsupported-`replaceState` test now genuinely exercising the guard (mutation-verified). Approving with `build_and_test` still running — merge on green.

fdevans:
Root-caused and fixed the CI failures (4 Selenium suites, ~39 failing tests, all `TimeoutException`s waiting for the URL to reach `/command/run` while actually sitting on `/activity?adhoc=true`).

`activityList.vue` isn't exclusive to the standalone Activity page — it's also embedded on the ad hoc **Command** page (`adhocNext.gsp`), which sets `activityPageHref` to the Activity page's own URL and bootstraps `query = { adhoc: true }` even though the browser is actually on `/command/run`. `syncQueryToUrl()` blindly called `history.replaceState` on mount regardless of which page it was actually embedded in, clobbering the Command page's address bar to the Activity page's URL — which is exactly what broke every Selenium test waiting to land on `/command/run`.

Fixed by only syncing the URL when `window.location.pathname` already matches the Activity page's own path. Added a regression test that reproduces the exact scenario (mounts with the URL set to `/command/run`) — confirmed it fails without the guard and passes with it.

This is a good concrete example of exactly the kind of interaction risk flagged in the PR description as needing UI-team review before merge — glad CI caught it rather than it shipping.

fdevans:
Follow-up on CI: the previous fix resolved 3 of 4 failing Selenium suites. The last one (`T Selenium - Execution & Other`) had a different, more specific failure: `ActivitySpec`'s "default max #max shows #expect rows" tests navigate directly to `/project/x/activity?max=N` and expect that param to survive page load, but it was getting silently dropped.

Cause: `syncQueryToUrl()` was reusing the `activityHref` computed property for the address-bar sync, which builds its URL from only `this.query` (the filter fields) — it has no notion of `max`/`offset`/etc. Every sync call was replacing the *entire* query string with just the filter params, discarding anything else already there.

Fixed by merging into the existing `URLSearchParams` instead of rebuilding the query string from scratch: remove only the keys this component actually manages, then re-set the current filter values, leaving unrelated params (like `max`) untouched. Verified by reverting to the old approach and confirming the new test reproduces the exact `max`-goes-missing symptom, then restored the fix.

Pushed — waiting on this CI run now.

PR #10549, opened 2026-09-02
[RUN-4886] Gate node/npm engine Renovate group behind Dependency Dashboard approval

luismalamoc:
## Tell us about your PR

Maintenance / CI cost reduction for Renovate configuration.

Renovate runs independently on `rundeck` and `rundeckpro`, and node/npm versions (`.nvmrc` + `engines.npm`) must move in lockstep across both repos. Renovate cannot open a single cross-repo PR, so the auto-created "node and npm engine versions" PRs (branch `renovate/node-and-npm-engine-versions`) can never be merged unilaterally — they only consume CircleCI credits on every rebase.

Added `"dependencyDashboardApproval": true` to the existing `"node and npm engine versions"` package rule in `renovate.json`, and extended the rule's description to document the cross-repo lockstep rationale. Renovate now lists the update under "Pending Approval" on the Dependency Dashboard instead of creating a branch/PR. When a coordinated bump is due, approve the checkbox on both repos' dashboards in the same session, merge the rundeck PR first, then the rundeckpro one.

A mirror PR applies the identical change to `rundeckpro` (the two `renovate.json` files are intentionally kept in sync).

- A `schedule` on the group: reduces noise but still auto-creates unmergeable PRs.
- A single cross-repo PR: not supported by Renovate (per-repository operation; submodule contents are not scanned).

**Additional context**

Validated with `npx --package renovate -- renovate-config-validator renovate.json` (passes). No change needed for the per-project `.npmrc` files — they are already identical across the npm projects and Renovate honors them.

## Release Notes

N/A — CI/dependency-bot configuration only, no product change.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -121,10 +121,11 @@
       "enabled": true
     },
     {
-      "description": "Group Node.js version (.nvmrc) with the npm engine pin so they update together",
+      "description": "Group Node.js version (.nvmrc) with the npm engine pin so they update together. Gated behind Dependency Dashboard approval: node/npm must move in lockstep across the rundeck and rundeckpro repos (Renovate cannot open a single cross-repo PR), so auto-created PRs can never merge unilaterally and only burn CI. Approve the checkbox on BOTH repos' dashboards in the same session, merge rundeck first, then rundeckpro.",
       "matchManagers": ["nvm", "npm"],
       "matchDepNames": ["node", "npm"],
-      "groupName": "node and npm engine versions"
+      "groupName": "node and npm engine versions",
+      "dependencyDashboardApproval": true
     },
     {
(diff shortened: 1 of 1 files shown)


smartinellibenedetti (approved):
I'd like for us to see if there's a way to get renovate to update the node version to match what is available within the docker images we use

luismalamoc:
@smartinellibenedetti thanks for the approval. On matching node to what the CI docker image ships, I looked into it:

- Renovate has no datasource for "the node version bundled in `cimg/openjdk:17.0.20`", so it cannot track that automatically. The only config-level knobs are static (`allowedVersions` on the node group), which would need a manual update every time the image changes.
- In core's CircleCI config only `test-gui` installs node from `.nvmrc` via nvm; `build-rundeck` and `test-gradle` already use whatever node the `cimg/openjdk` image provides. So we are effectively running two node versions in the same pipeline today.

The clean fix is on the CI side rather than in Renovate: either drop the nvm step and rely on the image's node everywhere (then `.nvmrc` becomes purely a local-dev hint), or install from `.nvmrc` in every job so the image's node is irrelevant. That is a separate change from this PR; happy to open a ticket for it if you agree with either direction.

Also FYI: per your comment on the rundeckpro side, that PR was repurposed to gate core-shared npm packages too, so the two `renovate.json` files now intentionally diverge (the shared-package rule only makes sense on the consumer side).

Issue #10321, opened 2026-07-09
The text box in the script editor is very small


C64Axel:
**Describe the bug**
The script editor in the workflow is small and cannot be enlarged.
It's hard to edit longer scripts there.

**My Rundeck detail**

* Rundeck version: 5.20.10,6.0.0
* install type: deb
* OS Name/version: debian 13 trixie
* DB Type/version: mariadb 11.8.6

**To Reproduce**
1. Go to 'Edit this Job'
2. Click on 'Workflow'
3. Click on 'Script Execute an inline script'

**Expected behavior**
It should be possible to enlarge the editor's text window

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Edge,Chrome,Firefox
 - Version: 150.0.4078.48,150,148

<img width="1280" height="461" alt="Image" src="https://github.com/user-attachments/assets/bede4724-8986-4487-9062-da1476ab4b35" />

fdevans:
This should be fixed in 6.0, have you tried that?

C64Axel:
> This should be fixed in 6.0, have you tried that?

Yes, tried 5.20 and 6.0.0 see "Rundeck Version" above.
And different Browsers.

cadams-entrata:
I have noticed that this only happens when the "Conditional Logic Steps" feature is enabled. It also changes entry boxes to white in dark mode. When `rundeck.feature.earlyAccessJobConditional.enabled` is set to `false`, all of the boxes are normal.

fdevans:
It is related to the NextUI toggle I believe.  I'm pretty sure there's a fix in the queue already for it.

rmueck:
The only one that helps me was: Inserting this in `/etc/nginx/sites-available/rundeck`

~~~nginx
location / {
        # ... your existing proxy headers ...

        # Fix small script editor box in Rundeck 5.x UI
        sub_filter '</head>' '<style>.ace_editor { min-height: 450px !important; }</style></head>';
        sub_filter_once on;
    }
~~~
Then reload Nginx (sudo systemctl reload nginx). This injects a standard CSS override that forces all inline script textareas to stay at least 450 pixels tall.

Note: Non of the variables in rundeck-config.properties worked for me 
~~~conf
rundeck.feature.nextUi.enabled=false
rundeck.feature.earlyAccessJobConditional.enabled=false
~~~

C64Axel:
Same on Version 6.1.0

fdevans:
[...]

PR #10551, opened 2026-09-03
[RUN-4890] Make ACE editor manually resizable when minLines is unset

smartinellibenedetti:
## Tell us about your PR

The original GSP-based script/plugin-config editor supported manual drag-to-resize. When it was converted to `AceEditorVue.vue`, the resize behavior became wonky — CSS `resize` fights with Ace's own auto-grow-to-content sizing, producing a jarring "jump" on resize or on typing a new line.

As a workaround, #10137 ([RUN-4277](https://pagerduty.atlassian.net/browse/RUN-4277)) introduced admin-configurable `minLines`/`maxLines` props on `AceEditorVue.vue`, letting Ace auto-size the editor instead of relying on manual resize. #10536 ([RUN-4862](https://pagerduty.atlassian.net/browse/RUN-4862)) later just raised the out-of-the-box default `minLines` from 12 to 20, since most users weren't aware the setting existed and kept seeing a tiny editor.

This PR re-introduces working manual resize by making it mutually exclusive with Ace's `minLines`/`maxLines` auto-sizing, instead of only supporting one or the other project-wide, and makes resizable the new out-of-the-box default.

- `AceEditorVue.vue` now wraps its root element in a `<div>` that gets a `resizable` class when `minLines === 0`.
- When resizable, `minLines`/`maxLines` are no longer passed to Ace's `setOptions` at all, so Ace's own auto-sizing doesn't fight the manual CSS resize. A `ResizeObserver` keeps Ace's internal renderer in sync when the container is dragged to a new size.
- When `minLines` is set to a non-zero value, behavior is unchanged from #10137/#10536: Ace manages height via `minLines`/`maxLines` auto-sizing, and manual resize is disabled.
- Added `.resizable .ace_editor { resize: vertical; overflow: auto; height: 320px; }` to `ace-editor.scss`, scoped to the new wrapper class. Height lives on `.ace_editor` itself (not the wrapper), so the wrapper grows to contain a manually-resized editor instead of the enlarged editor overlapping subsequent form fields.
- Changed the out-of-the-box default `minLines` from 20 back to **0** (manually resizable) in `RundeckConfigBase.java`, `base.gsp`, and the System Configuration description in `FeatureFlagConfigurable.groovy` — the config description now explains that 0 means the editor is manually resizable rather than auto-sized to a fixed line count. Admins can still opt back into the fixed-height auto-sizing behavior by configuring a non-zero `guiAceEditorMinLines`.
- Updated `AceEditorVue.spec.ts` to cover both the resizable (`minLines === 0`) and non-resizable (`minLines > 0`) code paths, the `resizable` class toggling, and the `ResizeObserver` wiring. Fixed a stale assertion in `pluginPropEdit.spec.ts` that still expected the old default of 20.

Considered fully removing Ace's `minLines`/`maxLines` auto-sizing (from #10137) in favor of manual resize only, but that would break existing consumers/admins relying on the configured auto-fit sizing. Making the two modes mutually exclusive (gated on whether `minLines` is set) keeps both available without them conflicting, per Ace's own limitation that it doesn't handle auto-sizing and manual resize at the same time.

**Additional context**
Verified manually via the job creation Script step editor: with `minLines` unset (default), the editor is resizable and dragging it larger correctly pushes subsequent form fields down without overlap or jumping. With `minLines` set to a non-zero value (e.g. 20), the editor auto-fits its content with no resize handle, matching #10137/#10536 behavior for that mode.

- Related: [RUN-4277](https://pagerduty.atlassian.net/browse/RUN-4277) (#10137), [RUN-4862](https://pagerduty.atlassian.net/browse/RUN-4862) (#10536), #10321

## Release Notes

The ACE code editor is now manually resizable via drag handle by default (0 minimum lines). Admins can restore the previous auto-fit behavior by configuring a non-zero minimum line count under System Configuration → GUI.

[RUN-4277]: https://pagerduty.atlassian.net/browse/RUN-4277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RUN-4862]: https://pagerduty.atlassian.net/browse/RUN-4862?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RUN-4277]: https://pagerduty.atlassian.net/browse/RUN-4277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/java/com/dtolabs/rundeck/core/config/RundeckConfigBase.java
+++ b/core/src/main/java/com/dtolabs/rundeck/core/config/RundeckConfigBase.java
@@ -510,10 +510,10 @@ public class RundeckConfigBase {
         Enabled activityDefaultTimeFilter = new Enabled();
         Enabled vueKeyStorage = new Enabled(true);
         Enabled pluginGroups = new Enabled(true);
-        // RUN-4277/#10321: default raised from 12 to 20 — Ace no longer supports a manual
-        // resize handle (upstream removed it), so a taller out-of-the-box default is the
-        // supported way to make the inline-script/code editor more usable without configuration.
-        int guiAceEditorMinLines = 20;
+        // RUN-4862: minimum number of visible lines in the ACE code editor. A value of 0
+        // (the default) makes the editor manually resizable via drag handle instead of
+        // auto-sizing to a fixed number of lines.
+        int guiAceEditorMinLines = 0;
         int guiAceEditorMaxLines = 0;
 
         @Data
--- a/rundeckapp/grails-app/views/layouts/base.gsp
+++ b/rundeckapp/grails-app/views/layouts/base.gsp
@@ -134,7 +134,7 @@
             logo:'${g.appLogo()}',
             logocss:'${g.appLogocss()}',
             appRundeckGatewayUrl: '${g.appRundeckGatewayUrl()}',
-            aceEditorMinLines: ${cfg.getInteger(config:'feature.guiAceEditorMinLines', default:20)},
+            aceEditorMinLines: ${cfg.getInteger(config:'feature.guiAceEditorMinLines', default:0)},
             aceEditorMaxLines: ${cfg.getInteger(config:'feature.guiAceEditorMaxLines', default:0)}
         },
         hideVersionUpdateNotification: '${session.filterPref?.hideVersionUpdateNotification}',
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/pluginPropEdit.vue
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/pluginPropEdit.vue
@@ -221,11 +221,9 @@
                 prop.options['codeSyntaxSelectable'] === 'true' &&
                 !renderReadOnly
               "
-              height="200"
-              width="100%"
               :read-only="renderReadOnly"
               :context-variable-suggestions="scriptTypeContextVariables"
-              :min-lines="aceEditorMinLines"
+              :min-lines="aceEditorMinLines > 0 ? aceEditorMinLines : undefined"
               :max-lines="aceEditorMaxLines"
             />
           </ui-socket>
@@ -440,10 +438,7 @@ import {
   WorkflowStepType,
 } from "../utils/contextVariableUtils";
 
-// RUN-4277/#10321: default raised from 12 to 20 — Ace no longer supports a manual resize
-// handle (upstream removed it), so a taller out-of-the-box default is the supported way to
-// make the inline-script/code editor more usable without configuration.
-const ACE_EDITOR_DEFAULT_MIN_LINES = 20;
+const ACE_EDITOR_DEFAULT_MIN_LINES = 0;
 const ACE_EDITOR_DEFAULT_MAX_LINES = 0;
 
 interface Prop {
@@ -602,7 +597,8 @@ export default defineComponent({
       return this.appMeta.aceEditorMinLines ?? ACE_EDITOR_DEFAULT_MIN_LINES;
     },
     aceEditorMaxLines(): number {
-      const raw = this.appMeta.aceEditorMaxLines ?? ACE_EDITOR_DEFAULT_MAX_LINES;
+      const raw =
+        this.appMeta.aceEditorMaxLines ?? ACE_EDITOR_DEFAULT_MAX_LINES;
       return raw === 0 ? Infinity : raw;
     },
   },
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/tests/pluginPropEdit.spec.ts
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/tests/pluginPropEdit.spec.ts
@@ -60,11 +60,11 @@ describe("pluginPropEdit aceEditor computed props", () => {
   });
 
   describe("aceEditorMinLines", () => {
-    it("passes minLines of 20 to AceEditorVue when appMeta has no aceEditorMinLines", async () => {
+    it("passes minLines of 0 to AceEditorVue when appMeta has no aceEditorMinLines", async () => {
       mockGetRundeckContext.mockReturnValueOnce({ appMeta: {} });
       const wrapper = await createCodeWrapper({ prop: codeProp });
 
-      expect(wrapper.findComponent(AceEditorVue).props("minLines")).toBe(20);
+      expect(wrapper.findComponent(AceEditorVue).props("minLines")).toBe(0);
     });
 
     it("passes the configured minLines to AceEditorVue when appMeta provides it", async () => {
--- a/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/utils/AceEditorVue.vue
+++ b/rundeckapp/grails-spa/packages/ui-trellis/src/library/components/utils/AceEditorVue.vue
@@ -1,5 +1,7 @@
 <template>
-  <div :id="identifier" ref="root"></div>
+  <div :class="{ resizable: minLines === 0 }">
+    <div :id="identifier" ref="root"></div>
+  </div>
 </template>
 <script lang="ts">
 import { defineComponent, PropType } from "vue";
@@ -31,7 +33,7 @@ export default defineComponent({
     },
     minLines: {
       type: Number,
-      default: 12,
+      default: 0,
     },
     maxLines: {
       type: Number,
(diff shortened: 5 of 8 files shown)


smartinellibenedetti:
> ### 🔵 Needs a closer look
> `AceEditorVue.vue` does not fully handle runtime transitions between resizable and auto-sized modes (ResizeObserver wiring and Ace option state can become inconsistent if `minLines` changes after mount).
> 
> Review details
> ### Suppressed comments (2)
> **Previously missed (1)** — in code that hasn't changed since the last review.
> 
> **rundeckapp/grails-spa/packages/ui-trellis/src/library/components/utils/tests/AceEditorVue.spec.ts:86**
> 
> * This test doesn’t actually guarantee `minLines`/`maxLines` are omitted from _all_ `setOptions` calls: `toHaveBeenCalledWith(expect.not.objectContaining(...))` will pass as long as at least one call matches, even if another call included the forbidden keys. Iterating over `mockSetOptions.mock.calls` makes the assertion robust against future additional `setOptions` invocations.
> 
> **rundeckapp/grails-spa/packages/ui-trellis/src/library/components/utils/AceEditorVue.vue:116**
> 
> * The component’s behavior becomes inconsistent if `minLines` toggles at runtime (e.g., `> 0` → `0`): the `resizable` class updates automatically, but the existing Ace `minLines/maxLines` options are not cleared and the `ResizeObserver` is never attached/detached reactively. This can leave Ace auto-sizing enabled while CSS resize is enabled, reintroducing the “jump” you’re trying to avoid, and can also leave the resize observer running when switching back to non-resizable.
> 
> ```
>     minLines: function (val: number): void {
>       if (val && !this.isResizable) {
>         this.editor!.setOptions({ minLines: val });
>       }
>     },
>     maxLines: function (val: number): void {
>       if (!this.isResizable) {
>         this.editor!.setOptions({ maxLines: val });
>       }
> ```
> 
> * **Files reviewed:** 8/8 changed files
> * **Comments generated:** 0 new
> * **Review effort level:** Lite

Min/maxLines being stale is a scenario that isn't applicable to us at the moment

fdevans (approved):

luismalamoc:
## Review notes

Code looks correct and CI is green. Two things I'd like addressed or explicitly acknowledged before merge, plus some small simplifications.

### Needs attention

**1. Docs are out of date and no docs PR is linked.**
`rundeck/docs` (`docs/administration/configuration/config-file-reference.md`) still documents `rundeck.feature.guiAceEditorMinLines` with default `12` and describes it as "minimum number of visible lines". This PR changes the default to `0` and gives that value a new meaning (manually resizable). Since this is user-facing, a docs PR should accompany it.

**2. The default change affects every ACE editor in the app, not just plugin config forms.**
The `height`/`width` props on `AceEditorVue.vue` are never applied to the DOM (pre-existing dead props), so `minLines` was effectively the only thing controlling editor size. `AceEditor.vue` does not pass `min-lines`, so all of its consumers now go from auto-grow (≥ 20 lines) to a fixed 320px box with a resize handle:

- `OptionEdit.vue`, `DetailsEditor.vue`, `EditProjectFile.vue` (readme/motd), `EditProjectNodeSourceFile.vue`, `WorkflowGraph.vue`
- `pluginPropView.vue` (read-only)
- rundeckpro: `JobStepDisplay.vue` (GenAI, read-only) and `TourManager.vue`

The PR description only mentions verifying the Script step. A read-only viewer showing a resize handle, and a long node-source YAML file that used to grow to fit its content and now becomes a 320px scroll box, are UX changes worth a quick visual pass and a mention in the release notes. The System Configuration description ("rendered inside plugin configuration forms") is also no longer accurate.

### Simplifications (non-blocking)

- `pluginPropEdit.vue`: `:min-lines="aceEditorMinLines > 0 ? aceEditorMinLines : undefined"` is redundant. `undefined` falls back to the prop default of `0`, which yields the same result. Passing `aceEditorMinLines` directly is enough.
- `AceEditorVue.vue`: the template uses `minLines === 0` while the `isResizable` computed exists for exactly that. Use `isResizable` in the template.
- `AceEditorVue.vue`: in the `minLines` watcher, `val && !this.isResizable` can be just `!this.isResizable`, since `isResizable` already implies `val !== 0`.
- `ace-editor.scss`: `.resizable { width: 100%; position: relative }` and `width: 100%` on `.ace_editor` have no effect on block-level divs. `height`, `resize`, and `overflow` are all that's needed.
- `.resizable` is a very generic class name in a global stylesheet (`app.scss`). No collision today, but renaming to `.ace-resizable` is free insurance.
- The new comment in `RundeckConfigBase.java` references RUN-4862, but this work is RUN-4890. We also avoid ticket references in source files; the comment reads fine without it.

### Copilot comments status

1. **Stale `pluginPropEdit.spec.ts` assertion** — addressed in `735084e`. Thread can be resolved.
2. **Runtime transitions between resizable and auto-sized modes** — not addressed in code, but the author's reply ("not applicable to us at the moment") is valid: `appMeta` is static per page load and no consumer changes `minLines` at runtime. Agree with not building it. The reply is on the PR rather than the thread, so Copilot re-flagged it; worth replying inline and resolving.
3. **`resize: both` vs `vertical`** — addressed by updating the PR description.
4. **Suppressed: `options` could carry `minLines`/`maxLines`** — not applicable, no consumer passes `:options`.
5. **Suppressed: weak `toHaveBeenCalledWith(expect.not.objectContaining(...))` assertion** — not addressed. `setOptions` is only called once in `mounted` today, so the test is correct but fragile. Nit.

### Follow-up (out of scope)

The dead `height`/`width` props on `AceEditorVue.vue` and the consumers passing them (`height="120px"`, `"500"`, etc.) should be cleaned up in a separate PR.

luismalamoc (changes requested):
Requesting changes per the review notes above (https://github.com/rundeck/rundeck/pull/10551#issuecomment-5531197939): missing docs PR for the new default/semantics of `guiAceEditorMinLines`, and the default change reaches every `AceEditor.vue` consumer (including read-only viewers and rundeckpro components), which needs a visual pass and a release-notes/System Config description update. The simplifications listed there are optional.

fdevans:
Docs PR was merged to 6.3.0 branch for when that version releases

smartinellibenedetti:
> ### Needs attention
> **1. Docs are out of date and no docs PR is linked.** `rundeck/docs` (`docs/administration/configuration/config-file-reference.md`) still documents `rundeck.feature.guiAceEditorMinLines` with default `12` and describes it as "minimum number of visible lines". This PR changes the default to `0` and gives that value a new meaning (manually resizable). Since this is user-facing, a docs PR should accompany it.

PR was merged to 6.3 branch. As Forrest mentioned, it will be released alongside with the release.

> **2. The default change affects every ACE editor in the app, not just plugin config forms.** The `height`/`width` props on `AceEditorVue.vue` are never applied to the DOM (pre-existing dead props), so `minLines` was effectively the only thing controlling editor size. `AceEditor.vue` does not pass `min-lines`, so all of its consumers now go from auto-grow (≥ 20 lines) to a fixed 320px box with a resize handle:
> 
> * `OptionEdit.vue`, `DetailsEditor.vue`, `EditProjectFile.vue` (readme/motd), `EditProjectNodeSourceFile.vue`, `WorkflowGraph.vue`
> * `pluginPropView.vue` (read-only)
> * rundeckpro: `JobStepDisplay.vue` (GenAI, read-only) and `TourManager.vue`

This comment is inaccurate and not actionable. Height and width aren't dead props, the original logic watches them to trigger a resize. Changing that in this PR risks of introducing breaking changes without any added benefits. Could you please share a screenshot if you found any editors that actually decreased in size?

> The PR description only mentions verifying the Script step. A read-only viewer showing a resize handle, and a long node-source YAML file that used to grow to fit its content and now becomes a 320px scroll box, are UX changes worth a quick visual pass and a mention in the release notes. The System Configuration description ("rendered inside plugin configuration forms") is also no longer accurate.

This comment is inaccurate, the reason this issue keeps being revisited is that users want the ability to resize the field, which was the original behaviour before the Vue conversation. For users that do want the field to grow to fit the content, they can leverage the min/max lines behaviour, which still works and can be set.  

> ### Simplifications (non-blocking)
> * `ace-editor.scss`: `.resizable { width: 100%; position: relative }` and `width: 100%` on `.ace_editor` have no effect on block-level divs. `height`, `resize`, and `overflow` are all that's needed.

Just a heads up, this is the actual fix of the problem. The library requires the parent to have these styles in order to correctly calculate the size with the `resize: both`. Otherwise the editor wouldn't calculate the size correctly.

fdevans (approved):

luismalamoc (approved):
Approving. The docs PR (rundeck/docs#1954) is merged to the 6.3.0 branch, and the wider scope of the default change is a product decision that's already been made, so nothing blocking remains.

Optional nit for a follow-up: the System Configuration description still says the setting applies to editors "rendered inside plugin configuration forms", while the default now affects every ACE editor in the app. The other simplifications from my earlier comment stay optional.

PR #10554, opened 2026-09-03
[RUN-4861] Support bare JSON array bodies in JsonUtil for MenuController#listExport

### LINKED ISSUE #10468: JSON errors when browsing projects

fdevans:
## Tell us about your PR

Bugfix. Fixes #10468. The legacy (non-NextUI) job list page POSTs the visible job ids as a bare JSON array body to `POST /menu/listExport`, e.g. `["7add463a-...","3b66b5b7-..."]`, rather than a JSON object. `JsonUtil.parseRequestBody` (added as part of the Grails 7 migration, replacing the old `request.JSON` binding) always deserializes into `Map.class`, so this array body fails to parse (`MismatchedInputException: Cannot deserialize value of type LinkedHashMap from Array value`). This is logged as an ERROR on every legacy job list page view, and `listExport()` silently falls back to `idlist=null` instead of the intended comma-joined id list, since the parse failure is swallowed by safe navigation (`data?.join(...)`).

Added `JsonUtil.parseRequestBodyAsList(request)`, sharing the existing reader/fallback logic (extracted into `readRequestBodyContent`) but deserializing into `List.class`, and switched `MenuController#listExport` to use it instead of `parseRequestBody`. The other call sites of `parseRequestBody` are untouched since they all expect JSON objects.

Considered making `parseRequestBody` accept a target class/type parameter instead of adding a second method, but that would change the signature (and behavior) for the 9 other existing call sites that all expect a JSON object — a much larger blast radius for a fix that only needs to affect this one array-bodied endpoint.

**Additional context**

Added unit tests for `JsonUtil.parseRequestBody`/`parseRequestBodyAsList`, and extended `MenuControllerSpec` with a "list Export" case using a non-empty array body, asserting `idlist` is populated correctly (the existing test only used an empty array and didn't assert on `idlist`, so it didn't catch this regression). `JsonUtilSpec`: 8/8 passing. `MenuControllerSpec`: 157/158 passing (1 pre-existing unrelated skip).

This recreates the fix from #10535, which was closed after its branch was accidentally built on top of an unrelated commit from a different PR — this branch is cut cleanly from `main` with a single commit containing only this fix.

## Release Notes

Fixed the legacy (non-NextUI) job list page so that job ids are correctly parsed from the `POST /menu/listExport` request body, instead of silently failing to parse and logging an error on every page view.

### DIFF
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
@@ -3737,7 +3737,7 @@ if executed in cluster mode.
             // The request body here is a bare JSON array of job ids (not an object), sent by the
             // legacy (non-NextUI) job list page — see RUN-10468.
             def data = com.dtolabs.rundeck.util.JsonUtil.parseRequestBodyAsList(request)
-            def nextScheduled = data?.join(",")?.replaceAll(/"/, '')
+            def nextScheduled = data?.join(",")
             def query = new ScheduledExecutionQuery()
             query.idlist = nextScheduled
             query.projFilter = params.project
--- a/rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
+++ b/rundeckapp/grails-app/controllers/rundeck/controllers/MenuController.groovy
@@ -3734,7 +3734,9 @@ if executed in cluster mode.
         def results=[:]
         if(request.format=='json' ) {
             // Grails 7: Parse body using Jackson instead of request.JSON
-            def data = com.dtolabs.rundeck.util.JsonUtil.parseRequestBody(request)
+            // The request body here is a bare JSON array of job ids (not an object), sent by the
+            // legacy (non-NextUI) job list page — see RUN-10468.
+            def data = com.dtolabs.rundeck.util.JsonUtil.parseRequestBodyAsList(request)
             def nextScheduled = data?.join(",")?.replaceAll(/"/, '')
             def query = new ScheduledExecutionQuery()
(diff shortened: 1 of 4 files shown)


ronaveva (approved):
LGTM

PR #10555, opened 2026-09-03
[RUN-4891] Improve the ruleset workflow panel by adding scroll and keyboard support

ddarby-hike:
## Tell us about your PR

Bugfix + follow-up UX enhancements. [RUN-4891](https://pagerduty.atlassian.net/browse/RUN-4891): In the NextUI job editor, opening the Ruleset workflow strategy's "Rules" side panel for a job with a large ruleset (~60 lines) only displayed about 20 lines with no scrollbar, making the rest of the ruleset inaccessible/uneditable through the UI.

*Scrolling fix:*
The Ace editor backing the Rules panel is configured with `maxLines: Infinity`, so it auto-grows to fit all content instead of scrolling internally. The containing flex pane (`.rdtabs__pane`) had `flex-grow: 1` but no `min-height: 0`, so it also grew to match the oversized content instead of respecting its bounded space, and the overflow was silently clipped by an ancestor `overflow: hidden` with no scrollbar ever appearing.

Added `min-height: 0; overflow-y: auto;` to the scoped `:deep(.rdtabs__pane)` rule in `WorkflowGraph.vue`, so the pane respects its flex-allotted space and scrolls internally when content overflows. This rule only affects the Tabs instance rendered inside `WorkflowGraph.vue` (verified no other component relies on `.rdtabs__pane`'s layout).

Also cleaned up all pre-existing ESLint warnings in this file while it was open (dead code, missing prop defaults, unused component registrations, keyboard accessibility on toolbar buttons). Left one warning intentionally unaddressed: `vue/enforce-style-attribute` on the global `<style>` block, since it targets attribute selectors on JointJS-generated SVG nodes that are injected outside Vue's template — scoping it would silently break the diagram's interactive highlighting.

*Resizable divider (follow-up):*
The Rules panel width was hardcoded to 600px with no way to adjust the split between the graph canvas and the panel. Added a draggable, keyboard-resizable divider (`role="separator"` with ARIA value attributes) between them:
- Panel width is clamped between 25% and 75% of the graph widget's total width (computed from the container's measured `clientWidth`).
- Default width is 60% of the container width, measured on mount — it always resets to that on open (no persistence).
- Text selection is disabled during drag so dragging over the Ace editor doesn't select its text; listeners are cleaned up on mouseup and on unmount.

*Gutter width (follow-up):*
The Rules editor's line-number gutter looked much wider than the ~2-3 digits it actually needs. Ace's default gutter padding (`padding-left: 19px`, plus `padding-right: 13px` when fold widgets are enabled, which they are by default) reserves space for annotation/fold icons this editor never shows (its syntax worker is disabled, so no error/warning markers ever appear). Disabled fold widgets on this editor and tightened the remaining `.ace_gutter-cell` padding, scoped to `#wf_editor` only so no other Ace editor instance elsewhere in the app is affected.

*Review feedback follow-up:*
- Added a focus style to the scale-to-fit (`fa-crosshairs`) and edit (`fa-pen`) icon buttons — they previously had no visible focus indicator when tabbed to. New `.workflow-graph-icon-btn:focus` scoped rule matches standard button focus styling (`color: var(--font-color); box-shadow: var(--default-states-color) 0 0 0 3px;`).
- Removed the "Beta!" badge that was rendered on the edit button, along with its now-unused `graph.badge.beta` i18n key.

*Regression fix (found during manual testing after review):*
After the above changes were reviewed, manual testing found the Rules panel was back to only showing ~20 lines regardless of content — the scroll fix appeared to stop working. Root cause: commit `07edb10832` (RUN-4890, "Make ACE editor manually resizable when minLines is unset") landed on `main` and was pulled into this branch via a merge, changing `AceEditorVue.vue`'s `minLines` default from `12` to `0` and redefining `minLines === 0` to mean "manually resizable, fixed 320px box" instead of "auto-grow to fit content." The mid-level `AceEditor.vue` wrapper that `WorkflowGraph.vue` actually uses never forwarded `minLines`/`maxLines` to the underlying editor, so it silently picked up the new `0` default and the Rules editor dropped into the fixed 320px resizable box instead of auto-growing.

Fixed by forwarding `minLines`/`maxLines` through `AceEditor.vue`, with defaults (`0`/`Infinity`) matching `AceEditorVue.vue`'s own defaults exactly — so any `AceEditor.vue` consumer that doesn't specify a value gets the same manually-resizable editor as a bare `AceEditorVue.vue` instance, instead of a third, disconnected default. `WorkflowGraph.vue`'s Rules editor explicitly opts into `min-lines="20"`, since a fixed-height resizable box isn't the right fit for a full ruleset script editor — it now starts at 20 visible lines and grows further as needed. `EditProjectFile.vue` (the only other `AceEditor.vue` consumer) is unaffected and keeps the resizable default unless it opts in the same way.

Considered fixing the height/width application in the lower-level `AceEditorVue.vue` wrapper (its `height`/`width` props are never actually applied to the DOM) and/or bounding Ace's `maxLines`, but that component is shared by many other consumers across the codebase, so a change there carries broader blast radius than this task needs. The `.rdtabs__pane` fix is scoped entirely to this component; the `minLines`/`maxLines` forwarding fix touches only `AceEditor.vue`, the narrower mid-level wrapper with just two consumers. Similarly considered persisting the user's chosen panel width (e.g. localStorage) but resetting to the 60% default on each open was preferred. Also tried centering the "Rules" tab heading in the panel header, but per feedback reverted it — left justified as it was originally.

**Additional context**
Verified with `eslint` (0 errors, only pre-existing/expected warnings) and the Jest suites for `job/workflow-designer/tests/`, `utils/tests/AceEditor.spec.ts` (new), and `utils/tests/AceEditorVue.spec.ts` — all passing. This is a CSS/interaction change with no test file created for `WorkflowGraph.vue` itself (none existed previously; it's a large jointjs canvas component outside the scope of this fix). The resizable divider and gutter-width change have not been manually verified in a live browser — no browser automation tool was available in the development session.

## Release Notes

- Fixed the NextUI job editor's Ruleset strategy "Rules" panel not scrolling when the ruleset has more lines than fit the visible area.
- Added a resizable divider between the graph canvas and the Rules panel, and narrowed the Rules editor's line-number gutter.
- Added a focus indicator to the graph's icon buttons and removed the leftover "Beta!" badge from the edit button.

[RUN-4891]: https://pagerduty.atlassian.net/browse/RUN-4891?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rundeckapp/grails-spa/packages/ui-trellis/src/library/components/job/workflow-designer/WorkflowGraph.vue
```diff
@@ -13,10 +13,29 @@
...
+              class="btn btn-default"
+              role="button"
+              tabindex="0"
+              :aria-label="$t('graph.action.edit')"
+              :title="$t('graph.action.edit')"
+              @click="edit"
+              @keydown.enter="edit"
+              @keydown.space.prevent="edit"
+            >
               <i class="fa fa-pen" style="transform: translate(0, -5px)" />
```
smartinellibenedetti:
```suggestion
              <i class="fa fa-pen" />
```

I know that the style tag was added before your changes, could you please remove it?

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/job/workflow-designer/WorkflowGraph.vue
```diff
@@ -29,25 +48,33 @@
                   color: var(--warning-color);
                 "
               >
-                Beta!
+                {{ $t("graph.badge.beta") }}
               </div>
```
smartinellibenedetti: The translation is correct, but wanted to ask if you could delete this div and the entry in i18n as well. The ruleset should have left the beta stage releases ago 😅

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/job/workflow-designer/WorkflowGraph.vue
```diff
@@ -983,5 +1108,25 @@ export default defineComponent({
...
   flex-grow: 1;
+  min-height: 0;
+  overflow-y: auto;
+}
+
+.workflow-graph-resizer {
+  width: 6px;
+  cursor: col-resize;
+  flex-shrink: 0;
+  z-index: 101;
```
smartinellibenedetti, line 1108: Nit: Could we use a pre-existing z-index within the project please? would like to avoid adding new values for the z-index to make it easier to maintain the code.

smartinellibenedetti (changes requested):
Changes are looking good, I've tested and was able to both resize the bar and navigate with the keyboard.

There's just 2 small adjusments that would be great if you can address them:

1 - could you please add a style for focus for the divs that render the `fa fa-crosshairs` and `fa fa-pen` icons? It can be the same style as a button would show (color: var(--font-color); box-shadow: var(--default-states-color) 0 0 0 3px;) 

2 - could you please remove the `beta!` label? I know it existed before this PR, but it would be super helpful if you can remove it as part of these changes. I've left 2 comments with the changes to be done

I'll update the ticket and this PR description to explain the full extent of the changes since it does a bit more than just fixing the scroll, but overall thank you for the fix and the keyboard logic!

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/job/workflow-designer/WorkflowGraph.vue
ddarby-hike: Addressed in 2fe33b1f: removed the beta badge `<div>` (the fa-pen icon itself is unchanged) and added a focus style for the icon buttons per the review feedback.

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/job/workflow-designer/WorkflowGraph.vue
ddarby-hike: Addressed in 2fe33b1f — removed the div and the `graph.badge.beta` i18n entry. Agreed, long overdue :)

ddarby-hike:
Thanks for the review! Pushed 2fe33b1f addressing both items:

1. Added a focus style (`color: var(--font-color); box-shadow: var(--default-states-color) 0 0 0 3px;`) to the scale-to-fit and edit icon buttons.
2. Removed the "Beta!" badge from the edit button, along with the now-unused `graph.badge.beta` i18n key.

PR description updated to reflect the additional scope.

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/job/workflow-designer/WorkflowGraph.vue
ddarby-hike: Looked into this — the project's only established z-index scale is the Bootstrap-derived one in `_variables.scss` (`$zindex-navbar/dropdown/modal/tooltip/popover`), which starts at 1000 and is meant for navbar/dropdown/modal/tooltip/popover layers, not local component overlays like this. Every other `z-index: 100`-ish value in the codebase (including the sibling side panel a few lines above this one) is likewise its own independent hardcoded value, not a shared token — so there isn't an existing value in this range to point to.

The `101` here specifically needs to render above the side panel's `z-index: 100` (both are flex items, so z-index applies without `position` being set) so the resizer isn't visually covered by the panel's box-shadow. Reusing `100` for both would put them at equal z-index, and since the resizer comes first in the DOM, the panel would paint over it again — reintroducing the issue this was added to avoid.

Given there's no existing value that fits, I'll leave this as-is unless you had a specific value in mind — happy to adjust if so.

smartinellibenedetti (approved):
Approved, thank you once again!

ddarby-hike:
Pushed 4fd6dfd fixing a regression found during manual testing: the Rules panel had gone back to showing only ~20 lines.

Root cause: `07edb10832` (RUN-4890) landed on `main` and got merged into this branch, changing `AceEditorVue.vue`'s `minLines` default from `12` to `0` — where `0` now means "manually resizable, fixed 320px box" instead of "auto-grow to fit content." `AceEditor.vue`, the wrapper `WorkflowGraph.vue` uses, never forwarded `minLines`/`maxLines` to the underlying editor, so it silently picked up the new default and lost the auto-grow behavior this PR's scroll fix depends on.

Fixed by forwarding `minLines`/`maxLines` through `AceEditor.vue` with defaults matching the prior behavior (`12`/`Infinity`). This also fixes the same silent regression for `EditProjectFile.vue`, the only other consumer of that wrapper. Added `AceEditor.spec.ts` (didn't exist before) covering the forwarding.

PR description updated with the details.

PR #10563, opened 2026-09-08
[RUN-4852] Stop SCM git-export from wiping workdir on Create Branch re-init

Jesus-Osuna-M:
## Tell us about your PR

Bugfix for [RUN-4852](https://pagerduty.atlassian.net/browse/RUN-4852): on 6.1.0, SCM git-export with **Create Branch** enabled and **Branch ≠ Base Branch** destroys its working directory on every plugin initialize. That is a regression from RUN-4018 (clone `baseBranch` first). `ScmLoader` re-inits about every 20s, so serialized-but-uncommitted job files disappear, Export shows **No changes**, and after failed retries the loader can persist `enabled=false`.

- Skip the base-branch-first clone when the workdir is already checked out on the configured export branch, so a loader re-init only fetches instead of `removeWorkdir()` + re-clone.
- After creating the export branch, check it out locally and treat a non-OK push result as a configuration error (instead of leaving HEAD on the base branch and wiping on the follow-up `cloneOrCreate`).

The RUN-4018 path is unchanged for a **fresh** workdir: still clone `baseBranch` first, then create the missing export branch.

- Stop calling `loadPluginWithConfig(..., initialize=true)` on every loader tick. That would also stop the wipe storm, but it is a larger lifecycle change and does not fix the first initialize after a config save.
- Never wipe on any branch mismatch (always checkout). That is safer long-term, but it would change existing "re-init with a new branch" behavior that currently expects a clean re-clone.

**Additional context**
Customer: Syntax Systems USA (SFDC 00851745). Workaround: disable Create Branch once the export branch exists on the remote, or set Branch = Base Branch.

Verified locally: `:plugins:git-plugin:test` for `GitExportPluginSpec` and `BaseGitPluginSpec`.

## Release Notes

SCM git-export no longer deletes its working directory on every initialize when Create Branch is enabled and the export branch differs from the base branch. Uncommitted job changes stay visible for export instead of reporting "No changes".

[RUN-4852]: https://pagerduty.atlassian.net/browse/RUN-4852?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/BaseGitPlugin.groovy
```diff
@@ -696,13 +723,14 @@ class BaseGitPlugin {
 
         try {
             createCommand.call()
+            git.checkout().setName(newBranch).call()
```
luismalamoc, line 726: The checkout runs before the push is validated. If the push is rejected, `createBranch` throws, but HEAD is already on `newBranch`. On the next `ScmLoader` tick, `workdirCheckedOutOn(base, branch)` returns `true`, the base-first clone is skipped, and `initialize` succeeds silently with a local branch that does not exist on the remote. The "configuration error" surfaces exactly once and then hides until the export push fails.

Suggest moving `git.checkout().setName(newBranch).call()` after the `failedUpdates` check. With HEAD still on `baseBranch`, the next tick takes the base-first path (`cloneOrCreate` with `branch = baseBranch` only fetches, no wipe), reaches `createBranch` again, and re-throws until the config or remote permissions are fixed. Two-line reorder.

plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/GitExportPlugin.groovy
```diff
@@ -162,6 +144,40 @@ class GitExportPlugin extends BaseGitPlugin implements ScmExportPlugin {
...
 
+    /**
+     * Opens or clones the export workdir.
+     * <p>
+     * When Create Branch is enabled, a <em>fresh</em> workdir is cloned from
+     * {@code baseBranch} first so a missing export branch can be created (RUN-4018).
+     * If the workdir is already checked out on the export branch, that base-first
+     * clone is skipped: {@code cloneOrCreate} treats a branch mismatch as a reason
+     * to delete and re-clone, which would wipe serialized-but-uncommitted job files
+     * on every ScmLoader initialize (RUN-4852).
```
luismalamoc, line 155: Please drop the `(RUN-4018)` and `(RUN-4852)` references from the Groovydoc. Ticket keys belong in PR and commit titles, not in source files. The original comment block did not carry them, and the explanation reads fine without them.

luismalamoc (changes requested):
Fix traced end to end: with `createBranch` and `branch != baseBranch`, the old `setup` wiped the workdir twice per `ScmLoader` tick (base-first clone saw the export branch and re-cloned, then the follow-up `cloneOrCreate` saw `baseBranch` and re-cloned again). `workdirCheckedOutOn` short-circuits that path, the fresh-workdir behavior from the base-branch-first change is unchanged, and the push status check mirrors `PushAction`. The regression test in `GitExportPluginSpec` reproduces the customer scenario exactly.

Two small changes requested inline:

1. `BaseGitPlugin.createBranch`: move the checkout after the push validation so a rejected push keeps failing on every tick instead of surfacing once and then passing silently.
2. `GitExportPlugin.cloneOrCreateForExport`: remove ticket keys from the Groovydoc.

Note, no action needed: changing the remote URL while `createBranch` is on and the export branch is missing on the new remote makes `cloneOrCreate` wipe and fail with "Failed cloning" for one tick; the next tick sees no workdir and takes the fresh path, so it self-heals.

On the Copilot thread about "Non existent": that message is pre-existing text moved by this PR, used in three places and asserted verbatim in `GitExportPluginSpec`. Fine to resolve as out of scope.

plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/BaseGitPlugin.groovy
Jesus-Osuna-M: Good catch, confirmed and fixed in c8fff0a.

I traced the path you describe and it reproduces exactly: with the checkout ahead of the push validation, tick 1 throws but leaves HEAD on `dev2`, so on tick 2 `workdirCheckedOutOn(base, 'dev2')` returns `true`, the base-first clone is skipped, `cloneOrCreate` finds a matching branch and only fetches, `getFullBranch()` matches so the "branch differs" block never runs, and `initialize` completes against a branch that exists only locally.

Changes:
- `git.checkout().setName(newBranch).call()` now runs after the `failedUpdates` check.
- Put it in its own try/catch rather than leaving it bare — previously a checkout failure was reported as "Failed creating branch", which was misleading.

One detail worth recording, since it is what makes the reorder safe on the retry path: `branchCreate` already sets `.setForce(true)`, so when tick 2 re-enters `createBranch` with the local `dev2` left over from tick 1, it resets to `origin/master` instead of throwing `RefAlreadyExistsException`. The error therefore stays `Failed push to remote` on every tick rather than mutating into a confusing "branch already exists".

Added a regression test for it, `BaseGitPluginSpec > createBranch leaves HEAD on the base branch when the push fails`, since this case had no coverage. I verified it is not vacuous: with the original ordering restored it fails, and with the fix it passes.

Test run on the fix:
- `BaseGitPluginSpec` + `GitExportPluginSpec`: 150/150, 0 skipped
- Full `:plugins:git-plugin:test`: 252/252, 0 failures, 0 skipped

The success path is unchanged: HEAD is still on the export branch before the follow-up `cloneOrCreate`, so the workdir is not wiped, and `re initialize with createBranch keeps uncommitted files when already on export branch` still passes.

plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/GitExportPlugin.groovy
Jesus-Osuna-M: Done in c8fff0a — both ticket keys are out of the `cloneOrCreateForExport` Groovydoc, and the explanation reads the same without them.

luismalamoc (approved):
Both requested changes are in c8fff0a: the checkout now runs after the push validation with its own error message, and the ticket keys are out of the Groovydoc. The new `createBranch leaves HEAD on the base branch when the push fails` test covers the retry path and is not vacuous. Approving.

PR #10583, opened 2026-09-11
[RUN-4893] Fix SCM import job state races and stale cluster state

Jesus-Osuna-M:
## Tell us about your PR

This is a bugfix for [RUN-4893](https://pagerduty.atlassian.net/browse/RUN-4893). It covers two defects found in SCM Git Import while running a two-node cluster:

1. SCM Git Import can return HTTP 500 when the background cluster loader mutates `jobStateMap` while request handling iterates it.
2. A node can keep import state for jobs another node already deleted. That node lists those jobs as pending imports even though they exist neither in the database nor in the remote, and a later import of the same file can fail with `A plugin returned an error: Component: No row with the given identifier exists: [rundeck.ScheduledExecution#<id>]`.

*Concurrent iteration*

Added a synchronized snapshot helper that preserves `jobStateMap` iteration order. The four import read paths now iterate stable snapshots instead of the live synchronized-map wrapper.

*Cluster state consistency*

Job change events only reach listeners in the JVM that handled the change, so each node is expected to reconcile its SCM import cache from the shared database. Two things prevented that from working:

- Reconciliation happened only inside `clusterFixJobs`, after metadata and Git work, and was skipped entirely when the project had no jobs left.
- The job lists it relied on were read through Hibernate's query cache. That cache is local to each JVM and is not invalidated by another node's writes, so a node could read rows that another node had deleted.

Changes:

- Added a local-only `reconcileJobState` hook to `ScmImportPlugin`, defaulting to a no-op, and implemented it in `GitImportPlugin` by pruning `jobStateMap` and `ImportTracker`. It performs no remote SCM operations.
- The import loader reconciles right after reading the job list, before metadata lookup, fetch, pull and status refresh, so a stale node is cleaned even if a later step fails.
- Import status requests reconcile as well, including projects whose job list is empty, while Git fetch and pull stay loader-owned so status polling adds no network work.
- The authoritative project job list, job lookups by UUID (`ScheduledExecution.getByIdOrUUID`) and `PluginMeta` reads bypass the query cache. The UUID lookup is what `ScheduleComponent` uses when persisting job components, which is where the import failure above surfaced.

- Replacing the map with `ConcurrentHashMap` was rejected because it changes ordering and weakens iteration consistency for both import and export.
- Holding the map monitor during the repository tree walk was rejected because it would block loader writes across Git I/O.
- Relaying job change events over cluster messaging was rejected. Messages can be lost or delayed and cannot restore state after a restart, so database-driven reconciliation remains the correctness mechanism.
- Disabling Hibernate's query cache globally, or adding distributed cache invalidation, would address this class of problem beyond SCM. Both are broader changes that need their own measurement and are intentionally outside this PR.

**Additional context**

RUN-4894 tracks the separate loader performance issue that increases exposure to the race. That optimization is intentionally outside this PR.

### Evidence collected before the fix

Two nodes on ports 4440 and 4441, cluster mode enabled, sharing one MySQL database and one Git remote.

Stale import state, with the database holding 9 jobs and the remote holding 10 files, so exactly one file should have been listed:

| Node | Items listed for import | Stale entries |
|---|---|---|
| 4440 | 210 | 200 |
| 4441 | 310 | 300 |

Project status was correct on both nodes (`1 unimported file(s) found`), while the action list carried entire batches of jobs that no longer existed anywhere. The stale entries differed per node, each node holding the batch it had not deleted itself, which is the signature of a per-JVM cache that another node's writes never invalidate.

The same cache produced a hard failure on the action input endpoint:

```
ERROR errors.GrailsExceptionResolver
  ObjectNotFoundException occurred when processing request:
  [GET] /api/60/project/p1/scm/import/action/import-jobs/input
No row with the given identifier exists: [rundeck.ScheduledExecution#1528]
    at org.hibernate.cache.internal.QueryResultsCacheImpl.get
    at org.hibernate.loader.Loader.getResultFromQueryCache
    at org.hibernate.loader.Loader.listUsingQueryCache
    at org.hibernate.loader.criteria.CriteriaLoader.list
```

And on the job component persist path, reached while importing a file from the node that had not performed the delete:

```
WARN services.ScheduledExecutionService - Job Component  persist error:
No row with the given identifier exists: [rundeck.ScheduledExecution#2030]
    at rundeck.services.ScheduledExecutionService.getByIDorUUID
    at rundeckpro.schedules.components.ScheduleComponent.persist
    at org.rundeck.app.components.RundeckJobDefinitionManager.persistComponents
    at rundeck.services.ScheduledExecutionService.saveComponents
```

### Cluster validation after the fix

Round 1, cross-node re-import, the case that previously failed:

| Step | Result |
|---|---|
| Import 10 files on 4440 | Database 10, both nodes `CLEAN` with 10 items |
| Delete one job on 4440 (id 2066) | Database 9 |
| Read 4441 | `1 unimported file(s) found`, 9 `CLEAN` plus 1 `IMPORT_NEEDED`, no stale entries |
| Import that file from 4441 | Succeeded, job recreated with id 2074 |

The identifier changing from 2066 to 2074 is the exact condition that previously triggered the component persist failure above.

Round 2, full lifecycle with deletions driven from 4440:

| Step | 4440 | 4441 |
|---|---|---|
| Import 10 | `CLEAN`, 10 | `CLEAN`, 10 |
| Delete the files from the remote | `DELETE_NEEDED`, 10 | `DELETE_NEEDED`, 10 |
| Delete the 10 jobs | `CLEAN`, 0 items | `CLEAN`, 0 items |
| Add one new file | 1 to import, 0 stale | 1 to import, 0 stale |

Round 3, same lifecycle with the roles reversed, importing and deleting from 4441:

| Step | 4440 | 4441 |
|---|---|---|
| Import 11 from 4441 | `CLEAN`, 11 | `CLEAN`, 11 |
| Delete one job on 4441 (id 2098), re-import from 4440 | Listed exactly 1, import succeeded as id 2106 | — |
| Delete the files from the remote | `DELETE_NEEDED`, 11 | `DELETE_NEEDED`, 11 |
| Delete the 11 jobs from 4441 | `CLEAN`, 0 items | `CLEAN`, 0 items |

No `ObjectNotFoundException` and no component persist errors appeared in either node's log during these rounds.

Temporary instrumentation was added during the investigation and removed before this commit. It confirmed that reconciliation is invoked from both the loader and the status request, that it dispatches to the plugin implementation rather than the interface default, and that a single plugin instance serves both the loader and web requests on each node.

### Automated verification

- `./gradlew :plugins:git-plugin:test` and `./gradlew :rundeckapp:test` for the affected specs: 667 tests, no failures, covering `GitImportPluginSpec`, `ImportTrackerSpec`, `ScmControllerSpec`, `ScmServiceSpec`, `ScmLoaderServiceSpec`, `JobMetadataServiceSpec` and `ScheduledExecutionServiceSpec`
- `./gradlew build -x check`
- `./gradlew codenarcComplexity`, no new violations

## Release Notes

Fixed intermittent HTTP 500 responses from SCM Git Import when import status is read concurrently with the cluster SCM loader. Fixed SCM Git Import listing jobs that were already deleted on another cluster node, and the import failure that followed when re-importing those files.

[RUN-4893]: https://pagerduty.atlassian.net/browse/RUN-4893?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/GitImportPlugin.groovy
```diff
@@ -244,7 +244,7 @@ class GitImportPlugin extends BaseGitPlugin implements ScmImportPlugin {
             }
         }
 
-        Map jobsCache = jobStateMap.collectEntries {key, value -> [value.path, value]}
+        Map jobsCache = snapshotJobStateMap().collectEntries {key, value -> [value.path, value]}
```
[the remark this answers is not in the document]
Jesus-Osuna-M: Good catch. Addressed in 323e8ef162: `getStatusInternal` now retains the job-state snapshot used to build `jobsCache`, resolves the tracked job ID once, and reads/reuses the state from that snapshot instead of performing two live `jobStateMap.get` calls. I also added a regression test that simulates the entry disappearing between live reads; it fails with the previous implementation and passes with the snapshot reuse. The full git-plugin suite (247 tests) and `./gradlew build -x check` pass.

rundeckapp/grails-app/services/rundeck/services/ScmService.groovy
```diff
@@ -1605,10 +1605,31 @@ class ScmService {
...
+     * Reconcile import state and perform the full cluster fix when jobs exist.
+     *
+     * @param auth authorization context
+     * @param project project name
+     * @param jobs authoritative current project jobs
+     */
+    void fixImportStatus(UserAndRolesAuthContext auth, String project, List<ScheduledExecution> jobs){
+        def plugin = getLoadedImportPluginFor project
+        if(plugin) {
+            plugin.reconcileJobState((jobs ?: [])*.extid.findAll { it } as Set<String>)
```
luismalamoc: **Blocker: `jobs` is not authoritative on this path, so this reconcile evicts every other job in the project.**

`reconcileJobState` removes from `jobStateMap` and `ImportTracker` every job ID that is *not* in the set it receives. `fixImportStatus` is reached through `importStatusForJobs(..., runClusterFix = true)`, and several callers pass a subset of the project's jobs:

- `ScmController.groovy:2104` (single job status, `[scheduledExecution]`)
- `ScmController.groovy:2595` (diff page, `[job]`)
- `ScheduledExecutionService.groovy:5349` (job show dropdown, `[scheduledExecution]`)
- `JobScmMetadataComponent.groovy:126` (job browse, `importStatusForJobIds(project, authContext, ids, true, ...)` with only the IDs on the current page)

In cluster mode, every job page view or job list render will therefore wipe the node-local import state and tracking for all jobs not in that subset. Project status flips to "N files need import" (paths become `trackedItemIsUnknown`) and `getJobStatus` for the other jobs falls back to `refreshJobStatus`, which is the per-job git history walk this ticket is trying to avoid. The loader repopulates on its next cycle, so the state flaps continuously under normal UI traffic.

Suggested fix: reconcile only where the job list is known to be complete. That is already the case in `ScmLoaderService.groovy:506` and in `ScmController.groovy:993` (`reconcileImportJobState` with `listJobsForProjectUncached`). Remove the `reconcileJobState` call here and the one inside `GitImportPlugin.clusterFixJobs`, and restore the `jobs && jobs.size() > 0` guard in `importStatusForJobs` so an empty subset does not trigger the cluster fix at all.

rundeckapp/grails-app/services/rundeck/services/ScmService.groovy
```diff
@@ -1366,7 +1366,7 @@ class ScmService {
     Map<String, JobImportState> importStatusForJobs(String project, UserAndRolesAuthContext auth, List<ScheduledExecution> jobs,  boolean runClusterFix = true, Map<String, Map> jobsPluginMeta = null) {
         def status = [:]
         def clusterMode = frameworkService.isClusterModeEnabled()
-        if(jobs && jobs.size()>0 && clusterMode && runClusterFix ){
+        if(clusterMode && runClusterFix ){
```
luismalamoc: Please restore the `jobs && jobs.size() > 0` guard. With the reconcile moved out of `fixImportStatus` (see the comment below), an empty list has nothing to do here, and keeping the guard means callers that pass a subset never pay for a remote fetch or pull on a status read.

plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/GitImportPlugin.groovy
```diff
@@ -764,8 +768,27 @@ class GitImportPlugin extends BaseGitPlugin implements ScmImportPlugin {
...
+        cachedJobIds.removeAll(authoritativeJobIds)
+        cachedJobIds.each { String jobId ->
+            jobStateMap.remove(jobId)
+            importTracker.untrackJob(jobId)
+        }
+    }
+
 
     Map clusterFixJobs(ScmOperationContext context, List<JobScmReference> jobs, Map<String,String> originalPaths){
+        reconcileJobState(jobs.collect { it.id }.findAll { it } as Set<String>)
```
luismalamoc: Same problem as in `ScmService.fixImportStatus`: `jobs` here is whatever the caller passed, and `fixImportStatus` passes single-job lists from the job show page, diff page and job browse. This line then evicts the state of every other job in the project.

It is also redundant on the loader path, which already calls `reconcileJobState` with the full uncached list at `ScmLoaderService.groovy:506` right before calling `clusterFixJobs`. Please drop this call and let the two authoritative call sites (loader and `apiProjectStatus`) own reconciliation.

rundeckapp/src/test/groovy/rundeck/services/ScmServiceSpec.groovy
```diff
@@ -894,6 +895,41 @@ class ScmServiceSpec extends Specification implements ServiceUnitTest<ScmService
         1 * plugin.getJobStatus(_,_)>> Mock(JobState)
     }
 
+    def "fix import status reconciles an empty project without remote cluster fix"() {
```
luismalamoc: This test encodes the behavior that causes the eviction (reconcile with whatever list `fixImportStatus` receives). Once the reconcile moves out of `fixImportStatus`, this test should go, and a replacement that guards the regression would be: `importStatusForJobs(project, auth, [oneJob], true)` in cluster mode must call `0 * plugin.reconcileJobState(_)`.

plugins/git-plugin/src/test/groovy/org/rundeck/plugin/scm/git/GitImportPluginSpec.groovy
```diff
@@ -294,6 +295,157 @@ class GitImportPluginSpec extends Specification {
...
+        when:
+        plugin.jobChanged(event, job)
+
+        then:
+        !plugin.jobStateMap.containsKey('123')
+        plugin.importTracker.trackedPaths().empty
+        plugin.importTracker.trackedPath('123') == null
+    }
+
+    def "cluster reconciliation removes state for jobs deleted on another node"() {
```
luismalamoc: If the reconcile call is removed from `clusterFixJobs`, this test no longer applies as written. The scenario is already covered by "local reconciliation removes stale jobs and preserves current jobs" below, so this one can be deleted rather than rewritten.

plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/ImportTracker.groovy
```diff
@@ -75,9 +75,55 @@ class ImportTracker {
...
+        trackedPathsMap.remove(jobId)
+        path
+    }
+
+    /**
+     * Return a stable snapshot of all tracked job IDs.
+     *
+     * @return tracked job IDs
+     */
+    Set<String> trackedJobIdSet() {
```
luismalamoc, line 134: Suggestion, same class of bug as the ticket: `trackedPaths()` at line 34 does `trackedCommits.keySet() + trackedJobIds.keySet()`, which iterates two live `synchronizedMap` views without holding their monitors. `getStatusInternal` calls it twice (`GitImportPlugin.groovy:233` and `:237`) on the request path while the loader's `refreshJobStatus` calls `trackJobAtPath`, so the same `ConcurrentModificationException` and HTTP 500 can still surface from here. Wrapping each `keySet()` copy in `synchronized (map)`, as this new method already does, closes it in four lines.

rundeckapp/grails-app/controllers/rundeck/controllers/ScmController.groovy
```diff
@@ -989,6 +989,11 @@ Since: v15''',
                 def jobsPluginMeta = scmService.getJobsPluginMeta(params.project, isExport)
                 //relaod all jobs to get project status
                 scmService.exportStatusForJobs(params.project, authContext, jobs.schedlist, true, jobsPluginMeta)
+            } else {
+                scmService.reconcileImportJobState(
+                    params.project,
+                    scheduledExecutionService.listJobsForProjectUncached(params.project)
```
luismalamoc, line 995: This is the right place to reconcile (the list is complete and uncached). Suggestion: reconcile only needs UUIDs, but this hydrates every `ScheduledExecution` in the project on each status poll, and the UI polls this endpoint. A projection avoids that, for example:

```groovy
ScheduledExecution.executeQuery(
    'select se.uuid from ScheduledExecution se where se.project = :project',
    [project: project],
    [cache: false]
)
```

exposed as `listJobUuidsForProjectUncached(project)` next to the existing helper, with `reconcileImportJobState` taking the `Set<String>` directly. The loader still needs full entities, so it keeps `listJobsForProjectUncached`.

plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/BaseGitPlugin.groovy
```diff
@@ -96,6 +96,17 @@ class BaseGitPlugin {
         this.commonConfig = commonConfig
     }
 
+    /**
+     * Creates a stable copy of the job state map for safe iteration.
+     *
+     * @return a snapshot preserving the map's iteration order
+     */
+    protected Map<String, Map> snapshotJobStateMap() {
+        synchronized (jobStateMap) {
```
luismalamoc, line 105: Suggestion: `GitExportPlugin.getStatusInternal` still does `jobStateMap.find { ... SynchState.LOADING }` on the live map (`GitExportPlugin.groovy:288`). Same race, and this helper is already available to it. One-line swap to `snapshotJobStateMap().find { ... }`.

luismalamoc (changes requested):
Thanks for the thorough write-up and the cluster validation, the evidence section made this easy to follow. The snapshot work for the `ConcurrentModificationException` is solid and the earlier Copilot finding about the second live `jobStateMap.get` is confirmed fixed with a good regression test.

Requesting changes for one functional regression introduced by the cluster-reconciliation half of the PR:

**Blocker: `reconcileJobState` is invoked with partial job lists.** `ScmService.fixImportStatus` and `GitImportPlugin.clusterFixJobs` both reconcile against whatever `jobs` they receive, and `importStatusForJobs(..., runClusterFix = true)` is called with single-job or per-page subsets from the job show page, the diff page, the job action dropdown and the job browse metadata component. In cluster mode each of those requests will evict the import state and tracking of every other job in the project, flipping the project status and forcing per-job git history recomputation until the loader's next cycle. Details and the suggested fix are inline; in short, keep reconciliation in the two places where the list is complete and uncached (`ScmLoaderService` and `apiProjectStatus`) and restore the empty-list guard in `importStatusForJobs`.

Non-blocking suggestions inline: `ImportTracker.trackedPaths()` still iterates live synchronized maps on the same request path, `GitExportPlugin` still does a live `find` on `jobStateMap`, and the status endpoint could reconcile from a UUID projection instead of hydrating every job per poll.

Copilot only reviewed the first three files at commit `4b984b5`; the cluster commits were never re-reviewed. Worth re-requesting once the blocker is addressed.

Jesus-Osuna-M:
Thanks for the careful review — the blocker is real and I reproduced it. Pushed in 3f16b305ad.

## Blocker: `reconcileJobState` with partial job lists

Confirmed, and worse than partial eviction in one case. With an empty list, `cachedJobIds.removeAll(emptySet)` removes nothing from the removal set, so the whole project's `jobStateMap` and `ImportTracker` get wiped. That is reachable: `importStatusForJobIds` filters unresolvable UUIDs with `.grep { it }`, so on a node holding stale state the list can collapse to empty, and its only caller is `JobScmMetadataComponent` with `runClusterFix = true`.

Applied your suggested fix:

- Removed the reconcile from `ScmService.fixImportStatus` and from `GitImportPlugin.clusterFixJobs`. Both removals are needed — leaving it in `clusterFixJobs` would reintroduce the eviction through `fixImportStatus` even with the service-level call gone.
- Restored the `jobs && jobs.size() > 0` guard in `importStatusForJobs`.
- Reconciliation now lives only at the two authoritative call sites, `ScmLoaderService` and `apiProjectStatus` via `reconcileImportJobState`.
- Documented the complete-list precondition on `ScmImportPlugin.reconcileJobState`, since the contract is what makes a subset call a bug rather than an optimization.

Tests: dropped both tests you flagged and added the two guards you proposed — `0 * plugin.reconcileJobState(_)` for a single-job list in cluster mode, plus one for the restored empty-list guard.

### One correction on the guard's rationale

You wrote that keeping the guard means "callers that pass a subset never pay for a remote fetch or pull on a status read". That is not what it does: a one-job list is non-empty, passes the guard, and still reaches `clusterFixJobs` → `fetchFromRemote` plus a possible `gitPull`. The guard only filters the empty case. Subsets pay the same remote I/O they paid before this PR. Making per-job status reads skip remote I/O would be a separate change, and arguably a design question, since `clusterFixJobs` exists to sync the node.

## Non-blocking suggestions

Applied both thread-safety ones:

- `ImportTracker.trackedPaths()` now copies each `keySet()` under its own monitor, using `LinkedHashSet` to preserve the previous iteration order.
- `GitExportPlugin.getStatusInternal` uses `snapshotJobStateMap()`.

Not applied: the UUID projection for `apiProjectStatus`. It is a real improvement, but it adds a service method and changes `reconcileImportJobState`'s signature, and I would rather not widen the blocker fix. Happy to add it here if you prefer, or file it as a follow-up.

## Added: same root cause on the export path

While validating the above on a two-node cluster I hit this on the export side, and it is the same query-cache mechanism as this ticket's second defect:

```
org.hibernate.ObjectNotFoundException: No row with the given identifier exists: [rundeck.ScheduledExecution#2134]
  at org.hibernate.cache.internal.QueryResultsCacheImpl.get
  at org.hibernate.loader.Loader.listUsingQueryCache
  at rundeck.controllers.ScmController.getViewExportActionItems(ScmController.groovy:1291)
  at rundeck.controllers.ScmController.apiProjectActionInput(ScmController.groovy:1204)
```

`getViewExportActionItems` and `performScmAction` read the project job list with `ScheduledExecution.findAllByProject(project)`, through the query cache. With jobs being deleted on the other node, **44 of 180** requests to `/scm/export/action/project-commit/input` returned HTTP 500 — and all 44 landed on the node that had *not* performed the deletes, which is the per-JVM stale-cache signature.

Both call sites now use `listJobsForProjectUncached`. Same scenario after the change: **160 of 160 HTTP 200**, zero `ObjectNotFoundException`. The export tests in `ScmControllerSpec` mock the uncached helper, so a return to the cached read fails on a missing invocation.

I pulled this into the PR rather than splitting it because it is two lines, reuses the helper this PR already adds, and you had already flagged the export path as carrying the same problems. Say the word if you would rather have it separate.

## Verification

- `ScmServiceSpec` 75, `ScmControllerSpec` 25, `ScheduledExecutionServiceSpec` 507, `ScmLoaderServiceSpec` 29, `JobMetadataServiceSpec` 2 — 638 tests, no failures
- `:plugins:git-plugin:test` — 249 tests, no failures
- Two-node cluster, cluster mode enabled, shared MySQL and one Git remote:
  - All four partial-list triggers (`apiJobStatus`, `jobs/browse?meta=scmImport`, job show page, diff page) leave project status and item state unchanged
  - Cross-node delete: the node that did not delete reports `1 unimported file(s) found`, 9 `CLEAN`/tracked and 1 `IMPORT_NEEDED`/untracked, no stale entries; re-importing that file from it succeeds and recreates the job
  - Empty project (0 jobs, 0 remote files): both nodes `CLEAN` with 0 items
  - Import status and action input under load: 116 of 116 HTTP 200, no `ConcurrentModificationException`

rundeckapp/grails-app/services/rundeck/services/ScmService.groovy
Jesus-Osuna-M: Fixed in 3f16b305ad. Removed the reconcile from `fixImportStatus`, and restored the `jobs && jobs.size() > 0` guard in `importStatusForJobs`. Reconciliation now happens only at `ScmLoaderService` and `apiProjectStatus`, both of which hold the complete uncached list.

Worth adding that the empty-list case was worse than partial eviction: `cachedJobIds.removeAll(emptySet)` removes nothing from the removal set, so the entire project's `jobStateMap` and `ImportTracker` were wiped. That is reachable, since `importStatusForJobIds` drops unresolvable UUIDs with `.grep { it }` and its only caller passes `runClusterFix = true`.

rundeckapp/grails-app/services/rundeck/services/ScmService.groovy
Jesus-Osuna-M: Restored in 3f16b305ad.

One correction on the rationale: the guard does not stop subsets from paying for a remote fetch or pull. A one-job list is non-empty, passes the guard, and still reaches `clusterFixJobs` → `fetchFromRemote` plus a possible `gitPull`. It only filters the empty case, and that remote I/O is unchanged from before this PR. Making per-job status reads skip remote work would be a separate change.

plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/GitImportPlugin.groovy
Jesus-Osuna-M: Removed in 3f16b305ad. Agreed on both counts, harmful and redundant.

It also had to go together with the service-level call rather than instead of it: leaving this one in place would have reintroduced the same eviction through `fixImportStatus`, since that still calls `clusterFixJobs` with whatever subset it received.

rundeckapp/src/test/groovy/rundeck/services/ScmServiceSpec.groovy
Jesus-Osuna-M: Done in 3f16b305ad. Replaced with the guard you proposed plus one for the restored empty-list case:

- `importStatusForJobs(project, auth, [oneJob], true)` in cluster mode → `0 * plugin.reconcileJobState(_)`
- `importStatusForJobs(project, auth, [], true)` → `0 * plugin.reconcileJobState(_)` and `0 * plugin.clusterFixJobs(_, _, _)`

plugins/git-plugin/src/test/groovy/org/rundeck/plugin/scm/git/GitImportPluginSpec.groovy
Jesus-Osuna-M: Deleted in 3f16b305ad. It was asserting the failure mode itself: `clusterFixJobs(context, [], [:])` followed by `plugin.jobStateMap.isEmpty()`. The scenario stays covered by "local reconciliation removes stale jobs and preserves current jobs".

plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/ImportTracker.groovy
Jesus-Osuna-M: Applied in 3f16b305ad. Each `keySet()` is now copied while holding its own map monitor.

One deliberate difference from `trackedJobIdSet()`: I used `LinkedHashSet` instead of `HashSet`, to preserve the iteration order that the previous `Set` union produced, since the result feeds the tree-walk comparison in `getStatusInternal`.

rundeckapp/grails-app/controllers/rundeck/controllers/ScmController.groovy
Jesus-Osuna-M: Not applied, and I would like your call on this one. It is a real improvement, but it adds a service method and changes `reconcileImportJobState`'s signature, and I would rather not widen the blocker fix.

plugins/git-plugin/src/main/groovy/org/rundeck/plugin/scm/git/BaseGitPlugin.groovy
Jesus-Osuna-M: Applied in 3f16b305ad — `GitExportPlugin.getStatusInternal` now uses `snapshotJobStateMap()`.

Validating that path on a two-node cluster also surfaced a separate defect on the export action input endpoint: `getViewExportActionItems` reads the project job list through Hibernate's query cache, so with jobs being deleted on the other node 44 of 180 requests returned HTTP 500 with `ObjectNotFoundException`. Same root cause as this ticket's second defect. Fixed in the same commit and detailed in the PR comment.

luismalamoc (approved):
Re-reviewed at `3f16b305`. Everything requested is in and verified against the branch:

- `reconcileJobState` is now only reached from `ScmLoaderService` and `apiProjectStatus` via `reconcileImportJobState`, both with the complete uncached list. Removed from `fixImportStatus` and `clusterFixJobs`, and the interface Javadoc now states the complete-list precondition.
- Empty-list guard restored in `importStatusForJobs`. Good catch that the empty case was a full wipe, not just partial eviction.
- `ImportTracker.trackedPaths()` copies under each monitor, `GitExportPlugin.getStatusInternal` uses the snapshot helper.
- Tests: the two tests encoding the eviction are gone, replaced by the partial-list and empty-list guards asserting `0 * plugin.reconcileJobState(_)`.
- The export `findAllByProject` reads moved to `listJobsForProjectUncached` are two lines on a helper this PR already adds, backed by the 44/180 to 160/160 measurement. Fine to keep here.

You are right about the guard rationale: it only filters the empty case, subsets still pay the same remote fetch as before this PR. Thanks for the correction.

UUID projection for `apiProjectStatus`: agreed, follow-up rather than here.

Approving.

rundeckapp/grails-app/controllers/rundeck/controllers/ScmController.groovy
luismalamoc: Follow-up is the right call. Keep this PR scoped to the blocker fix; a separate ticket for the UUID projection is fine.

PR #10596, opened 2026-09-14
[RUN-4947] Fix rundeck-cli session auth broken behind forwarded-headers proxy

### LINKED ISSUE #10584: [Bug]: Local user auth broken in 6.2.1
### Rundeck Version

6.2.1

### Operating System

Dockerhub image

### Java Version

17.0.20

### Describe the Bug

Something seems to have broken with user auth between 6.1.0 and 6.2.1.

On 6.1.0, I can do operations with rundeck-cli, authenticated with $RD_USER and $RD_PASSWORD set, and $RD_URL=http://localhost:4440/

On 6.2.1, rundeck-cli authenticates OK but operations result in:
$ rd keys update -p keys/github/access-token -t password -f token
Error: (unauthenticated) is not authorized for: /api/29/storage/keys/github/access-token
[code: unauthorized; APIv59]
Authorization failed: 403 Forbidden

If I set the RD_PASSWORD to something wrong, I get org.rundeck.client.api.LoginFailed , so the password is accepted, but somehow the session becomes unauthenticated after login?

This doesn't seem to affect header-based auth from the reverse-proxy. Human users with browsers are still OK.

### Reproduction Steps

Use rundeck-cli with user/pass on 6.2.1, RD_URL set to localhost

### Expected Behavior

rundeck-cli works

### Logs / Stack Trace

```shell
[2026-09-09T13:55:35,684] INFO  security.RundeckPreauthenticationRequestHeaderFilter - Roles header X-Forwarded-Roles
[2026-09-09T13:55:35,684] INFO  security.RundeckPreauthenticationRequestHeaderFilter - Roles received null
[2026-09-09T13:55:35,684] INFO  security.RundeckPreauthenticationRequestHeaderFilter - User header X-Forwarded-Email
[2026-09-09T13:55:35,684] INFO  security.RundeckPreauthenticationRequestHeaderFilter - User / UUID recieved null
[2026-09-09T13:55:35,698] INFO  web.requests "GET /user/login" 127.0.0.1 http  form 1 ? [] (rd-cli-tool/2.2.0 (rundeck-cli/2.2.0))
[2026-09-09T13:55:35,720] INFO  security.RundeckPreauthenticationRequestHeaderFilter - Roles header X-Forwarded-Roles
[2026-09-09T13:55:35,720] INFO  security.RundeckPreauthenticationRequestHeaderFilter - Roles received null
[2026-09-09T13:55:35,720] INFO  security.RundeckPreauthenticationRequestHeaderFilter - User header X-Forwarded-Email
[2026-09-09T13:55:35,720] INFO  security.RundeckPreauthenticationRequestHeaderFilter - User / UUID recieved null
[2026-09-09T13:55:35,802] INFO  audit.AuditLoggerPlugin - Audit Event: AuditEvent {Timestamp=Wed Sep 09 13:55:35 GMT 2026, ActionType='login_success', UserInfo={username='gh', userRoles=[github]}, RequestInfo={serverHostname='rundeck', serverUUID='a14bc3e6-75e8-4fe4-a90d-a16dcc976bf6', sessionID='node01lf09lsowrtt61ks9mjpqhby6590', userAgent='rd-cli-tool/2.2.0 (rundeck-cli/2.2.0)'}, ResourceInfo={resourceType='user', resourceName='gh'}}
[2026-09-09T13:55:35,809] INFO  security.RundeckPreauthenticationRequestHeaderFilter - Roles header X-Forwarded-Roles
[2026-09-09T13:55:35,809] INFO  security.RundeckPreauthenticationRequestHeaderFilter - Roles received null
[2026-09-09T13:55:35,809] INFO  security.RundeckPreauthenticationRequestHeaderFilter - User header X-Forwarded-Email
[2026-09-09T13:55:35,809] INFO  security.RundeckPreauthenticationRequestHeaderFilter - User / UUID recieved null
[2026-09-09T13:55:35,815] INFO  web.requests "GET /user/login" 127.0.0.1 http  form 1 ? [] (rd-cli-tool/2.2.0 (rundeck-cli/2.2.0))
[2026-09-09T13:55:35,827] INFO  security.RundeckPreauthenticationRequestHeaderFilter - Roles header X-Forwarded-Roles
[2026-09-09T13:55:35,827] INFO  security.RundeckPreauthenticationRequestHeaderFilter - Roles received null
[2026-09-09T13:55:35,827] INFO  security.RundeckPreauthenticationRequestHeaderFilter - User header X-Forwarded-Email
[2026-09-09T13:55:35,827] INFO  security.RundeckPreauthenticationRequestHeaderFilter - User / UUID recieved null
[2026-09-09T13:55:35,832] ERROR interceptors.AuthorizationInterceptor - (unauthenticated) UNAUTHORIZED for storage/apiKeys
```

### Database

None

### Confirmation

- [x] I have verified this occurs on a version within Rundeck's [supported window](https://docs.rundeck.com/docs/history/release-calendar.html) (released within the last 12 months).

ncofreortiz-hub:
Jira: https://pagerduty.atlassian.net/browse/RUN-4947

## Tell us about your PR

Bugfix. Fixes #10584.

`rundeck-cli` authenticated with `RD_USER`/`RD_PASSWORD` (session/cookie auth) worked on 6.1.0 but broke on 6.2.1: login succeeds, but every subsequent API call (e.g. `rd keys update`) fails with `Error: (unauthenticated) is not authorized for: /api/29/storage/keys/...` / `403 Forbidden`. Header-based auth from a reverse proxy and browser sessions were unaffected.

Root cause: 6.2.x added logic in `JettyServletContainerCustomizer.applySecureSessionCookieDefault()` that auto-marks the session cookie `Secure` whenever `grails.serverURL` starts with `https://`. In reverse-proxy / header-auth deployments (TLS terminated upstream, `X-Forwarded-*` headers), `grails.serverURL` is the proxy's public HTTPS URL, not the scheme this Jetty listener itself actually serves — Rundeck's own port is typically still plain HTTP for direct/local access. A client hitting that port directly over HTTP (like `rd-cli` or a health check) receives a `Secure` cookie on login, which any spec-compliant HTTP client (browsers, OkHttp used by rundeck-cli) correctly refuses to resend over a non-TLS connection. Every following request then carries no session, so `SetUserInterceptor`/`AuthorizationInterceptor` treat it as anonymous → 403.

Reproduced end-to-end against `rundeck/rundeck:6.2.1` in Docker with `RUNDECK_GRAILS_URL=https://...` + `RUNDECK_SERVER_FORWARDED=true` (the same shape as header-auth/reverse-proxy deployments): login over plain HTTP returned `Set-Cookie: JSESSIONID=...; Secure`, and the next request was rejected exactly like the report (`(unauthenticated) is not authorized for: ...`).

`applySecureSessionCookieDefault()` now skips the auto-`Secure` default when `useForwardHeaders` is enabled (the same flag this class already uses for `X-Forwarded-*` support), since `grails.serverURL`'s scheme isn't a reliable signal for this listener's own scheme in that mode. Standalone (non-proxied) HTTPS deployments keep the existing auto-`Secure` behavior unchanged. Operators can still force the flag explicitly via `server.servlet.session.cookie.secure`.

Verified the fix resolves the issue by rebuilding the WAR with the patch and re-running the same Docker repro: cookie no longer carries `Secure`, and both `/api/.../system/info` and `/api/.../storage/keys/...` succeed (200 / expected 404 for a nonexistent key) instead of 403.

- Basing the decision on the Jetty factory's own SSL config instead of `grails.serverURL`: more "correct" in theory, but the factory isn't configured with SSL either in the TLS-terminated-upstream case, so it wouldn't change the outcome and would be a larger behavioral change.
- Requiring operators to always set `server.servlet.session.cookie.secure` explicitly: already possible today (the auto-default already backs off if it's set), but doesn't fix the regression for existing deployments that upgrade without noticing the new default.

**Additional context**

- New/failing-first test: `JettyServletContainerCustomizerSpec` — "applySecureSessionCookieDefault leaves cookie secure unset when serverUrl is https but this Jetty listener is behind a forwarded-headers proxy".
- No changes to default (non-proxied) behavior.

## Release Notes

Fixed a regression where Rundeck could mark the session cookie `Secure` even when this server's own listener is plain HTTP (deployments behind a TLS-terminating reverse proxy with forwarded headers enabled), which broke session/cookie-based authentication for clients like `rundeck-cli` connecting directly to that HTTP port.


Jesus-Osuna-M (approved):
Looks good — this is the right fix for #10584 / RUN-4947. Skipping the factory-level `Secure` default when `useForwardHeaders` is on restores rundeck-cli session auth over the listener's plain HTTP port, and standalone HTTPS is unchanged. Operators can still force `server.servlet.session.cookie.secure=true`.

One clarification worth adding to the release notes (the Docker repro's "cookie no longer carries Secure" is only true for that direct-HTTP path):

Behind a reverse proxy, the session cookie is still marked `Secure` when the incoming request is treated as HTTPS (`X-Forwarded-Proto: https`). Direct HTTP to Rundeck's own port (e.g. rundeck-cli) will not get `Secure`, which is required for session auth to work there. If forwarded headers are enabled but `X-Forwarded-Proto` is missing, browser cookies will also lack `Secure`; set `server.servlet.session.cookie.secure=true` in that case.

PR #10597, opened 2026-09-14
[RUN-4532] Remediate Hardcoded Default Encryption Password in Rundeck Package Installs

ncofreortiz-hub:
## Tell us about your PR

Security fix for [RUN-4532](https://pagerduty.atlassian.net/browse/RUN-4532) (source report [PS-1681](https://pagerduty.atlassian.net/browse/PS-1681)). Rundeck RPM/DEB package installs ship with a publicly known default key-storage encryption password (`default.encryption.password`). The postinstall scripts are meant to randomize it, but:
- Both scripts only used `openssl rand -hex 8` — 64 bits of entropy, below the 128-bit modern minimum.
- DEB additionally gated the whole randomization block on `if [ -f "$DIR/rundeck-config.properties" ]`, a conditional RPM doesn't have.

Docker and WAR installs are not affected (different config-injection mechanisms) and are out of scope.

- `packaging/lib/rpm/scripts/postinst.sh` and `packaging/lib/deb/scripts/postinst`: changed `openssl rand -hex 8` to `openssl rand -hex 16` (128 bits) in both.
- `packaging/lib/deb/scripts/postinst`: removed the `if [ -f ... ]` guard so randomization runs unconditionally, matching RPM. The `sed` substitutions are unchanged and still only match a line whose value is literally `default.encryption.password`, so re-running postinst on an already-randomized config on upgrade remains a no-op.
- `packaging/test/docker/installcommon/scripts/deb-tests.sh` and `rpm-tests.sh`: added a `test_encryption_password` check to the existing Docker-based install-test harness, asserting the installed config no longer contains `default.encryption.password` and that the substituted password is a 32-character hex string.

Full spec: `.claude/specs/RUN-4532.md`. Out of scope for this PR (per spec, confirmed with ticket owner): auditing/rotating secrets on already-deployed instances, provisioning/documentation guidance, and splitting the two storage converters into independently-generated passwords (both continue to share one generated value, only its entropy changed).

- Splitting the two storage converters (`rundeck.storage.converter.1` / `rundeck.config.storage.converter.1`) into independently-generated passwords was considered but scoped out — the ticket's required remediation actions only call for the entropy fix and DEB/RPM parity, not a password-count change.
- Keeping DEB's `if [ -f ... ]` guard and only bumping entropy was considered, but the ticket explicitly asks for DEB to be unconditional like RPM; dpkg guarantees the config file exists by the time `postinst` runs, so removing the guard doesn't change observed behavior on a normal install/upgrade.

**Additional context**

No changes to `packaging/lib/common/etc/rundeck/rundeck-config.properties`, RPM's `.rpmnew` upgrade-file resolution logic, or Docker/WAR install paths (confirmed unaffected during investigation).

## Release Notes

Increased the entropy of the auto-generated key-storage encryption password on RPM/DEB package installs from 64 to 128 bits, and made the DEB postinstall script randomize it unconditionally (matching RPM).

[RUN-4532]: https://pagerduty.atlassian.net/browse/RUN-4532?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PS-1681]: https://pagerduty.atlassian.net/browse/PS-1681?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.claude/plans/RUN-4532.md
Jesus-Osuna-M, line 1: do no commit this file please

.claude/specs/RUN-4532.md
Jesus-Osuna-M, line 1: do no commit this file please

Jesus-Osuna-M (approved):

PR #10602, opened 2026-09-15
[RUN-4476] Fix plugin upload security issues

carlosrfranco:
## Tell us about your PR

Bugfix. Closes four security issues in the plugin upload/install flow (`PluginController`, `PluginValidator`), reported via HackerOne and confirmed in an internal code review (RUN-4476):

1. `.groovy` plugin uploads skipped validation entirely (`PluginValidator.validate()` returned `true` unconditionally), allowing arbitrary Groovy code to be loaded and executed as the Rundeck service account.
2. Uploaded/installed filenames were used unsanitized to build filesystem paths (temp file and final `libextDir` copy), allowing path traversal outside the intended directories.
3. The ACL check authorized against the generic `system` resource with `[admin, ops_admin]` instead of the `plugin` resource kind, so `OpsAdmin` (a license-tier role documented as having no access to application data) could install plugins and get code execution — inverting the intended privilege ordering relative to `AppAdmin`.
4. `installPlugin()` treated a `pluginUrl` starting with `/` as a local `file:` URI, giving an authorized user a local-file-read/inclusion primitive.

- `PluginValidator`: removed the `.groovy` branch so `.groovy` uploads fall through to the existing failure path. `.jar`/`.zip` remain the supported, validated packaging formats.
- `PluginController`: added a `resolveSafePluginFile()` helper (reusing the zip-slip-guard idiom already used elsewhere in the codebase) applied at all three attacker-influenced path-construction sites: the upload temp file, the URL-install temp file, and the `libextDir` copy.
- `PluginController`: `uploadPlugin()`/`installPlugin()` now authorize `install` on `RESOURCE_TYPE_PLUGIN` instead of `[admin, ops_admin]` on `RESOURCE_TYPE_SYSTEM`, matching `ArtifactController`/`RepositoryController`.
- `PluginController#installPlugin()`: now requires the `pluginUrl` to parse with an `http`/`https` scheme, rejecting bare `/`-prefixed paths and any other scheme (including `file:`).
- Removed a UI hint that referenced Groovy plugin installation, now stale given fix 1.

- For fix 1, considered AST-parsing `.groovy` sources to check for a `@Plugin`-annotated class. Rejected: `core`'s Groovy dependency is `testImplementation`-only (adding it to main would be a disproportionate footprint increase), and an annotation-presence check is trivially spoofable, so it wouldn't provide real security value.
- For fix 3, considered keeping `[ACTION_ADMIN, ACTION_OPS_ADMIN]` in the OR-list alongside the new resource type (as `ArtifactController.authorized()` does). Rejected: tracing how `OpsAdmin`/`AppAdmin` deny-rules are compiled shows this would likely be a no-op for the exact bug being fixed, since nothing denies the generic `admin`/`ops_admin` action specifically on `plugin`-kind resources. The ticket's own recommendation is to authorize on `plugin` kind *instead of* the `system`/`admin` check, which is what's implemented.

**Additional context**
- `Features.PLUGIN_SECURITY`'s default is intentionally left unchanged (still opt-in/disabled by default) — flipping it would disable plugin upload/install out of the box, which is out of scope for this fix.
- Unit tests only (Spock + JUnit); no integration/E2E changes.
- Jira: RUN-4476

## Release Notes

Fixed several security issues in the plugin upload/install flow: `.groovy` plugin files are no longer accepted without validation, uploaded/installed filenames are sanitized against path traversal, plugin install/uninstall authorization now correctly checks the `plugin` resource kind, and `installPlugin` no longer accepts local `file:` URLs.


carlosrfranco:
Addressed the Copilot review:

1. **Stale \`jar/zip/groovy\` upload label** — updated \`plugin.form.pluginFile.label\` to \`jar/zip\` across all 5 locale bundles (en, de, fr, ja, nl), and removed the now-orphaned \`plugin.form.groovyNote\` entries (the GSP reference was already removed in the original commit).

2. **Nav menu ACL mismatch** — \`_sysConfigNavMenuJson.gsp\`'s \`pluginInstall\` check allowed `[install, admin, app_admin]` on the `plugin` kind, while `PluginController` only allows `install`. Narrowed the nav menu check to `install` only, so it no longer shows the "Upload Plugin" link to users who'd then get `Unauthorized`.

3. **Did not** add `admin`/`app_admin` back into `PluginController`'s authorization check as suggested. Traced the `AppAdmin` license-tier user class (`user-classes.json`) and confirmed it carries an explicit deny-rule for `install`/`uninstall` specifically on the `plugin` resource kind — but generic `admin`/`app_admin` actions on that kind aren't covered by that deny, and are commonly granted broadly by a site's own ACL policy. Adding them as a fallback here would let `AppAdmin` route around the specific restriction the license tier is designed to enforce, reopening the same class of bug this PR fixes for `OpsAdmin`. Fixed the actual UI-consistency issue at its source (item 2) instead.

rundeckapp/grails-app/controllers/rundeck/controllers/PluginController.groovy
```diff
@@ -698,6 +726,27 @@ Since: v49''',
...
+     * path separators or {@code ..} segments, or that would otherwise resolve outside
+     * {@code baseDir} once normalized.
+     *
+     * @param baseDir the directory the resolved file must live in
+     * @param name the untrusted plugin file name
+     * @return the resolved {@link File} within {@code baseDir}
+     * @throws IllegalArgumentException if {@code name} is blank or would escape {@code baseDir}
+     */
+    private File resolveSafePluginFile(File baseDir, String name) {
+        if (!name || name.contains('/') || name.contains('\\') || name.contains('..')) {
```
luismalamoc, line 740: **Blocker:** the name `.` still passes this check. It contains no `/`, `\` or `..`, and `new File(baseDir, ".")` normalizes to `baseDir` itself, so the `startsWith` guard below also passes. Then `tmpFile.delete()` fails silently (directory) and `tmpFile << file.inputStream` throws `FileNotFoundException (Is a directory)` with no catch, so both `uploadPlugin` (filename `.`) and `installPlugin` (`http://host/.`) return a 500 instead of `plugin.error.invalid.filename`.

Reproduced locally with plain Java: `rejectedByContains=false escapes=false isDir=true delete()=false` and the write throws.

This is the open Copilot thread; the helper is unchanged in the follow-up commit. Simplest fix: replace the normalize/startsWith block with `if (resolved.isDirectory()) throw new IllegalArgumentException(...)` (see next comment) and add `"."` to the `where:` block of the traversal spec.

rundeckapp/grails-app/controllers/rundeck/controllers/PluginController.groovy
```diff
@@ -698,6 +726,27 @@ Since: v49''',
...
+            throw new IllegalArgumentException("Invalid plugin file name: ${name}")
+        }
+        File resolved = new File(baseDir, name)
+        if (!resolved.toPath().normalize().startsWith(baseDir.toPath().normalize())) {
```
luismalamoc, line 744: **Suggestion:** this check is unreachable. Once the name is guaranteed to contain no `/`, `\` or `..`, `new File(baseDir, name)` can never resolve outside `baseDir`. Either drop it, or swap it for the check that actually adds value here and also closes the `.` case above:

```groovy
File resolved = new File(baseDir, name)
if (resolved.isDirectory()) {
    throw new IllegalArgumentException("Invalid plugin file name: ${name}")
}
return resolved
```

rundeckapp/grails-app/controllers/rundeck/controllers/PluginController.groovy
```diff
@@ -598,8 +598,8 @@ Since: v49''',
         AuthContext authContext = rundeckAuthContextProcessor.getAuthContextForSubject(session.subject)
         boolean authorized = rundeckAuthContextProcessor.authorizeApplicationResourceAny(
             authContext,
-            AuthConstants.RESOURCE_TYPE_SYSTEM,
-            [AuthConstants.ACTION_ADMIN,AuthConstants.ACTION_OPS_ADMIN]
+            AuthConstants.RESOURCE_TYPE_PLUGIN,
+            [AuthConstants.ACTION_INSTALL]
```
luismalamoc, line 602: **Suggestion (applies here and at line 650):** authorizing on `[ACTION_INSTALL]` alone diverges from two things that already exist:

- the authorization docs list `admin` as "Full access" on the `plugin` resource kind;
- `RepositoryController.authorized()` / `ArtifactController.authorized()` accept `[admin, ops_admin, install]` on the same kind, so a user with `admin` on `plugin` can install from the marketplace but not via upload.

Recommend `[AuthConstants.ACTION_INSTALL, AuthConstants.ACTION_ADMIN]` here, at line 650, and in `_sysConfigNavMenuJson.gsp`.

On the rationale given for rejecting the Copilot fallback suggestion: it holds for `app_admin` but not for `admin`. The `AppAdmin` user class carries `deny-actions: [admin, ops_admin]` with `system-types: ["*"]`, so the `admin` action is already denied on the `plugin` kind for that class. Adding `ACTION_ADMIN` cannot reopen the AppAdmin path; adding `ACTION_APP_ADMIN` would, so leave that one out.

Related, and the reason this PR does not fully close the `OpsAdmin` case on its own: that class has no deny rule for `install`/`uninstall` on the `plugin` kind (unlike `AppAdmin`). With a wildcard grant like the shipped `admin.aclpolicy` (`resource: allow: '*'`), an `OpsAdmin` user still passes `install` on `plugin` after this change, and still passes `ops_admin` on the marketplace path. The companion fix is a deny rule for `install`/`uninstall` on kind `plugin` in the `OpsAdmin` class definition, mirroring the existing `AppAdmin` one. Details in the review summary.

rundeckapp/grails-app/views/menu/_sysConfigNavMenuJson.gsp
```diff
@@ -27,7 +27,7 @@
 <g:set var="pluginInstall" value="${auth.resourceAllowedTest(
         type: AuthConstants.TYPE_RESOURCE,
         kind: AuthConstants.TYPE_PLUGIN,
-        action: [AuthConstants.ACTION_INSTALL, AuthConstants.ACTION_ADMIN, AuthConstants.ACTION_APP_ADMIN],
+        action: [AuthConstants.ACTION_INSTALL],
```
luismalamoc, line 30: **Suggestion:** keep this list identical to whatever `PluginController` ends up checking. If the controller adopts `[ACTION_INSTALL, ACTION_ADMIN]`, mirror it here so the menu link and the endpoint agree. Note `grails-repository/grails-app/views/artifact/index.gsp` still gates `repocaninstall` on `[install, admin, ops_admin]`; out of scope for this PR, but worth a follow-up so the two install UIs stop disagreeing.

core/src/test/java/com/dtolabs/rundeck/core/plugins/PluginValidatorTest.java
```diff
@@ -0,0 +1,63 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.dtolabs.rundeck.core.plugins;
+
+import java.io.File;
+import java.io.FileWriter;
+
+import junit.framework.Assert;
```
luismalamoc, line 21: **Blocker (convention):** new tests should be Spock, not JUnit, and `junit.framework.Assert` is the JUnit 3 API. `core` already has Spock wired up and there is a `PluginValidationSpec.groovy` in this same package. Two `def "..."` features there (`.groovy` rejected, unknown extension rejected) cover this and let us delete this file.

rundeckapp/src/test/groovy/rundeck/controllers/PluginControllerSpec.groovy
```diff
@@ -485,13 +488,53 @@ class PluginControllerSpec extends Specification implements ControllerUnitTest<P
...
+        request.addFile(new GrailsMockMultipartFile("pluginFile",maliciousName,"application/octet-stream",pluginInputStream))
+        controller.uploadPlugin()
+
+        then:
+        1 * controller.rundeckAuthContextProcessor.getAuthContextForSubject(_)
+        1 * controller.featureService.featurePresent(_) >> false
+        1 * controller.rundeckAuthContextProcessor.authorizeApplicationResourceAny(_, AuthConstants.RESOURCE_TYPE_PLUGIN, [AuthConstants.ACTION_INSTALL]) >> true
+        response.text == '{"err":"Invalid plugin file name"}'
+        !new File(uploadTestTargetDir, "evil.zip").exists()
+        !new File(uploadTestBaseDir.parentFile, "evil.zip").exists()
```
luismalamoc, line 528: **Suggestion:** this assertion is always true and does not check what it implies. `RELATIVE_PLUGIN_UPLOAD_DIR` is `var/tmp/pluginUpload`, so `../evil.zip` would have landed in `<baseDir>/var/tmp/`, and `../../../../tmp/evil.zip` in `<baseDir>/../tmp/`, never directly in `uploadTestBaseDir.parentFile`. The `response.text` assertion above already proves the request was rejected before any write; drop this line (and the `uploadTestTargetDir` one is fine to keep).

core/src/main/java/com/dtolabs/rundeck/core/plugins/PluginValidator.java
```diff
@@ -35,8 +35,6 @@ public static boolean validate(File pluginFile) {
             } catch (Exception iex) {
                 log.error("Error loading plugin.", iex);
             }
-        } else if(pluginFile.getName().endsWith(".groovy")) {
-            return true;
         } else {
             log.error("File: ${pluginFile.getName()} is not a valid Rundeck plugin.");
```
luismalamoc, line 39: **Suggestion:** pre-existing, but every rejected `.groovy` upload now lands in this branch. `${pluginFile.getName()}` is not interpolated in a Java string, so the log line prints the literal placeholder. One-liner while the file is open:

```java
log.error("File: {} is not a valid Rundeck plugin.", pluginFile.getName());
```

rundeckapp/grails-app/controllers/rundeck/controllers/PluginController.groovy
```diff
@@ -611,7 +611,13 @@ Since: v49''',
         }
         def file = request.getFile('pluginFile')
         ensureUploadLocation()
-        File tmpFile = new File(frameworkService.getRundeckFramework().baseDir,RELATIVE_PLUGIN_UPLOAD_DIR+"/"+file.originalFilename)
+        File tmpFile
+        try {
```
luismalamoc, line 615: **Optional:** this `try/catch IllegalArgumentException` block is repeated three times (here, `installPlugin`, `validateAndCopyPlugin`). If the helper returned `null` for an invalid name instead of throwing, each call site collapses to `if (!tmpFile) { renderErrorCodeAsJson(...); return }`. Not blocking; take it or leave it.

luismalamoc (changes requested):
## Review summary — NEEDS_WORK

The four fixes are the right shape (drop `.groovy` uploads, sanitize names at all three sites, rescope the ACL to the `plugin` kind, reject non-http(s) URLs) and the diff is focused. Three things need to land before this merges; the rest are suggestions. Inline comments carry the details.

### Blockers

1. **`.` bypasses `resolveSafePluginFile` and produces a 500** (`PluginController.groovy:740`). Verified locally. This is the still-open Copilot thread; the helper was not touched in the follow-up commit. Fix is one line (`resolved.isDirectory()`), plus `"."` in the traversal spec's `where:` block.
2. **The `OpsAdmin` case is not fully closed by the OSS change alone.** The `OpsAdmin` user class has no deny rule for `install`/`uninstall` on the `plugin` kind (the `AppAdmin` class does). With a wildcard grant such as the shipped `admin.aclpolicy` (`resource: allow: '*'`), an `OpsAdmin` user still passes the new `install`-on-`plugin` check, and still passes `ops_admin` on the marketplace path. The rescope here only closes configurations that grant `ops_admin` narrowly on `system`. The companion change is a deny rule for `install`/`uninstall` on kind `plugin` in the `OpsAdmin` class definition, mirroring the existing `AppAdmin` rule. That is a real change on the commercial side, so the paired PR there stops being test-only.
3. **`PluginValidatorTest.java` is JUnit (3-style `Assert`)**. Convention is Spock; `PluginValidationSpec.groovy` already exists in the same package. Move the two cases there and delete the file.

### Suggestions

- Authorize on `[ACTION_INSTALL, ACTION_ADMIN]` (controller lines 602/650 and `_sysConfigNavMenuJson.gsp`). `admin` is documented as full access on the `plugin` kind and the marketplace controllers already honor it. Safe for `AppAdmin`, whose class denies the `admin` action on every system type; the rationale for rejecting the Copilot fallback is correct for `app_admin` only.
- The `normalize().startsWith()` guard is unreachable after the separator/`..` checks; replace it with the `isDirectory()` check.
- `PluginControllerSpec.groovy:528` asserts a path the traversal could never have reached; drop it.
- `PluginValidator.java:39` logs a literal `${pluginFile.getName()}`; now hit by every rejected `.groovy` upload.
- Optional: the three identical `try/catch IllegalArgumentException` blocks collapse if the helper returns `null`.

### Checked and fine

- Title, template, ticket link, base branch, scope.
- Sanitizer covers `..`, `/`, `\`, `InvalidPathException` (subclass of `IllegalArgumentException`) and the `libextDir` copy site. `http`/`https` scheme check closes the `file:` and bare-path cases.
- Label/`groovyNote` cleanup from the Copilot review is complete across all five bundles; no leftovers at HEAD.
- `MockWebServer` is already a `testImplementation` dependency of `rundeckapp`.
- `docs.rundeck.com` plugin-install page never documented GUI `.groovy` upload, so no docs PR is needed; `release-notes/include` label is present.

### Copilot thread status

- `.` filename thread: **not addressed**, still open — blocker 1 above.
- Stale `jar/zip/groovy` label and `groovyNote`: **addressed** in 88a9ad4.
- Restore `admin`/`app_admin` fallbacks: declined by author. Partially right on both sides: `app_admin` would indeed let `AppAdmin` route around the deny rule; `admin` would not. See the ACL suggestion above.

rundeckapp/grails-app/controllers/rundeck/controllers/PluginController.groovy
carlosrfranco: Traced this precisely (including the existing `UserClassAuthProviderExtensionSpec` coverage) and confirmed both parts:

- Added `ACTION_ADMIN` alongside `ACTION_INSTALL` here, at the `installPlugin` check, and in `_sysConfigNavMenuJson.gsp` — matching `ArtifactController`/`RepositoryController`. Confirmed safe against `AppAdmin` via `UserClassAuthProviderExtensionSpec.groovy:394-410`: its blanket `deny-actions: [admin, ops_admin]` rule (with `system-types: ["*"]`) applies across every resource kind — not just `AuthResources.appTypes` — because `AuthConstants.TYPE_RESOURCE` is unconditionally prepended before that list is iterated, and the resulting rule carries no kind filter. So `ACTION_ADMIN` is already denied for `AppAdmin` on `RESOURCE_TYPE_PLUGIN`; adding it as a fallback can't reopen that path. Left `ACTION_APP_ADMIN` out, since that action isn't covered by the same deny and would reopen it, as you noted.
- On the `OpsAdmin` gap: confirmed it's real and confirmed it can't be closed from this side. `OpsAdmin`'s user class has no deny-rule for `install`/`uninstall` on `kind: plugin` (unlike `AppAdmin`, which has exactly that rule), so with the shipped `admin.aclpolicy`'s wildcard `resource: allow: '*'` grant, an `OpsAdmin` user still passes `install` on `plugin` directly — regardless of what this controller checks. The fix is a deny-rule addition to `OpsAdmin`'s definition in `plugins/rundeckpro-license/src/main/resources/user-classes.json`, which lives in the commercial `rundeckpro` repo, not this OSS `rundeck` PR. Tracking that as a separate follow-up rather than folding it into this change.

rundeckapp/grails-app/controllers/rundeck/controllers/PluginController.groovy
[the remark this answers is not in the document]
luismalamoc: Fixed in 522fc4b: the unreachable `normalize().startsWith()` guard was replaced with `resolved.isDirectory()`, and `"."` is now covered in both the upload and URL-install traversal specs. Resolving.

rundeckapp/grails-app/controllers/rundeck/controllers/PluginController.groovy
```diff
@@ -698,6 +726,27 @@ Since: v49''',
...
+     * Resolves a plugin file name against a base directory, rejecting names that contain
+     * path separators or {@code ..} segments, or that would otherwise resolve outside
+     * {@code baseDir} once normalized.
+     *
+     * @param baseDir the directory the resolved file must live in
+     * @param name the untrusted plugin file name
+     * @return the resolved {@link File} within {@code baseDir}
+     * @throws IllegalArgumentException if {@code name} is blank or would escape {@code baseDir}
+     */
+    private File resolveSafePluginFile(File baseDir, String name) {
```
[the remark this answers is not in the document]
luismalamoc: Agree, please apply. The helper only touches `String`/`File` APIs, the file already uses `@CompileStatic` on `sendResponse` and the line-327 helper, and it is the project-wide Groovy convention. Nothing in the body needs dynamic dispatch, so this should be a one-line change with no behavior impact.

rundeckapp/grails-app/controllers/rundeck/controllers/PluginController.groovy
luismalamoc: Verified in 522fc4b: `isDirectory()` check in place, `"."` added to both traversal specs. Resolving.

rundeckapp/grails-app/controllers/rundeck/controllers/PluginController.groovy
luismalamoc: Applied in 522fc4b. Resolving.

rundeckapp/grails-app/views/menu/_sysConfigNavMenuJson.gsp
luismalamoc: Mirrored to `[ACTION_INSTALL, ACTION_ADMIN]` in 522fc4b, matches the controller. Resolving; the `artifact/index.gsp` note stays as a separate follow-up.

core/src/test/java/com/dtolabs/rundeck/core/plugins/PluginValidatorTest.java
luismalamoc: Done in 522fc4b: JUnit file deleted, both cases live in `PluginValidationSpec.groovy`. Resolving.

rundeckapp/src/test/groovy/rundeck/controllers/PluginControllerSpec.groovy
luismalamoc: Dropped in 522fc4b. Resolving.

core/src/main/java/com/dtolabs/rundeck/core/plugins/PluginValidator.java
luismalamoc: Fixed in 522fc4b. Resolving.

rundeckapp/grails-app/controllers/rundeck/controllers/PluginController.groovy
luismalamoc: Not taken, which is fine since it was optional. Resolving.

rundeckapp/grails-app/controllers/rundeck/controllers/PluginController.groovy
luismalamoc: Thanks for tracing the `AppAdmin` part, that matches what I see in the spec (`RESOURCE_TYPE_PLUGIN | ACTION_ADMIN | REJECTED_DENIED`). The `[install, admin]` change is good.

On the `OpsAdmin` gap I have to push back on "tracked separately". Two problems:

1. There is no follow-up. No ticket in RUN references it, nothing is linked to RUN-4476, and the paired rundeckpro PR (#5048) still only carries the `.gitmodules` change, with the submodule pointer still on 88a9ad4 rather than this commit.
2. Being in the other repo is not a reason to defer it. The paired rundeckpro PR already exists for this exact ticket; that is where the deny rule goes. It is one `deny-rules` entry on the `OpsAdmin` class (`install`/`uninstall` on `kind: plugin`, a copy of the `AppAdmin` one) plus the matching rows in `UserClassAuthProviderExtensionSpec`. That turns #5048 from a throwaway test PR into a real companion PR, which is the standard flow when a fix spans both repos.

The HackerOne finding and the PR description both claim the `OpsAdmin` code-execution path is closed. Without the deny rule that claim only holds for ACLs that grant `ops_admin` narrowly on `system`; with the shipped wildcard policy the path stays open. So the ticket cannot close on this PR alone.

Request: add the `OpsAdmin` deny rule and spec rows to #5048, bump its submodule pointer to the final commit here, and drop the `(TEST)` / draft status there. I will review both together. Keeping this thread open until that lands.

luismalamoc (changes requested):
## Re-review of 522fc4b — still NEEDS_WORK, two items left

Everything code-level from the first pass is verified and resolved: `.` rejected via `isDirectory()` with both specs extended, `[ACTION_INSTALL, ACTION_ADMIN]` in the controller and nav menu, JUnit test moved into `PluginValidationSpec`, vacuous assertion dropped, log placeholder fixed. Thanks for the fast turnaround and for tracing the `AppAdmin` deny semantics; that matches the spec rows.

### Still open

1. **`OpsAdmin` deny rule belongs in the paired rundeckpro PR, not a future ticket.** The reply on the ACL thread says it is "tracked separately", but nothing exists: no RUN ticket, nothing linked to RUN-4476, and rundeckpro#5048 is unchanged (still `.gitmodules` only, submodule pointer still on 88a9ad4). Option 2 from my first review was to add the rule to #5048 and turn it into a real companion PR; that option was not addressed. It is one `deny-rules` entry on `OpsAdmin` (`install`/`uninstall` on `kind: plugin`, same shape as the `AppAdmin` rule) plus the corresponding `UserClassAuthProviderExtensionSpec` rows. Without it the PR description's claim that the `OpsAdmin` code-execution path is closed only holds for ACLs that grant `ops_admin` narrowly on `system`; with the shipped wildcard `admin.aclpolicy` it stays open. Please add it to #5048, bump that PR's submodule pointer to the final commit here, and drop the draft/`(TEST)` status so both can be reviewed as a unit.
2. **Copilot's new finding is valid, please apply it:** annotate `resolveSafePluginFile` with `@CompileStatic`. The file already does this for `sendResponse` and the line-327 helper, the body only uses `String`/`File` APIs, and it is the project convention. One line, no behavior change.

Resolved every thread that 522fc4b closed, including Copilot's original `.` thread. The ACL thread and the new `@CompileStatic` thread stay open until the two items above land.

PR #10603, opened 2026-09-15
[RUN-4896] Warn when no LogFileStorage plugin is configured at startup

elquimeras:
## Tell us about your PR

Bugfix (observability). `LogFileStorageService.afterPropertiesSet()` returns silently when `rundeck.execution.logs.fileStoragePlugin` does not resolve at bean-initialization time. In that case the storage consumer, the retrieval consumer and the periodic incomplete-request scheduler are never started, while log storage requests keep accumulating in unbounded queues. There was no log line, metric or exception, so an operator only sees a growing `/system/logstorage/incomplete` backlog with zero successes, failures or retries, and a thread dump with no log storage threads (internal ticket RUN-4896, Rundeck 6.1.0; one instance reached 55K+ queued executions before the cause was found).

- Replace the commented-out `System.err.println` in the early-return branch with a single `log.warn` that names the property and states that log storage consumers stay disabled until it is set. The wording is intentionally source-neutral (it does not tell the operator to edit `rundeck-config.properties`), because a follow-up change will start the consumers when the property is set later through DB-backed System Configuration. Level is WARN, not ERROR: running without a log file storage plugin is a valid OSS setup.
- Add three Spock features to `LogFileStorageServiceSpec` covering `afterPropertiesSet()` for the first time:
  - plugin configured: exactly one `TaskRunner` per executor, periodic scheduler registered only for the `periodic` resume strategy;
  - plugin not configured (`null` and empty): nothing started, no exception;
  - log capture with a log4j2 `WriterAppender` (same pattern as `UserActionServiceSpec`): exactly one WARN when unconfigured, none when configured, message names the property and does not mention `rundeck-config.properties`.

`LogFileStorageServiceSpec` goes from 82 to 88 tests; the full `:rundeckapp:test` suite passes (6181 tests). Verified on a real `bootRun` with the property unset (exactly one WARN at startup) and set (no WARN).

- ERROR level: rejected, it would be noise for every OSS install without a log storage plugin.
- Marking `FILE_STORAGE_PLUGIN` with `restart true`: rejected, a restart does not help when the value lives only in System Configuration, and the follow-up change makes the property effective at runtime.
- Fixing the root cause here (re-evaluating the configuration after DB-backed System Configuration loads): deliberately split into a follow-up PR so this diagnostic can be backported to 6.1.x and 6.2.x on its own.

**Additional context**

- Internal ticket: RUN-4896.
- Follow-up: start the consumers on `AppEvents.APP_CONFIG_CHANGED` when the property is set via System Configuration (separate PR, rebases on top of this one).
- Backport candidates: 6.1.x, 6.2.x.
- In unit tests, `ServiceUnitTest` triggers `afterPropertiesSet()` at bean init with a null `configurationService`, so the WARN now appears once per spec test in test output. Harmless; production injects the service before init callbacks.

## Release Notes

Log a WARN at startup when no log file storage plugin is configured (`rundeck.execution.logs.fileStoragePlugin`), instead of silently leaving log storage consumers disabled.


elquimeras:
Closing in favour of the consolidated PR #10638, which carries these exact commits (cherry-picked linearly onto current `main`) plus the review follow-up from #10614. The team asked for a single PR instead of the two-part stack. Review context stays here; please continue on #10638.

PR #10605, opened 2026-09-15
[RUN-4869] Prettier config, prettier/prettier promoted to error and one-shot reformat (ui-trellis)

luismalamoc:
**Jira tickets**: RUN-4869 (Prettier config and scripts) and RUN-4871 (one-shot reformat, `prettier/prettier` promoted to `error`).

## Tell us about your PR

Frontend tooling maintenance for `rundeckapp/grails-spa/packages/ui-trellis`. Today Prettier runs through `eslint-plugin-prettier` as a `warn` (1642 `prettier/prettier` warnings printed on every build, nothing failing) and there is no `.prettierrc`, so the formatting standard is implicit. This PR makes the standard explicit, reformats the tree once, and promotes the rule to `error` so the build fails on unformatted code again.

Review commit by commit:

1. **Config + scripts** (RUN-4869)
   - New `.prettierrc` listing Prettier 3 defaults explicitly (double quotes, semicolons, `trailingComma: all`, 80 columns, 2 spaces, LF, `vueIndentScriptAndStyle: false`) so the standard does not drift when Prettier changes its defaults.
   - `.prettierignore` reduced to `build`, `lib`, `node_modules` (drops a dangling `!.storybook` negation).
   - `package.json`: added `format` (`prettier --write "src/**/*.{ts,tsx,js,vue,scss,css}"`) and `format:check`.
   - Package `.gitignore`: `!.prettierrc` because the root `.gitignore` hides every dotfile.
2. **Mass reformat** (RUN-4871): output of `npm run format`, no manual edits. 161 files under `src/`, plus 4 CSS files after review.
3. **`prettier/prettier` from `warn` to `error`** (RUN-4871): `eslint-plugin-prettier` and `plugin:prettier/recommended` stay; the rule severity changes so `npm run lint` and the webpack builds (`eslint-webpack-plugin`, `failOnError`) fail on real Prettier violations. `eslint-config-prettier` stays via the recommended preset. An earlier revision of this PR removed the plugin instead; reverted per review.

- Removing the ESLint plugin and running Prettier only via npm scripts: rejected in review, it left nothing able to fail the build on formatting.
- Empty `.prettierrc`: satisfies "add a config file" but documents nothing and still floats with the Prettier version.
- Widen scope to `md`/`mdx`/`json`/`html`: overlaps the Spotless `text-files` format in rundeckpro and `mdx` needs a plugin not allowed by the Renovate allowlist. Out of scope.

**Additional context**

- Editor / hook / fast CI enforcement is tracked in RUN-4870.
- Verification on this branch: `npm run format:check` passes; `npx eslint src` reports 0 errors and 1574 warnings, identical to `main` once the 1642 `prettier/prettier` warnings are gone; `npm run test:unit` 120 suites / 1356 tests pass; `npm run build:library` succeeds. Regression check: a single-quote import fails ESLint with `Replace 'vue' with "vue"  prettier/prettier`. Frontend also verified on a staging instance with the CI image built from the companion rundeckpro branch.
- Merge-conflict exposure of the reformat commit is limited to branches touching `ui-trellis/src`. A quick merge window is appreciated.
- Related rundeckpro PR applies the same change to `rundeckpro-ui-components`; the two are independent.

## Release Notes

None (internal tooling).


luismalamoc:
Addressing the Copilot review:

- **CSS not covered by the format globs**: agreed. `css` is now part of `format` / `format:check` (`src/**/*.{ts,tsx,js,vue,scss,css}`) and the 4 drifted CSS files under `src/library/theme` are reformatted in a follow-up commit. `format:check` passes.
- **`toggleSwitch.scss` "behavior change"**: false positive from diff alignment. The declarations before and after are identical (`.p-toggleswitch-handle`, `.p-toggleswitch-slider`, `&.p-disabled .p-toggleswitch-slider` gray-200, `&.p-toggleswitch-checked` with disabled blue-200 and default blue-500). Prettier only re-indented 4 spaces to 2 and added the missing semicolon on the first declaration; no selector or value was added, removed or moved. The commit is still the raw output of `npm run format`.

Branch also updated with `main`.

smartinellibenedetti (changes requested):
Thanks for getting the Prettier config and reformat in. `.prettierrc`, `.prettierignore`, `format`/`format:check` are exactly what we needed, and the reformat itself looks clean.

The tickets had drifted from what we'd actually agreed on though, so I went ahead and updated the RUN-4869/RUN-4871 descriptions to match. The part that still needs fixing in this PR: `.eslintrc.js` drops `eslint-plugin-prettier` and `plugin:prettier/recommended` altogether, and the `prettier/prettier` rule is just deleted instead of promoted. What we'd agreed was to take it out of "warn": promote it to `error` once the reformat landed, so the build fails on real Prettier violations again, same as it already does for other ESLint errors via `eslint-webpack-plugin`'s `failOnError`. Deleting the rule and the plugin does the opposite. Now nothing can ever fail the build on this.

I tested it to be sure. Dropped a single-quote import into `executionShow.ts` with your current `.eslintrc.js`, ran `npx eslint` and a full `npm run dev`, both clean: 0 errors/warnings, build succeeds. Put back just the two removed lines (`"prettier/prettier": "warn"` and `plugin:prettier/recommended`), nothing else, and ESLint catches it right away: `Replace './i18n.js' with "./i18n.js";  prettier/prettier`. So as it stands, nothing checks formatting anywhere: not the editor, not `npm run dev`, not `npm run build`.

No objections to decoupling Prettier from ESLint in the long run. But whatever the final setup ends up being, ESLint still needs to keep catching everything else it catches today (the other open warnings, RUN-4875 to RUN-4877 and the rest), independent of what happens to Prettier. These ~150/1642 drifted files piled up in the first place because nothing forced anyone's hand into keeping things formatted, so whatever we land on needs to actually close that gap, not just relocate it. We shouldn't be waiting until CI to find out formatting is wrong either; our pipelines are too slow for that (close to an hour for rundeck, closer to an hour and a half for rundeckpro). Preference would be the editor handles it, that's Prettier's own recommendation, but happy to keep discussing the mechanism in RUN-4870. Can this stay out of scope here until we've settled that?

For this PR: keep `eslint-plugin-prettier` + `plugin:prettier/recommended`, promote `prettier/prettier` from `warn` to `error` now that the reformat's in. `eslint-config-prettier` stays either way, unaffected.

luismalamoc:
Done as requested:

- `eslint-plugin-prettier` and `plugin:prettier/recommended` are back in `.eslintrc.js` (and in `package.json` / lockfile, same pins as `main`); `prettier/prettier` is now `error` instead of `warn`. `eslint-config-prettier` stays via the recommended preset.
- Branch merged with `main` (npm deps batch, uiv bump, laravel-mix removal). The merged code was run through `npm run format` so `format:check` still passes.
- Verified on this branch: `npm run format:check` clean; `npx eslint src` 0 errors, 0 `prettier/prettier`, 1574 warnings (same as `main`); unit tests 120 suites / 1356 tests; `npm run build:library` succeeds. Repro from your review: a single-quote import in `src/app/components/ui/main.ts` now fails ESLint with `Replace 'vue' with "vue"  prettier/prettier` (1 error), and therefore the webpack build.

Agreed on keeping the enforcement mechanism (editor on save, optional hook, parallel-only CI step) in RUN-4870; I added a description there with the options and the no-critical-path constraint.

smartinellibenedetti (approved):

luismalamoc:
Heads-up, unrelated to Prettier but pushed on this branch so CI can go green: `main` currently breaks the rundeckpro Selenium test `NodeWizardSpec.add edit delete node using wizard` (fails on rundeckpro `main` too, e.g. pipeline 25261).

**Root cause** (commit `0a31b9e7be` here): #10539 (RUN-4866) added `isStaticTextProp()` to skip `displayType: STATIC_TEXT` props in `prepareInputs()` and `exportInputs()`. `PropertyBuilder.autogenInstanceId()` renders `AutogenInstanceId` props with `displayType: STATIC_TEXT` too, so the Node Wizard's `wizard-id` (its only property) is now treated as static text:

- in create mode the generated id is no longer seeded into `inputValues`, so `pluginConfig` never emits `update:modelValue`, `ProjectPluginConfig` never flags `modified`, and the node sources page never shows "You have unsaved changes" after adding a Node Wizard source (that is the exact wait the test times out on);
- `exportInputs()` drops `wizard-id` from the saved config, so the source would be persisted without its id.

**Fix**: `isStaticTextProp()` returns false for `prop.type === "AutogenInstanceId"`. Two unit tests added in `pluginConfig.spec.ts` (seeded id emitted in create mode; existing id kept in edit mode); both fail on `main` and pass with the fix. Full ui-trellis suite: 121 suites / 1370 tests green.

cc @fdevans @ddarby-hike, happy to move this to its own PR if you prefer it separate from the formatting change.

luismalamoc:
Correction: this PR was already merged when I pushed the fix, so the push recreated the deleted branch. The fix now lives in its own PR: #10619. The recreated `RUN-4869-prettier-config-and-format` branch can be deleted; nothing depends on it.

PR #10608, opened 2026-09-15
[RUN-4975] Quote exec command references in place, not the whole argument

### LINKED ISSUE #10293: Rundeck sshj-ssh plugin no longer working with double quotes
After upgrading to Rundeck 5.20.0; it looks like the ssj-ssh plugin no longer works with double-quoted commands in the job steps.  For example, the following used to work:
<exec>sudo "sh /u01/scripts/sre/myscript.sh ${option.parm1}"</exec>
 
We had to change it to the following in order for it to work:
<exec>sudo sh /u01/scripts/sre/myscript.sh ${option.parm1}</exec>

With double-quotes, the job returns:
 
[sudo] password for user1:
sudo: sh /u01/scripts/sre/myscript.sh userval1: command not found
[user1@serverxyz ~]$ echo $?
1
 
Note: This is on Oracle Linux 9 server.

fdevans:
## Tell us about your PR

Bugfix. Fixes #10293 and #10027, both regressions from the command injection fix in PR #10003 (RUN-4175, Rundeck 6.0).

That fix quoted an *entire* exec-command argument whenever it contained a `${option.*}`/`${globals.*}`/`${data.*}` reference. That's correct when the whole argument *is* the reference, but breaks jobs that embed a reference inside a larger, job-authored argument (grouped by the author's own quotes), e.g.:

```
sudo "sh /u01/scripts/sre/myscript.sh ${option.parm1}"
```

Rundeck wraps the *whole* expanded argument in shell quotes, turning it into a single literal token (`'sh /u01/scripts/sre/myscript.sh userval1'`) that `sudo`/the shell can't execute — `sudo: sh /u01/scripts/sre/myscript.sh userval1: command not found` (#10293). When the job author had already wrapped just the reference in their own quotes (#10027's PowerShell example, `-TenantPath '${option.TenantPath}'`), the same whole-argument quoting instead produces mangled, double-escaped output like `''"'"'CLIENT_Name'"'"''`.

Root cause: `ExecArgList.buildCommandForNode()`'s shared-context overload (used by every real node command execution, via `ExecutionServiceImpl`) expanded the whole argument first, then quoted the entire result if the argument was flagged "quoted" — an all-or-nothing decision per *argument*, not per *reference*.

Fix: when an argument is flagged for quoting, quote each substituted reference's *value* in place, as it's substituted, instead of quoting the fully-expanded argument afterward. This still fully blocks command injection via the reference's value — the only untrusted part — without touching the surrounding job-authored text.

The existing CVE regression suite (`ExecCommandInjectionTest`, covering pipe/redirect/substitution/AND/semicolon injection on both Unix and Windows) continues to pass unchanged. Added new tests for the embedded-reference, already-quoted, and injection-still-blocked cases, plus a Windows variant. The separate deprecated `buildCommandForNode(Map, String)` overload (unrelated legacy callers) is untouched — its whole-argument-quoting behavior is preserved exactly as before.

Also removed dead code found while fixing this: in `CommandNodeStepPlugin.groovy`, an early substitution pass (`SharedDataContextUtils.replaceDataReferencesInObject` on a `String[]`) was a complete no-op — that method has no handling for array input (only `String`, `Map`, `Collection`), so it silently returned the array unchanged on every call. Removed it; the plugin now only decides which arguments need quoting (based on whether they contain a reference), leaving actual substitution to `ExecArgList.buildCommandForNode()` where the fix lives.

Considered simply not quoting an argument when the reference is embedded in a larger string (rather than being the whole argument). Rejected: this would silently reopen the original command-injection vulnerability for the (likely common) case of a reference embedded in a larger literal argument, since the untrusted value would then flow through completely unescaped.

**Additional context**

Verified with a full run of the core execution test suite (54 test classes, 0 failures) plus the entire `script-node-step-plugin` module, and an empirical reversion check: reverting the per-reference-quoting change causes exactly the 5 new/updated tests targeting this behavior to fail, reproducing the same malformed output shapes reported in #10293 and #10027, confirming the tests actually catch the regression.

## Release Notes

Fixed a regression from the 6.0 command-injection fix (RUN-4175) where an exec command argument that embeds an option/data reference inside a larger, job-authored string (e.g. `sudo "sh script.sh ${option.name}"`, or a reference the job author had already wrapped in their own quotes) could be mis-quoted, producing a broken or garbled command. References are now quoted in place, value-only, preserving the surrounding job-authored text while still fully protecting against command injection via the reference's value.


fdevans:
Update: after opening this, I checked whether the fix's design was safe for *every* caller of the code path it touches (given this repo's history of quoting fixes needing follow-up corrections — RUN-4416, RUN-4587). It wasn't quite: `ScriptBasedRemoteScriptNodeStepPlugin` (backing all bundled script-type plugins) resolves `${...}` references itself before building the `ExecArgList`, so the per-reference quoting in the first commit had nothing left to match for that caller — which meant quoting was silently skipped entirely for it, not just mis-applied.

Pushed a follow-up commit that falls back to quoting the whole argument when no `${...}` reference is actually present in the string received (i.e. it was already substituted upstream), preserving the original behavior for that caller exactly, while keeping the in-place fix for callers that hand this code raw references (Command/Script/Script-file steps). Added a regression test for the specific silent-bypass scenario and verified it against a revert of just that commit.

fdevans:
Confirmed — verified this empirically before fixing, and it reproduced exactly as described (`-TenantPath ''x; whoami''`, semicolon unquoted).

Pushed a fix (082548c) that detects when a reference is directly flanked by matching author-supplied single quotes, and in that case escapes only the value's own embedded `'` characters for insertion *inside* the existing quotes (the standard `'\''` technique) instead of wrapping it in a second, conflicting pair. Added regression tests for both this exact scenario and a value containing its own embedded single quote.

Scoped this to Unix single-quote flanking only, since that's the reported and confirmed case — noted in code comments that Windows-side flanking-quote handling is a separate, unaddressed gap if it comes up.

The two suppressed comments (`quoteAppliedDuringExpand` losing quoting when no reference is present, and bypassing `featureQuotingBackwardCompatible`) were also both real — already fixed in the two commits pushed just before this one.

PR #10612, opened 2026-09-16
[RUN-4649] Add @PluginOutput annotation and capture exposed property values for conditional logic

carlosrfranco:
## Tell us about your PR

Enhancement. Part of RUN-4649 (M4 Job authoring: Conditional Logic, epic RUN-4244). Introduces a `@PluginOutput` annotation so a plugin property can be marked as exposed for reference by a conditional step's condition.

- New `@PluginOutput(group=, name=, description=)` / `@PluginOutputs` annotations (`com.dtolabs.rundeck.plugins.descriptions`), following the existing `@PluginProperty`/`@RenderingOption` pattern; usable on a field that already carries `@PluginProperty`.
- `PluginAdapterImpl.propertyFromField` reads `@PluginOutput` via reflection and includes it in the built `Property` (via a new `Property.getOutputMetadata()` default method + `PluginOutputMetadata` value object, threaded through `PropertyBuilder`), so it's available through the plugin description/metadata APIs.
- `PluginApiService.pluginPropertyMap` / `ApiProviderProp` expose the new metadata through the existing plugin metadata REST API.
- `StepPluginAdapter.executeWorkflowStep` now writes the resolved value of any `@PluginOutput` property into the step's `SharedDataContext` (`ContextView.step(N)`, configured group/name), reading it off the plugin instance's field via reflection when the property was bound directly to a field (the common case) rather than the leftover instance-config map.

Considered threading the output metadata through every `PropertyUtil.forType` overload/concrete `Property` implementation; instead used a small decorator in `PropertyBuilder.build()` so the change stays isolated to the one path (`propertyFromField`) that needs it, per the design doc's guidance to keep this backwards compatible.

**Additional context**
This is a backend-only foundation ticket. No frontend/`ConditionsEditor.vue` changes. A companion rundeckpro-side PR rewrites `"<Step Label> - <Property Name>"` condition keys into the internal `${stepNum:group.name}` form that consumes this data (plugins/workflow-conditional-step).

## Release Notes

Adds a `@PluginOutput` annotation allowing plugins to mark properties as exposed values, consumable by future Conditional Logic UI/condition work.


core/src/main/java/com/dtolabs/rundeck/core/execution/workflow/steps/StepPluginAdapter.java
```diff
@@ -141,6 +141,15 @@ public StepExecutionResult executeWorkflowStep(final StepExecutionContext execut
...
         }
+
+        // Captured after executeStep() returns successfully, not before: a property that carries
+        // only @PluginOutput (no @PluginProperty) is a computed value that the plugin sets on its
+        // own field *during* execution (e.g. an ID created by the step), so it doesn't exist yet
+        // beforehand. Capturing here also means a step that fails/throws exposes no output.
+        // Uses executionContext.getOutputContext() (the same channel used by log-filter and
+        // node-executor output capture) rather than getSharedDataContext() directly, since that's
+        // what the workflow engine merges forward into subsequent steps' conditional evaluation.
+        PluginOutputCapture.captureOutputMetadataValues(executionContext.getOutputContext(), description, plugin, config);
```
Jesus-Osuna-M: **Blocker:** this runs after the `try`/`catch (Throwable)` above, so it isn't covered by the plugin error handling.

`PluginOutputCapture.resolvePropertyValue` catches only `IllegalAccessException`. `field.setAccessible(true)` throws `InaccessibleObjectException` or `SecurityException` (both unchecked), and `value.toString()` runs arbitrary plugin code. Any of those escapes `executeWorkflowStep` and fails a step whose `executeStep` already returned successfully.

Suggest wrapping the capture itself (ideally inside `PluginOutputCapture.captureOutputMetadataValues`, so both adapters get it) in a try/catch that logs at `Constants.DEBUG_LEVEL` and swallows. Capture is best-effort metadata; it shouldn't be able to turn a successful step into a failure.

core/src/main/java/com/dtolabs/rundeck/core/execution/workflow/steps/node/NodeStepPluginAdapter.java
```diff
@@ -199,6 +199,14 @@ public NodeStepResult executeNodeStep(final StepExecutionContext context,
...
                                           node);
         }
+        // Captured after executeNodeStep() returns successfully, not before, mirroring
+        // StepPluginAdapter. context.getOutputContext() is already scoped to this node by the
+        // dispatcher (SequentialNodeDispatcher/ParallelNodeDispatcher build a per-node
+        // ExecutionContext whose output context defaults to ContextView.nodeStep(stepNum,
+        // node.getNodename())), so no explicit node-scoping is needed here: each node's captured
+        // value lands in its own slot, and a later node-step conditional evaluated on this same
+        // node will resolve back to this value without any node-name syntax.
+        PluginOutputCapture.captureOutputMetadataValues(context.getOutputContext(), description, plugin, config);
```
Jesus-Osuna-M: Same as `StepPluginAdapter.java:152` — outside the `try`/`catch (Throwable)`, so a reflection failure here fails a node step that already executed successfully. Handling it once inside `PluginOutputCapture.captureOutputMetadataValues` would cover both call sites.

core/src/main/java/com/dtolabs/rundeck/core/plugins/configuration/PluginOutputCapture.java
```diff
@@ -0,0 +1,126 @@
...
+     * off the plugin's own field of the same name, which is where {@code PluginAdapterImpl} sourced
+     * the property's name/title from when it built the Description for such a field.
+     */
+    private static Object resolvePropertyValue(
+            final Property property,
+            final Object pluginInstance,
+            final Map<String, Object> config
+    )
+    {
+        if (config != null && config.containsKey(property.getName())) {
```
Jesus-Osuna-M: The config map is consulted before the field for every property, including `isOutputOnly()` ones.

`fa736fa7` hides output-only properties from the config GUI, but `PluginAdapterUtility.configureProperties` still has no `outputOnly` awareness, so a job defined via API/YAML with a `configuration` key matching the field name would land in `config` and shadow the value the plugin actually computed during execution — silently, and the captured "output" would be the job-authored string instead.

For an output-only property the config map can never hold a legitimate value, so it'd be safer to skip it:

```java
if (!property.isOutputOnly() && config != null && config.containsKey(property.getName())) {
    return config.get(property.getName());
}
```

core/src/main/java/com/dtolabs/rundeck/core/plugins/configuration/PluginOutputCapture.java
```diff
@@ -0,0 +1,126 @@
...
+        }
+        Field field = new PluginAdapterImpl().fieldForPropertyName(property.getName(), pluginInstance);
+        if (field == null) {
+            field = fieldByName(pluginInstance.getClass(), property.getName());
+        }
+        if (field == null) {
+            return null;
+        }
+        try {
+            field.setAccessible(true);
```
Jesus-Osuna-M: `setAccessible(true)` throws `InaccessibleObjectException` (JDK 17+ module access) and `SecurityException` — both unchecked, so neither is covered by the `catch (IllegalAccessException)` below. Combined with the call sites being outside the adapters' error handling, this is the concrete path by which a successful step ends up reported as failed. Broadening to `catch (RuntimeException | IllegalAccessException)` here, or catching around the whole capture loop, would close it.

Jesus-Osuna-M (changes requested):
Reviewed at `b6b6417`. The `outputOnly` + `PluginApiService` filter added in `fa736fa7` resolves the config-UI exposure concern, and `PluginOutputCapture` is a good extraction that keeps the workflow-step and node-step paths honest. One blocker left, plus a few smaller items.

## Blocker

**Output capture runs outside the plugin error handling.** At both call sites (`StepPluginAdapter.java:152`, `NodeStepPluginAdapter.java:209`) the capture happens *after* the try/catch that converts plugin failures into a `StepExecutionResult`/`NodeStepResult`. `resolvePropertyValue` only catches `IllegalAccessException`, so an unchecked exception from reflection (`setAccessible` -> `InaccessibleObjectException`/`SecurityException`) or from a value's `toString()` escapes `executeWorkflowStep`/`executeNodeStep` uncaught — turning a step that already ran successfully into a hard failure. Capture is best-effort metadata; it should never be able to fail a step that worked. Details inline.

## Suggestions

- `resolvePropertyValue` consults the config map before the field for *every* property, including `isOutputOnly()` ones — see inline on `PluginOutputCapture.java:92`.
- `new PluginAdapterImpl()` is allocated per property inside `resolvePropertyValue`; it's stateless, so a `private static final` instance would do.
- `OutputMetadataProperty` doesn't override `equals`/`hashCode`, so a decorated property no longer compares equal to the undecorated one it wraps. Nothing in this PR depends on it; worth a note in case descriptions ever get deduped.
- New files in `core/` carry `Copyright 2016 SimplifyOps, Inc.` (copied boilerplate). The new files on the companion rundeckpro PR use `Copyright 2024 Rundeck, Inc.`.
- The Release Notes section is filled in but the `release-notes/include` label isn't applied.

## Looks good

Javadoc is complete throughout; tests are Spock and cover the interesting cases (output-only fields, multi-output, node-step capture). I traced the capture channel against `BaseWorkflowExecutor:670` — the step's output context is already bound to `ContextView.step(c)` and `combineResultData` propagates it forward, so the mechanism is sound. No secrets, no injection surface, no migrations, no API version impact.

Reviewed together with rundeckpro/rundeckpro#5055 (RUN-4649).

PR #10614, opened 2026-09-16
[RUN-4896] Start log storage consumers when the plugin is configured via System Configuration

elquimeras:
## Tell us about your PR

Bugfix. `LogFileStorageService` evaluates `rundeck.execution.logs.fileStoragePlugin` only once, in `afterPropertiesSet()`. When the value is provided later, for example saved only in the DB-backed System Configuration of Rundeck Enterprise (which loads after all beans have initialized), the storage consumer, the retrieval consumer and the periodic incomplete-request scheduler never start. Log storage requests then accumulate in unbounded queues indefinitely; a restart does not help because the ordering is always the same (internal ticket RUN-4896, Rundeck 6.1.0, 55K+ queued executions on one instance).

This PR is stacked on #10603, which added the startup WARN for the unconfigured case. Base branch is `RUN-4896-logstorage-warn-unconfigured`; it will be retargeted to `main` once #10603 merges.

- Move the consumer/scheduler startup from `afterPropertiesSet()` into `startConsumersIfConfigured()`, guarded by an `AtomicBoolean.compareAndSet` so it runs at most once per service lifetime, and only when the plugin name resolves.
- Add `@Subscriber(AppEvents.APP_CONFIG_CHANGED) onAppConfigChanged(Set<String> keys)` that re-evaluates the configuration and starts the consumers if they are not running yet. `AppEvents.APP_CONFIG_CHANGED` is the core event already consumed by `DynamicMailSender`, `AuditEventsService` and `DelegateStorageTree`; Rundeck Enterprise publishes it after refreshing `ConfigurationService`, both on the initial load of DB-backed configuration and on every later refresh (on all cluster members). In OSS nothing publishes it, so OSS behaviour is unchanged.
- The handler logs INFO when it starts the consumers late (`Log storage consumers started: rundeck.execution.logs.fileStoragePlugin is now configured (plugin: ...)`), returns early once started, and catches any `Throwable` (the grails event bus dispatches asynchronously; a failure must never affect the publisher).
- `afterPropertiesSet()` keeps the WARN from #10603.

Tests (`LogFileStorageServiceSpec`, 88 -> 96): late start once (periodic/delayed strategies), no start while the plugin is unresolved, at most one start across init and repeated events, two concurrent events start once, exactly one INFO on late start and none when started at init, handler failure logged and contained. Full `:rundeckapp:test` suite green (6295 tests).

Verified on an Enterprise stack built from this branch: saving the property only in System Configuration starts the consumers without restart (INFO logged); after a restart with the value only in the database, the startup WARN is followed ~10 s later by the INFO once the database configuration loads.

- A second `ContextRefreshedEvent` listener in the service: rejected, ordering against the Enterprise configuration loader is not guaranteed without a shared `@Order` contract, and it would not cover values saved later at runtime.
- Lazy start on the first storage/retrieval request: rejected, adds synchronization to hot paths and leaves the periodic resume scheduler unstarted until a request arrives.
- Filtering the event on the property key: not needed, the re-evaluation is a single config lookup and the event is rare; the guard makes it idempotent.

**Additional context**

- Internal ticket: RUN-4896. Follows #10603 (same ticket, part 1).
- The idempotency guard matters: `SimpleAsyncTaskExecutor` does not reuse threads and blocks the submitter at its concurrency limit, so duplicate never-ending `TaskRunner`s would leak threads and eventually block.
- Stopping or swapping consumers when the plugin changes after start is out of scope.
- Follow-up (not in this PR): the System Configuration UI shows a generic "Restart is Required" banner for properties added in the Custom section; after this change no restart is needed for this property.
- Backport candidates together with #10603: 6.1.x, 6.2.x.

## Release Notes

Log storage consumers now start when `rundeck.execution.logs.fileStoragePlugin` becomes configured after startup (for example through System Configuration), instead of requiring the property in `rundeck-config.properties` at boot.


Jesus-Osuna-M (changes requested):
Reviewed the diff and verified the design against the code. Approving in principle — the mechanism is sound — with one substantive item I'd like addressed first.

**What I verified**

- Event contract matches the existing subscribers: `Set<String> keys` is the same signature as `DynamicMailSender.configurationChanged` and `AuditEventsService.onRundeckConfigurationChange`.
- The publisher ordering holds: `ConfigServiceRefresher` fires `eventBus.notify(APP_CONFIG_CHANGED, allKeys)` in step 6, *after* `resetAppCfg()` in step 5, so `configurationService` already returns the fresh value when the handler runs.
- OSS behaviour is genuinely unchanged — there is no publisher of `APP_CONFIG_CHANGED` anywhere in this repo.
- The log4j2 appender mutation in the new specs is safe here because `rundeckapp/build.gradle` sets `maxParallelForks = 1`. Worth remembering if in-JVM parallelism is ever enabled.

---

**Please fix: the guard is set before the work, so a transient failure disables the consumers permanently**

In `startConsumersIfConfigured()`:

```groovy
if (!consumersStarted.compareAndSet(false, true)) return false
// ... all the work that can throw happens after this point
```

If anything throws after the CAS — `logFileStorageTaskExecutor.execute` raising `TaskRejectedException` on thread-creation failure, or `logFileStorageTaskScheduler.scheduleAtFixedRate` on a scheduler already shut down by `cleanup()` — the flag stays `true`. Every subsequent event then returns early at `if (consumersStarted.get()) return` and nothing ever retries.

That reproduces exactly the bug this PR fixes (queues growing with no consumers), only with one ERROR line instead of silence. The `"a failure while starting consumers ... is logged and contained"` feature currently encodes that behaviour without asserting any recovery. Low probability, but permanent and contrary to the PR's own goal.

Reset-on-failure would be the natural fix:

```groovy
if (!consumersStarted.compareAndSet(false, true)) return false
boolean ok = false
try {
    // ... work
    ok = true
    return true
} finally {
    if (!ok) consumersStarted.set(false)
}
```

One caveat: on a *partial* failure (storage consumer submitted, retrieval throws) a retry would submit a second storage `TaskRunner`, which is precisely the thread leak the idempotency guard exists to prevent. So the reset needs either per-step granularity or individually idempotent submissions.

If that looks like more machinery than it's worth, the minimal alternative is to keep the flag sticky and make the ERROR actionable — state explicitly that log storage consumers will stay down until restart, since no retry will occur. Either is fine by me; silently unrecoverable is the part I'd rather not ship.

---

**Minor / optional**

1. `afterPropertiesSet()` resolves the plugin name twice — once directly, once inside `startConsumersIfConfigured()`. `if (!startConsumersIfConfigured()) { log.warn(...) }` is equivalent at init (it cannot already be started) and drops the duplication.
2. `catch (Throwable)` also swallows `Error` (OOM, StackOverflow). `Exception` covers the stated goal of not affecting the publisher.
3. `startConsumersIfConfigured()` is public (no modifier in Groovy). The specs only call `afterPropertiesSet` and `onAppConfigChanged`, so it can be `private` rather than widening the bean's surface.
4. Late start sets `concurrencyLimit` after `logFileTaskExecutor` may already have been used — `CompletableFuture.supplyAsync(supplier, logFileTaskExecutor)` uses it independently of the consumer. Tasks submitted while the limit was unbounded don't increment `ConcurrencyThrottleSupport`'s counter but do decrement it once the limit is active, so the counter can drift negative and the effective throttle ends up looser than configured. Benign (no deadlock), but new to the deferred path.
5. `FILE_STORAGE_PLUGIN_KEY` in the spec hardcodes a string the same test already references as `LogFileStorageService.FILE_STORAGE_PLUGIN.key` — worth using the constant.

elquimeras:
Thanks for the careful review, the sticky-guard point was right and I've gone with the per-step option rather than the actionable-ERROR one.

**Guard granularity.** `startConsumers()` is now `private synchronized` and keeps one flag per step (`storageConsumerStarted`, `retrievalConsumerStarted`, `periodicDequeueScheduled`). A retry after a failure submits only what is still missing, so a partial failure can't duplicate a running `TaskRunner`. The `AtomicBoolean` stays as the fast-path check and is set only once every step completed. New spec feature: first change fails while submitting the retrieval consumer, second change completes it; asserts exactly 1 storage `execute`, 2 retrieval `execute`, 1 `scheduleAtFixedRate`.

**Minors, all applied:**
1. `afterPropertiesSet()` resolves the plugin once; the WARN stays specific to "not configured" and never fires for a start failure.
2. `catch (Exception)` instead of `Throwable`; the ERROR now says the missing consumers will be retried on the next configuration change.
3. Start method is `private`.
4. Concurrency limits moved to `applyExecutorConcurrencyLimits()`, applied at init regardless of the plugin, so a late start never touches the limit of an executor that `supplyAsync` already used. The "not configured at init" feature now asserts both limits are set.
5. Spec uses `LogFileStorageService.FILE_STORAGE_PLUGIN.key`; the hardcoded constant is gone.

`LogFileStorageServiceSpec` 96 -> 97, full `:rundeckapp:test` green.

elquimeras:
Closing in favour of the consolidated PR #10638, which carries these exact commits (cherry-picked linearly onto current `main`) plus the review follow-up from #10614. The team asked for a single PR instead of the two-part stack. Review context stays here; please continue on #10638.

PR #10619, opened 2026-09-17
[RUN-4866] Keep AutogenInstanceId props out of the STATIC_TEXT exclusion in pluginConfig

luismalamoc:
## Tell us about your PR

Bugfix, regression from #10539 (RUN-4866). Since 2026-09-17 `main` breaks the rundeckpro Selenium test `NodeWizardSpec.add edit delete node using wizard` (fails on rundeckpro `main` too, pipeline 25261): after adding a Node Wizard node source, the project node sources page never shows "You have unsaved changes".

**Root cause**: #10539 added `isStaticTextProp()` to skip `displayType: STATIC_TEXT` props in `prepareInputs()` and `exportInputs()`. `PropertyBuilder.autogenInstanceId()` renders `AutogenInstanceId` props with `displayType: STATIC_TEXT` as well, so the Node Wizard's `wizard-id` (its only property) is now treated as display-only text:

- in create mode the generated id is no longer seeded into `inputValues`, so `pluginConfig` never emits `update:modelValue`, `ProjectPluginConfig` never flags `modified` and the page never shows the unsaved-changes banner;
- `exportInputs()` drops `wizard-id` from the saved config, so the source would be persisted without its id.

`isStaticTextProp()` returns `false` for `prop.type === "AutogenInstanceId"`. The STATIC_TEXT exclusion keeps working for real display-only props (WorkflowStrategy `info`), and autogenerated ids are seeded and exported again.

Two unit tests added to `pluginConfig.spec.ts`: a generated id is seeded and emitted in the config payload in create mode; an existing id is kept in the exported config in edit mode. Both fail on `main` and pass with the fix.

- Checking `staticTextDefaultValue` instead of the type: `loadPluginData()` sets it for AutogenInstanceId too, so it does not discriminate.
- Changing `PropertyBuilder.autogenInstanceId()` to stop using STATIC_TEXT: would change how every autogen id renders in the form; out of proportion for this regression.

**Additional context**

- Verification: `npx jest pluginConfig.spec.ts` 41/41; full ui-trellis suite 121 suites / 1370 tests; `npm run format:check` and `npx eslint` clean on the two files.
- The rundeckpro side needs no change; rundeckpro/rundeckpro#5050 will go green once this is in `main`.

## Release Notes

Fixed: adding a plugin whose only property is an autogenerated id (e.g. the Node Wizard node source) did not mark the form as modified and dropped the id from the saved configuration.


luismalamoc:
Addressed the Copilot comment (https://github.com/rundeck/rundeck/pull/10619#discussion_r4038218233): `autogenProp` is now a factory returning a fresh object per test, since `loadPluginData()` mutates prop objects in place. Spec still 41/41.

fdevans (approved):

elquimeras (approved):

PR #10620, opened 2026-09-17
[RUN-4980] Fix Custom Fields value-loss bug in DynamicFormPluginProp

ddarby-hike:
## Tell us about your PR

Bugfix (RUN-4980). The "Custom Fields" (`DYNAMIC_FORM`) plugin-property widget in `ui-trellis` (used by PagerDuty's "Send a Change Event" step, and also by Jira/ServiceNow/RoiMetrics plugins) only synced a typed field value to the saved job config on the input's `change` (blur) event, not on every keystroke. Typing a value and saving the job without first clicking/tabbing away from that field silently reverted it to empty — the config was saved with the value from whenever the field was first added, not what was actually typed.

- `DynamicFormPluginProp.vue`: replaced `v-model` + `@change` with `:value` + `@input`, so the field's value (and the parent's `modelValue`) update on every keystroke instead of only on blur.
- Small UX improvements to the "Add Field" modal noticed while working in this area: added help text under Field Key and Field Label clarifying what's required (Key) vs. optional (Label — defaults to Key; Description — auto-generated if left blank), and reordered Field Key above Field Label.
- `pluginPropEdit.vue`: moved the property's own description text to render *above* the Custom Fields widget instead of after it. `DYNAMIC_FORM`-type properties render no label of their own, so the description was the only context a user had, and it used to read last instead of first.

Considered adding a second `@input` listener alongside the existing `v-model`, but that relies on Vue's listener-merge ordering between the implicit `v-model` handler and an explicit one on the same native event, which isn't guaranteed. Replacing `v-model` with an explicit `:value`/`@input` handler that updates the value and re-emits in one step avoids that ambiguity entirely.

**Additional context**
Added regression tests for both fixes in `DynamicFormPluginProp.spec.ts` and `pluginPropEdit.spec.ts`, and verified each fails against the pre-fix code before confirming it passes with the fix. 20/20 tests passing across both spec files.

Companion rundeckpro PR: RUN-4979 (PagerDuty plugin fixes that surfaced this bug via the "Send a Change Event" step's Custom Fields UI).

## Release Notes

Fixed a bug where a value typed into a plugin's "Custom Fields" editor could be silently lost if the job was saved without first clicking away from the field.


rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/tests/DynamicFormPluginProp.spec.ts
```diff
@@ -140,6 +140,281 @@ describe("DynamicFormPluginProp.vue", () => {
...
+  it("emits the updated value on every keystroke, without waiting for blur", async () => {
+    // Regression test: the field value used to be synced to the parent
+    // (and therefore to the saved job/step config) only on the input's
+    // 'change' event, which fires on blur. Typing a value and saving
+    // without first clicking/tabbing away silently dropped it. The value
+    // must now be emitted on 'input' directly.
+    const wrapper = createWrapper();
+    await flushPromises();
+    const inputField = wrapper.find('[data-testid="field-input-0"]');
+    await inputField.setValue("Updated Value");
```
luismalamoc: **Blocker:** this regression test doesn't catch the bug it's meant to guard. In `@vue/test-utils` 2.5.0, `setValue()` on a text input triggers `input` **and then `change`** (see `DOMWrapper.setValue`: "trigger `change` for `v-model.lazy`"). The pre-fix code (`v-model` + `@change="changeField"`) therefore emits too, so this test passes with or without the fix.

Dispatch only `input`:
```ts
const inputField = wrapper.find('[data-testid="field-input-0"]');
(inputField.element as HTMLInputElement).value = "Updated Value";
await inputField.trigger("input");
```
Please also double-check it actually fails against the pre-fix component.

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/tests/DynamicFormPluginProp.spec.ts
```diff
@@ -140,6 +140,281 @@ describe("DynamicFormPluginProp.vue", () => {
...
+    expect(wrapper.find('[data-testid="field-label-help"]').text()).toContain(
+      "message_fieldLabelHelp",
+    );
+    expect(wrapper.find('[data-testid="field-key-help"]').text()).toContain(
+      "message_fieldKeyHelp",
+    );
+  });
+
+  it("associates the Field Key and Field Label inputs with their labels and help text, and marks Key required", async () => {
+    // Copilot review on RUN-4980: the labels had no `for`, the inputs no
```
luismalamoc: **Blocker:** please don't reference the ticket or the review process in file contents (`Copilot review on RUN-4980`). Ticket keys belong in the PR title and commit messages only; "Copilot review" is history, not documentation. Same in this file at lines 252, 285, 327, 352, 366, 389.

While at it, please trim these test comments down to one line (or drop them): the test name should say what's being checked, and the *why* lives in the commit message.

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/tests/pluginPropEdit.spec.ts
```diff
@@ -163,4 +165,44 @@ describe("pluginPropEdit", () => {
...
+
+    const descriptions = wrapper.findAllComponents(PluginDetails);
+    const widget = wrapper.findComponent(DynamicFormPluginProp);
+
+    expect(descriptions.length).toBe(1);
+    expect(descriptions[0].props("description")).toContain(
+      "Any custom fields to be included in the change event",
+    );
+    expect(widget.exists()).toBe(true);
+    // Copilot review on RUN-4980: DynamicFormPluginProp derives its modal's
```
luismalamoc: **Blocker:** same as in `DynamicFormPluginProp.spec.ts`, remove the `Copilot review on RUN-4980` reference from file contents, and shorten the comment.

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/tests/pluginPropView.spec.ts
```diff
@@ -451,6 +451,51 @@ describe("PluginPropView", () => {
       expect(pairs[1].text()).toContain("8080");
     });
 
+    it("falls back to the key when a stored custom field has no label — Copilot review on RUN-4980", async () => {
```
luismalamoc: **Blocker:** the ticket key is part of the test name here (and at line 478). Please drop `— Copilot review on RUN-4980` from both test names; ticket refs go in the PR title and commits only.

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/DynamicFormPluginProp.vue
```diff
@@ -209,6 +265,17 @@ export default defineComponent({
       type: String,
       required: true,
     },
+    // An instance-unique prefix for the modal's control/help ids, e.g.
```
luismalamoc: **Suggestion:** we're on Vue 3.5.42, so `useId()` gives an instance-unique, SSR-safe id for free. That would replace the `idPrefix` prop, the `fieldIdPrefix` computed with its regex sanitizing, the `:id-prefix` binding in `pluginPropEdit.vue`, and the two id-collision tests:
```ts
setup() {
  return { uid: useId() };
},
```
then `:id="`${uid}-key-input`"` etc. `useId` is Composition-only, so a minimal `setup()` is allowed under our Vue rules.

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/DynamicFormPluginProp.vue
```diff
@@ -269,22 +351,46 @@ export default defineComponent({
...
+          // label as-is, so normalize it here too, not just on creation.
+          if (!value.label || value.label.trim() === "") {
+            value.label = value.key;
+            normalizedLabel = true;
+          }
           return value;
         });
         this.customFields = parsedFields;
+        if (normalizedLabel) {
+          this.refreshPlugin();
```
luismalamoc: **Suggestion:** I'd remove this load-time normalization (and its test). After the later rounds, all three renderers already fall back to the key: this editor's own label (line 13), `pluginPropView.vue` and `PluginTagLib.groovy`. That makes this redundant, and it has a side effect: `refreshPlugin()` emits `update:modelValue` on mount, so just opening a job with a legacy field marks the config as changed.

If you drop it, make line 13 trim-aware to match the other renderers: `field.label?.trim() || field.key`.

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/DynamicFormPluginProp.vue
```diff
@@ -269,22 +351,46 @@ export default defineComponent({
...
+      this.invalidKey = false;
       this.modalAddField = true;
     },
     addField() {
       let field = {} as CustomField;
       this.duplicate = false;
+      this.invalidKey = false;
+
+      const key = this.useOptions ? this.selectedField?.value : this.newField;
+      if (!key || key.trim() === "") {
```
luismalamoc: **Suggestion:** the key is validated trimmed but stored untrimmed (`key: this.newField`), so `" env"` passes, gets saved with the space, and isn't caught as a duplicate of `"env"`. Trim once here and use that value for both the check and the stored key.

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/DynamicFormPluginProp.vue
```diff
@@ -269,22 +351,46 @@ export default defineComponent({
...
+        this.invalidKey = true;
+        return;
+      }
 
       if (this.useOptions) {
         if (this.selectedField !== null) {
```
luismalamoc: **Suggestion (nit):** this `if (this.selectedField !== null)` is dead now, since the guard above already returns when there's no selected value. It can go.

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/DynamicFormPluginProp.vue
```diff
@@ -94,53 +100,103 @@
...
               <div class="col-md-8">
                 <input
-                  v-model="newLabelField"
+                  :id="`${fieldIdPrefix}-key-input`"
+                  v-model="newField"
                   type="text"
                   :class="['form-control']"
-                  data-testid="field-label-input"
+                  required
+                  aria-required="true"
```
luismalamoc: **Suggestion (nit):** `aria-required="true"` is redundant, because native `required` already exposes the required state to assistive tech.

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/DynamicFormPluginProp.vue
```diff
@@ -218,12 +285,27 @@ export default defineComponent({
...
       duplicate: false,
+      invalidKey: false,
       newField: "",
       newLabelField: "",
       newFieldDescription: "",
       selectedField: { value: "", label: "" },
     };
   },
+  computed: {
+    // A page can render more than one DynamicFormPluginProp (one per
```
luismalamoc: **Suggestion:** the comments in this component are much longer than the code they describe. This one is 8 lines for a 3-line computed, the `idPrefix` prop has 6, `addField` has 4 for a ternary, and the one at 360 is 4 more. Most of them retell the review history ("used to", "previously", "not just on creation"). Please cut each to one line about what the code does, or remove it. The history belongs in the commit messages, which already cover it well.

rundeckapp/grails-spa/packages/ui-trellis/src/app/utilities/locales/en_US.js
```diff
@@ -70,6 +75,13 @@ const messages = {
   message_fieldKeyAppendedDescription: "{0} (Field key: {1})",
   message_fieldFilter: "Type to filter a field",
   message_empty: "Can be empty",
+  // A dedicated key rather than reusing message_empty: other locale
```
luismalamoc: **Suggestion:** a 5-line comment explaining why a key *wasn't* reused is review history, not something a reader of the catalogue needs. Please remove it; the commit message already explains it.

rundeckapp/grails-spa/packages/ui-trellis/tests/unit/jobeditAutocomplete.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+function loadAutocompleteBase(jQueryMock: unknown): AutocompleteBase {
+  const source = fs.readFileSync(
+    path.resolve(
+      __dirname,
+      "../../../../../grails-app/assets/javascripts/jobedit.js",
+    ),
+    "utf8",
+  );
+  const start = source.indexOf("function autocompleteBase(");
+  const end = source.indexOf("\nfunction _initJobPickerAutocomplete");
```
luismalamoc: **Suggestion:** this test is fragile for what it covers. It slices `jobedit.js` by string position up to the *next* function's name and `eval`s it with `new Function`, so renaming or reordering `_initJobPickerAutocomplete` breaks it without any change in behavior. All of that guards a single `dispatchEvent` line. I'd drop the file. If you want coverage, a Selenium check that picks an autocomplete suggestion and saves the job would test the real integration.

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/pluginPropEdit.vue
```diff
@@ -390,7 +405,13 @@
       ></slot>
     </template>
 
-    <div v-if="prop.desc" class="col-sm-10 col-sm-offset-2 help-block">
+    <div
+      v-if="
+        prop.desc &&
+        !(prop.options && prop.options['displayType'] === 'DYNAMIC_FORM')
```
luismalamoc: **Suggestion:** `prop.options && prop.options['displayType'] === 'DYNAMIC_FORM'` now appears three times in this template (and the `plugin-details` block is duplicated). An `isDynamicForm` computed would make this `v-if="prop.desc && !isDynamicForm"`.

rundeckapp/grails-app/taglib/rundeck/PluginTagLib.groovy
```diff
@@ -201,7 +201,10 @@ class PluginTagLib {
                 out << '<div class="customattributes"></div>'
                 fields.each{field->
                     out << '<span class="configpair">'
-                    out << '<span title="">'+field.label.encodeAsSanitizedHTML()+': </span>'
+                    // A truthy/elvis check alone lets a whitespace-only label ("  ") through
+                    // instead of falling back to the key.
+                    def fieldLabel = field.label?.trim() ? field.label : field.key
```
luismalamoc: **Suggestion:** Copilot's earlier comment asked to cover whitespace-only labels in *both* render paths. `pluginPropView.vue` got tests, but this taglib fallback has none, and there's no `PluginTagLib` spec yet. That said, since this is a one-line elvis, I'm fine either way. Nit: `field.label?.trim() ?: field.key` is the idiomatic Groovy form, and the 2-line comment can go.

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/DynamicFormPluginProp.vue
```diff
@@ -1,7 +1,6 @@
 <template>
   <div id="fieldcustomeditor" class="col-sm-12">
     <input ref="hiddenFieldInput" type="hidden" :name="name" />
```
luismalamoc, line 3: **Suggestion (nit):** the `<hr />` removal isn't mentioned in the PR description. Is it intentional (because the description now renders above the widget)? If so, a note in the description would help QA.

luismalamoc (changes requested):
Thanks for the thorough follow-through on every Copilot round. The core fix (`@change` → `@input`) and the autocomplete `dispatchEvent` bridge are right.

**Blockers**
1. The keystroke regression test doesn't guard the bug: VTU's `setValue()` fires `change` too, so it passes against the pre-fix code (inline on `DynamicFormPluginProp.spec.ts`).
2. Ticket/review references (`Copilot review on RUN-4980`) in file contents, including two test names. Please keep them in PR titles and commits only.

**Main ask: less prose.** Six review rounds left a lot of multi-line comments that retell *why the code changed* ("used to", "previously", "Copilot review"). Please cut code and test comments to one line about what the code does, or delete them. The commit messages already carry the history.

**Simplifications (suggestions)**
- `useId()` (Vue 3.5) instead of the `idPrefix` prop + `fieldIdPrefix` + sanitizing + 2 tests.
- Drop the load-time label normalization: all renderers already fall back to the key, and it emits on mount (marks the job config changed on open).
- Drop the `jobedit.js` source-slicing test; it breaks on unrelated refactors.
- Small ones inline: trim the stored key, dead `selectedField` check, redundant `aria-required`, `isDynamicForm` computed.

The layers added across rounds overlap (the label fallback now lives in five places). Removing a few of them makes this PR smaller *and* easier to trust.

luismalamoc:
One process tip for the next revision and future PRs: consider running these two Claude Code skills before opening or updating a PR:

- **[caveman](https://github.com/juliusbrussee/caveman)** for terse, to-the-point English in code comments, test descriptions, commit messages and PR text. It would help a lot with the verbose, history-style comments flagged in this review.
- **[ponytail](https://github.com/DietrichGebert/ponytail)** (`/ponytail-review`) to catch over-engineered implementations early, such as the `idPrefix` plumbing when `useId()` already exists, or overlapping fallback layers. Its focus is "what can we delete", which pairs well with Copilot's review, since Copilot tends to push toward adding more.

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/DynamicFormPluginProp.vue
ddarby-hike: Yes, it was intentional and I added a note in the PR about thank.

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/tests/DynamicFormPluginProp.spec.ts
ddarby-hike: Fixed: the test now sets the element value directly and triggers only `input` (not `setValue()`, which also fires `change`). Verified it fails against the pre-fix `@change` handler and passes with the fix. dc9527a3356433e95637a637d749b531fcadd797

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/tests/DynamicFormPluginProp.spec.ts
ddarby-hike: Fixed: removed the ticket/review references from test names and comments in this file, and trimmed the surrounding comments to one line or dropped them. dc9527a3356433e95637a637d749b531fcadd797

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/tests/pluginPropEdit.spec.ts
ddarby-hike: Fixed: removed the ticket/review reference and shortened the comment. dc9527a3356433e95637a637d749b531fcadd797

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/tests/pluginPropView.spec.ts
ddarby-hike: Fixed: dropped the ticket reference from both test names. dc9527a3356433e95637a637d749b531fcadd797

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/DynamicFormPluginProp.vue
ddarby-hike: Fixed: replaced the `idPrefix` prop + `fieldIdPrefix` computed with `useId()`, and removed the `:id-prefix` binding in pluginPropEdit.vue. Collapsed the two id-collision tests into one that mounts two instances under the same app, since `useId()` scopes uniqueness per app instance. dc9527a3356433e95637a637d749b531fcadd797

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/DynamicFormPluginProp.vue
ddarby-hike: Fixed: removed the load-time label normalization and its test, and made the template fallback trim-aware (`field.label?.trim() || field.key`) to match the other renderers. dc9527a3356433e95637a637d749b531fcadd797

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/DynamicFormPluginProp.vue
ddarby-hike: Fixed: the Field Key is now trimmed once, and that value is reused for both the validation check and the stored key/label fallback. dc9527a3356433e95637a637d749b531fcadd797

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/DynamicFormPluginProp.vue
ddarby-hike: Fixed: removed the dead `if (this.selectedField !== null)` check. dc9527a3356433e95637a637d749b531fcadd797

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/DynamicFormPluginProp.vue
ddarby-hike: Fixed: removed the redundant `aria-required="true"`. dc9527a3356433e95637a637d749b531fcadd797

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/DynamicFormPluginProp.vue
ddarby-hike: Fixed: trimmed the history-retelling comments throughout this file down to one line or removed them. dc9527a3356433e95637a637d749b531fcadd797

rundeckapp/grails-spa/packages/ui-trellis/src/app/utilities/locales/en_US.js
ddarby-hike: Fixed: removed the comment. dc9527a3356433e95637a637d749b531fcadd797

rundeckapp/grails-spa/packages/ui-trellis/tests/unit/jobeditAutocomplete.spec.ts
ddarby-hike: Fixed: deleted the file. dc9527a3356433e95637a637d749b531fcadd797

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/pluginPropEdit.vue
ddarby-hike: Fixed: added an `isDynamicForm` computed and used it at both call sites. dc9527a3356433e95637a637d749b531fcadd797

rundeckapp/grails-app/taglib/rundeck/PluginTagLib.groovy
ddarby-hike: Fixed: switched to `field.label?.trim() ?: field.key` and removed the 2-line comment. dc9527a3356433e95637a637d749b531fcadd797

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/DynamicFormPluginProp.vue
```diff
@@ -322,8 +378,9 @@ export default defineComponent({
         }
 
         field = {
-          key: this.newField,
-          label: this.newLabelField,
+          key,
```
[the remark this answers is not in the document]
ddarby-hike: Fixed: the generated description now uses the trimmed `key` instead of the untrimmed `this.newField`, so it matches the stored key. dec9de15ad

rundeckapp/grails-spa/packages/ui-trellis/src/library/components/plugins/DynamicFormPluginProp.vue
```diff
@@ -64,11 +63,18 @@
...
           data-testid="duplicate-warning"
-          ref="duplicateWarningRef"
           ><b>Warning!</b> {{ $t("message_duplicated") }}.</alert
         >
+        <alert
+          v-if="invalidKey"
+          ref="invalidKeyWarningRef"
+          type="warning"
+          data-testid="invalid-key-warning"
+          ><b>Warning!</b> {{ $t("message_fieldKeyRequired") }}.</alert
```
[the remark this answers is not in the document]
ddarby-hike: Fixed: moved "Warning!" into the locale catalogue as `message_warning` and render it through `$t()` in both alerts. dec9de15ad

ddarby-hike:
Fixed: `useId()` is only unique within a single Vue app, and the legacy dynamic-form entry point (`src/app/pages/dynamic-form/main.js`) creates a separate `createApp()` per `.dynamic-form-vue` element. Set a distinct `app.config.idPrefix` per mount so ids stay unique even with multiple dynamic-form properties on one page. 4aa970795f

luismalamoc (approved):
All blockers and suggestions are addressed. Thanks for the quick, thorough turnaround. The diff reads much cleaner now.

Non-blocking follow-up: `useId()` is unique only per Vue app. You handled that in `dynamic-form/main.js`, but `job/editor/main.js` (one `createApp` per section), `project-config/main.js` and `project-nodes-config/main.js` also mount several apps per page without an `app.config.idPrefix`. If two sections render a Custom Fields widget, their modal ids can collide. The impact is a11y only (`for`/`aria-describedby`), and the fix is a one-liner per `createApp`. Fine to do it here or in a follow-up.

rundeckapp/grails-spa/packages/ui-trellis/src/app/pages/dynamic-form/main.js
```diff
@@ -26,5 +26,8 @@ for (let i = 0; i < els.length; i++) {
   });
   app.use(VueCookies);
   app.use(i18n);
+  // useId() is only unique within one app; each .dynamic-form-vue element gets
+  // its own app instance, so give each a distinct prefix to avoid id collisions.
+  app.config.idPrefix = `dynamic-form-${i}`;
```
luismalamoc: **Suggestion (non-blocking):** the same per-app prefix is needed wherever a page mounts several apps that can render `DynamicFormPluginProp`:

- `job/editor/main.js:102` (one `createApp` per section, inside `section.elements.forEach`)
- `project-config/main.js:20`
- `project-nodes-config/main.js:29`

Without it, two sections that each render a Custom Fields widget both get ids starting at `v-0-…`, and the modal's `for`/`aria-describedby` can point to the other section's input. Fix is one line after each `createApp`, e.g.:
```js
app.config.idPrefix = `job-editor-${section.name}-${index}`;
```
Fine here or as a follow-up.

luismalamoc (commented):
Inline detail for the non-blocking follow-up mentioned in the approval.

rundeckapp/grails-spa/packages/ui-trellis/src/app/pages/dynamic-form/main.js
ddarby-hike: Fixed: added a distinct `app.config.idPrefix` after each `createApp()` in `job/editor/main.js`, `project-config/main.js`, and `project-nodes-config/main.js`. 0d2ade27d9

PR #10622, opened 2026-09-17
[RUN-4870] Run prettier format:check in the T Unit UI job before Jest

luismalamoc:
## Tell us about your PR

CI enhancement. Since `prettier/prettier` became an ESLint `error` in `ui-trellis`, an unformatted file fails the webpack build, but that feedback only arrives from the hour-long `Build` CircleCI job (`eslint-webpack-plugin`, `failOnError`). The package already ships a `format:check` npm script (`prettier --check "src/**/*.{ts,tsx,js,vue,scss,css}"`) that nothing in CI called.

`rundeck_gui_tests` in `scripts/circleci/build-functions.sh`, which backs the root-level parallel `T Unit UI` job, now runs `npm ci`, then `npm run format:check`, then `npm run test:unit` instead of `npm run ci:test:unit`. Same CircleCI step, single `npm ci`, no new job, no `config.yml` change, no new dependency (`prettier` 3.9.6 is already a devDependency). `setup.sh` runs with `set -ueo pipefail`, so a non-zero `prettier --check` fails the step. The `ci:test:unit` npm script is untouched (still used by the Gradle `runNpmTest` task).

Result: an unformatted file in `ui-trellis/src` fails `T Unit UI` within the first minutes of the workflow instead of at the end of `Build`. `T Unit UI` stays at the root of `build_and_test` with no `requires`, so the critical path is unchanged.

- A separate CircleCI step named "Format check": more visible in the UI, but it would need a second `npm ci` (or splitting install from tests). Not worth ~1-2 min extra in the job.
- `prettier --check .` instead of the `src/` glob: would cover config and test files outside `src/` that were never reformatted, so `main` would fail on day one.
- Editor format-on-save config or a pre-commit hook: dropped from the ticket, each developer configures their own editor.

**Additional context**

Verified locally on this branch: `npm ci && npm run format:check` exits 0; adding a probe file with a single-quoted import makes `format:check` exit 1 naming the file. `bash -n` passes on `build-functions.sh`.

## Release Notes

None, CI-only change.


smartinellibenedetti:
Closing this PR for now as there's a bug that once solved might make this change unnecessary.

PR #10623, opened 2026-09-22
[RUN-0000] Convert user/list.gsp to Vue behind nextUi

smartinellibenedetti:
## Tell us about your PR

Enhancement. Converts the legacy `/user/list` (User Profiles) admin page from GSP/KO to Vue, following the project's standard gsp-to-vue conversion pattern.

- Rebuilt the page as a Vue app (`UserListPage.vue` → `UserTable.vue` → `UserListRow.vue`/`UserDetailPanel.vue`), preserving the legacy page's exact behavior: user list with edit/new-profile links, per-row expandable detail panel, and a Groups column (with help tooltip) shown only for the logged-in user's own row.
- Gated the Vue rendering behind the existing `nextUi` URL-param/cookie/system-flag mechanism (`uiType == 'next'` in the GSP) instead of a dedicated one-off feature flag, so it's consistent with how the rest of the app's converted pages are gated. Legacy KO/GSP rendering remains the server default.
- Registered `user:list` in `NextUiInterceptor` so the page correctly picks up the site-wide `nextUiMode` flag and the per-user `nextUi` cookie preference, not just an explicit `?nextUi=true` query param.
- Added Jest coverage for every new Vue component, and Selenium coverage (`UserListSpec`, `@UiModeFlag(status = NEXT_UI)`) driving both the `nextUi=true` and default (legacy) rendering paths via `UiModes.nextUiAndDefault()`.

An earlier iteration of this conversion used a dedicated app-wide `vueUserList` feature flag (always-on, no per-request override). That was replaced with the standard `nextUi` mechanism used elsewhere in the app, since it gives per-user/per-request control and matches the established pattern other converted pages use, rather than introducing a one-off flag.

**Additional context**
Verified live in a running instance (both `?nextUi=true` and the legacy default path, including toggling the `nextUi` cookie) — expander/detail-panel/tooltip/NOT SET behavior is identical between the two rendering paths, and no new console/network errors were introduced.

**Open question — page discoverability**
While researching this page's history for this conversion, we found `/user/list` has had no link pointing to it anywhere in the app since a nav-cleanup commit in 2013 (`037c71d8d7`, "Remove link to user profiles") — it's only ever been reachable by direct URL since then. It was never formally deprecated or slated for removal; the route, controller, and view kept working and getting incidental maintenance the whole time.

We also confirmed it isn't redundant with anything else in OSS: `menu/userSummary` (the page that later took over its old nav slot) is read-only monitoring/activity data with no edit or create capability, and `/user/profile` is self-service (own account only). `/user/list` + `/user/edit` + `/user/create` remain the only OSS pages that let an app-admin edit another user's profile or pre-create a placeholder profile before that user's first login.

Given that, this PR converts the page as-is rather than deleting it, but leaves open whether it should be re-linked into the nav for discoverability, or intentionally left as a direct-URL-only admin utility. Flagging for maintainer input before merge.

## Release Notes

Converted the `/user/list` (User Profiles) admin page to Vue, behind the existing Next UI early-access toggle.


smartinellibenedetti:
Closing for now — holding off while we evaluate whether this page should stay, be merged into User Manager, or be retired. Will reopen or re-submit once that's settled.

PR #10628, opened 2026-09-22
[RUN-4950] Fix cross-project execution log disclosure via project import

ncofreortiz-hub:
## Tell us about your PR

Security bugfix (RUN-4950, HackerOne #3960872 / PS-1700, CWE-863). Rundeck project import only rewrote an imported execution's `outputfilepath` to a project-scoped value when the referenced log file was physically present in the import archive. If a crafted archive referenced a path not in the archive (e.g. an absolute path to another project's log file on disk), the raw archive value was persisted unchanged, and import still reported success. `LogFileStorageService` then served that stored path directly at read time with no containment check and no ownership verification — only the requester's own-project ACL was checked. Net effect: any user with import+admin rights on one project could read arbitrary other projects' execution logs (which routinely contain secrets printed to stdout).

Two independent layers of defense:

1. `ProjectService.importExecutionsToProject`: when the archive-supplied `outputfilepath` isn't backed by a file actually present in the archive (and isn't a remote-storage pointer), the raw value is now dropped (set to `null`) instead of being persisted verbatim. Existing error reporting (`execerrors` / logs) for the "no matching outfile" case is unchanged.
2. `LogFileStorageService.getFileForExecutionFiletype`: added a containment check (`isStoredPathSafeForProject`) that rejects a stored `outputfilepath` which canonicalizes into a *different* project's directory under the shared local logs root, falling back to the normal generated path for the execution's own project instead. A path outside the shared logs root entirely (legacy pre-1.6 custom storage locations) is left untouched, preserving existing supported behavior.

Added regression tests (TDD — confirmed RED before the fix in both cases):
- `ProjectServiceSpec`: a forged execution XML with an absolute `outputfilepath` pointing outside the importing project, with no matching file in the archive — asserts the persisted `Execution.outputfilepath` is never the forged value.
- `LogFileStorageServiceSpec`: an execution whose stored path resolves into another project's log directory under the shared root — asserts the service falls back to the execution's own generated path.

Considered resolving the archive-supplied path to the importing project's own log-storage root instead of dropping it, but since there's no legitimate local file to move/rename into that location (it wasn't found in the archive), nulling the field is simpler and avoids fabricating a path that still doesn't point at real content. Considered a strict "must be under this project's log dir" containment check at read time for *all* stored paths, but an existing test (`getFileForExecutionFiletype partial`) demonstrates intentional support for arbitrary custom storage locations from pre-1.6 upgrades outside the standard project directory structure; the fix scopes the containment check to the shared logs root specifically, so it closes the cross-project traversal without breaking that legacy behavior.

**Additional context**

Root cause locations from the report:
- `Execution.groovy:451` — `outputfilepath` copied verbatim from archive XML (unchanged; this is also used for normal, non-import execution creation)
- `ProjectService.groovy:1838-1882` — path only rewritten if a matching log file is found in the zip; otherwise the raw value was persisted (fixed)
- `LogFileStorageService.groovy:1188-1310` — stored path used directly at read time, no containment/ownership check (fixed)

Full build (`./gradlew build -x check`) and both touched Spock spec files pass in full, with no new complexity or lint concerns.

## Release Notes

Fixed a security issue where project import could be used to bypass per-project ACLs and read another project's execution logs, by trusting an archive-supplied log file path that wasn't verified against the import archive's actual contents.


Jesus-Osuna-M (changes requested):
Solid approach overall — it closes both layers the ticket's engineering notes called for, and the reasoning in "Alternatives considered" about the pre-1.6 custom storage locations is the right instinct. Two blockers before this can go in, plus three points that need an explicit decision rather than a code change.

Blockers

1. CI is red — T Unit Gradle fails with an NPE in two pre-existing tests

rundeck.services.LogFileStorageServiceTests.testGetFileForExecutionFiletypeLegacyLogFile
rundeck.services.LogFileStorageServiceTests.testGetFileForExecutionFiletypeLegacyStateFile

java.lang.NullPointerException: Cannot invoke method getFrameworkProperties() on null object
  at LogFileStorageService.getLocalLogsDir(LogFileStorageService.groovy:1291)
  at LogFileStorageService.isStoredPathSafeForProject(LogFileStorageService.groovy:1214)
  at LogFileStorageService.getFileForExecutionFiletype(LogFileStorageService.groovy:1192)

This isn't only a test-setup problem. The useStoredPath branch previously had no dependency on frameworkService or on framework.logs.dir; isStoredPathSafeForProject() now calls getLocalLogsDir() on every invocation that has a stored path, and getLocalLogsDir() throws IllegalStateException when framework.logs.dir is unset (LogFileStorageService.groovy:1290-1293). The catch (IOException ignored) covers neither that nor the NPE, so the exception escapes to the caller.

Two suggestions, either is fine:

Only resolve the logs root when it's actually needed, and decide the exception policy deliberately (catch Exception, and document which way it fails) instead of inheriting it from the narrow IOException catch.
Stub frameworkService in the two legacy tests.

Worth noting the pipeline fail-fast left T Functional API and all T Integration * jobs at not_run, so there's no CI signal above the unit layer yet.

2. Wrong Jira key — RUN-4790 instead of RUN-4950

Commit title: [RUN-4790] Fix cross-project execution log disclosure via project import (the PR title is correct)
LogFileStorageServiceSpec.groovy, new test Groovydoc: "Regression test for RUN-4790 / ..."
ProjectServiceSpec.groovy, same

RUN-4790 is a different ticket, so this cross-links the wrong issue and breaks traceability on the security fix.

Needs a decision, not necessarily a change

3. Layer 2 doesn't match acceptance criterion #2 as written. The AC says the resolved path must be validated as within the execution's own project's log directory. LogFileStorageService.groovy:1213-1219 exempts any path that falls entirely outside <framework.logs.dir>/rundeck:

if (!canonicalStored.path.startsWith(rootPrefix)) {
    //outside the shared logs root entirely: not a cross-project path
    return true
}

The reported scenario is covered, and I agree the existing getFileForExecutionFiletype partial case (/tmp/test/logs/blah/1.rdlog, useStored=true) shows the legacy behaviour is intentional — so this is a reasonable trade-off, not a mistake. But it does mean layer 2 is not an independent guarantee, which is what the AC asks for. Either get explicit sign-off on the deviation, or gate the legacy exemption behind a config flag. Please record whichever we pick on RUN-4950.

4. Item 3 of the engineering plan is still open. The raw new File(...) on the stored path remains in ExecutionController.groovy:255 and Execution.outputfilepathSizeAsString() (Execution.groovy:354-355). Both only expose file size rather than content, so severity is low — but if we're deliberately leaving them, let's say so rather than leave it implicit.

5. No functional/API-level regression test, which the ticket asked for alongside the unit tests, and no note yet on AC #4 (verification against the PoC in the linked ticket).

Minor

6. The reflection in ProjectServiceSpec looks avoidable. Groovy ignores private for dynamic dispatch — the existing integration tests call projectService.importExecutionsToProject([temp], [:], "test", null, ...) directly (ProjectServiceTest.groovy:156, 190, 288). The hardcoded getDeclaredMethod(..., ArrayList, Map, Object, IFramework, Object, Object, Map, Object) will break silently (no compile error) on any signature change.

7. Weak assertion. imported.outputfilepath != forgedPath would also pass if the code wrote some other wrong value. The ticket asks for null or the generated path for the destination project; asserting that directly — plus imported.project == 'sandbox' — pins the behaviour down.

8. Misleading warn message. When isStoredPathSafeForProject returns false from the exception path, the log still says "resolves into another project's log directory", which won't be what happened.

9. No save(flush: true) after e.outputfilepath = null. Low risk — GORM dirty-checking flushes it at session commit, same as the existing happy path — but the ticket's notes called for it explicitly, so worth a conscious "not needed because ..." if we skip it.

10. Branch is 2 commits behind main — worth a rebase before merge.

PR #10632, opened 2026-09-22
[RUN-4976] Recover legacy Jasypt content that is Base64-encoded or stale-flagged plaintext

ncofreortiz-hub:
## Tell us about your PR

Bugfix. [RUN-4976](https://pagerduty.atlassian.net/browse/RUN-4976): a customer (MeteoSwiss) upgrading Rundeck 5.20.1 -> 6.2.0 hit a crash-loop:

```
org.rundeck.plugin.encryption.EncryptionException: Legacy Jasypt decryption failed
Caused by: javax.crypto.IllegalBlockSizeException: last block incomplete in decryption
Caused by: org.bouncycastle.crypto.DataLengthException: last block incomplete in decryption
```

Two distinct, interacting defects:
1. `LegacyJasyptDecryptor` always treats stored bytes as raw binary `[salt][ciphertext]`. The customer's `scm-export.properties` bytes aren't block-aligned once interpreted that way, so BouncyCastle throws.
2. Separately, the customer's `project.properties` rows are stored as genuine plaintext but still carry a stale `jasypt-encryption:encrypted=true` flag inherited from the old 5.x plugin. `ModernEncryptionConverterPlugin` unconditionally tries to decrypt anything flagged as encrypted, so it also throws on already-plaintext content.

Both exceptions propagated uncaught through `ProjectManagerService.loadProjectConfigResource()`, breaking the cached project loader (crash-loop at boot, or repeated request failures for that project).

- `LegacyJasyptDecryptor.decrypt()`: when the raw-binary interpretation fails, fall back to Base64-decoding the input and retrying before giving up. Recovers this data shape regardless of exactly how it got that way.
- `ModernEncryptionConverterPlugin.decryptLegacy()`: when legacy decryption still fails and the raw bytes look like plaintext (printable ASCII, e.g. starting with `#`), return the content unchanged with a `WARNING` log instead of throwing. This targets the stale-flag case without touching the (separate, out-of-scope) metadata-merge hazard that likely caused the stale flag in the first place.
- `ProjectManagerService.loadProjectConfigResource()`: if `project.properties` still can't be read/decrypted after the above, log a `WARNING` and degrade to an empty config instead of propagating the exception and breaking the whole project load.
- SCM config (`scm-export.properties`/`scm-import.properties`) is intentionally **not** given the same swallow-and-continue treatment: if genuinely undecryptable after the Base64 fallback, it still fails loudly, since SCM integration state should not silently disappear.
- Added regression tests generated with the **real** `org.jasypt` library (`StandardPBEByteEncryptor`), closing a test-coverage gap where existing tests only validated self-consistency against Rundeck's own `LegacyJasyptEncryptor` reimplementation.
- Added a test proving genuinely corrupt/undecryptable content still throws unchanged (confirms SCM config stays loud).

- Pausing all code changes until the customer's exact pre-upgrade `encryptorType`/algorithm/provider config is confirmed. Rejected: the Base64 fallback and plaintext-detection guard are both safe, low-risk, non-breaking additions that unblock the customer's upgrade now, independent of the still-open question of exactly why their data ended up in this shape (tracked as a follow-up comment on the ticket).
- Fixing `GormStorageDataProvider.update()`'s metadata merge-not-replace behavior, a plausible related mechanism for stale encryption flags. Left out of scope for this fix; noted in the ticket for follow-up.

**Additional context**

Full spec and root-cause writeup: see the RUN-4976 Jira ticket comments. Related: RUN-4512 (legacy-flag-vs-aes-gcm-flag precedence, a different variant of the metadata/content-mismatch class) and RUN-4336 (removed the real `org.jasypt` dependency in favor of the `encryption-core` reimplementation).

## Release Notes

Fixed an issue where upgrading from a pre-6.x Rundeck install could crash-loop or silently drop project configuration when legacy Jasypt-encrypted storage content was Base64-encoded, or when stale encryption metadata flags were left over from an older Rundeck version.

[RUN-4976]: https://pagerduty.atlassian.net/browse/RUN-4976?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rundeckapp/grails-app/services/rundeck/services/ProjectManagerService.groovy
```diff
@@ -459,9 +459,23 @@ class ProjectManagerService implements ProjectManager, ApplicationContextAware,
...
+            //reading the resource throw (e.g. legacy Jasypt decrypt failure) even after the
+            //storage-layer plaintext fallback. Degrade to an empty config instead of breaking
+            //the whole project load.
+            log.warn("Failed loading project properties from storage: ${resource.path}: could not read/decrypt contents, continuing with empty config: " + e.message, e)
+            return new LoadedConfig(
+                    config      : properties,
+                    lastModified: resource.contents.modificationTime,
+                    creationTime: resource.contents.creationTime
+            )
+        }
```
luismalamoc, line 478: **Blocker — this catch opens a permanent data-loss path.**

`mergeProjectProperties()` and `setProjectProperties()` both read `oldprops` through this same `loadProjectConfigResource()` and then call `storeProjectConfig()`. With this catch in place:

1. Any decrypt failure — a misconfigured converter password, a corrupt AES-GCM record, not just the legacy stale-flag case — silently yields an empty config.
2. The next write (Edit Nodes, API config PUT, any merge) stores *only the new props*, re-encrypted with AES-GCM.
3. The original, still-recoverable config is overwritten for good.

Before this PR the exception blocked that write, so the data stayed intact. This is the same silent-empty-config loss described in Finding 4 of the ticket, now widened to every encrypted project.

The ticket's actual `project.properties` case (plaintext content + stale `jasypt-encryption:encrypted` flag) is already handled by the plaintext fallback in `ModernEncryptionConverterPlugin.decryptLegacy()`, so this catch isn't needed for the fix. Suggest removing it (and its spec) and letting genuine decrypt failures keep failing loudly.

plugins/aes-gcm-encryption-plugin/src/main/java/org/rundeck/plugin/encryption/ModernEncryptionConverterPlugin.java
```diff
@@ -216,6 +230,28 @@ protected byte[] transform(byte[] data) {
...
+        for (int i = 0; i < sampleSize; i++) {
+            int b = data[i] & 0xFF;
+            if (b == '\n' || b == '\r' || b == '\t') {
+                continue;
+            }
+            if (b < 0x20 || b > 0x7E) {
+                return false;
+            }
+        }
+        return true;
```
luismalamoc, line 252: **Blocker — the printable-ASCII heuristic also matches Base64, which breaks the "SCM config stays loud" guarantee.**

The Base64 alphabet is entirely printable ASCII, so any Base64 payload passes `looksLikePlaintext()`. Take a Base64-shaped `scm-export.properties` (exactly the shape this PR targets) with a wrong password or algorithm:

1. The raw decrypt fails, and the Base64-decode retry in `LegacyJasyptDecryptor` also fails.
2. `looksLikePlaintext()` returns `true`, so the raw Base64 text is returned as "plaintext".
3. `ScmPluginConfig.loadFromStream()` (`PluginConfigService.loadScmConfig`) parses it as a single key with an empty value.
4. SCM ends up silently disabled, with no error.

The smaller fix is to replace the 512-byte loop with a check on the properties header:

```java
private static boolean looksLikePlaintext(byte[] data) {
    return data != null && data.length > 0 && data[0] == '#';
}
```

Both `project.properties` (`#text/x-java-properties`) and `Properties.store()` output start with `#`, and `#` is not in the Base64 alphabet. Keep the check *after* the decrypt failure, as it is now: a real ciphertext whose salt starts with `#` (1/256) would otherwise be returned raw.

plugins/aes-gcm-encryption-plugin/src/test/groovy/org/rundeck/plugin/encryption/BackwardCompatibilityIntegrationSpec.groovy
```diff
@@ -358,6 +358,49 @@ class BackwardCompatibilityIntegrationSpec extends Specification {
...
+        then: "the plaintext is recovered unchanged instead of throwing"
+        readAllBytes(result) == plaintextProperties
+    }
+
+    def "RUN-4976: genuinely corrupt, non-plaintext-shaped content still throws (SCM config stays loud)"() {
+        given: "corrupt/undecryptable binary content, simulating a real scm-export.properties failure"
+        def plugin = createModernPlugin()
+        def path = Mock(Path)
+        def corrupt = new byte[40]
+        Arrays.fill(corrupt, (byte) 0xFF)
```
luismalamoc, line 392: **Blocker (test gap for the comment on `looksLikePlaintext`)** — `0xFF` bytes are non-printable, so this test never reaches the case that actually leaks. Please add a case with a *valid Base64* legacy payload decrypted with the wrong password, and assert that reading it still throws. It fails with the current heuristic and passes with the `data[0] == '#'` check.

plugins/aes-gcm-encryption-plugin/src/test/groovy/org/rundeck/plugin/encryption/LegacyJasyptDecryptorSpec.groovy
```diff
@@ -196,4 +211,62 @@ class LegacyJasyptDecryptorSpec extends Specification {
...
     }
+
+    /**
+     * RUN-4976: MeteoSwiss's scm-export.properties bytes are not block-aligned as raw
+     * Jasypt binary output, but decrypt correctly once Base64-decoded first. Reproduces
+     * the "last block incomplete in decryption" crash-loop with a Base64-wrapped payload.
+     */
+    def "decrypt recovers Base64-encoded legacy Jasypt payload via fallback"() {
+        given:
+        def password = "meteoswiss-scm-password"
```
luismalamoc, line 222: **Blocker — customer name in a public OSS repo.** `MeteoSwiss` appears here (comment + `"meteoswiss-scm-password"` fixture) and in `BackwardCompatibilityIntegrationSpec` (Scenario 7 comments). This repo is public, so please use neutral wording (e.g. "a 5.x upgrade where...") and a generic fixture password. The PR description also names the customer and case details and should be scrubbed as well.

Same for the `RUN-4976` references in code comments, test names and log messages: ticket keys belong in the PR title/commits, not in file contents.

plugins/aes-gcm-encryption-plugin/src/main/java/org/rundeck/plugin/encryption/ModernEncryptionConverterPlugin.java
```diff
@@ -207,7 +207,21 @@ private HasInputStream decryptLegacy(HasInputStream input) {
...
+                        return getLegacyDecryptor().decrypt(getResolvedPassword(), data);
+                    } catch (EncryptionException e) {
+                        if (looksLikePlaintext(data)) {
+                            // Pre-existing 5.x condition (RUN-4976): the jasypt-encryption:encrypted
+                            // flag was set on this record, but the content was never actually
+                            // encrypted. Decrypting already-plaintext bytes can never succeed, so
+                            // recover the original content instead of failing every read.
+                            logger.warn("readResource: content flagged jasypt-encryption:encrypted=true "
+                                    + "failed to decrypt but looks like plaintext; returning raw content "
+                                    + "unchanged (see RUN-4976).");
```
luismalamoc, line 220: Two things on this log line:

- **Blocker (with the customer-name comment):** `(see RUN-4976)` in a production log message. Please drop the ticket reference.
- **Suggestion:** this fires on *every* read. The project cache refreshes every minute and the stale flag isn't cleared until the next write, so you get one WARN per project per minute, forever. That's far less than the 155k lines from the ticket, but still noise. Consider logging the path once (e.g. a `Set` of already-warned paths) or logging at debug after the first warn.

plugins/aes-gcm-encryption-plugin/src/test/groovy/org/rundeck/plugin/encryption/LegacyJasyptDecryptorSpec.groovy
```diff
@@ -196,4 +211,62 @@ class LegacyJasyptDecryptorSpec extends Specification {
...
+        def plaintext = "scm.export.enabled=true\nscm.export.branch=main".bytes
+        def rawEncrypted = jasyptEncrypt(plaintext, password, "PBEWITHSHA256AND128BITAES-CBC-BC", "BC", 1000)
+        def base64Wrapped = Base64.encoder.encode(rawEncrypted)
```
luismalamoc, line 225: **Suggestion:** this fixture is our own `LegacyJasyptEncryptor` output wrapped in `Base64.encoder`. Now that `org.jasypt` is on the test classpath, generating it with the real `StandardPBEStringEncryptor` (the triage hypothesis for the Base64 shape) is a one-line change and proves compatibility with actual library output rather than with our own assumption of it.

encryption-core/build.gradle
```diff
@@ -32,6 +32,7 @@ repositories {
 
 dependencies {
     implementation("org.bouncycastle:bcprov-jdk18on:${bouncyCastleVersion}")
+    implementation("org.slf4j:slf4j-api:${slf4jVersion}")
```
luismalamoc, line 35: **Suggestion:** this new `slf4j-api` dependency on `encryption-core` exists only for the single `warn` in `LegacyJasyptDecryptor`. The caller (`ModernEncryptionConverterPlugin`) already has a logger, so the decryptor could stay log-free (e.g. the caller logs when a Base64 recovery happens, or the decryptor exposes that it used the fallback). That keeps the core library dependency-free.

luismalamoc (changes requested):
Thanks for the thorough investigation. The Base64 fallback in `LegacyJasyptDecryptor` looks right, and adding real-`org.jasypt` fixtures closes a real coverage gap. Requesting changes for three blockers (details inline):

1. **`ProjectManagerService.loadProjectConfigResource()` catch**: swallowing decrypt failures into an empty config lets the next `mergeProjectProperties`/`setProjectProperties` overwrite the original config permanently, and this applies to any encrypted project (e.g. a wrong converter password), not just the legacy case. The plaintext fallback in the plugin already covers the ticket's `project.properties` scenario, so please remove this catch.
2. **`looksLikePlaintext()` matches Base64**: a Base64-shaped SCM config that still fails to decrypt gets returned as raw text and SCM is silently disabled, which contradicts the "SCM stays loud" guarantee. A `data[0] == '#'` check is smaller and excludes Base64. Please also add a test for a valid-Base64 payload with the wrong password.
3. **Customer name and ticket refs in a public repo**: please remove the customer name (tests, fixtures, PR description) and the `RUN-4976` refs in code comments, test names and log messages.

The suggestions (real `StandardPBEStringEncryptor` fixture, log-once, no slf4j in `encryption-core`) are non-blocking.

On scope: once 1 is removed and 2 is simplified, the diff actually gets smaller.

luismalamoc:
One more note on the customer-name blocker: please **squash-merge** this PR once the fixes are in. Otherwise the intermediate commits, including the ones that still contain the customer name in code, tests and fixtures, end up in `main`'s history. Removing the name in a follow-up commit is not enough on its own.

Also worth noting: the existing commit message (`ce733deb19`) names the customer in its body. GitHub's default squash message concatenates the individual commit messages, so when squashing, please edit the final commit message to remove the customer name as well. Rewriting the branch history (amend/rebase + force-push) before merging would also keep it out of the PR's commit list.

rundeckapp/grails-app/services/rundeck/services/ProjectManagerService.groovy
ncofreortiz-hub: Removed. You're right that this widened the blast radius to any encrypted project, not just the stale-flag case -- dropped the catch (and its test) in d83424a041, back to propagating decrypt failures loudly. The plaintext fallback in `ModernEncryptionConverterPlugin` already covers the ticket's actual project.properties scenario.

plugins/aes-gcm-encryption-plugin/src/main/java/org/rundeck/plugin/encryption/ModernEncryptionConverterPlugin.java
ncofreortiz-hub: Good catch. Replaced with the `data[0] == '#'` check you suggested in d83424a041, and added a regression test (a Base64-wrapped legacy payload with the wrong password) proving it now still throws instead of being returned as plaintext.

plugins/aes-gcm-encryption-plugin/src/test/groovy/org/rundeck/plugin/encryption/BackwardCompatibilityIntegrationSpec.groovy
ncofreortiz-hub: Added in d83424a041: `BackwardCompatibilityIntegrationSpec` now has a case that encrypts with a different password than the plugin is configured with, Base64-wraps it, and asserts the read still throws.

plugins/aes-gcm-encryption-plugin/src/test/groovy/org/rundeck/plugin/encryption/LegacyJasyptDecryptorSpec.groovy
ncofreortiz-hub: Scrubbed in d83424a041 -- removed the customer name and case-specific fixture password from both spec files and the PR description, and dropped the in-code RUN-4976 references (kept only in the commit/PR).

plugins/aes-gcm-encryption-plugin/src/main/java/org/rundeck/plugin/encryption/ModernEncryptionConverterPlugin.java
ncofreortiz-hub: Both addressed in d83424a041: dropped the ticket reference from the log line, and added a per-path dedupe set so the WARNING logs once per path instead of every read (subsequent reads log at debug).

plugins/aes-gcm-encryption-plugin/src/test/groovy/org/rundeck/plugin/encryption/LegacyJasyptDecryptorSpec.groovy
ncofreortiz-hub: Done in d83424a041 -- the Base64-fallback test fixture is now generated with the real `org.jasypt` library's `StandardPBEStringEncryptor` rather than hand-wrapping our own reimplementation's output in Base64.

encryption-core/build.gradle
ncofreortiz-hub: Agreed, done in d83424a041 -- removed the `slf4j-api` dependency and the logger from `encryption-core`. The decryptor no longer logs; the caller (which already had a logger) owns the plaintext-fallback WARNING.

PR #10633, opened 2026-09-22
[RUN-4996] Fix grails.serverURL not visible early enough to link-generating beans

### LINKED ISSUE #10352: Rundeck 6.0.1: Broken link in Email notification
**Describe the bug**
After upgrading from Rundeck 5.20 to 6.0.1 I noticed that the link to the execution in the notification mail is broken. The link is `http://project/$MYPROJECT/execution/show/$EXECUTIONID`.

In other words: The host is missing from the url and the protocol is `http` eventhough my instance uses `https`.

**My Rundeck detail**

* Rundeck version: 6.0.1 on Temurin Java 25
* install type: deb
* OS Name/version: Debian Bookworm
* DB Type/version: postgres

**To Reproduce**
Just configure a notification mail for a job and check the link in the mail received.
I got the issue for both success and failure mail.

**Expected behavior**
The mail should contain a valid link to the execution.

**Desktop:**
 - OS: Debian Testing
 - Browser Firefox
 - Version 140.12.0esr

fdevans:
## Tell us about your PR

Bugfix. Fixes #10352.

`grails.serverURL` set only via a custom `rundeck-config.properties`/`.groovy` file (i.e. via `-Drundeck.config.location`, not `application.yml`) was invisible to any bean that reads config eagerly at bean-definition build time — most notably the default `grailsLinkGenerator` bean. This is because that registration previously only happened from `Application#setEnvironment()`, an ordinary `EnvironmentAware` callback that fires only once `Application` itself is instantiated as a bean — which is after Grails has already built its plugin bean definitions during `ApplicationContext` refresh.

Since notification emails are sent asynchronously with no live HTTP request to fall back on for host/scheme, generated links lost their host and scheme entirely when `grails.serverURL` was set this way.

Added `RundeckConfigEnvironmentPostProcessor`, a Spring Boot `EnvironmentPostProcessor` registered via `META-INF/spring.factories`, which performs the same property-source registration during Spring Boot's `prepareEnvironment()` phase — a phase guaranteed to complete before any `ApplicationContext` or bean definitions exist.

The registration logic itself was extracted out of `Application#setEnvironment()` into a new static `Application.loadRundeckPropertySources(Environment)` method, called from both the new post-processor and the existing `setEnvironment()` callback (kept for backward compatibility — re-registering an identically-named property source via `addFirst()` is a safe, idempotent no-op).

Added a regression test (`RundeckConfigEnvironmentPostProcessorTest`) that drives a real `GrailsApp` through the actual Spring Boot startup sequence up through environment preparation, using a temp `rundeck-config.properties` file with a custom `grails.serverURL`, and asserts the property is visible at that point — before any bean definitions could have read it. Verified empirically both ways: with the post-processor disabled, the test fails (`grails.serverURL` reads `null` at that point, reproducing the issue); with it enabled, the test passes. The existing `Application` test suite (9 tests) still passes unchanged.

**Behavior change:** `loadRundeckPropertySources()` now aborts startup (`IllegalStateException`) if pre-bootstrap initialization fails, instead of silently continuing on a partially-initialized config. Previously an operator with a broken config could see a confusing downstream NPE instead of the actual cause; they now get a clear, immediate failure.

Considered moving all of `setEnvironment()`'s work earlier via other lifecycle hooks, but `EnvironmentPostProcessor` is the standard, documented Spring Boot extension point specifically for this purpose (see e.g. Spring Boot's own `ConfigDataEnvironmentPostProcessor`), and keeping `setEnvironment()`'s existing registration intact minimizes risk/blast radius rather than removing it outright.

**Additional context**
This isn't specific to `.properties` vs `.groovy` config format — it's a bootstrap-ordering issue affecting anything set only via the config-location file rather than `application.yml`. Root-cause and verification details also posted to #10352.

## Release Notes

Fixed an issue where `grails.serverURL` set via a custom `rundeck-config.properties`/`.groovy` file was not applied early enough in startup, causing notification email links to be generated without a host or scheme. Startup now also fails fast with a clear error if pre-bootstrap initialization fails, rather than continuing on a broken config.


Jesus-Osuna-M (changes requested):
A few things I'd tighten before merge:

**Javadoc is too long.** The `loadRundeckPropertySources` doc (~30 lines) and the `RundeckConfigEnvironmentPostProcessor` class doc (~30 lines) cover substantially the same ground, and the rest of `Application.groovy` uses 3–5 line Javadoc. Keep the full explanation in one place and cross-reference it from the other.

**The test can hide the real failure cause.** `RundeckConfigEnvironmentPostProcessorTest.groovy:116` — `catch (Throwable ignoredAbort) {}` swallows everything. If the post-processor itself throws later on (e.g. the new `IllegalStateException`), the test fails on `captor.capturedEnvironment != null` with no cause attached. Capture the throwable and rethrow it when `capturedEnvironment == null`.

**Pin the captor's ordering explicitly.** It currently works because an unordered `ApplicationListener` defaults to `LOWEST_PRECEDENCE` and so runs after `EnvironmentPostProcessorApplicationListener` (`HIGHEST_PRECEDENCE + 10`) on the same event. Correct, but implicit — having `EnvironmentCaptor` implement `Ordered` returning `LOWEST_PRECEDENCE` makes that intentional rather than incidental.

**Temp dir leak.** `File.createTempDir()` + `deleteOnExit()` won't remove a non-empty directory. Add `tmpCfgDir.deleteDir()` to the `cleanup:` block.

**Hardcoded property-source name.** The first test asserts the literal `"rundeck.config.location"`; use `RundeckInitConfig.SYS_PROP_RUNDECK_CONFIG_LOCATION`, which is where that name actually comes from.

**Behavior change isn't in the PR body.** Prebootstrap failure going from "continue silently" to "abort" is the right decision, but it's not mentioned under *Describe the solution* or in *Release Notes*. Worth one line — an operator with a broken config who used to get a confusing NPE now gets a clear abort instead.

fdevans:
Thanks @Jesus-Osuna-M, addressed all of these:

- Trimmed the duplicated Javadoc — the full "why early registration matters" explanation now lives only in `RundeckConfigEnvironmentPostProcessor`'s class doc; `Application#loadRundeckPropertySources()` cross-references it and keeps only the invariants unique to that method (dbCreate ordering, prebootstrap-success check).
- The `GrailsApp` lifecycle test no longer swallows the real failure: it captures whatever `app.run()` throws and rethrows it as the assertion's cause if the environment was never captured, instead of just failing on a null `capturedEnvironment`.
- `EnvironmentCaptor` now explicitly implements `Ordered` (`LOWEST_PRECEDENCE`) rather than relying on an unordered listener's implicit default.
- Fixed the temp directory leak — `deleteDir()` in cleanup instead of relying on `deleteOnExit()`, which doesn't remove non-empty directories.
- Replaced the hardcoded `"rundeck.config.location"` literal with `RundeckInitConfig.SYS_PROP_RUNDECK_CONFIG_LOCATION`.
- Updated the PR body (solution + release notes) to call out the prebootstrap-failure-aborts-startup behavior change explicitly.

Both test suites still pass (`ApplicationTest` 9/9, `RundeckConfigEnvironmentPostProcessorTest` 2/2). Re-requesting your review.

Jesus-Osuna-M (approved):
Approving — all six points from my previous review are addressed, and correctly:

- Javadoc trimmed on `loadRundeckPropertySources` (~30 → ~15 lines), with the post-processor class doc kept as the single source of truth.
- `catch (Throwable)` now captures the throwable and attaches it as the cause of an `AssertionError` when `capturedEnvironment` is null.
- `EnvironmentCaptor` implements `Ordered` explicitly.
- `tmpCfgDir.deleteDir()` in cleanup, with the ineffective `deleteOnExit()` calls removed.
- Property-source name now comes from `RundeckInitConfig.SYS_PROP_RUNDECK_CONFIG_LOCATION`.
- Behavior change documented in the PR body and Release Notes.

I also reviewed the `prebootstrapSucceeded` change in 35a1973 from scratch, since it came from a different review. The design holds up: `RundeckWebAppInitializer.onStartup()` calls `runPrebootstrap()` and ignores its return value, so on the containerized war path a failed prebootstrap used to leave `rundeckConfig` non-null but broken — the old `rundeckConfig == null` guard short-circuited and you got a downstream NPE. The flag makes that abort with the real error instead. Good regression test for it too.

Note `build_and_test` was still in progress when I approved — worth letting the containerized integration jobs finish before merging, since this changes startup ordering.

Four optional nits, none blocking:

1. `ApplicationTest."RunPreboostrap"` now mutates `prebootstrapSucceeded` as a side effect of `runPrebootstrap()` with no cleanup. On the happy path it leaks `true` (harmless), but if that spec ever fails it leaks `false` JVM-wide and every later spec reaching `loadRundeckPropertySources` throws `IllegalStateException` — one real failure becomes a cascade. A `cleanup: Application.prebootstrapSucceeded = null` would also make it consistent with the other specs, which all snapshot/restore.

2. `!prebootstrapSucceeded` unboxes a `Boolean`. Not reachable with null today, and `Application` isn't `@CompileStatic` so Groovy truth applies — but it becomes an NPE the day someone does add `@CompileStatic`. `prebootstrapSucceeded != Boolean.TRUE` is equivalent and safe under both.

3. The `if`/`throw` inside the `then:` block works (Spock only rewrites top-level expression statements), but reads more naturally at the end of `when:`.

4. `TestAppInitializer` calls `new InitializeRundeckPreboostrap().run()` directly rather than through `runPrebootstrap()`, so it never sets the flag. Likely moot — that one-arg `environmentPrepared(ConfigurableEnvironment)` isn't the Spring Boot 3 interface method (note the commented-out `//@Override`), so it looks like dead code. Worth confirming rather than assuming; if it is live, you'd get a double prebootstrap in tests.

